PR #22, opened 2020-03-22
Fix typo .check -> .is

mmiszy:


vriad:
This is my fault. I saw that another library puts their type guard behind a `.check()` method and I liked that name a lot better than `.is()`. So I started to switch the documentation over from `.is` to `.check`, must've gotten distracted halfway through. For the moment `.check` should be considered the "canonical" API and `.is` should be considered deprecated (but still supported until a v2 is released).

PR #44, opened 2020-05-03
Update Yum comparison

aaronjensen:
Hi there, I was just made aware of yup so I wanted to take a look at how it compared to zod and I found this.

It looks like properties within yup now default to non-optional.

The remaining point is interesting, but the conclusion of "But it's not acceptable that an object that's assignable to the inferred TypeScript type can fail validation by the validator it was inferred from." is probably unrealistic for a library like yup, which also validates the values, not just the shape. 

One would need a type system capable of expressing dependent types to really meet that criticism. The reality is that yup validates both the types and the values simultaneously. I think it definitely could improve the required array type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1040,20 +1040,7 @@ Doesn't support static type inference. 😕
 
 Yup is a full-featured library that was implemented first in vanilla JS, with TypeScript typings added later.
 
-Yup supports static type inference, but unfortunately the inferred types aren't actually correct. Currently, the yup package treats all object properties as optional by default:
-
-```ts
-const schema = yup.object({
-  asdf: yup.string(),
-});
-schema.validate({}); // passes
-
-type SchemaType = yup.InferType<typeof schema>;
-// returns { asdf: string }
-// should be { asdf?: string }
-```
-
-Yup also mis-infers the type of required arrays.
+Yup mis-infers the type of required arrays.
 
 ```ts
 const numList = yup


README.md
```diff
@@ -1040,20 +1040,7 @@ Doesn't support static type inference. 😕
...
-
-```ts
-const schema = yup.object({
-  asdf: yup.string(),
-});
-schema.validate({}); // passes
-
-type SchemaType = yup.InferType<typeof schema>;
-// returns { asdf: string }
-// should be { asdf?: string }
```
aaronjensen, line 1053: After re-reading this I'm not entirely sure the original statement made sense, so I'm not sure what you were saying initially.

It does indeed return `{asdf: string}` and if you specify `nonRequired()`, it returns `{asdf?: string}`. In the paragraph above, it says "Currently, the yup package treats all object properties as optional by default" which is the inverse of what's true as far as I can tell.

I don't know that I see the current default behavior of yup as a concern.

Here's their explanation https://github.com/jquense/yup#why-does-infertype-not-default-to-nonrequired

aaronjensen: Okay, I'm sorry, this took a little time to digest and I'm with you now. You're saying that the library itself considers them optional, but that the type marks them as non-optional. They don't totally explain their reasoning in the link I provided above. In the end, I think this is a valid criticism as it's going to be surprising to some.

I'll go ahead and close this out, sorry for the noise!

vriad:
Yep, you got it. 

I think you do raise a valid criticism of the language I use, specifically this bit: "But it's not acceptable that an object that's assignable to the inferred TypeScript type can fail validation by the validator it was inferred from".

Back when I wrote that there was a one-to-one mapping between all Zod and TypeScript features, but just a few days ago I released the first support for refinement types (ie custom validations) so it's now possible for a value to pass static validation and fail runtime validation in Zod. This is why I held out for so long on custom validators, but it any real codebase you're going to need to enforce types that can't be expressed in TS's type system. I modified the language accordingly 👍

aaronjensen:
@vriad I just came across https://github.com/jquense/yup/issues/593#issuecomment-605717756 

That may be a good idea to add into the comparison... It looks like I'm going to possibly be switching back to zod over this.

aaronjensen:
@vriad as of the new type definitions just released: `@types/yup` 0.29: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44589

It is no longer the case that things are typed as required even though they're optional, so that's good.

Unfortunately, they completely mangled the type signature, so it's not readable.

CleanShot 2020-05-21 at 21 12 33@2x

vriad:
Yeah, it's an extremely odd choice of defaults. Crazy to me that they're trying to used the term `required` to mean "non-empty" instead of "not undefined"...

I'll update the Yup comparison to reflect these changes to the types - thanks for pointing it out!

aaronjensen:
@vriad no problem.

And yeah... to be honest i'm not even sure what `required` means because some things say it can be null, but it in practice it must be non-empty. 

If it weren't for the fact that `yup` supports adding validation messages (which makes it work nicely with things like `react-hook-form`), then I'd probably switch us back to zod at this point. 

Of course, my sense is still that form validations are a separate concern, so I have mixed feelings on this.

vriad:
Yeah, I just started using react-hook-form for a new project and the built-in Yup support is...frustrating. It would be nice if they accepted any object with a proper set of methods + error handling conventions. Might look into that as a PR next week.

aaronjensen:
They do via `validationResolver`, though the next version will do away with built-in yup support and instead have separate packages for different libraries: https://github.com/react-hook-form/react-hook-form-resolvers

PR #54, opened 2020-05-28
WIP: add customized errors

blackrabbit99:
The PR allows to extend error for each type of validation using custom object so consumers will have flexible tool to distinguishing and handling errors
```
try {
    z.object({ points: z.number().error({e:1}), dots: z.boolean().error({e:2}) })
    .parse({
      points: '1',
      dots: 1,
    });
  } catch (e) {
    expect(e.errors[0].details).toEqual({e:1});
    expect(e.errors[1].details).toEqual({e:2});
  }
```


package.json
```diff
@@ -18,7 +18,9 @@
     "url": "https://github.com/vriad/zod/issues"
   },
   "homepage": "https://github.com/vriad/zod",
-  "dependencies": {},
+  "dependencies": {
+    "yarn": "^1.22.4"
```
tuchk4, line 22: why do we need `yarn` at the deps?

src/ZodError.ts
```diff
@@ -40,6 +41,16 @@ export class ZodError extends Error {
     ]);
   };
 
+  static fromObject = (details: object) => {
+    return ZodError.create([
+      {
+        path: [],
```
tuchk4, line 47: is path always equals to `[]`? 
Or is it updated somewhere else?

src/__tests__/object.test.ts
```diff
@@ -60,3 +60,16 @@ test('nonstrict', () => {
...
+
+test('object error', () => {
+  try {
+    z.object({ points: z.number().error({e:1}), dots: z.boolean().error({e:2}) })
+    .parse({
+      points: '1',
+      dots: 1,
+    });
+  } catch (e) {
+    expect(e.errors[0].details).toEqual({e:1});
```
tuchk4, line 72: Should we add tests for the `error.path`?

 ```
 expect(e.errors[0].path).toEqual('points');
````

Comment (tuchk4) [reply src/__tests__/object.test.ts:72]: As I understand correctly, `e.errors[0].message` always equals to `''`?

File: src/parser.ts [hunk 4]
```diff
@@ -7,6 +7,14 @@ export type ParseParams = {
   seen: { schema: any; objects: any[] }[];
 };
 
+const ZodErrorHandler = (messageOrObject: string | object): void => {
```
tuchk4, line 10: I am not familiar with the `zod` codestyle rules, but it is better to name functions int he camelCase style

```
zodErrorHandler
```

src/parser.ts
```diff
@@ -7,6 +7,14 @@ export type ParseParams = {
...
+  if(typeof messageOrObject === 'string') {
+    throw ZodError.fromString(messageOrObject);
```
tuchk4, line 12: maybe leave throw outside of the `zodErrorHandler` to not to break error stack trace?

src/parser.ts
```diff
@@ -29,42 +37,42 @@ export const ZodParser = (schemaDef: z.ZodTypeDef) => (obj: any, params: ParsePa
 
   switch (def.t) {
     case z.ZodTypes.string:
-      if (typeof obj !== 'string') throw ZodError.fromString(`Non-string type: ${typeof obj}`);
+      if (typeof obj !== 'string') ZodErrorHandler(def.customError || `Non-string type: ${typeof obj}`);
```
tuchk4, line 40: maybe pass `customError` as closure?

```
const zodErrorHandler = createZodErrorHandler(def.customError);
```

src/types/base.ts
```diff
@@ -80,6 +81,11 @@ export abstract class ZodType<Type, Def extends ZodTypeDef = ZodTypeDef> {
     return this;
   };
 
+  error(customError: object) {
```
tuchk4, line 84: I think it should also accept `customError` as `string`

package.json
blackrabbit99: removed. I have no yarn installed in global scope so i made a mistake. thank you

src/ZodError.ts
blackrabbit99: this attribute is filling differently so initialisation looks correctly

src/__tests__/object.test.ts
blackrabbit99: I have added ability to override message or use default. test updated

src/parser.ts
blackrabbit99: I have following styles in the repo looks like it is ok

src/parser.ts
blackrabbit99: Good point to discuss from one side more clear stack trace from other side code duplication :(

src/parser.ts
blackrabbit99: I will overthink it do you mean it will allow to clean stacktrace ?

src/ZodError.ts
```diff
@@ -40,6 +41,16 @@ export class ZodError extends Error {
...
+        message: '',
```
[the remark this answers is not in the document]
blackrabbit99: Resolved by ability to override message

src/types/base.ts
blackrabbit99: added functionality and test

tuchk4:
any updates?

vriad:
Large changes to the error handling system were published with zod@1.8. This approach wasn't what I ended up going with; it didn't allow the granularity of customization I was looking for. But needless to say thanks @blackrabbit99 for your great work. 

Check out the new Error Handling guide at https://github.com/vriad/zod/blob/master/ERROR_HANDLING.md

PR #79, opened 2020-07-07
Generic Type Support

stevengeeky:
And additional cleanup :) 🧹🧼

Support for generic types has been added in the form of `z.generic()` which extends an existing union type. Usage is as follows:

```typescript
const $Cat = z.object({
  type: z.literal('cat'),
  ability: z.literal('meow'),
});
const $Dog = z.object({
  type: z.literal('dog'),
  ability: z.literal('bark'),
});
const $Animal = z.union([$Cat, $Dog]);

const $AnimalAbility = z.generic($Animal, ($A) => $A.shape.ability);
type Ability = z.infer<typeof $AnimalAbility>; // "meow" | "bark"
```


vriad:
Hey Steven,

Thanks for submitting this PR and for your interest in Zod 👍

A lot of the below is feedback on submitting PRs in the future. I don't mean to be critical, just trying to explain how maintainers think about things. I'm always appreciative of people who make contributions and I like a lot of the changes you made.

PRs like this are kind of a nightmare. Because you did so much in one PR, it makes it really [hard for me](https://medium.com/@hugooodias/the-anatomy-of-a-perfect-pull-request-567382bb6067) to review the changes. Nearly every file has been changed, including all the test files. This means that I can't simply run the tests and have confidence that everything is still working as expected (since you may have accidentally made a breaking change and modified the tests to accommodate it). I'm not saying you did did this, but it requires a lot of manual verification for me to be absolutely sure.

I would also recommend running changes to the build/testing/linting pipeline by the maintainer first. Those things are part of the workflows we use to evaluate PRs so making changes to them make things harder on us. I ended up rolling back most of your tooling changes (gts, running tests from the build folder, changes to tsconfig). Things like getting rid of the namespaces and changing the name of the build folder just add unnecessary noise. 

In the future I strongly recommend splitting changes like this into multiple PRs.

A few other nitpicks:
- you left some commands in `package.json` as `npm.cmd` which only works on Windows I believe
- there's no need to Zod to exclusively throw ZodErrors. The ZodError class is only used to provide additional information about parsing issues. In other methods (e.g. ZodObject.pick), you can throw normal errors if an invalid input is provided
- I think the core ZodGeneric functionality would be better implemented as a method on ZodUnion, perhaps called `.distribute`. As I understand it, it doesn't necessitates a new class. Let me know if you disagree.
- In the readme I mention that .parse returns the exact instance you pass into it. Your approach using `parsedObject` violates that.

I like the improvements to pick and omit. Unfortunately I've already rewritten those methods since last pushed to master. Same story for parser.ts. By extension I didn't end up merging many of these changes because I'd already made such significant edits on my end. 

Given that I don't love the concept behind ZodGeneric, I've already rewritten large chunks of parser.ts/object.ts in the last few weeks, and all the noise from the tooling changes, I'm afraid I won't be able to merge this - there's just too many moving parts. 

I think the repo could benefit significantly from a better linting setup. If you wanted to submit a new PR that provides that (ideally using pure eslint, not gts) I would likely merge it. But hold off on that until I publish a new version to GitHub (shoudl be soon) because I've made some significant edits since the last push to master.

PR #107, opened 2020-08-02
Test that all errors are collected even if schema instance is reused.

brabeji:
It seems that some errors are omitted when reusing schema instance. Adding a simple test.


brabeji:
This happens specifically with `nulls`. From what I was able to guess, it's because they're marked as seen and parsed only once per given schema. The possible fix might be to mark only `object`s but I'm not 100% positive about that from a quick look at the code.

vriad:
Yep that's exactly what's happening. I'll improve that behavior in zod@1.11.

PR #108, opened 2020-08-03
Fix confirm example in README.md

ivan-kleshnin:
The current one has an invalid key name: `confirm vs confirmPassword` mismatch typo.


vriad:
Thanks 👍

PR #111, opened 2020-08-05
Add regexp validation for strings

pierreis:
This pull request adds a helper for validating strings against a regular expression. Usage is as follows:

    z.string().regexp(/^moo+$/, "Moo validation failed");

It also changes the `z.string.uuid()`, `z.string().url()` and `z.string().email()` to use a common implementation.


o-alexandrov:
How about letting users to pass an array of RegEx'es instead of a single one.

For example, email validation imho shouldn't have a single RegEx.
There won't be a valuable error message based on the following:
```ts
const emailRegex = /^((([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+(\.([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+)*)|((\x22)((((\x20|\x09)*(\x0d\x0a))?(\x20|\x09)+)?(([\x01-\x08\x0b\x0c\x0e-\x1f\x7f]|\x21|[\x23-\x5b]|[\x5d-\x7e]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])|(\\([\x01-\x09\x0b\x0c\x0d-\x7f]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF]))))*(((\x20|\x09)*(\x0d\x0a))?(\x20|\x09)+)?(\x22)))@((([a-z]|\d|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])|(([a-z]|\d|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])([a-z]|\d|-|\.|_|~|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])*([a-z]|\d|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])))\.)+(([a-z]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])|(([a-z]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])([a-z]|\d|-|\.|_|~|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])*([a-z]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])))$/i;
```

Instead, it's more useful to do something like below:
```ts
/**
 * Can check regexes here:
 * @see https://regexr.com/4hq79
 * Inspired by https://emailregex.com/ for the JavaScript case
 */
export const regexEmail = [
  {
    msg: `Minimum 5 chars`,
    exp: /^.{5,}$/,
  },
  {
    msg: `Maximum 254 chars`,
    exp: /^.{1,254}$/,
  },
  {
    msg: `Need @ symbol`,
    exp: /@/,
  },
  {
    msg: `Invalid part before @`,
    exp: /^[a-zA-Z0-9_\-\.@]+$/,
  },
  {
    msg: `Invalid part after @`,
    exp: /.*@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/,
  },
]
```

pierreis:
Thanks, that is a great example!

In this case, I would probably implement this by chaining regex calls, each with their own parameters, which are cumulative.

That is not necessarily more verbose than creating arrays of objects, but significantly simplifies the signature of the method 🙂

o-alexandrov:
Yes, I agree, it's more of a user preference. 
I'd vote for a **list of regex'es** to be passed though as it's simpler to reuse an array of regex'es between frontend and backend than exporting individual regexes to be used. (using `zod` on the frontend is unnecessary for us and, based on my observations, for many)

Surely, in our case we could create a helper to be used on the backend that loops over array and creates a chain dynamically, but I guess the argument for reusability is strong enough to be heard :)

o-alexandrov:
Another argument, some regex'es should be checked after another.
If you have to define the order with chaining, I believe it might hurt in other ways too:
- for example, you reuse regex'es between properties, then you would have to probably:
  - create unreadable helpers that enforce certain order

vriad:
Thanks @pierreis this is a great PR. Gonna lump this in with a release that goes out this weekend. Though I think I'll rename the method to `.regex` — it seems to be winning the popularity war, JavaScript be damned: https://stackoverflow.com/questions/3058246/which-is-the-correct-shorthand-regex-or-regexp

@o-alexandrov are you imagining that the user would be sharing a serialized array of regexes between the client and server? Since true codesharing between the client and server is so widespread now (yarn workspaces, etc), you should be able to share the actual Zod schema easily enough. I'd rather just choose the friendliest API (the chainable one) instead of designing around a use case that's less and less common. 👍

In any case, gonna merge this and we can overload the function to accept an array later if there's real demand.

o-alexandrov:
In my last several projects where we had to provide the user with clear errors, all of the regex'es were split into multiple ones
  - there was no attribute that needed regex validation, but had a single regular expression to express potential issues
    - w/ the example above on email or password validation, you pretty much always should provide with a list of expressions to provide your user with clear instructions

You could [try an example](https://anime.club/auth/sign-in) from my latest project on how we reuse validations between the frontend and backend.

---

Codesharing `zod` itself isn't necessary for the frontend.
Why? Because all you need there is looping over RegEx'es on textfield change.
- bundling zod on the frontend doesn't make sense, unless you provide your user w/ ability to not fill out the form and you have to check on form submission

pierreis:
I tend to think that any “complex” validation logic such as chains of regexes should be implemented in a custom function. I see many edge cases that would make any non-trivial use case complicated:

- Do we want to fail fast, or to return an error per regex?
- What if I want to not only set the message, but also the params?

I would favor a simple API over complicating it.

That however does highlight one of the current limitations of refinements: the inability of a refinement delegate to pass a modified error instead of having it defined at the root. One possible implementation of this, that would greatly simply implementing what you mention, would be to enable a delegate to throw a custom ZodError instead of returning a false value.

vriad:
@pierreis I just proposed a similar solution for custom refinements here: https://github.com/vriad/zod/issues/100#issuecomment-669048991

It definitely makes sense to provide `makeError` in refinements as well.

vriad:
@pierreis what do you mean "defined at the root"? If you mean you want to customize the error path, you can currently do that: https://github.com/vriad/zod#custom-validation

Having the ability to throw fully custom errors is still useful of course.

pierreis:
You can surely change the path, but it has to be define when creating the refinement, not at execution time.

For example, I would like to be able to overwrite the error code or set the error message from within the delegate function, preventing me to write a refinement that introduces anything more than very simple logic, that is not sufficient in all cases. I have found quite a few situations in which this is problematic, and requires passing this information inside the returned parameters and transforming the error outside of Zod.

vriad:
You'll be able to set the error code to a different ZodErrorCode but you won't be able to set it to an arbitrary string. For that you should use ZodErrorCode.CustomError and pass whatever data you care about in `params` (like you said). 

Do you want to be able to set the path at execution time? I'm not clear what your use cases are.

vriad:
@pierreis Merged in zod@1.11. Thank you! 🚀

Issue #120, opened 2020-08-18
Too strict url check


danielepolencic:
The following URL fails the validation:

```
http://nautil.us/issue/88/love--sex/the-hard-problem-of-breakfast
```

However, the same URL is accepted without any issue by the URL class:

```js
> new URL('http://nautil.us//issue/88/love--sex/the-hard-problem-of-breakfast').toString()
'http://nautil.us//issue/88/love--sex/the-hard-problem-of-breakfast'
```

I'm not familiar with the intricacies of validating URLs, so this might be a non-issue. But I thought about reporting it since I can use the URL just fine.

Using zod 1.10.1

vriad:
The first URL is parsing fine for me. The URL you're passing into `new URL` is different from the first: it contains two slashes after `nautil.us`. Nautilus still manages to serve the correct page, but in general double slashes won't work as expected in URLs.

I hadn't considered using `new URL` as a way of checking for valid URLs...I like that idea! I would accept a PR that switches URL validation over to that approach, since it feels more Javascript-native.

vriad:
Merged your PR and published in zod@1.11. Thanks Daniele!

PR #122, opened 2020-08-19
Switched from regex to URL constructor when validating URLs

danielepolencic:
This is meant to solve #120


vriad:
@danielepolencic Merged and published in zod@1.11. Thank you! 🚀

PR #124, opened 2020-08-20
fixed babel-jest import bug (#125)

typeetfunc:
I get an error when accessing `nullable` method when using `babel-jest`. I've discovered that the error is due to the ZodUnion file not being imported correctly.

`TypeError: Cannot read property 'create' of undefined` at ZodNumber.ZodType.nullable (node_modules/zod/src/types/base.ts:114:62)

I use explicit import of barrels files for fix this problems

Relates: 
https://github.com/vriad/zod/issues/125
https://github.com/vriad/zod/issues/119 
https://github.com/vriad/zod/commit/c6fff3a0e9be02da32dda0dc9e58fd853fb628a3

### DIFF
--- a/src/helpers/partialUtil.ts
+++ b/src/helpers/partialUtil.ts
@@ -1,4 +1,4 @@
-import * as z from '..';
+import * as z from '../index';
 
 export namespace partialUtil {
   export type RootDeepPartial<T extends z.ZodTypeAny> = {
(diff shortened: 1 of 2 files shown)


vriad:
Fixed this (and several other occurrences of the same issue) in zod@1.11. Thanks!

PR #132, opened 2020-08-31
Example of `dynamicRefinement` to validate by dynamic constraints.

brabeji:
I have a scenario in which some validation constraints are fetched from a server. The proposed changes might solve the use case without introducing a "validation context" feature.

There is an evident performance tradeoff - I'm creating (potentially lots of) schemas dynamically during validation to keep the API as simple as possible. Nonetheless, I think there are cases when such a solution is sufficient. What do you think?

Thanks again for the lib!


vriad:
A similar approach is now available in Zod 2 👍

brabeji:
Great news, I love it!

PR #139, opened 2020-09-02
feat: make nonstrict object props "unknown" rather than "any"

ForbesLindesay:
This makes them remain type safe, while allowing you to be flexible if you are parsing/validating large JSON objects and you only care about a few fields.

BREAKING CHANGE: un-validated fields on "nonstrict" objects are `unknown` instead of `any`


vriad:
Thanks for this PR. Zod 2 is now in beta and resolved this problem differently. Objects are "nonstrict" by default (in that they allow unknown keys) but there is also no index signature whatsoever on object schemas unless you use `.catchall()`.

PR #141, opened 2020-09-03
use setPrototypeOf polyfill

mad-gooze:
Object.setPrototypeOf breaks in old browsers (like, legacy Android && ios)

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,9 @@
     "url": "https://github.com/vriad/zod/issues"
   },
   "homepage": "https://github.com/vriad/zod",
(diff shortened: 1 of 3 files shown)


vriad:
Thanks for this PR. Seems like browser compatibility (something I didn't consider much when building Zod) is becoming more of an issue for people.

I ended up using a workaround from [this issue](https://stackoverflow.com/a/48342359) instead of using a polyfill. I want to keep zod at zero dependencies. Published in zod@1.11.5. Thanks again!

mad-gooze:
thx @vriad

Issue #152, opened 2020-09-15
Specific Optional Type


sylvanaar:
I was just playing around with various things, and I noticed optional-ness is reapplied to things depending on how you write it, for example if I write it this way (see below), `message` is "double optional" since that is what the code says it is. But double optional doesn't have any meaning - so it can be flattened.

so i have this union which makes sense given the code below

`    message: z.ZodUnion<[z.ZodUnion<[z.ZodString, z.ZodUndefined]>, z.ZodUndefined]>;`

can it become:  `    message: z.ZodUnion<[z.ZodString, z.ZodUndefined]>;`

Optional is not just a union with undefined though, it also includes "missing" / "not there at all". So there seems to be something that is not being captured currently.

Unions also seem to be a weak point in the error messaging which is otherwise top notch. Having a specific optional - lets you give the error message for the optional type instead of the union which i think is a pretty big deal.

Right now all unions (so all optionals) give the same error:

```
[Error: 1 validation issue(s)

     Issue #0: invalid_union at message
     Invalid input
   ]
```
vs a non-optional of the same

```
[Error: 1 validation issue(s)

     Issue #0: invalid_type at message
     Expected string, received number
   ]
```

This difference only make sense if you know that optional was being represented by a union with undefined in which we no longer know which option in the union is our type.

Code i was experimenting with is below for reference

```ts
const ErrorBase = z
    .object({
        message: z.string(),
        stack: z.string(),
    })
    .partial();

export const ErrorType = ErrorBase.extend({
    name: z.string(),
    code: z.number(),
    context: z.record(z.any()),
    cause: ErrorBase,
}).partial();
```

sylvanaar:
Will you accept a PR?

vriad:
Go ahead and submit the PR. My initial reaction to this concept was negative because I don't like the redundancy of representation. But I'm warming up to the idea. Your work on this looks excellent!

vriad:
> Optional is not just a union with undefined though, it also includes "missing" / "not there at all". So there seems to be something that is not being captured currently

Practically speaking how do you want Zod's behavior to change? Zod already handles this correctly:

```ts
z.object({
  name: z.string().optional()
}).parse({}) // passes

// inferred type is { name?: string }
// note the question mark
```

sylvanaar:
> Practically speaking how do you want Zod's behavior to change? Zod already handles this case correctly:
> 
> ```ts
> z.object({
>   name: z.string().optional()
> }).parse({}) // passes
[...]

PR #155, opened 2020-09-16
✨ Feature: First-Class Optional Types

sylvanaar:
This feature introduces a first class optional type. Described more fully in #152 

This is a WIP - mostly done except for some typing issues having to do with infinite recursion on `ZodAnyType` All tests pass in TS 4.0 - but on 3.7 errors are present,

I would like feedback on the PR before I call it complete - hence it is a draft.

1. The existing union type with undefined had some shortcomings. It did not capture whether something could be missing as well as undefined. These issues can be important in objects and tuples.
2. The error messages returned from the union based optionals were generic and confusing.
3. An optional of an optional returned a union of a union before - with this pr it returns the original optional

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
   },
   "devDependencies": {
     "@types/jest": "^25.1.4",
+    "@types/node": "14.10.3",
     "jest": "^25.1.0",
     "make-coverage-badge": "^1.2.0",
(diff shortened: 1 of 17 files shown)


colinhacks:
Great work on this! This same functionality was introduced in Zod 2 around the same time as this PR.

Issue #161, opened 2020-09-22
Zod 2 beta 5: parse() adds optional fields with 'undefined' as a value


grreeenn:
Hi @vriad,

I've upgraded to Zod 2 beta and it works cool.

The first bug I've met now, is that if you send an object without certain optional fields, `.parse()` adds them with undefined value.

For instance, my schema looks like that (simplified):
```ts
export const SNamedEntity = z.object({
  _id: string().refine(validateUUID),
  name: z.string().max(100),
  createdAt: string().refine(validateDateString).optional(),
  updatedAt: string().refine(validateDateString).optional(),
  isActive: boolean().optional(),
});
```
And the object I send to it has only `_id` and `name`. However, when I set a breakpoint after I `.parse()` this object with the with the above schema, I see that all of the optional fields are present there now: 

```js
{ 
  _id: '5f6418f13b179e3da326a1aa',
  name: "John Doe",
  createdAt: undefined,
  updatedAt: undefined,
  isActive: undefined,
}
```

It freaks out Mongo, it thinks that I'm trying to overwrite its internal createdAt field with undefined and throws me an error.

Interestingly, if I send a valid date in createdAt field, mongo is totally allright with that (it ignores it as it should). Only `undefined` trips it :)

grreeenn:
actually, I'm sorry - it's `.safeParse()` that does it, not `.parse()`

sylvanaar:
https://github.com/vriad/zod/blob/a014ff6d041801e13021bc21fd359b3e20318ffd/src/parser.ts#L369

Probably should be

```ts
        const keyValidator = dataKeys.includes(key)
```

o-alexandrov:
the issue exists in beta 8 as well

vriad:
I'm having trouble replicating this:

```ts
export const SNamedEntity = z.object({
  _id: z.string(),
  name: z.string().max(100),
  createdAt: z
    .string()
    .optional()
    .refine(() => true),
  updatedAt: z
    .string()
    .optional()
    .refine(() => true),
  isActive: z.boolean().optional(),
});

const result = SNamedEntity.safeParse({
  _id: 'asdf',
  name: 'adfasdfkjasdf',
});

console.log(JSON.stringify(result, null, 2));
```

results in

```
{
  "success": true,
  "data": {
    "_id": "asdf",
    "name": "adfasdfkjasdf"
  }
}
```

In the future please provide a full code example that a maintainer can just copy/paste to see the issue to avoid back and forths like this.

[...]

PR #164, opened 2020-09-23
Fix for missing values in objects #161

sylvanaar:


vriad:
I fixed this issue (using a different approach) in zod@2.0.0-beta.9 👍

PR #165, opened 2020-09-23
Use .default() value only when schema is null or undefined

### LINKED ISSUE #162: [Discussion] zod2 - default() condition
The current implementation replaces all falsy values to the default.   
I was surprised by this. I expected that only `undefined` is checked - but maye I'm just biased because I am used to Joi: [Joi.default()](https://joi.dev/api/?v=17.2.1#anydefaultvalue)

There are also cases where we don't want this.  
* a number may have a default of `5`, but `0` could also be a valid value
* and also `null` can sometimes be a valid value (different from the default)

The options that we have:  
1. `default()` could have a parameter to decide which to use
1. we could have 3 different functions: `defaultIfFalsy()`, `defaultIfUndefined()`, `defaultIfUndefinedOrNull()`
1. or just document in detail how `default()` function works and force the users of the lib to create custom transformers for the other cases

marlonbernardes:
**Before this change**:
```ts
const schema = z.object({
  sandbox: z.boolean().default(true)
});

// Good - returns { sandbox: true } - which makes sense as it's not supplied
schema.parse({}); 

// Bad! - returns { sandbox: true } even though I've provided a  value for it.
schema.parse({ sandbox: false }); 
```

**After this change**:
```ts
const schema = z.object({
  sandbox: z.boolean().default(true)
});

// Good - returns { sandbox: true } - which makes sense as it's not supplied
schema.parse({}); 

// Good - returns { sandbox: false } as a value was supplied to the sandbox property
schema.parse({ sandbox: false }); 
```

As I'm using the null coalescing operator it means either `null` or `undefined` values will cause the default value to be used.
We might want to limit this only to `undefined` values so it matches the same behaviour of `joi` and `yup`. Let me know what you think.

Fixes #162

### DIFF
--- a/README.md
+++ b/README.md
@@ -1595,7 +1595,7 @@ const stringWithDefault = z.transformer(
 );
 ```
 
-Equivalently you can express this using the built-in `.default()` method, available on all Zod schemas.
+Equivalently you can express this using the built-in `.default()` method, available on all Zod schemas. A default value will be used if the schema is `null` or `undefined`.
 
 ```ts
 z.string().default('default value');
--- a/src/__tests__/transformer.test.ts
+++ b/src/__tests__/transformer.test.ts
@@ -59,6 +59,27 @@ test('default', () => {
   expect(data).toEqual('asdf');
 });
 
+test('default when property is null or undefined', () => {
+  const data = z.object({
+    foo: z.boolean().nullable().default(true),
+    bar: z.boolean().default(true)
+  }).parse({ foo: null });
+
+  expect(data).toEqual({ foo: true, bar: true });
+});
+
+test('default with falsy values', () => {
+  const schema = z.object({
+    emptyStr: z.string().default('def'),
+    zero: z.number().default(5),
+    falseBoolean: z.boolean().default(true)
+  })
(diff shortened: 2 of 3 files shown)


vriad:
Merged a version of this. The default value should only be applied if the value is `undefined` not `null`. Thanks for doing this! Great PR.

ricardo-devis-agullo:
How can I use default for nullable values too (like [fromNullable ](https://gcanti.github.io/io-ts-types/modules/fromNullable.ts.html)type does in io-ts)?

Right now for every value I have to inline something like this
```typescript
const defaultValue = 'MYDEFAULTVALUE';
const schema = z.object({
  nullableField: z.string().nullish().transform(x => x == null ? defaultValue : x);
});
```

Is there another way of doing this? Or some sort of function that I can reuse keeping the static type inference?

ricardo-devis-agullo:
I think I found something that works
```typescript
function fromNullable<O, D, I>(schema: z.ZodType<O, D, I>, defaultValue: O) {
  return schema.nullish().transform((val) => (val == null ? defaultValue : val));
}
```

Issue #179, opened 2020-10-01
Failed async refinement causes other `issues` to not be returned


ntilwalli:
When using the new `safeParseAsync` an async validation failure causes the `issues` array to only contain the async failure, the other errors are not included in the returned value.  Is this expected? The failing test is posted below.
```
import * as z from 'zod'

test('testing zod', () => {
  expect.assertions(2)
  const base = z.object({
    hello: z.string(),
    foo: z.number()
      .refine(async (data) => {
        return false
      }, {message: "invalid"})
  })
  const result = base.safeParseAsync({hello: 3, foo: 3})

  return result.then(r => {
    expect(r.success).toBe(false)
    if (r.success === false) 
      expect(r.error.issues.length).toBe(2) // <--- `issues` has length of 1 here
  });
});
```

sylvanaar:
I think your test should run them `async` vs `sync` and compare the run count - rather than using a magic number.

ntilwalli:
```
import * as z from 'zod'

test('testing zod', () => {
  const base = {
    hello: z.string(),
    foo: z.number()
  }

  const testval = {hello: 3, foo: 3}
  const result1 = z.object({...base, foo: base.foo.refine(data => false)}).safeParse(testval)
  const result2 = z.object({...base, foo: base.foo.refine(async data => false)}).safeParseAsync(testval)
  
  const r1 = result1
  return result2.then(r2 => {
    if (r1.success === false && r2.success === false) 
      expect(r1.error.issues.length).toBe(r2.error.issues.length) // <--- r1 has length 2, r2 has length 1
  });
});
```

vriad:
Fixed in zod@2.0.0-beta.12 thanks to some great work by @ntilwalli. Thanks Nikhil!!!

PR #187, opened 2020-10-09
fix: add PseudoPromise catch for use in ZodType.object validation

ntilwalli:
This PR fixes the async multi-error accumulation bug (#179) by adding a `catch` method to the `PseudoPromise` class and using it in while parsing object schemas.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -375,20 +375,22 @@ export const ZodParser = (schema: z.ZodType<any>) => (
         if (!keyValidator) continue;
 
         // check if schema and value are both optional
-        try {
-          keyValidator.parse(undefined, {
-            ...params,
-            path: [...params.path, key],
-          });
 
-          // const keyDataType = getParsedType(data[key]);
-          if (!Object.keys(data).includes(key)) {
-            // schema is optional
-            // data is undefined
-            // don't explicity add undefined to outut
-            continue;
-          }
-        } catch (err) {}
+        // const keyDataType = getParsedType(data[key]);
+        if (!Object.keys(data).includes(key)) {
+          try {
+            const output = keyValidator.parse(undefined, {
+              ...params,
+              path: [...params.path, key],
+            });
+            if (output === undefined) {
+              // schema is optional
+              // data is undefined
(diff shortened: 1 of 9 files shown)


src/parser.ts
```diff
@@ -943,26 +969,44 @@ export const ZodParser = (schema: z.ZodType<any>) => (
...
-        ERROR.addIssue(
-          makeError({
-            code: ZodIssueCode.custom,
-            message: 'Invalid',
-          }),
-        );
+      let resolvedValue = await PROMISE.getValueAsync();
+ 
+      if (resolvedValue !== INVALID) {
+        let someError: boolean = false;
```
vriad, line 975: I don't like that the custom checks short-circuit once an error is encountered. In general Zod tries to surface as many errors as possible upfront, instead of doling them out one by one. Is there some other reason you implemented it this way?

vriad:
WOW this is some great work. I started trying to fix #179 before I saw this PR and I arrived at the exact same solution. Wonderful! I've already merged this locally and made some some stylistic changes. 

I left one comment regarding the `someError` short-circuiting logic. I'm switching back to a `Promise.all` approach that runs all the custom checks in parallel and getting rid of the short-circuit. Unless you had a compelling reason for doing it that way?

src/parser.ts
ntilwalli: My motivation here was async validations which hit a server, which are costly and I don't want to run them unless the sync validations succeed.  So (for example) if a username doesn't fit the expected regex, skip checking if the username already exists.

I see the value of surfacing all the errors up front.  Could we make it an optional setting, so it defaults to running them all in parallel, but we could opt-in to running them serially if we want?

ntilwalli:
Nice, I'm glad we arrived at the same solutionI I replied to the code comment. I consider avoiding unnecessary async validations to be useful.  What do you think about offering serial custom validation checking as an opt-in?

src/parser.ts
vriad: Sure I can implement this. How does `runAsyncValidationsInSeries` sound as a parameter name?

PR #189, opened 2020-10-11
feat: add the current path as a refinement `check` parameter

ntilwalli:
This is a follow-up PR to (#187). It's not intended for merge, but just to ask the question, what is the proper way to integrate the path into a custom refinement error message?  I've made a PR which injects the `params.path` from the parser into the refinement `check` args.  Is there a better way?  For example if I want to throw a ZodError in my custom refinement with the code of `ZodIssueCode.invalid_string` it requires the `path` to be set but in the current signature the refinement validation function doesn't have access to the path that is currently being validated.  This PR does the work of passing that information down from the parser.

This PR also includes a change which swaps out parallel custom refinements validations (`Promise.all`) for a serial approach so failed earlier validations will not cause later validations to be run.  This is useful, for example, when mixing sync and async validations. When a earlier-in-the-queue sync validation fails, a more costly async validations should not be run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1596,7 +1596,7 @@ const stringWithDefault = z.transformer(
 );
 ```
 
-Equivalently you can express this using the built-in `.default()` method, available on all Zod schemas.
+Equivalently you can express this using the built-in `.default()` method, available on all Zod schemas. A default value will be used if the schema is `null` or `undefined`.
 
 ```ts
 z.string().default('default value');
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ Some other great aspects:
 - Immutability: methods (i.e. `.optional()` return a new instance
 - Concise, chainable interface
 - Functional approach: [parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)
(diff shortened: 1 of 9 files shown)


vriad:
I added this in zod@2.0.0-beta.10 using a similar approach. Here's a demonstration of how to achieve this using the more advanced `.refinement` method (which is slightly more advanced than `.refine`)

```ts
const foo = z.string().refinement(
  _val => false,
  (_, ctx) => {
    return {
      code: ZodIssueCode.custom,
      message: `bad path: ${ctx.path.join('.')}`,
    };
  },
);

const data = z.object({ foo });

const t1 = await data.spa({ foo: 'asdf' });
console.log(t1);
```

If you want to use the path in your validation logic, you can use the even more powerful `._refinement` method:

```ts
const noNested = z.string()._refinement((_val, ctx) => {
  if (ctx.path.length > 0) {
    ctx.addIssue({
      code: ZodIssueCode.custom,
      message: `schema cannot be nested. path: ${ctx.path.join('.')}`,
    });
  }
});

const data = z.object({
  foo: noNested,
});

// passes
noNested.parse('asdf');

// fails
data.parse({ foo: 'asdf' });
```

PR #208, opened 2020-10-31
Generate both CommonJS and ES modules

cbenz:
Related to #175 

Based on this article: https://blog.logrocket.com/publishing-node-modules-typescript-es-modules/

I added a new tsconfig file to generate a new ESM bundle. I changed the output directory from `lib/src` to `lib/{cjs,esm}/src`. I added a command to `npm run build` (that could have been parallelized with [concurrently](https://www.npmjs.com/package/concurrently) but maybe it's overkill).

I tested `npm run build` and the ES modules worked with `npm link` from a local application that uses ES modules with [Snowpack](https://www.snowpack.dev/) (and I confirm that the CommonJS build did not work because of some error with those generated lines: `Object.defineProperty(exports, "__esModule", { value: true });`).

Obviously I did not tested `npm publish`, so I hope I did not break the publication process.

Note: I did not understand the need to include `tsconfig.package.json` in the `lib/src` directory, so I kept it as-is in the `lib/{cjs,esm}/src` directories.


vriad:
Thanks for this wonderful PR! I made some small changes overriding and cleaned up all the tsconfigs — there were several redundant settings. Hopefully I didn't break anything.

Turns out there's no reason why the `tsconfig.package.json` was getting copied into the `lib` directory 🤷‍♂️ So I got rid of that.

`npm publish` worked fine. Please me know if the ESM module is working as expected! Once I get confirmation, I'll update the README.

Is there any specific instructions for how to consume the ES6 module? Do you have to use `import * as z from 'zod/lib/esm'`? Or does it work automatically in a system that supports ES6 modules?

cbenz:
I can confirm that it worked for me now, doing `import * as z from "zod"`. No need to reference `zod/lib/esm` thanks to the `"module": "./lib/esm/src"` key in `package.json`.

vriad:
Excellent! Thanks Christophe!

Issue #205, opened 2020-10-29
Slow parsing performance


mgilbir:
I've just started using this library to check that parsed JSON objects conform to the types I define. I'm really enjoying it, but I'm afraid I've hit a performance snag.

Everything works fine with small tests but when I ran it on a slightly bigger input file (~10MB) I noticed that it was really slow.

I've tried the latest version and the beta, and `check` vs `parse`, with similar results. The beta is faster, but I don't see a big difference between `check` and `parse`.

After profiling the `check` version in the beta, I'm seeing calls to `.check` taking in the 500ms-1100ms range per object. 
![image](https://user-images.githubusercontent.com/167588/97613435-fc568d00-1a18-11eb-9d53-89899c718724.png)

Are those numbers typical, or am I doing something wrong with the schema definitions?

My schema definitions look like:
```
const EntryJsonSchema = z.object({
  a: z.string().optional().nullable(),
  b: z.string().optional().nullable(),
  id: z.string().optional().nullable(),
  creation: z.string().optional().nullable(),
  content: ContentJsonSchema.optional().nullable(),
  labels: z.string().array().optional().nullable(),
  answers: AnswerJsonSchema.array().optional().nullable(),
  results: ResultJsonSchema.array()
    .optional()
    .nullable(),
});

const ContentJsonSchema = z.object({
  id: z.string().optional().nullable(),
  title: z.string().optional().nullable(),
  version: z.union([z.number(), z.string()]).optional().nullable(),
});

const AnswerJsonSchema = z.object({
  key: z.string().optional().nullable(),
  value: z.any().optional().nullable(),
});

const ResultJsonSchema = z.object({
  key: z.string().optional().nullable(),
  value: z.any().optional().nullable(),
});
```

I'm really hoping that there is a way to speed it up, as this is too expensive for the use case where I'll have to process files with 100k+ objects.

Thanks!

vriad:
Honestly I've never really stress tested Zod with payloads that large. I suspect there is lots of low-hanging fruit that would fastly improve performance. 

I'm a little confused though. Zod is taking 500+ms to parse _each object_? How big is each object? Is it possible you're hitting memory limits of the Nodejs process? Without a sample data set it's hard for me to duplicate this.

mgilbir:
Thanks for taking the time to think about this. 

[...]

PR #218, opened 2020-11-10
delay costly makeError to when it is needed

mgilbir:
In the context of issue #205, this change makes parsing 4x faster in my tests.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1,11 +1,19 @@
 import * as z from './types/base';
 import { ZodDef } from './index';
-import { ZodError, ZodErrorCode, ZodSuberror, ZodSuberrorOptionalMessage } from './ZodError';
+import {
+  ZodError,
+  ZodErrorCode,
+  ZodSuberror,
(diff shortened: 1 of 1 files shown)


vriad:
Can you summarize what you did here? Everything was prettified with a shorter `printWidth` so it's hard to find your changes.

vriad:
Ah I just saw you're other comment. Basically your moving the `makeError` definition out of the parsing function. Makes sense! I'll play around with this.

mgilbir:
Agh, sorry! I only saw now how messy the diff got due to applying prettier. I was rushing to send this between meetings and I didn't double check the diff before pushing 😞

I moved the `makeError` out fo the parsing because in the profiler I saw that the `ZodError` generation was always present and taking a significant amount of time on JSON blobs that never error. Doing less work on a heavy-hit function was bound to speed things up.

I didn't see any other obvious things in the profiler, so I didn't dig deeper.

colinhacks:
Merged into Zod v1 and v2-beta. Thanks @mgilbir, great work on this!

Issue #227, opened 2020-11-17
Repo restructuring to remove circular dependencies


o-alexandrov:
### Related issues

- https://github.com/colinhacks/zod/issues/175#issuecomment-724316421
- https://github.com/colinhacks/zod/pull/226

### RFC

- [ ] `base`s shouldn't import other modules
  - for example, `src/types/base.ts` should be reconsidered
- [ ] api reused in other modules should be modularized, for example:
  - `src/parser.ts` should be split into two files
    - `src/parser/index.ts` & `src/parser/types.ts` so the circular dependency between ZodError & parser is removed
      - and a similar approach could be done on other instances of circular dependency

colinhacks:
I don't know how to do this without putting everything into one file. Circular dependencies are supposed to be allowed as long as you're careful about loading order. There's a reason `base.ts` imports all those things, it needs them to provide methods like `.optional()`. Perhaps something like this would solve the problem: https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de

o-alexandrov:
> Circular dependencies are supposed to be allowed as long as you're careful about loading order.

- I have to disagree. It is theoretically possible to add builds/bundlers' configs and add bundling of `zod` to `tests`. It'd be an odd way, but one that would guarantee the building/bundling w/ zod is safe.
  - whereas if we remove circular deps completely, we can forget about a possibility of having such issue & not to test it

I believe the two points in the RFC should completely remove all circular deps we have in the project.
- I'll attempt to create a PR with necessary restructuring to examine after #231 is closed or if it hangs for more than 2 weeks.
- the only difficulty I predict to have is with classes, I will probably have to modularize some parts of interdependent classes

Personally, I don't believe it's possible to get rid of circular deps with the `internal` pattern.
- I created a #231 PR though indicating the problem isn't resolved using such pattern
  - I like the ease of using `madge` to observe the issue is still present

colinhacks:
Zod has been building/bundling and working fine for a long time with circular dependencies. The most recent issue with bundling is #222 and it was due to a dumb mistake on my part (using def.catchall.constructor.name to see if an object schema's catchall is a ZodNever instance).

[...]

PR #226, opened 2020-11-17
Remove circular dependencies

o-alexandrov:
It lets us not to guess where the circular dependencies are.
Instead, we could have a tool that reports it to us (pls see a screenshot below).

I saw your [comment](https://github.com/colinhacks/zod/issues/175#issuecomment-724316421
) and wanted to help:
> It seems to be related to some tricky circular dependency issues that are doing to take some time to figure out, so I'm removing the ESM module in zod@2.0.0-beta.21. 

Screen Shot 2020-11-17 at 12 11 20 AM

Issue #225, opened 2020-11-17
Batch repo enhancements


o-alexandrov:
Let me please propose batch of changes to the DX, so it'd be easier for others to contribute.
Please let me know what you would approve, so I'd know what to help you with :)

### Todo (Not started yet)

- [ ] add [automated releases generation](https://medium.com/p/44c015f38fd6) and make commits messages more strict
  - as a user of this library, it was a bit time consuming to browse commits to realize what changed between `beta` versions

### Completed
- [x] use `typescript-eslint` instead of `tslint`
  - it's been almost 2 years since the standard in TS linting is `ESLint`, [more info](https://eslint.org/blog/2019/01/future-typescript-eslint)
- [x] add `.log` files to `.gitignore`
  - for example, there's already a mistakenly uploaded `yarn-error.log`
- [x] bump prettier & use community defaults
  - let's use the defaults, so it's more likely to please higher % of contributors
- [x] ~~add `.npmignore` to exclude unrelated files from the deployed package~~
  - improve the `files` + `exclude` instead
- [x] remove empty `dependencies` object from package.json
- [x] add `husky` to add `git` hooks
  - it's important to automatically prevent issues like in #218 from happening, when there are bunch of unrelated formatting changes
    - at least automatic code formatting should be in place
- [x] add official funding references of your GitHub sponsors link to the `package.json`
  - I suggest to start with the use described in [this PR from react-virtual](https://github.com/tannerlinsley/react-virtual/pull/78/files)

colinhacks:
This all sounds great to me. I'm very open to a PR for these things. I'm also open to giving you commit access to the repo since you've been contributing to Zod for a long time, let me know 👍

o-alexandrov:
Thank you. If you kindly grant commit access, I plan to carefully follow existing patterns in naming of commits, branches, and will always ask before doing any action for the first time.

I'm glad to contribute these changes in the next 48 hours.

o-alexandrov:
Realized there are two other improvements, copied to the list above:
- add official funding references of your GitHub sponsors link to the `package.json`
  - I suggest to start with the use described in [this PR from react-virtual](https://github.com/tannerlinsley/react-virtual/pull/78/files)
- add [automated releases generation](https://medium.com/p/44c015f38fd6) and make commits messages more strict
  - as a user of this library, it was a bit time consuming to browse commits to realize what changed between `beta` versions

jameschensmith:
[...]

Issue #215, opened 2020-11-06
ESM module in Zod 2 beta causing runtime error


lukeramsden:
With Zod 2.0.0-beta.20:

![Screenshot](https://i.imgur.com/0kYAYgU.png)

With Zod 1.11, no error.

Only seems to happen when using the object type
```ts
// no crash
z.string();
z.date();
// crashes
z.object({});
z.object({
  id: z.string(),
});
```

alfonmga:
It's crashing for me too. Zod works fine on local but when I deploy my Next.js app to production (Vercel) it crashes with the error message: `TypeError: Prototype value can only be an object or null`.

Affected versions: `2.0.0-beta.20` (latest), `2.0.0-beta.19` and `2.0.0-beta.18`.

It works fine on `2.0.0-beta.17`, so looking at the commits pushed after `2.0.0-beta.18` ESM modules changes could be the origin of the issue.

o-alexandrov:
Issue is most probably unrelated to `react native`, but related to ESM introduction only.
Yes, exactly, zod worked fine on `2.0.0-beta.16`, but tried to upgrade to `.20` and saw an issue:
```
Object prototype may only be an Object or null: undefined
```

vriad:
I removed the ESM module in `zod@2.0.0-beta.21`. If you upgrade your builds should work again.

I've also reproduced this issue. Now that I'm aware of this I'll be able to guarantee ESM compatibility before trying to ship an ESM module again. It's not trivial to fix so it'll take some time. There are certain internal circular imports inside Zod's source code that work fine with CommonJS modules but not with ESM.

vriad:
Gonna re-open because the root problem here isn't solved :(

vriad:
@aabedraba @alfonmga Actually I suppose this particular issue is solved, I'm re-opening #175 instead.

PR #231, opened 2020-11-19
[WIP] - attempt to use "internal" pattern to fix circular deps

o-alexandrov:
This PR attempts to use [internal pattern](https://medium.com/p/a04c987cf0de) that @colinhacks proposed to use in [this comment](https://github.com/colinhacks/zod/issues/227#issuecomment-729856633).

Unfortunately, it doesn't seem to resolve the problem.
Now, the circular dependency appears to be between `internal` <=> `module`s (that depend on other modules, thru the use of `internal`).

@mweststrate I apologize for pinging and please feel free to ignore. However, it'd be wonderful to know the view of the author of this pattern. Thank you :)

Result:
```sh
$ yarn madge src/index.ts --circular
Processed 44 files (701ms) 

✖ Found 37 circular dependencies!

1) PseudoPromise.ts > internal.ts
2) internal.ts > ZodDef.ts > types/any.ts
3) internal.ts > ZodDef.ts > types/array.ts
4) internal.ts > ZodDef.ts > types/bigint.ts
5) internal.ts > ZodDef.ts > types/boolean.ts
6) internal.ts > ZodDef.ts > types/date.ts
7) internal.ts > ZodDef.ts > types/enum.ts
8) internal.ts > ZodDef.ts > types/function.ts
9) internal.ts > ZodDef.ts > types/intersection.ts
10) internal.ts > ZodDef.ts > types/lazy.ts
11) internal.ts > ZodDef.ts > types/literal.ts
12) internal.ts > ZodDef.ts > types/map.ts
13) internal.ts > ZodDef.ts > types/nativeEnum.ts
14) internal.ts > ZodDef.ts > types/never.ts
15) internal.ts > ZodDef.ts > types/null.ts
16) internal.ts > ZodDef.ts > types/nullable.ts
17) internal.ts > ZodDef.ts > types/number.ts
18) internal.ts > ZodDef.ts > types/object.ts
19) internal.ts > ZodDef.ts > types/optional.ts
20) internal.ts > ZodDef.ts > types/promise.ts
21) internal.ts > ZodDef.ts > types/record.ts
22) internal.ts > ZodDef.ts > types/string.ts
23) internal.ts > ZodDef.ts > types/transformer.ts
24) internal.ts > ZodDef.ts > types/tuple.ts
25) internal.ts > ZodDef.ts > types/undefined.ts
26) internal.ts > ZodDef.ts > types/union.ts
27) internal.ts > ZodDef.ts > types/unknown.ts
28) internal.ts > ZodDef.ts > types/void.ts
29) internal.ts > ZodError.ts
30) internal.ts > codegen.ts
31) internal.ts > defaultErrorMap.ts
32) internal.ts > helpers/maskUtil.ts
33) internal.ts > helpers/objectUtil.ts
34) internal.ts > helpers/partialUtil.ts
35) internal.ts > isScalar.ts
36) internal.ts > parser.ts
37) internal.ts > types/base.ts
```


mweststrate:
Didn't check the PR, but the pattern doesn't _solve_ circular dependencies
(they are inherit to the module system), but rather to be able to control
them by making loading order predicatable. So just make sure that things
that are required in the module initialization of another module are
required _first_. So if a class is used as a base class of another class,
make sure to put it on top of the internal file. If you have 2 modules that
require each other _during inititialisation_ you will have to split those
up into smaller modules.

o-alexandrov:
Thank you very much for the quick answer and my apologies, I was too overhyped when read your article :D
- even though I read it in full, it slipped from me that the only thing we are doing using this pattern is:
  - introducing a manageable way to order modules' initialization

@colinhacks please let me know your opinion on how should we proceed to fix the original problem (#215) this PR attempts to solve:
- unpredictable breaks in the **production** (!) bundles

Arguments:
- while it maybe relatively easy to apply `internal` pattern to few circular dependencies in a project
  - the number (40+) of modules in `zod` makes it too complicated to organize even analyzing the dependencies' tree
- order of modules is not actually guaranteed (at least not in webpack 5) https://github.com/webpack/webpack/issues/11871
  - I understand this bug will be probably resolved, but do we want to live with such possibility in other builders/bundlers?
- even if we fix the problem in #215, what will inform us in the future, that the order is still valid
  - we'd either have to add a lot of unusual (crazy) tests or we risk to hit this issue again
- I suggest to **completely fix circular dependencies** by applying strategy described in #227
  - and add `madge` to `husky` automated linting checks, described in #225 
    - that way we completely get rid of the #215 issue in all builders/bundlers and we'll be protected in the future

o-alexandrov:
Closing in favor of #226

PR #237, opened 2020-11-27
Correct expected type name in bigint parse errors

jeremyBanks:


colinhacks:
Good catch! That error has been there for a long time 😬  I suppose `bigint` doesn't get much utilization...

PR #246, opened 2020-12-03
Isolated modules

jeremyBanks:
For the discussion of Deno support at #209, this draft PR demonstrates changes to support TypeScript's `isolateModules` mode.

The meaningful changes are limited to changing some `export`s to `export type`s, and importing Jest's test helpers explicitly instead of relying on global context. Isolated modules mode requires TypeScript to be able to understand each module in isolation, without that context.

However, this required raising the minimum supported version from TypeScript 3.7 to 3.8, which in turn requires updating several other dependencies, so most of this PR is meaningless changes from Prettier.

The compatibility hit might not make sense for Zod 2.0, but perhaps it would make sense some time in the future.


jeremyBanks:
replaced with non-draft PR targeting v3 branch: https://github.com/colinhacks/zod/pull/269

PR #247, opened 2020-12-03
Deno support, given isolated modules

jeremyBanks:
For the discussion in #209, this draft PR brings together isolated modules support (#246), @wilmveel's test refactoring, a simpler Deno transformation script (I'd rather not use the complicated dependency, it turns out our needs can be covered by a small script), and GitHub actions to automatically rebuild the Deno version and test across Deno and Node.


jeremyBanks:
this will be replaced with a PR targeting v3

PR #253, opened 2020-12-05
Add husky & lint-staged

jameschensmith:
Adds [Husky](https://github.com/typicode/husky) & [lint-staged](https://github.com/okonet/lint-staged) as new dev dependencies. This addition will initially start formatting each staged file coming in. Once each file is formatted, though, upcoming PRs should be focused on the content without the unnecessary bloat of formatting unformatted code.

Relates to #225.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,9 @@
   },
   "devDependencies": {
     "@types/jest": "^25.1.4",
+    "husky": "^4.3.4",
     "jest": "^25.1.0",
+    "lint-staged": "^10.5.3",
     "make-coverage-badge": "^1.2.0",
     "nodemon": "^2.0.2",
     "prettier": "^1.19.1",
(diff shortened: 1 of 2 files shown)


o-alexandrov:
Looks great, btw, if you wanted to use Husky v5, you could, since `zod` is an open source project.
I think using v4 is still great, as maybe there are some unreported issues in the fifth version of Husky.

o-alexandrov (approved):

jameschensmith:
@o-alexandrov, I was thinking the same thing! Great minds think alike 😋  I tested both v4 and v5 before making the decision to stick with v4 for now. v5 includes a new `.husky/` directory, which Husky then configures all git hooks to point to that directory using the git configuration `core.hooksPath`. When going back to v4 and being confused for a bit, this was a silent killer before I found it 😅  It also requires you to install `pinst` and set up some additional scripts. I think we should give v5 a bit more time 🙂

PR #261, opened 2020-12-08
Added support backing information

o-alexandrov:
### Related issues

- #225

### Changes

Based on the documentation [described here](https://github.com/nodejs/package-maintenance/blob/main/docs/PACKAGE-SUPPORT.md#support-backing)
Please be informed there are a lot of different options, so it might be worth checking the document yourself @colinhacks as I might have missed something valuable.

P.S. we haven't fully resolved the issue with formatting yet, as you can see I added a new line in the `package.json` changes.
I was thinking about fixing it, but decided to keep it so it's clearer why I create a new PR on the remaining aspects of formatting

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,11 @@
   },
   "homepage": "https://github.com/colinhacks/zod",
   "funding": "https://github.com/sponsors/colinhacks",
+  "support": {
+    "backing": {
+      "npm-funding": true
+    }
+  },
   "keywords": [
     "typescript",
     "schema",


colinhacks:
Where is this documented?

o-alexandrov:
@colinhacks it's [here](https://github.com/nodejs/package-maintenance/blob/main/docs/PACKAGE-SUPPORT.md#integration-into-packagejson). Citing:
> A new field called support is added to package.json. The contents of this field will be one of the following:

We are using object definition:
> object, with repository key matching the schema for the existing repository entry, and optional path field that complies with the format described above under a string

so by now, we end up with:
```
support: {
}
```

Then, let's look at the [backing](https://github.com/nodejs/package-maintenance/blob/main/docs/PACKAGE-SUPPORT.md#support-backing) details:
> Each object consists of an identifier as defined below and an optional URL (except for entries with type npm-funding which must have a value of true) with additional information.

so that's how we ended up with:
```
"support": {
  "backing": {
    "npm-funding": true
  }
}

o-alexandrov:
I believe `npm-funding: true` hooks the previously added `funding` option in package.json with either existing or to-be-existing at least npm ways to notify about funding opportunities.
Unfortunately, I'm not developing libraries like yours, so I don't have experience actually using it.
But as a user of zod, I'm interested in increasing your ability to work on this project, so I wanted to maximize the conversion here of your funding :D

colinhacks:
Excellent! I'd never seen that document 😛

Merging now.

PR #262, opened 2020-12-08
Adds linting & Fixes IDE-related inconsistencies

o-alexandrov:
### Related issues

- #225

### Changes

- Adds linting with ESLint
  - 2 notable aspects
    - we also autofix `import & export` statements order
    - I disabled some rules, as we probably need to discuss first whether we should strictly follow the `recommended` configs of ESLint & typescript-eslint
- Adds .editorconfig to resolve IDE-related inconsistencies
  - [prettier ignores .json](https://github.com/prettier/prettier/issues/322) and I saw some us of overriding each other with a different end of line IDE setting in package.json (similar issues were also in other files)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,46 @@
+module.exports = {
+  env: { browser: true, node: true },
+  root: true,
+  parser: "@typescript-eslint/parser",
+  plugins: [
+    "@typescript-eslint",
+    "import",
+    "simple-import-sort",
+    // 'prettier' commented as we don't want to run prettier through eslint because of performance degradation
+  ],
+  extends: [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "prettier/@typescript-eslint", // Uses eslint-config-prettier to disable ESLint rules from @typescript-eslint/eslint-plugin that would conflict with prettier
+  ],
+  rules: {
+    /**
+     * eslint-plugin-import @see https://github.com/benmosher/eslint-plugin-import
+     */
+    "import/order": 0, // turn off in favor of eslint-plugin-simple-import-sort
+    "import/no-unresolved": 0,
+    "import/no-duplicates": 1,
+    /**
+     * eslint-plugin-simple-import-sort @see https://github.com/lydell/eslint-plugin-simple-import-sort
+     */
+    "sort-imports": 0, // we use eslint-plugin-import instead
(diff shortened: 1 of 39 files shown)


.eslintrc.js
```diff
@@ -0,0 +1,46 @@
...
+    "import/no-duplicates": 1,
+    /**
+     * eslint-plugin-simple-import-sort @see https://github.com/lydell/eslint-plugin-simple-import-sort
+     */
+    "sort-imports": 0, // we use eslint-plugin-import instead
+    "simple-import-sort/imports": 1,
+    "simple-import-sort/exports": 1,
+    /**
+     * @typescript-eslint/eslint-plugin @see https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin
+     */
```
o-alexandrov, line 31: @colinhacks please let me know whether we should follow the recommended rules below that I temporarily disabled

package.json
```diff
@@ -27,11 +27,13 @@
     "inference"
   ],
   "scripts": {
+    "fix:format": "prettier --write \"src/**/*.ts\"",
+    "fix:lint": "eslint --fix --ext .ts ./src",
+    "fix": "yarn fix:lint && yarn fix:format",
```
o-alexandrov, line 32: @colinhacks please let me know, whether you agree with:
- organizing sub scripts with a `:` separator

src/types/string.ts
```diff
@@ -13,6 +13,7 @@ export interface ZodStringDef extends z.ZodTypeDef {
   };
 }
 
+// eslint-disable-next-line
```
o-alexandrov, line 16: @colinhacks I had to disable this line, because there are multiple different eslint complaints about the regex here
I wanted to first confirm whether you approve modifications of this regex based on eslint recommendations

o-alexandrov (commented):
Left some comments to ease the review

package.json
colinhacks: Works for me!

src/types/string.ts
colinhacks: If they don't change the behavior then I'm good with it. I assume the linter wouldn't recommend changes that substantially impact the regex.

.eslintrc.js
colinhacks: Whatever you think is best here, I'm not opinionated about linting.

PR #266, opened 2020-12-10
Resolve rollup build issues

aaronjensen:
This appears to resolve my rollup issue, though it is likely a coincidence. In any case, it's a start towards extracting things and making it more possible to remove circular references, or at least lessen their impact.

This is perhaps a starting point that may unblock some folks and hopefully won't conflict too much with @o-alexandrov's work in #226


o-alexandrov:
@aaronjensen thank you, you partially resolved circular references that I want to resolve completely later this week.
No problems, only gratitude 😄

src/parser.ts
```diff
@@ -2,13 +2,15 @@ import { defaultErrorMap, ZodErrorMap } from "./defaultErrorMap";
 import { INVALID, util } from "./helpers/util";
 import { ZodDef, ZodNever, ZodPromise } from "./index";
 import { NOSET, PseudoPromise } from "./PseudoPromise";
-import * as z from "./types/base";
```
o-alexandrov, line 5: I really like that you also remove a case like this, when everything is imported (it was asking for circular dependencies' issue)

src/switcher.ts
```diff
@@ -1,62 +1,64 @@
 import { util } from "./helpers/util";
-import * as z from "./index";
```
o-alexandrov, line 2: and here too, thank you!

src/types/array.ts
```diff
@@ -3,10 +3,11 @@
 // import { ZodUnion } from './union';
 import { ZodIssueCode } from "../ZodError";
 import * as z from "./base";
```
o-alexandrov, line 5: I'll resolve it in my PR

o-alexandrov (approved):
I am grateful you kindly resolved some aspects that trigger `circular dependencies` issue:
- some `import * as ...` imports from modules that reexport modules which in turn could have been accessed directly
  - I understand it's time consuming to resolve all such usage, I'll try to make it consistent in my PR on other circular dependencies removal.
- and you modularized some shared logic, so it is accessed directly

@colinhacks could you please let us know whether you approve these changes, as they negatively affect the DX when you work on the library.
- it is easier to `import * as z` in the internal api, but this is the main part of the problem with circular references

colinhacks:
I still don't understand how to plan to remove circular dependencies. Subclasses of ZodType (e.g. ZodOptional) are imported into base.ts and are used in method definitions. You can't get around that.

o-alexandrov:
Execution order before and after bundling won’t be guaranteed as mentioned here https://github.com/colinhacks/zod/pull/231#issuecomment-730373217 but would have been relatively consistent (would break from time to time).

But if don’t do resolutions like in this PR, then there would be too many circular dependencies issues that it’s impossible to create a non-breaking order (I have played with it in the mentioned PR above).

The issue with interdependent classes is very rare in this project, I believe there only one or two such instances.
I believe we’d be able to resolve the interdependent classes issue either by:
- modularizing some parts of the classes
- or we could always mask it with an execution order “fix”, which would be a lot easier when there are only a couple of such issues.

o-alexandrov:
@colinhacks btw, yes, I was experiencing that issue https://github.com/colinhacks/zod/issues/215#issuecomment-723549501

colinhacks:
A couple things:

1. I don't think https://github.com/webpack/webpack/issues/11871 is related to *loading order*, it's just about the key ordering of the returned object when you use `import * as`. As I understand it, the file loading order is predictable and consistent regardless of build system (although different between CommonJS and ESM module systems of course).

2. michelwestrate clarified that this only applied to files that are mutually dependent *during initialization*. As an example, ZodOptional is dependent on ZodType during initialization (since it extends ZodType) but ZodType isn't dependent on ZodOptional in the way Michel is describing. It's the same reason this works fine:

```ts
const foo = ()=>{
  bar();
}

const bar = () => {
  foo();
};
```
I'm happy getting rid of the `import * as` statements that are used internally, I'm happy to be more disciplined about only importing what I need. 

I'm not willing to get rid of the methods that currently rely on cyclical imports, like `.optional()` and `.nullable()`. I can't tell if you were asking that or not.

If you have a way to reproduce #215 please share! I don't know how to start fixing the ESM loading order issues because I don't have a way to test it.

o-alexandrov:
1. I meant, in the future, if you use `webpack` to bundle (& minify) `zod` for users; as of now, it will mess up the order.
    - any other bundler might have the same issue at some point or even webpack again, after they resolve that issue
2. I agree

Regarding reproduction, it happened to me with webpack 5.
I resolved all circular dependencies' issues except one that has six instances, please refer to https://github.com/colinhacks/zod/pull/226#issuecomment-743727445

PR #269, opened 2020-12-13
Deno/Isolated Modules support, and GitHub Actions to test and fix 

jeremyBanks:
Adds support for TypeScript's "isolated modules" mode, at the expense of bumping the minimum version for Zod v3 from TypeScript 3.7 to TypeScript 3.8. This changes type-only exports to use the `export type` keyword, and imports Jest's test helpers explicitly, so the TypeScript files can be compiled in isolation, instead of requiring shared context.

Adds a GitHub Actions workflow to automatically regenerate `deno/lib` as well as run our other fixers (eslint and prettier) automatically if a commit is pushed without running them locally.

Adds a GitHub Actions workflow to build and test again our maximum and minimum supported versions of TypeScript on Node, and also test on Deno.

This is still **not** compatible with `deno bundle` due to the bug https://github.com/denoland/deno/issues/8680 (previously discussed around https://github.com/colinhacks/zod/pull/209#issuecomment-743696950). I experimented with changing export names to work around it, but the number of names that needed to be changed was significant, and the bundle still seemed to be invalid due to other errors, so I reverted those changes. For now we're only testing for compatibility with the primary `deno run`/`deno install` workflow (only using `tsc`), not the less-stable `deno bundle`/`deno compile` workflow (which uses both `tsc` and `swc`).


jeremyBanks:
I haven't tested it yet, but Deno appears to have merged a fix for the SWC bundler issue, so it should work in the next release.

colinhacks:
I merged this PR with a bunch of major structural changes from #226. I'm not sure how to verify that everything is still working as expected for Deno users. All the tests pass and everything compiles nicely with `"isolatedModules: true" so I went ahead and pushed the updated version to `zod@next`. @jeremyBanks can you give it a try and look through the auto-generated deno/lib to make sure things are working? 

Testing the SWC bundler bug fix would also be great 👍

jeremyBanks:
@colinhacks Great! I imported the v3 branch into an existing project and was able to use it in a few places without encountering any errors; it looks good to me. 👍

Unfortunately, the Deno/SWC bundler is still broken. The previous error has been fixed, the output is now syntactically valid, but it's renaming exported types in an inconsistent way that creates reference errors. I'll file a new issue with Deno for the new error once I have a chance to reduce it. 🤞

jeremyBanks:
Filed bundler bug at https://github.com/denoland/deno/issues/8978. not sure if my reduced case is the exact same root cause, but it's probably at least a nearby bug.

PR #292, opened 2021-01-14
Performance improvements

reinismu:
image
To
image

1. Only create `ZodError` object when we need it.
That was what I was going for at the beginning
2. Don't extend `Error` from `ZodError`
Even when we create `ZodError` only when things fail it is called a lot if we use unions. So I looked deeper and realized that `Error` creation is expensive. Let me know if you think this is breaking anything


src/ZodError.ts
```diff
@@ -123,15 +123,14 @@ export const quotelessJson = (obj: any) => {
   return json.replace(/"([^"]+)":/g, "$1:");
 };
 
-export class ZodError extends Error {
+export class ZodError {
```
lilred, line 126: `ZodError` has to extend `Error` in order to preserve stack trace information.

src/__tests__/promise.test.ts
```diff
@@ -1,5 +1,6 @@
 import { util } from "../helpers/util";
 import * as z from "../index";
+import { ZodError } from "../ZodError";
```
lilred, line 3: As per above, please revert this file

src/parser.ts
```diff
@@ -109,117 +109,112 @@ export const ZodParser = (schema: ZodType<any>) => (
...
 
   const HANDLE = (err: Error) => {
     if (err instanceof ZodError) {
-      ERROR.addIssues(err.issues);
+      ISSUES.push(...err.issues);
       return INVALID;
     }
-    throw ERROR;
+    throw new ZodError(ISSUES);
   };
```
lilred, line 120: This will lose stack trace information. I would recommend reverting your changes to this file, defining error as `let ERROR;`, and then replacing uses of `ERROR` with a `getErrorObject` function that instantiates `ERROR` when it is first called.

lilred (changes requested):
Very cool initiative. Once my comments are handled I would recommend a merge.

src/ZodError.ts
lilred: ...also so it can be caught by users naively using `instanceof Error`.

src/parser.ts
lilred: On second thought, @colinhacks do you figure the stack trace information is useful here? I'm not convinced.

src/ZodError.ts
reinismu: This is the core performance issue. Zod overuses try/catch in flow. Getting stack information is costly. Removing `Error` is quick fix. I guess the best way would be to change flow so it wouldn't be overused

lilred: >I guess the best way would be to change flow so it wouldn't be overused

I fully agree with you. And in fact it looks like your other changes are going a long way in that direction?

reinismu: Well I tried to remove as many error creations as I could. 
Performance issues in zod are really making me worried. Even our requests where we validate data goes up from 300ms to 1600ms.

lilred: I share your concern. We use Zod in production too and if possible we would like to drive its performance up drastically, at least in the happy path. However, removing the inheritance of `Error` would be a breaking chance that I doubt would be palatable to @colinhacks, never mind the users.

In lieu of this change, there are a few things we could do to ensure that `ZodError` is instantiated as rarely as possible. You've already done some of these things in this PR, which I'm thankful for.

My intuition is that the next step after merging your changes to `src/parser.ts` should be to run benchmarks and look at flame graphs, but this isn't something I have much experience with, nor the time to perform right now. Help is always appreciated. While we don't have benchmarks we can also look at making non-API-breaking optimizations.

lilred: I zoomed out and saw that you had benchmarks above, which is wonderful! Would you be able to run another benchmark with just your changes to `src/parser.ts`?

reinismu: That what I was doing at first. As it does help with types where there are no unions. It does almost nothing where they are. Every union fail creates and throws an error :/ 
That's why my benchmarks only improved when I got rid of `Error`

lilred: Oh wow, I didn't realize that was the problem. indeed, union parsing should collect issues without instantiating `Error`.

Let's give @colinhacks a few days to chime in in case there's a particular approach he'd like to take. I understand he's busy with work at the moment.

scotttrinh:
I imagine we want to port this fix into `v3` as well, right? Definitely hitting this in my current usage and would love to help get this merged!

lilred:
@scotttrinh I agree that this should probably go into v3.

I haven't been able to get a hold of @colinhacks, I understand he's very busy with work. I'm interested in trying to run a friendly fork, though I couldn't begin work on it until at least the week after the next.

reinismu:
In the end zod got just too slow for us. Never thought a library would in this day and age :/ We used it in parts of the frontend and users got very mad that website was super slow.

Switched to https://github.com/badrap/valita

scotttrinh:
Going to update the fork at zod-contrib/zod to latest v3, but was interested to hear from you @colinhacks about your prioritization of performance improvements. The main issue here is allocating an error for every run of `parse` even when it isn't used. And furthermore, subclassing `Error` in `ZodError` also has some non-trivial cost since `Error` comes with the stack (see comment in https://github.com/colinhacks/zod/pull/292#discussion_r567867700)

colinhacks:
@scotttrinh I'm aware of the Error issue. If you want to take a stab at it, that would be great. Definitely write the PR against the v3 branch. If there are other performance improvements you want to incorporate based on this PR's ideas, feel free to do that too.

scotttrinh:
@colinhacks 

Started working on this and want to report back a bit of what I'm seeing:

There are two big changes here:

1. Only allocate an error if there is an issue instead of on every call to parse.
2. Do not "sub-class" `Error` in `ZodError`

It seems like just getting rid of the sub-classing gives us a huge boost in both valid parses and invalid. Any thoughts?

<details>
<summary>
Currently: "object invalid 27,814 ops/sec"
</summary>

```
z.string: empty string x 48,183 ops/sec ±4.64% (82 runs sampled)
z.string: short string x 45,118 ops/sec ±5.73% (75 runs sampled)
z.string: long string x 50,951 ops/sec ±3.25% (81 runs sampled)
z.string: invalid: null x 36,886 ops/sec ±3.92% (84 runs sampled)
z.string: manual parser: long x 700,950,412 ops/sec ±3.64% (84 runs sampled)
z.object: empty: valid x 25,550 ops/sec ±3.01% (79 runs sampled)
z.object: empty: valid: extra keys x 25,125 ops/sec ±2.78% (86 runs sampled)
z.object: empty: invalid: null x 36,037 ops/sec ±3.59% (84 runs sampled)
z.object: short: valid x 13,711 ops/sec ±4.84% (80 runs sampled)
z.object: short: valid: extra keys x 13,239 ops/sec ±4.70% (80 runs sampled)
z.object: short: invalid: null x 34,528 ops/sec ±7.57% (81 runs sampled)
z.object: long: valid x 7,490 ops/sec ±3.97% (63 runs sampled)
z.object: long: valid: extra keys x 7,673 ops/sec ±4.08% (84 runs sampled)
z.object: long: invalid: null x 27,814 ops/sec ±8.56% (64 runs sampled)
```

</details>

<details>
<summary>
Only allocate on failure: "object invalid 27,761 ops/sec"
</summary>

```
z.string: empty string x 560,109 ops/sec ±6.24% (80 runs sampled)
z.string: short string x 573,299 ops/sec ±5.57% (80 runs sampled)
z.string: long string x 608,116 ops/sec ±4.59% (88 runs sampled)
z.string: invalid: null x 34,280 ops/sec ±4.15% (62 runs sampled)
z.string: manual parser: long x 705,877,940 ops/sec ±4.06% (86 runs sampled)
z.object: empty: valid x 45,039 ops/sec ±4.72% (66 runs sampled)
z.object: empty: valid: extra keys x 46,582 ops/sec ±5.01% (77 runs sampled)
z.object: empty: invalid: null x 26,996 ops/sec ±16.14% (63 runs sampled)
z.object: short: valid x 31,774 ops/sec ±5.54% (73 runs sampled)
z.object: short: valid: extra keys x 34,718 ops/sec ±5.30% (76 runs sampled)
z.object: short: invalid: null x 32,920 ops/sec ±8.31% (75 runs sampled)
z.object: long: valid x 17,128 ops/sec ±16.67% (64 runs sampled)
z.object: long: valid: extra keys x 23,249 ops/sec ±5.94% (76 runs sampled)
z.object: long: invalid: null x 27,761 ops/sec ±8.13% (69 runs sampled)
```

</details>

<details>
<summary>
Do not subclass Error: "object invalid 86,946 ops/sec"
</summary>

```
z.string: empty string x 540,451 ops/sec ±3.46% (86 runs sampled)
z.string: short string x 348,678 ops/sec ±17.25% (59 runs sampled)
z.string: long string x 436,669 ops/sec ±15.48% (82 runs sampled)
z.string: invalid: null x 111,825 ops/sec ±3.28% (86 runs sampled)
z.string: manual parser: long x 720,663,528 ops/sec ±3.03% (86 runs sampled)
z.object: empty: valid x 388,305 ops/sec ±4.97% (88 runs sampled)
z.object: empty: valid: extra keys x 306,263 ops/sec ±13.89% (75 runs sampled)
z.object: empty: invalid: null x 98,812 ops/sec ±14.33% (77 runs sampled)
z.object: short: valid x 153,457 ops/sec ±9.91% (79 runs sampled)
z.object: short: valid: extra keys x 149,708 ops/sec ±5.59% (83 runs sampled)
z.object: short: invalid: null x 111,474 ops/sec ±5.10% (81 runs sampled)
z.object: long: valid x 77,332 ops/sec ±3.48% (86 runs sampled)
z.object: long: valid: extra keys x 73,265 ops/sec ±4.34% (86 runs sampled)
z.object: long: invalid: null x 86,946 ops/sec ±12.95% (74 runs sampled)
```

</details>

<details>
<summary>
With both changes: "object invalid 121,532 ops/sec"
</summary>

```
z.string: empty string x 580,162 ops/sec ±4.84% (83 runs sampled)
z.string: short string x 576,156 ops/sec ±5.78% (82 runs sampled)
z.string: long string x 515,700 ops/sec ±9.86% (76 runs sampled)
z.string: invalid: null x 95,510 ops/sec ±6.18% (77 runs sampled)
z.string: manual parser: long x 689,104,213 ops/sec ±3.86% (82 runs sampled)
z.object: empty: valid x 427,090 ops/sec ±6.95% (88 runs sampled)
z.object: empty: valid: extra keys x 374,931 ops/sec ±4.91% (79 runs sampled)
z.object: empty: invalid: null x 114,452 ops/sec ±4.92% (81 runs sampled)
z.object: short: valid x 178,976 ops/sec ±3.61% (82 runs sampled)
z.object: short: valid: extra keys x 169,163 ops/sec ±3.80% (85 runs sampled)
z.object: short: invalid: null x 112,253 ops/sec ±3.63% (87 runs sampled)
z.object: long: valid x 78,291 ops/sec ±4.50% (88 runs sampled)
z.object: long: valid: extra keys x 81,841 ops/sec ±2.74% (90 runs sampled)
z.object: long: invalid: null x 121,532 ops/sec ±3.58% (90 runs sampled)
```

</details>

lilred:
I think both changes are desirable and can be performed together or separately. I wasn't sure about removing the subclassing on Error because that's a breaking change, but since 3.0 isn't out I guess we have that luxury.

scotttrinh:
See updated PR @ https://github.com/colinhacks/zod/pull/459

Issue #172, opened 2020-09-28
Support for Map and Set


lo1tuma:
Currently it is not possible to validate data structures that use `Map`, `WeakMap`, `Set` or `WeakSet`. Is there any chance to support those types, similar to how `z.record` or `z.array` works.

For example:

```js
const myMapSchema = z.map(z.string(), z.number());
const mySetSchema = z.set(z.string());
```

vriad:
PR welcome!

sylvanaar:
you can do something like this with the existing types:

```ts
const test = new Map<string, number>();

test.set("foo", 1);
test.set("bar", 2);

const testType = z.instanceof(Map).refine(map => {
    const keys = z.string();
    const vals = z.number();
    for (const [k,v] of map.entries()) {
        if(!vals.check(v) || !keys.check(k))
            return false;
    }
    return true;
})


testType.parse(test)        
```

vriad:
Support for maps just arrived in Beta 15 courtesy of @lo1tuma ✨

PR #298, opened 2021-01-25
Add Set / z.set() support 

akutruff:
#172


src/parser.ts
```diff
@@ -354,6 +357,46 @@ export const ZodParser = (schema: ZodType<any>) => (
...
+              ERROR.addIssues(err.issues);
+              return INVALID;
+            })
+            .then((item) => {
+              returnedSet.add(item);
+            });
+        })
+      ).then(() => {
+        return returnedSet;
+      });
```
lilred, line 399: This looks fairly expensive, but I see `Map` is done the same way. @lo1tuma was there any particular reason for the copy?

lilred (changes requested):
Sweet PR.

XxOsirisxX:
Hello :) 

When do you think we will be able to use the z.set() function ?

PR #322, opened 2021-02-23
fix implement return type

zhaoyao91:
function.implement should return a function which has the same type declaration with the function shema. but T extends ... is a weaker definition.

for example

```typescript
const MyFnSchema = z.function().args(z.number()).returns(args.number());

const myFn = MyFnSchema.implement(a => a*2 as any)
// Oops, it works, but myFn is with type (a: number) => any, not (a:number) => number
```

### DIFF
--- a/src/types/function.ts
+++ b/src/types/function.ts
@@ -57,7 +57,7 @@ export class ZodFunction<
     });
   };
 
-  implement = <F extends InnerTypeOfFunction<Args, Returns>>(func: F): F => {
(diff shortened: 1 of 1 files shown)


colinhacks:
This makes it possible to omit the call to `.returns()` and instead let TypeScript infer the return type of the function. This is useful if, for instance, you wish to validate the inputs to an API endpoint but let the return type be inferred. I consider this a feature.

PR #336, opened 2021-03-18
Fixed formatting in ERROR_HANDLING table

JuanM04:


colinhacks:
Awesome thanks for this!

I was relying on VSCode's automatic table formatting...guess I need to be more careful.

JuanM04:
The commented rows were causing issues with VSCode. As long as you keep the comments apart, it should be fine

colinhacks:
Ah, really good to know.

PR #348, opened 2021-03-19
Moves eslint-plugin-unused-imports to dev deps

ksmithut:
It seems like perhaps this was just a mistake of omitting the -D because can't seem to find any reference to using the plugin in the main source.


colinhacks:
So it is! Good catch! Fixed in alpha.10.

PR #363, opened 2021-03-25
feat(type): V3 allow pass option array direct to zod enum

xiaoyu-tamu:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -774,14 +774,14 @@ type FishEnum = z.infer<typeof FishEnum>;
 // 'Salmon' | 'Tuna' | 'Trout'
 ```
 
-You must pass the array of values directly into `z.enum()`. This does not work:
+You can also pass the array of values directly into `z.enum()`:
 
 ```ts
-const fish = ["Salmon", "Tuna", "Trout"];
+const fish = ["Salmon", "Tuna", "Trout"] as const;
 const FishEnum = z.enum(fish);
 ```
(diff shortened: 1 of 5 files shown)


colinhacks:
Good work here but I'm not convinced this is a good idea. A big part of the reason it was implemented this way is so users don't accidentally try this:

```ts
const fish = ["Salmon", "Tuna", "Trout"];
const FishEnum = z.enum(fish);
```

With your changes, this code works, but the resulting schema has an inferred type of `string`. It's an easy mistake to make, which is why I want it to be very clear that it's not allowed.

xiaoyu-tamu:
Yes, you're right. I believe this due to typescript’s current limitations.

Other validation lib also have this limitation.

```ts
const options = ['a', 'b'] ;

import * as yup from 'yup';
const a = yup.string().oneOf(options);
type A = yup.InferType<typeof a>;
// type A = string | undefined

import * as struct from 'superstruct';
const b = struct.enums(options);
type B = struct.Infer<typeof b>;
// type B = string
```


```ts
// cast to const

import * as yup from 'yup';
const a = yup.string().oneOf(options);
type A = yup.InferType<typeof a>;
// type A = "a" | "b" | undefined

import * as struct from 'superstruct';
const b = struct.enums(options);
type B = struct.Infer<typeof b>;
// type B = "a" | "b"
```

PR #375, opened 2021-03-31
Improve error message when err.path is empty

JoshuaWise:
Currently, the ZodError message looks buggy when `err.path` is empty. There should be some kind of explicit indication of an error occurring at the root-level value, instead of just showing an empty string.

Before:
```
Issue #1: unrecognized_keys at
```

After:
```
Issue #1: unrecognized_keys at [[root]]
```

### DIFF
--- a/src/ZodError.ts
+++ b/src/ZodError.ts
@@ -144,10 +144,9 @@ export class ZodError extends Error {
       '',
     ];
     for (const err of this.errors) {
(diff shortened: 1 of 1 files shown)


colinhacks:
Seems like a good solution! Note that in Zod 3 I've moved away from formatting this error entirely and just `JSON.stringify` `error.issues`.

PR #377, opened 2021-04-01
Make the refinement message function able to return multiple errors instead of just one

travisvalenti:
I had a use case where I needed to be able to output a set of issues from a refinement, rather than just one, and so I've added this change to make that happen.

I'm not sure but I believe this would also require an update to the documentation, and possibly another update in the Promise version of the refinement, before it would be ready to merge.

They're not required for my use case, but if you're interested in supporting this in the main project I can add those changes too.


colinhacks:
This is actually already possible in Zod 3 with the `superRefine` method. It's what `refine` uses under the hood. You can learn a bit about it here: https://github.com/colinhacks/zod/discussions/346#discussioncomment-504659

Issue #397, opened 2021-04-14
`.transform()` returning output types unioned with input types


djrollins:
Hi there,

I've been excitedly looking to into using Zod as it provides great type-safety compared to yup, but also doesn't require a programming style shift like io-ts, which isn't so palatable for my team.

I have unfortunately hit a bit of snag with type inference of `transform()` which is currently limiting our ability to adopt this, though I do wonder if if I am misunderstanding the intent of this function.

I am trying to create a reusable Schema for for the `Decimal` class from `decimal.js`, which is pretty pervasive throughout our code-base.

I'd like it to accept `number`, `string` and `Decimal` instances as input but ultimately transform them into `Decimals`

Here is my attempt:

```typescript
import * as z from "zod";
import { Decimal } from "decimal.js";

// const decimal = (): z.Schema<Decimal> =>
const decimal = (): z.Schema<Decimal | string | number> =>
  z
    .instanceof(Decimal)
    .or(z.string())
    .or(z.number())
    .refine((value) => {
      try {
        return new Decimal(value);
      } catch (error) {
        return false;
      }
    })
    .transform((value) => new Decimal(value));
```
My expectation here would be that `refine` would trigger an error during parsing if input could not be converted to a Decimal and `transform` would ensure that the output is a `Decimal`.

However if I make the return type of this function with `z.Schema<Decimal>` I get a type error because the inferred type is actually `z.Schema<Decimal | string | number>`.

Am I misunderstanding what `transform` is supposed to be doing here?

Further, it's a little sad that I have to try and construct the `Decimal` instance to validate the input and then have to do it again in the `transform` call. I do understand that `refine` does not actually change the type of the input so this makes sense. Something I'd like to see is something like the following where where `refine` (or some new function?) accepts a type guard that informs the output type. (I have no idea if this is possible though):

```typescript
// const decimalIdeal = (): z.Schema<Decimal> =>
const decimalIdeal = (): z.Schema<Decimal | string | number | undefined> =>
  z
    .instanceof(Decimal)
    .or(z.string())
    .or(z.number())
    .transform((value) => {
      try {undefined
        return new Decimal(value);
      } catch (error) {
        return undefined;
      }
    })
    .refine((value): value is Decimal => value instanceof Decimal);
```
Note that the return type also includes `undefined` in order to type check now. It was this that lead me to believe that `transform` is returning the output type union-ed with the input type.

Many thanks for this library and any help you may be able to provide!

EDIT:

[...]

Issue #392, opened 2021-04-12
Type of a Schema ?


vjau:
Hi,
i would like to write a function that would accept a schema as a parameter.
```ts
const func = (schema:z.Schema<string>)=>{
   const result = schema.safeParse("foo");
}
```
However, the compiler says that `schema` has type   `ZodSchema<string, zod.ZodTypeDef, string> | ((_: unknown) => boolean)`
What is the correct type to use for the function ?

fabien0102:
Works for me with `z.Schema` even with generic:

![image](https://user-images.githubusercontent.com/1761469/115110248-afd91c80-9f7a-11eb-81a6-a77b2e895ea3.png)

(I'm running the latest v3 version)

colinhacks:
Like @fabien0102 said, you should write this function with a generic type. Though we can do even better:

```ts
const func = <T extends z.ZodSchemaAny>(schema: T)=>{
   const result = schema.safeParse("foo");
};
```

`z.ZodSchemaAny` is just an alias for `z.ZodSchema<any,any,any>`. This way, the function can infer the subclass of ZodSchema (e.g. ZodString, ZodNumber, etc), whereas in Fabien's code the inferred type `T` will always be `ZodSchema` (the base class). Subtle but important difference, I should probably document this better.

PR #398, opened 2021-04-15
feat(types): swap order of Input and Def in ZodType

djrollins:
BREAKING CHANGE:

I recently raised an issue (#397) where I learned that I needed to
specify the `Input` type parameter for `Schema`/`ZodType` in order to
prevent it being inferred as a union of the input and ouput types of a
call to `.transform()`.

I missed figuring this out for myself as it required me understanding
and providing the `Def` parameter also as it comes before the `Input`
type in the type argument list.

On inspection it feels like it is important for `Def` to be inferred
correctly as it changes the interface of the schema type and therefore
as a user I probably shouldn't try and specify it. However, it does feel
like specifying the `Input` type is a reasonable thing to want to do.

This patch swaps the order of the `Def` and `Input` types in `ZodType`
so that `Input` can be specified whilst leaving `Def` to be inferred by
the compiler.

This would require specifying the same type twice in cases where you
would want to specify the `Def` type, so this also adds a
`ZodTypeSymmetric<T, Def = ZodTypeDef>` that sets the `Input` and
`Output` type parameters to `T`, to prevent this duplication.

I appreciate that is a a breaking change, but I believe it is a
worthwile quality-of-life improvement that would great to get this in
before 3.0 is properly released.

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -52,8 +52,8 @@ type ZodEffectsType<T extends ZodTypeAny> = T extends ZodEffects<
   : ZodEffects<T, T["_output"]>;
 export abstract class ZodType<
   Output,
-  Def extends ZodTypeDef = ZodTypeDef,
-  Input = Output
+  Input = Output,
+  Def extends ZodTypeDef = ZodTypeDef
 > {
   readonly _type!: Output;
   readonly _output!: Output;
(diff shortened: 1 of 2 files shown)


djrollins:
I've just realized this won't let the `Def` parameter be inferred, it will just be the default which is `ZodTypeDef`. This patch has significantly less value as a result and isn't really worth the breaking change. Closing.

colinhacks:
You're right, this wouldn't change anything about inference. If you are trying to write a generic function that accepts a ZodSchema, look at this example: https://github.com/colinhacks/zod/issues/392#issuecomment-826227242

It allows full inference of the Def and Input. In your case, the Input wasn't being "inferred as a union of the input and ouput types of a call to transform", it was just defaulting to the value of `Output`. You can see this in the class definition

```ts
export abstract class ZodType<
  Output,
  Input = Output // <- defaults
> { 
  // ...
}
```

As you can tell, the `Input` parameter was tacked on in a rather ad hoc way with the introduction of Transformers in Zod 2. It's a little awkward, but (as you said) not really worth changing.

PR #399, opened 2021-04-16
Update README.md

yolio2003:
just remove word 'a'


colinhacks:
Merged! Good catch :)

PR #404, opened 2021-04-16
export interface ZodObjectParams

zbauman3:
I'm using Zod to enforce the shape of other Zod schemas (for consistent api endpoint definitions). 

When I try:
```typescript
export const obj = z.instanceof( z.ZodObject );
```
I get: `Exported variable 'obj' has or is using name 'ZodObjectParams' from external module "/.../node_modules/zod/lib/src/types/object" but cannot be named.` This fixes  that.

### DIFF
--- a/src/types/object.ts
+++ b/src/types/object.ts
@@ -41,7 +41,7 @@ const objectDefToJson = (def: ZodObjectDef<any, any>) => ({
   ),
 });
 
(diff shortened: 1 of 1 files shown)


colinhacks:
I'm pretty sure this is already fixed in Zod 3 as well.

PR #420, opened 2021-05-01
[PoC] Add refineWith

oleg-codaio:
Adding a helper method that can be used to refine a value with a specified schema, in order to take advantage of Zod's built-in refinement functionality. This can be useful for refining a list of strings to a list of enums, for instance, or enforcing bounds on a number.

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -318,6 +318,7 @@ export abstract class ZodType<
   constructor(def: Def) {
     this._def = def;
     this.transform = this.transform.bind(this) as any;
+    this.refineWith = this.refineWith.bind(this) as any;
     this.default = this.default.bind(this);
   }
 
@@ -357,6 +358,30 @@ export abstract class ZodType<
       }) as any;
     }
     return returnType;
+  }
+
+  refineWith<NewOut, This extends this>(
+    returnType: ZodType<NewOut>
+  ): This extends ZodEffects<infer T, any>
+    ? ZodEffects<T, NewOut>
+    : ZodEffects<This, NewOut> {
+    let parsed: ReturnType<typeof returnType.safeParse>;
+    const refined: ZodEffects<any> = this._refinement((val, ctx) => {
+      parsed = returnType.safeParse(val);
+      if (!parsed.success) {
+        for (const issue of parsed.error.issues) {
+          ctx.addIssue(issue);
+        }
+      }
+    });
+
+    return refined.transform(() => {
+      if (!parsed.success) {
+        throw new Error(`Unreachable path.`);
+      }
+
+      return parsed.data;
+    }) as any;
   }
 
   default<T extends Input, This extends this = this>(
--- a/src/types.ts
+++ b/src/types.ts
@@ -318,6 +318,7 @@ export abstract class ZodType<
   constructor(def: Def) {
     this._def = def;
     this.transform = this.transform.bind(this) as any;
+    this.refineWith = this.refineWith.bind(this) as any;
     this.default = this.default.bind(this);
   }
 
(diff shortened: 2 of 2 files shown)


oleg-codaio:
Here's some more background on this: consider that you're using Zod to parse URL query params, which are defined as optional strings, and transform them into some usable type (e.g., enum, boolean, integer, etc.). Specifically, let's consider a param that represents the limit on the number of records that should be returned from a query endpoint. We'd currently need to write that like this:

```ts
z
  .string()
  .transform(v => v ? +v : undefined)
  .refine(v => v !== undefined && !Number.isInteger(v), {message: 'should be an integer'})
  .refine(v => v !== undefined && (v <= 0 || v > 100), {message: 'should be between 0 and 100'})
  .optional()
```

Instead, I'd love to be able to write:

```ts
z
  .string()
  .transform(v => v ? +v : undefined)
  .refineWith(z.number().int().min(1).max(100).optional())
  .optional()
```

This could be used for more complex validations too, such as transforming a comma-separated string of enums into an actual array of enums more elegantly (i.e., using `nativeEnum()`).

oleg-codaio:
See also: https://github.com/colinhacks/zod/discussions/345

oleg-codaio:
As an alternative to this, if `transform` handled Zod errors (instead of just returning some invalid symbol) that would work as well because we could call `.transform(val -> z.number().parse(val))`. Though that's a bit more verbose than the proposed syntax here.

cakoose:
Could also do both, e.g.:
1. Allow `.transform` to handle errors.
2. Add `.transformWith` as a convenience for using a Zod schema with `.transform`.

deno/lib/types.ts
```diff
@@ -359,6 +360,30 @@ export abstract class ZodType<
     return returnType;
   }
 
+  refineWith<NewOut, This extends this>(
+    returnType: ZodType<NewOut>
+  ): This extends ZodEffects<infer T, any>
+    ? ZodEffects<T, NewOut>
+    : ZodEffects<This, NewOut> {
+    let parsed: ReturnType<typeof returnType.safeParse>;
```
oleg-codaio, line 368: Actually I suppose there's a concurrency bug here due to this being in the scope of the schema rather than the specific parsing invocation of it. So `parsed` would have to be stashed somewhere within the context.

cdauth:
I stumbled across this pull request while searching for a solution to the same problem.

Until this is merged, I am using this function:

```typescript
/**
 * Transform the result of a zod scheme to another type and validates that type against another zod scheme.
 * The result of this function basically equals inputSchema.transform((val) => outputSchema.parse(transformer(val))),
 * but errors thrown during the transformation are handled gracefully (since at the moment, zod transformers
 * do not natively support exceptions).
 */
export function transformValidator<Output, Input1, Input2, Input3>(inputSchema: z.ZodType<Input2, any, Input1>, transformer: (input: Input2) => Input3, outputSchema: z.ZodType<Output, any, Input3>): z.ZodEffects<z.ZodEffects<z.ZodType<Input2, any, Input1>, Input2>, Output> {
    // For now we have to parse the schema twice, since transform() is not allowed to throw exceptions.
    // See https://github.com/colinhacks/zod/pull/420
    return inputSchema.superRefine((val, ctx) => {
        try {
            const result = outputSchema.safeParse(transformer(val));
            if (!result.success) {
                for (const issue of result.error.errors) {
                    ctx.addIssue(issue);
                }
            }
        } catch (err) {
            ctx.addIssue({
                code: z.ZodIssueCode.custom,
                message: err.message,
            });
        }
    }).transform((val) => outputSchema.parse(transformer(val)));
}
```

I'm using it for example to transform a JSON string to an object and then validating that object against a schema:

```typescript
const jsonValidator = transformValidator(z.string().optional(), (val) => val ? JSON.parse(val) : undefined, z.object({
    ...
});
```

Issue #355, opened 2021-03-23
Types not correctly inferred when using transformers (v3)


larsbs:
I'm trying to achieve one thing, basically, coalescing all `null` values into `undefined` when receiving data from an external provider. Since it's really annoying to write all the code necessary in `zod` to understand that again and again:

```ts
const myString = z.string().optional().nullable().transform((v) => v ?? undefined)
```

I was trying to write a wapper to do that automatically. It looks like the following:

```ts
function _optional<T>(zod: z.ZodType<T>): z.ZodTransformer<z.ZodNullable<z.ZodOptional<z.ZodType<T>>>, T | undefined> {
  return zod.optional().nullable().transform((v) => v ?? undefined);
}
```

However, this works well for simple types (for example, the one above):

```ts
const myString = _optional(ts.string());

// this gets correctly inferred as `string | undefined` instead of `string | null | undefined`
type t = z.infer<typeof myString>;
```

But it doesn't work for complex types like objects:

```ts
const UserSchema = _optional(z.object({
  email: _optional(z.string()),
  name: _optional(z.string()),
}));

type User = z.infer<typeof UserSchema>;
```

In the example above, the type `User` gets incorrectly inferred as: 

```ts
// notice the extra null typings
type User  = {
  email: string | null | undefined;
  name: string | null | undefined;
} | undefined
```

instead of 

```ts
type User  = {
  email: string | undefined;
  name: string | undefined;
} | undefined
```

like in the simple example at the beginning.

### Solutions

I could see two solutions, first one, see why this errors is happening (I don't know enough about zod internals as to answer this question myself just yet) or the second one, try to provide a simpler/nicer API for when you need some sort of default. For example:

```ts
const myString = z.string('default value');

const UserSchema = z.object({
  email: z.string(),
  name: z.string(),
}, { email: 'default email', name: 'default email' });
```

Basically meaning that every `zod` validator would accept an extra parameter that would be the default value in case the validated value is `null` or `undefined` (using `??`) and if that parameter is present, mark the validator automatically as `.optional().nullable()`

colinhacks:
You were really close to the solution here. Writing generic functions on top of Zod is hard - I've been meaning to write up a guide describing the best practices.

Here's a correct implementation of `_optional`:

```ts
function _optional<T extends z.ZodTypeAny>(zod: T): z.ZodEffects<z.ZodNullableType<z.ZodOptionalType<T>>, T['_output'] | undefined> {
[...]

Issue #391, opened 2021-04-12
Feature request: ".maybe()" type


ryami333:
Hi there, thanks for the excellent library, I'm using it on a variety of projects now! A problem I have come across recently is that while building schema for third-party REST APIs (ie. APIs which I have no control/influence over) I have to type non-mandatory properties like this:

```ts
/*
 * Generated from:
 * https://developer.zendesk.com/rest_api/docs/support/attachments#json-format
 */
export const attachmentResourceSchema = zod.object({
  // The content type of the image. Example value: "image/png"
  content_type: zod.string().nullable().optional(),

  // A full URL where the attachment image file can be downloaded
  content_url: zod.string().nullable().optional(),

  // If true, the attachment has been deleted
  deleted: zod.boolean().nullable().optional(),

  // The name of the image file
  file_name: zod.string().nullable().optional(),

  // Automatically assigned when created
  id: zod.number().nullable().optional(),

  // If true, the attachment is excluded from the attachment list and the attachment's URL can be referenced within the comment of a ticket. Default is false
  inline: zod.boolean().nullable().optional(),

  // The URL the attachment image file has been mapped to
  mapped_content_url: zod.string().nullable().optional(),

  // The size of the image file in bytes
  size: zod.number().nullable().optional(),

  // An array of attachment objects. Note that photo thumbnails do not have thumbnails
  thumbnails: zod.array(zod.unknown()).nullable().optional(),

  // A URL to access the attachment details
  url: zod.string().nullable().optional(),
})
```

The use of both `.nullable()` _and `.optional()` is apparently necessary (I've cross-checked against many payloads). Sometimes the payload contains those nullified properties, and other times it omits them entirely (ie. they are undefined). Could there be a `.maybe()` method which consolidates these two validations? I would be open to alternative naming.

Thanks!

ryami333:
Oh and before someone suggests using `.partial()` - not _all_ of the properties in my use-case actually maybe-typed. In the example I posted they happen to be, but there are quite a few schema in this system with some mandatory properties.

colinhacks:
Not a fan of this for reasons I describe briefly in #403 

In your particular case, you could implement a maybe function to avoid some verbosity:

```ts
const maybe = <T extends z.ZodTypeAny>(schema: T)=>{
  return schema.optional().nullable();
};

// equivalent
maybe(z.string());
z.string().optional().nullable();
```

PR #421, opened 2021-05-02
Standardize behavior of optionals + defaults

colinhacks:
#### 1. Default values are now implemented in separate `ZodDefault` class. 

Previously this logic was folded into ZodOptional.

#### 2. Standardized behavior of `.optional()` and `.nullable()` methods.

Previously calling `.optional()` on an instance of ZodOptional instance would return the current instance instead of redundantly nesting ZodOptional instances. This seems entirely unnecessary, feels too magical, and introduced complexity into the typings. This complexity makes it harder to write generic functions on top of Zod: https://github.com/colinhacks/zod/issues/355#issuecomment-805130057

#### 3. No `undefined` defaults

```ts
// before this was OK
z.string().optional().default(undefined)

// now it's not
z.string().optional().default('asdf'); // must be string
```

There's no point in setting `undefined` value as a  default value, it's a NOOP. This also lets us explicitly eliminate optionals from the inferred output types of ZodDefault:

```ts
export class ZodDefault<T extends ZodTypeAny> extends ZodType<
  util.noUndefined<T["_output"]>,
  ZodDefaultDef<T>,
  T["_input"] | undefined
> {
  // ...
}
```

#### 4. Calls to `.optional` always return schema with `T | undefined`

This is a consequence of #2.  Previously calling `.optional()` on `z.string().default()` would return the instance unchanged, since defaults values were implemented in the ZodOptional class. Now it introduces a new ZodOptional instance that wraps a ZodDefault instance. As such the type signature is now `string | undefined` instead of `string`.

This introduces a nice duality into chained calls to `default` and `optional`.

```ts
z.string().default('asdf'); // string
z.string().default('asdf').optional(); // string | undefined
z.string().default('asdf').optional().default('asdf'); // string
```

#### 5. Introduce .nullish()

```ts
// equivalent
z.string().nullish()
z.string().optional().nullable()
```


KATT:
> 1. Default values are now implemented in separate ZodDefault class.

No opinion

> 2. Standardized behavior of .optional() and .nullable() methods.

No opinion

> 3. No undefined defaults

Seems good as long as you don't **have to** pass a `default()` to `optional()` (but seems like you don't considering 4.)

Are you doing a `?? defaultValue` in the code, i.e `z.string().optional().nullable().default('x').parse(null)` return `x`? I think it should if not.

Can this also `.default()` used on complex objects too? 

> 4. Calls to .optional always return schema with T | undefined

Nice.


> 5. Introduce .nullish()

You know I love this (#391)


---

Overall, great changes 👏

alii (approved):
Looks fantastic, can't wait to use `.nullish`! :rocket:

colinhacks:
Glad you like it :) I was planning to just implement `.nullish` (basically a one-liner) then I got sucked down this rabbit hole :) Feeling good about it!

Currently defaults are only applied when the schema receives an `undefined` value. I'd be opening to adding a new subclass (NullishDefault?) or add flag inside ZodDefault to support `??`-like behavior. 

Regarding the base-class convenience method, this could be introduced as a new method (`nullishDefault()`?) or as a param on `default` (`.default('asdf', {nullish:true})`). Probably gonna do the new method though because I don't like params in general.

ejose19:
@colinhacks If a new method will be added (`nullishDefault`), then it will be great if a null-only default is provided as well, just to be consistent, and also since `T.nullable().default()` output type is `T | null` instead of just `T` (related #425)

iambryanhaney:
@colinhacks +1 for a `nullishDefault`.

I'm running into some impedance mismatches between Zod and React Hook Forms with regards to undefined / null handling.

RHF discourages the use of `undefined` for field states and reserves its use internally - explicitly setting a field to undefined will reset its state back to the form default.

Setting the field as `null` is the 'proper' way, but that means changing the schema from `optional()` to `nullish()`, effectively obviating `.default()` and causing consumers of the form to receive `number | null` type.

PR #434, opened 2021-05-13
Fix "Symbol is undefined" error

NumminorihSF:
We faced with issue, that in old browsers Symbol is not supported.
As I see, in the code, Symbol is used as unique value (not as unique key in an object).
So I added the check if Symbol exists and use an Object if not.


colinhacks:
I switched away from Symbol entirely, the `typeof Symbol === "function"` check just adds unnecessary complexity. Good PR, thanks!!

PR #438, opened 2021-05-15
feat(v3): string cuid validation

alii:
This PR adds cuid validation to a string. This is my first time contributing to Zod I'm sure I've probably left something out 😅


alii:
Any update on this 😀

lilred:
Maybe there's a discussion I missed so take this with a grain of salt:

I don't think the goal of Zod is to provide an exhaustive collection of validators. Rather, it's meant to be easy to extend with your own validators.

I'm not familiar with CUID. The [NPM package of the same name](https://www.npmjs.com/package/cuid) has under a million weekly downloads, so it's not exactly pervasive. *Based on what little I know,* I wouldn't recommend it for inclusion in the core Zod library.

colinhacks:
Despite appearances, this was merged in `zod@3.8`. Thanks @alii!

alii:
yay!

alii:
thanks colin for epic library
image

PR #455, opened 2021-05-19
make addQuestionMarks a homomorphic mapped type

### LINKED ISSUE #423: Add docstrings to inferred type
Given some code like:

```
import { z } from "zod";

const UserSchema = z.object({
  /**
   * The "ame" of the "usern" _not_ the "name" of the "user"!
   **/
  username: z.string(),
});

interface UserInterface {
  /**
   * The "ame" of the "usern" _not_ the "name" of the "user"!
   **/
  username: string;
}

type User = z.infer<typeof UserSchema>;

const foo: User = {
  username: "",
};

const bar: UserInterface = {
  username: "",
};
```

if I hover over `username: ""` of foo I don't see the docstring like I would when hovering over the username of bar

This is due to the fact that some of the mapped types are not quite homomorphic types (especially around addquestionmark) so typescript doesn't copy the metadata along to the new type. I think I might be able to change them into homomorphic types by using typescript 4.1's `[key in keyof T as ...]` syntax. 

Would you consider such a PR for merging? Are there any gotcha's that I should be aware of?

jauco:
This means that docstrings on the schema definition will transfer to the z.Infer generated type.

Fixes #423 

I can't think of a way to embed this into a test though. If anyone has any idea how to do so I'm all ears.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1153,16 +1153,14 @@ export namespace objectUtil {
   } &
     V;
 
-  type optionalKeys<T extends object> = {
-    [k in keyof T]: undefined extends T[k] ? k : never;
-  }[keyof T];
-
-  type requiredKeys<T extends object> = Exclude<keyof T, optionalKeys<T>>;
+  type CanBeUndefined<K extends keyof T, T> = undefined extends T[K] ? K : never;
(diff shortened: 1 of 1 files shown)


colinhacks:
Merged in 3.0.1. Thanks!

Issue #467, opened 2021-05-25
Complex type coercion


mildfuzz:
We have an issue where we are using Zod to apply strong typing practices to our API layer

Trouble we have is using complex object over GET, as everything becomes a string over the wire. 

Are we able to attempt deep object coercion with Zod? Something like:

```typescript
const schema = z.object({example: z.number()})

const data = {example: '100'}

const coersedData = scheme.coerce(data) // {example: 100}

const brokenData = {example: 'lemon'}

scheme.coerce(brokenData) // Error
```

scotttrinh:
You can use `refine` and `transform` for this. In your example, I think something like this should work:

```ts
import { z } from "zod";

const schema = z.object({
  example: z.string()
    .refine((maybeNumber: string) => /^\d+\.{0,1}\d*$/.test(maybeNumber)) // or whatever your domain wants here
    .transform((maybeNumber: string) => Number(maybeNumber)), // or parseInt or whatever
});

const data = { example: '100' };
const parsed = schema.parse(data);

const brokenData = { example: 'lemon' };
const safe = schema.safeParse(brokenData);
```

[TypeScript Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgLzgXzgMyhEcBEyEAJvgNwBQFAxhAHYDO8D1AFgKYgCGcAvCgDoIAIwBW7ajAAUCCnDjsAHl3AAbdgC5BTKMDoBzKQEo58uAKjtMe9lKncAnsPYA5AK4hnULTr36jfAB8cAD0AHoAOsQA1BECCAAMADQAjGhRAFQAJCECMOxM9lxOrh5eRgEhIXDQcADurFz5AG7sUHAOEG7txLhcevVcdDAMcByWpvJ5UEMMmNAgdo7O7p5tPjC6BgG8wateRSX7bRVJodW1YFxQDOwAksM17Q1N7K1QFGhGlDT0THDEJo8fhIJQqMDqLQAchSCQSUPQlFojHgVxu7GIfDgLA43AEaNuUkBMC43yoyP+whwAGt2HQACJArGg5RqTRwKHqED0BFoJF-ZhcTDsLE4zhcAQMIXsAAK10JVIgtIZQLJQA)

scotttrinh:
[...]

Issue #463, opened 2021-05-23
feature: prepare


bitlab-code:
What do you think about adding a new `.prepare()` method
- problem:
```typescript
const emailSchema = z.string().email()
emailSchema.parse(" aaa@bbb.ccc "); // throws Error
```
but - if trimmed - it's a valid email, I can accept it.

- solution:
 ```typescript
const emailSchema = z.string().email().prepare(v => v.trim())
// and more concisely
const trimmed = v => v.trim()
z.string(trimmed).email()
```

You can add a parameter (without breaking change) and/or a new method (i.e. prepare), both accept a value that is of type `T` just like transform

this is only a toy-example there're a lot of real use case

[edit]
this solution is inspired to .transform() but act before the validation.

bitlab-code:
@colinhacks I've took a look at code, I can PR if you are interested in this

colinhacks:
I was planning to implement this as a method called `preprocess`. Feel free to take a stab at this PR!

bitlab-code:
Hi @colinhacks,  opened a draft PR #468 .
but there is and issue with `optional`, `nullable` and `default` (don't pass the test) 😕

PR #468, opened 2021-05-25
Preprocess method

bitlab-code:
WIP: some tests fail.
There is and issue with `optional`, `nullable` and `default`.
Need help.

### DIFF
--- a/src/__tests__/preprocess.test.ts
+++ b/src/__tests__/preprocess.test.ts
@@ -1,0 +1,26 @@
+// @ts-ignore TS6133
+import { expect, test } from "@jest/globals";
+
+import { z } from "..";
+
+test("preprocess test suite", () => {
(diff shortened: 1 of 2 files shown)


lilred:
Can you describe the intended difference with `.transform`?

scotttrinh:
@lilred 

> Can you describe the intended difference with .transform?

I believe the idea here is that this runs _before_ hitting the input type parser, so you can change the input data. Discussed in [here](https://github.com/colinhacks/zod/issues/463#issuecomment-847469624) and [here](https://github.com/colinhacks/zod/issues/467)

lilred:
I removed the "draft" status on this PR expecting the CI to run but that didn't happen. I'll run the tests locally as soon as I have some time.

colinhacks:
A slightly different implementation of this feature was added in zod@3.8. Thanks @bitlab-code !

Issue #100, opened 2020-07-26
RFC: Transformations (e.g. casting/coercion)


colinhacks:
⚠️ Update: this approach has been abandoned in Zod 3 for a more Yup-like "transform chain" approach.

This is the proposed API for implementing data transformation functionality in Zod.

## Proposed approach

No need to over complicate things. The problem to solve is how Zod should handle/support transformations from type A to type B. So I propose the creation of a new class `ZodTransformer`. This class is a subtype of `ZodType` (so you can use it like any other schema — `.refine`, `.parse`, etc).

Internally, instances of this class will have these properties
- `input: ZodType`: an input schema
- `output: ZodType`: an output schema
- `transformer: (arg: T['_type']) => U['_type']`: a transformer function

There is only one transformer function, not a "transform chain" like in Yup (👋 @jquense). This makes it easier for Zod to statically type the function. Any sort of functional composition/piping can be done using libraries external to Zod.

You would create an instance using the `ZodTransformer.create` static factory method (aliased to `z.transformer`):

## Usage

Coercing a string into a number.

```ts
const stringToNumber = z.transform(z.string(), z.number(), (data)=>parseFloat(data));
type stringToNumber = z.infer<typeof stringToNumber>; // number

stringToNumber.parse("12") // => 12 (number)
```

## `.input/.output`

```ts
stringToNumber.input; // ZodString
stringToNumber.output; // ZodNumber
```

## `.transform` method

~~Every ZodTransform instance will have a `.transform` method. This method lets you easily chain transforms, instead of requiring many nested calls to `z.transform()`.~~

Every _ZodType_ instance (the base class for all Zod schemas) will have a `.transform` method. This method lets you easily create a ZodTransform, using your current schema as the input:

```ts
const trimAndMultiply = z.string()
  .transform(z.string(), x =>x.trim())
  .transform(z.number(), x => parseFloat(x))
  .transform(z.number(), num => num * 5);

console.log(trimAndMultiply.parse(' 5 ')); // => 25
```
## ~~`.toTransformer` function~~

⚠️ Edit: This section is no longer relevant since the `.transform` method has been moved to the base ZodType class instead of only existing on ZodTransform.

As you can see above, the first method call is `.transformer` (which is a factory function that returns a ZodTransform). All subsequent calls are to `.transform()` (a chainable method on the ZodTransformer class).

To make the syntax for defining chains of transforms more consistent, I propose a `toTransformer` function:

```ts
[...]

PR #470, opened 2021-05-26
Fix method name typo in the old syntax example

illarionvk:


colinhacks:
That's intentional, the original syntax for transforms was indeed `z.transformer(A, B, func)`: https://github.com/colinhacks/zod/issues/100

Thanks for keeping a sharp eye though 🤙

PR #483, opened 2021-06-02
Fix not considering the nil UUID a valid UUID

tstehr:
The nil UUID (`00000000-0000-0000-0000-000000000000`) is a valid UUID according to [RFC4122 (Section 4.1.7)](https://datatracker.ietf.org/doc/html/rfc4122#section-4.1.7). This PR adds it to the regex for UUID validation. It is added separately to avoid allowing `0` in locations where it is not permitted for other UUID versions.


colinhacks:
Great, thanks!

Issue #487, opened 2021-06-08
Add a custom context field to ParseParams


JonasZ95:
I would like to add an extra field called something like userContext to ParseParams which is passed to all parse and later refine calls. The idea behind this is allowing to set something like a API client as context which could be used with refine. Small example:

```
interface UserService {
  checkUniqueEmail(mail: string): Promise<boolean>;
}

const uniqueEmail = z.string().superRefine(async (mail, ctx) => {
  const userService = ctx.userContext as UserService;
  return await userService.checkUniqueEmail(mail);
});
```

colinhacks:
I'd strongly recommend against this pattern. Implementing schemas that depend on certain params being passed into `.parse` breaks encapsulation. Schemas should contain all the information/configuration they need to do their job. This is especially bad since there's no way for Zod/TypeScript to enforce this. Consider using factories & closures to achieve this:

```ts
const schemaForService = (userService: UserService)=>{
  return z.string().superRefine(async (mail) => {
    return await userService.checkUniqueEmail(mail);
  });
}
```

JonasZ95:
While I agree with you about how parse should work, It's also feeling wrong to recreate the whole schema. This is very relevant in React libs/apps which would either recreate the whole schema for each render call component is created or If you use 'useMemo'  each time an instance of the component is created. Since we also auto-generate the Zod schemas from our swagger specification, adding an extra field seemed to be the easiest way to do this. I tried to play around with a global cache, but It resulted in alot of boilerplate quote and the typescript compiler also had some problems properly deducing the types.

PR #488, opened 2021-06-08
Add UserContext field to ParseContext

JonasZ95:
As proposed in #487

### DIFF
--- a/deno/lib/helpers/parseUtil.ts
+++ b/deno/lib/helpers/parseUtil.ts
@@ -103,6 +103,7 @@ export const issueHelpers = (error: ZodError, params: ParseParams) => {
 
 export type ParseParams = {
   data: any;
+  userContext: any;
   path: (string | number)[];
   errorMap: ZodErrorMap;
   parentError: ZodError;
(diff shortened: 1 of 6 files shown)


deno/lib/helpers/parseUtil.ts
```diff
@@ -103,6 +103,7 @@ export const issueHelpers = (error: ZodError, params: ParseParams) => {
 
 export type ParseParams = {
   data: any;
+  userContext: any;
```
adriangroch, line 106: I've only stumbled across this PR and Issue coming from looking into an alternative to Yup as I'm not too impressed with the loose typing, but would it be worth using a generic type here over `any`?
```ts
export type ParseParams<T extends {}> = {
  userContext: T
 // ...
```
And also in `RefinementCtx`/`superRefine`?
That way you could do something along the lines of
```ts
.superRefine<UserContext>((arg, ctx) => {
  ctx.userContext.inc();
  return arg;
});
```
Which would relieve the need to cast `ctx.userContext as UserContext`

Issue #452, opened 2021-05-18
Missing index signature when using passthrough() and infer()


lo1tuma:
I’ve tried to update my projects from using zod v1 to zod v3. Unfortunately it seems like `passthrough()` does’t work in the same way as `nonstrict()` did before.

Example:

```ts
const foo = z.object({ foo: z.string() }).passthrough();
type Foo = z.infer<typeof foo>;
```

The inferred type `Foo` is now `{ foo: string; }`, but since I want to allow unrecognized keys I would expect something like `{ foo: string; [key: string]: unknown; }`.

Example with zod v1:

```
const foo = z.object({ foo: z.string() }).nonstrict();
type Foo = z.infer<typeof foo>;
```

The inferred type `Foo` is here `{ [x: string]: any, foo: string }`.

lo1tuma:
I’ve just learned about `catchall()` in zod v3, which adds an index signature to the inferred type. I guess this would work for my case.
The question is, what would be the expected behavior for `passthrough()` here?

colinhacks:
Fixed in 3.3.2. 

BTW Mathias, I just merged the  big performance enhancement PR in 3.3.2 so Zod's 200x faster now 😛

lo1tuma:
Awesome @colinhacks. Thank you!

gao-sun:
we're good with `.catchall(z.unknown())` (the correct type can be inferred):

```ts
const foo = z
  .object({
    id: z.string().optional(),
  })
  .catchall(z.unknown());

type Foo = z.infer<typeof foo>; // { [x: string]: unknown; id?: string; }
```

should `.passthrough()` has the same index signature? since their behavior looks the same. currently `.passthrough()` will not add `[x: string]: unknown;` to the inferred type.

PR #494, opened 2021-06-14
Stop inferring `unknown` fields as optional

lilred:
Addresses https://github.com/pkvince/medzy-monorepo/pull/295

This is a breaking change so let's wait until @colinhacks checks in

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### vNext
+
+- Object fields of type `unknown` are no longer inferred as optional.
 ### 3.2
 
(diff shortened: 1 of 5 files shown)


colinhacks:
Merged in 3.3!

pepakriz:
@colinhacks This improvement definitively breaks compatibility with `strictNullChecks=false`. It is not big deal, but just for you to know...

image

colinhacks:
Ah good catch. That's a bug that was introduced due to an unrelated fix: https://github.com/colinhacks/zod/issues/452

Fixed in 3.3.3.

pepakriz:
@colinhacks Tried with 3.3.4 and still, the same bug occurred when TS `strictNullChecks=false` is set. But as I said, no big deal

PR #523, opened 2021-07-06
Add type discriminator to zod type

rubber-duck-software:
Note: Unsure of the handling around ZodNonEmptyArray.


colinhacks:
Great PR, merged in 3.5.0. I added a `typeName` to `ZodEffects` and added a exhaustiveness unit test as well.

PR #524, opened 2021-07-07
Docs improvement

mattvagni:
When updating to version 3.3.4 typechecking failed for us since the signature of the superRefine function changed from returning `void` to returning `boolean | Promise<boolean>`.

This updates the code example to match the types + adds a note explaining the return type (like `.refine()`).

I'm not entirely sure if this was intentional given the documentation states that:

> If `ctx.addIssue` is NOT called during the execution of the function, validation passes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1362,6 +1362,7 @@ const Strings = z.array(z.string()).superRefine((val, ctx) => {
       inclusive: true,
       message: "Too many items 😡",
     });
+    return false;
   }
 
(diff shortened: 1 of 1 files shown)


colinhacks:
This error has since been rectified. `superRefine` briefly had an incorrect type signature, the result of a major rewrite of the Zod parsing engine. If you upgrade Zod this should resolve itself; there is no need to return a value from `superRefine`.

PR #526, opened 2021-07-08
Fix z.void() to reject null

andersk:
`z.void()` previously accepted `null` or `undefined`, but the TypeScript `void` type is only assignable from `undefined` (unless `strictNullChecks` is `false`, in which case every type is assignable from `null` regardless).

This unsoundness could be observed as follows:

```js
const x = z.string().or(z.void()).parse(null);
if (x !== undefined) {
  // TypeScript infers x is a string, but it’s actually null
  x.trim(); // oops
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 4.0
+
+- `.void()` now only accepts undefined, not null.
+
 ### 3.5
 
 - Add discriminator to all first-party schema defs
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ z.date();
 // empty types
 z.undefined();
 z.null();
-z.void(); // accepts null or undefined
+z.void(); // accepts undefined
 
 // catch-all types
 // allows any value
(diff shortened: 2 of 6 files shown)


lilred:
I wonder if the better solution wouldn't be to fix `z.void()` to in fact reject `null`. This would be a breaking change however. @colinhacks yay or nay?

colinhacks:
I'd prefer to keep the type signature as `void` (since getting `void` as the return type of `z.function` is the only reason this exists). I'm in favor of @lilred's solution. I think Zod's runtime behavior should conform to TypeScript's behavior in strict mode (as it does in many other cases).

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### 4.0
+
+- `.void()` now only accepts undefined, not null.
```
andersk, line 5: (Is this considered breaking enough to warrant a bump to 4.0?)

andersk (commented):
Sure, updated.

CHANGELOG.md
colinhacks: Nope

PR #545, opened 2021-07-19
Fix getter for maxLength

andparsons:
Fix getter on ZodString for maxLength, changing the def.checks from kind min to max


colinhacks:
Doh 🤦‍♂️

PR #478, opened 2021-05-31
fix setPrototypeOf in v3

mad-gooze:
Hi @colinhacks!

Please, review this PR - I've applied fix from https://github.com/colinhacks/zod/commit/7142f7d33a5071a589df798461f3e11b6311fd16

PR #547, opened 2021-07-20
Support legacy browsers

mad-gooze:
This PR adds ponyfills for `Object.keys`, `Object.setPrototypeOf` (duplicate of https://github.com/colinhacks/zod/pull/141 and https://github.com/colinhacks/zod/pull/478) and other functions which fixes Zod in legacy browsers like ie11.


deno/lib/helpers/parseUtil.ts
```diff
@@ -42,7 +42,7 @@ export const getParsedType = (data: any): ZodParsedType => {
     case "string":
       return ZodParsedType.string;
     case "number":
-      return Number.isNaN(data) ? ZodParsedType.nan : ZodParsedType.number;
+      return isNaN(data) ? ZodParsedType.nan : ZodParsedType.number;
```
mad-gooze, line 45: Number.isNaN is equal to isNaN because of typeof check.

deno/lib/helpers/util.ts
```diff
@@ -25,26 +25,35 @@ export namespace util {
...
     );
     const filtered: any = {};
     for (const k of validKeys) {
       filtered[k] = obj[k];
     }
-    return getValues(filtered);
+    return objectValues(filtered);
   };
 
-  export const getValues = (obj: any) => {
```
mad-gooze, line 38: getValues is a duplicate of objectValues.

src/helpers/parseUtil.ts
```diff
@@ -42,7 +42,7 @@ export const getParsedType = (data: any): ZodParsedType => {
     case "string":
       return ZodParsedType.string;
     case "number":
-      return Number.isNaN(data) ? ZodParsedType.nan : ZodParsedType.number;
+      return isNaN(data) ? ZodParsedType.nan : ZodParsedType.number;
```
mad-gooze, line 45: Number.isNaN is equal to isNaN because of typeof check.

src/helpers/util.ts
```diff
@@ -25,26 +25,35 @@ export namespace util {
...
     );
     const filtered: any = {};
     for (const k of validKeys) {
       filtered[k] = obj[k];
     }
-    return getValues(filtered);
+    return objectValues(filtered);
   };
 
-  export const getValues = (obj: any) => {
```
mad-gooze, line 38: `getValues` is a duplicate of `objectValues`, so I removed it.

src/types.ts
```diff
@@ -1867,8 +1867,10 @@ function mergeValues(
   if (a === b) {
     return { valid: true, data: a };
   } else if (aType === ZodParsedType.object && bType === ZodParsedType.object) {
-    const bKeysSet = new Set(Object.keys(b));
-    const sharedKeys = Object.keys(a).filter((key) => bKeysSet.has(key));
+    const bKeys = util.objectKeys(b);
```
mad-gooze, line 1870: Do not use Set for intersection as it is not supported in legacy mobile browsers (like Android 4.4.4).

deno/lib/helpers/parseUtil.ts
mad-gooze, line 45: `Number.isNaN` is equal to `isNaN` because of typeof check.

deno/lib/__tests__/object.test.ts
```diff
@@ -164,6 +164,7 @@ test("test that optional keys are unset", async () => {
     id: "asdf",
     set: undefined,
   });
+  // eslint-disable-next-line ban/ban
   expect(Object.keys(result)).toEqual(["id", "set"]);
```
lilred, line 168: Do ponyfills not work in Deno?

.eslintrc.js
```diff
@@ -44,5 +45,13 @@ module.exports = {
     "no-empty": "off",
     "no-useless-escape": "off",
     "no-control-regex": "off",
+
+    "ban/ban": [
+      2,
+      { "name": ["Object", "keys"], "message": "Object.keys() is not supported in legacy browsers, use objectKeys()" },
+      { "name": ["Object", "setPrototypeOf"], "message": "Object.setPrototypeOf() is not supported in legacy browsers" },
+      { "name": ["Number", "isNaN"], "message": "Number.isNaN() is not supported in legacy browsers" },
+      { "name": ["Number", "isInteger"], "message": "Number.isInteger() is not supported in legacy browsers" },
```
lilred, line 54: Would be nice to recommend alternatives for each of them

deno/lib/helpers/util.ts
```diff
@@ -25,28 +25,35 @@ export namespace util {
...
+      ? (obj: any) => Object.keys(obj) // eslint-disable-line ban/ban
+      : (object: any) => {
+          const keys = [];
+          for (const key in object) {
+            if (Object.prototype.hasOwnProperty.call(object, key)) {
+              keys.push(key);
+            }
+          }
+          return keys;
+        };
```
lilred, line 55: I'd be more comfortable with e.g. es-shims' implementation, unless I misunderstand how Deno works.

lilred (commented):
I really like the idea but I'm sketched out by the amount of `// eslint-disable-next-line` involved. I think more work needs to happens to actually remove these from the code base before this can be merged.

deno/lib/__tests__/object.test.ts
mad-gooze: they do - it's a test code, no need to use ponyfills here

deno/lib/helpers/util.ts
mad-gooze: what do you mean? 
I have no experience with deno, but looks like ponyfill is not needed there because Object.keys should be defined

lilred: If we know that a ponyfill is not needed then why are we checking if `Object.keys` is defined? (I have no experience with Deno either.)

lilred: @wilmveel or @jeremyBanks do you have thoughts on shimming in Deno? Is it necessary or not?

mad-gooze: > E: wait, is this generated code from the "normal" version? In that case disregard this comment.

yep

scotttrinh:
Instead of banning these explicit unsupported features, could we leverage https://github.com/amilajack/eslint-plugin-compat to make it explicit what we plan on supporting? There are plenty of other things that would break IE11 compatibility and trying to detect each of those things seems a bit tedious and error-prone. I'm also not sure that I totally agree that IE11 compatibility should be our responsibility, but I know that's a very debatable topic.

colinhacks:
Merged in zod@3.5.4. Great PR @mad-gooze.  

Happy to merge this now even if we don't have a great mechanism for catching future incompatibilities. As with most OSS< IE11 compatibility is best effort, and spending a lot of time coming up with a futureproof solution isn't worth anyone's time IMO. 

@mad-gooze thanks for doing this! Submit other issues/PRs if you encounter compat issues down the road and I'll happily merge them.

Issue #537, opened 2021-07-14
Add .promise() to ZodType


rubber-duck-software:
z.function() is remarkably useful for writing typesafe functions. Often these functions will be async, thus the return type will be a promise. Adding a .promise() method to the base class ZodType would enable syntax like 

```js
const getUsernameById z.function
    .args(number())
    .returns(z.string().promise())
    .implement(async id => {
        ...
        return username
    })
```

This would be a bit more ergonomic for the developer frequently writing async function definitions.

scotttrinh:
@rubber-duck-software 

How is that different from the existing `z.promise` type?

This seems to type-check just fine:

<img width="556" alt="image" src="https://user-images.githubusercontent.com/1682194/125842431-856297cb-9315-4732-aa2a-076e0152090d.png">

[TypeScript Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgLzgXzgMyhEcBEyEAJvgNwBQFAxhAHYDO8A5gKYwCqDrUdAhiFYAhAJ4BJYnAC8KAHSYArnWoxg9ABQBKCnF1xZfKMwbrksugpAAjHlu179Udgt4mzYHCGDdTsplGA6Zjt7PVlQMAAbVkE6GHU+BhFlOGBJKQA+RB0HPScYFzoCBW5eAVZyHN00bSA)

rubber-duck-software:
Purely a convenience method like `.array()`. Type checking remains unchanged. 

`.array()` adds readability. Often times we say "string array" so it is convenient to type "z.string().array()"

`.promise()` would fall under a similar category. We often think of an async function as returning a string, so we might start by typing `z.string()` but then finish by typing `.promise()` since the function is async leaving us with `z.string().promise()`.

scotttrinh:
@rubber-duck-software Thanks for the explanation. I don't have a strong opinion on this since I didn't even know about the `array` method on the various primitive types, but I definitely find "container" types to make more sense the other way: e.g. `Promise<string>`, `Array<string>`, `Set<string>`, etc, but I realize that it's somewhat stylistic.

Since it seems aligned with the existing API style, I'd be happy to review a PR that adds `promise` everywhere there is an `array` method.

PR #548, opened 2021-07-22
Add promise() to ZodType

rubber-duck-software:
Git Issue #537

### DIFF
--- a/README.md
+++ b/README.md
@@ -1497,7 +1497,7 @@ z.array(z.string());
 A convenience method for promise types:
 
 ```ts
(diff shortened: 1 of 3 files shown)


scotttrinh:
@rubber-duck-software

The pre-commit hook should've written a few more files (specifically Deno stuff), so I'm not sure if that's something that @colinhacks usually does when merging PRs, but we probably want to get those before we can merge this.

Would you mind updating the relevant part of the README as well?

scotttrinh (approved):
Looks good to me!

PR #557, opened 2021-07-26
Fix isInvalid equality check

### LINKED ISSUE #556: Zod not working when imported across packages
We ran into an issue when upgrading from 3.2.0 to 3.5.1, that when we're importing Zod defined schemas across our package boundaries (which just means that Zod is installed multiple times, once in each package that uses it) Zod breaks and the validation doesn't actually run.

I've put together a small repo to reproduce this: https://github.com/krivachy/zod-object-bug

The setup is:
- There are two packages in the `packages/` directory
- packageA contains a validation: [zEmail](https://github.com/krivachy/zod-object-bug/blob/main/packages/packageA/src/emailValidation.ts)
- packageB contains a model that uses zEmail: [models.ts](https://github.com/krivachy/zod-object-bug/blob/main/packages/packageB/src/models.ts)
- test fails when it's imported, but passes when it's locally defined: [index.test.ts](https://github.com/krivachy/zod-object-bug/blob/main/packages/packageB/index.test.ts)
- The same version of Zod is installed in both [packageA](https://github.com/krivachy/zod-object-bug/blob/main/packages/packageA/package.json) and [packageB](https://github.com/krivachy/zod-object-bug/blob/main/packages/packageB/package.json)

If we install Zod once at a root package.json level it seems to work, but we'd like to have our packages define their own dependencies.

Other libraries work perfectly fine this way, so I'm a bit confused as to why Zod breaks with this setup? Any advice would be much appreciated, thanks!

krivachy:
`===` comparisons only works when Zod is installed once in a monorepo setup with multiple packages using Zod.

This PR changes it to use structural type checking rather than referential equality checking.

The test reproduces the issue (i.e. fails before applying the change).

Closes #556

### DIFF
--- a/deno/lib/__tests__/parseUtil.test.ts
+++ b/deno/lib/__tests__/parseUtil.test.ts
@@ -1,0 +1,12 @@
+// @ts-ignore TS6133
+import { expect } from "https://deno.land/x/expect@v0.2.6/mod.ts";
+const test = Deno.test;
+
+import { INVALID, isInvalid } from "../helpers/parseUtil.ts";
+
+test("parseUtil isInvalid should use structural typing", () => {
+  // Test for issue #556: https://github.com/colinhacks/zod/issues/556
+  const newInstance: INVALID = Object.freeze({ valid: false });
+  expect(isInvalid(newInstance)).toBe(true);
+  expect(isInvalid(INVALID)).toBe(true);
+});
--- a/deno/lib/helpers/parseUtil.ts
+++ b/deno/lib/helpers/parseUtil.ts
@@ -198,7 +198,7 @@ export type ParseReturnType<T> =
   | ASYNC<SyncParseReturnType<T>>;
 
 export const isInvalid = (x: ParseReturnType<any>): x is INVALID =>
-  x === INVALID;
+  (x as any).valid === false;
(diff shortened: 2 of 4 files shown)


scotttrinh:
@krivachy Nice work sleuthing this. I'll double-check the benchmarks to ensure it's not a big regression, but I imagine it shouldn't be a noticeable increase in checking object reference equality vs checking an object's property boolean state. I'll comment here when I get some benchmarking results back.

krivachy:
Thanks :+1: Looking forward to have this PR in for the next release :)

scotttrinh:
I didn't notice any performance regressions on the benchmarks, so I think this is good to go!

scotttrinh (approved):

krivachy:
@scotttrinh This isn't super urgent for me, but just wondering why you haven't merged the PR yet? Does someone else also need to review this PR and merge it?

scotttrinh:
@krivachy I was hoping someone else would do the actual merge since I haven't done it since becoming a collaborator, but I'm not sure why I'm hesitant! I'll merge it now, thanks for the ping.

krivachy:
Thanks! This is a tiny PR, so shouldn't wreck havoc as your first merge 🤣

colinhacks:
@scotttrinh don't be hesitant! really glad this got merged.

Issue #564, opened 2021-07-28
Ability to `flatMap` on errors


adriangroch:
Like with `flatten()`, would be nice to have a `flatMap<U>(mapper: (string) => U )` to transform issues as needed.

Motivation for this is around the idea of being able to differentiate between custom-defined errors versus warnings in a form. Being able to flatten the errors, whilst also supplying the error code, or other `ZodIssue` specific information is a nice stepping stone to allow for that.

PR #565, opened 2021-07-28
Add flatMap capability to errors

adriangroch:
Proposed fix for #564


colinhacks:
Really great idea! Instead of introducing a new method, I just added this functionality inside the original `flatten` method. I think it's a simpler API and avoids duplicative code. 

Plus `flatMap` means something [very specific](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) in JS, and this doesn't perform a "flatMap" operation as most people understand it. 

Great PR! Thanks for adding documentation too 🤙

Issue #572, opened 2021-08-03
Stack Trace in Async Zod Function


rubber-duck-software:
Currently, async zod functions do not produce a useful stack trace when their validation fails. The trace generally contains only zod internals, looking something like this... 

```
Error: [
  {
    "code": "invalid_arguments",
    "argumentsError": {
      "issues": [
        {
          "code": "too_small",
          "minimum": 1,
          "inclusive": true,
          "type": "array",
          "path": [],
          "message": "Should have at least 1 items"
        }
      ]
    },
    "path": [],
    "message": "Invalid function arguments"
  }
]
    at new ZodError (/Users/USERNAME/PROJECT/node_modules/zod/lib/ZodError.js:75:28)
    at handleArgs (/Users/USERNAME/PROJECT/node_modules/zod/lib/types.js:1786:23)
    at ZodFunction.<anonymous> (/Users/USERNAME/PROJECT/node_modules/zod/lib/types.js:1817:49)
    at step (/Users/USERNAME/PROJECT/node_modules/zod/lib/types.js:59:23)
    at Object.next (/Users/USERNAME/PROJECT/node_modules/zod/lib/types.js:40:53)
    at fulfilled (/Users/USERNAME/PROJECT/node_modules/zod/lib/types.js:31:58)
```

Async function validation should be reworked slightly to produce a more usable stack trace

rubber-duck-software:
@colinhacks - could you take a look at this one? I think it resolves something you attempted in 3.7

mvasin:
Hi! I'm on version 3.20.2, and the problem persists. Here's the stack trace I receive:

```
/Users/user/project/node_modules/zod/lib/types.js:29
        const error = new ZodError_1.ZodError(ctx.common.issues);
                      ^
ZodError: [
  {
    "code": "invalid_type",
    "expected": "string",
    "received": "undefined",
    "path": [],
    "message": "Required"
  }
]
    at handleResult (/Users/user/project/node_modules/zod/lib/types.js:29:23)
    at ZodString.safeParse (/Users/user/project/node_modules/zod/lib/types.js:142:16)
    at ZodString.parse (/Users/user/project/node_modules/zod/lib/types.js:122:29)
    at Object.<anonymous> (/Users/user/project/src/config.ts:80:28)
    at Module._compile (node:internal/modules/cjs/loader:1159:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1213:10)
[...]

PR #573, opened 2021-08-03
Stack Trace in Async Zod Function

rubber-duck-software:
Git Issue #572


scolestock:
This would be great to get merged in order to improve the stack traces we get when return type parsing fails.  We're in the midst of migrating to Zod, but the stack traces we get in sentry.io are not enough to troubleshoot with.

rubber-duck-software:
This is a blocking issue for Marcato. We would like to use Zod in more places in our backend for runtime type safety, but runtime type safety is only helpful if we can track down the issues quickly. Currently we have an error with stack trace 

```
ZodError: [
  {
    "code": "invalid_arguments",
    "argumentsError": {
      "issues": [
        {
          "code": "invalid_type",
          "expected": "string",
          "received": "undefined",
          "path": [
            0,
            "filters",
...
  File "/var/app/current/node_modules/zod/lib/ZodError.js", line 75, col 28, in new ZodError
    var _this = _super.call(this) || this;
  File "/var/app/current/node_modules/zod/lib/types.js", line 1747, col 23, in handleArgs
    throw new ZodError_1.ZodError([issue]);
  File "/var/app/current/node_modules/zod/lib/types.js", line 1778, col 49, in ZodFunction.<anonymous>
    parsedArgs = _a.apply(void 0, _b.concat([_e.sent()]));
  File "/var/app/current/node_modules/zod/lib/types.js", line 59, col 23, in step
    op = body.call(thisArg, _);
  File "/var/app/current/node_modules/zod/lib/types.js", line 40, col 53, in Object.next
    function verb(n) { return function (v) { return step([n, v]); }; }
  File "/var/app/current/node_modules/zod/lib/types.js", line 31, col 58, in fulfilled
    function fulfilled(value) { try { step(generator.next(value)); } catch (e) { reject(e); } }

```

There is no telling where in our code base this error was thrown. It is useless to have an error message like this. Please merge this PR or recommend revisions to it soon.

scotttrinh:
@rubber-duck-software 

> Please merge this PR or recommend revisions to it soon.

Thanks for taking the time to PR — I wish I knew enough about the async validation code to feel comfortable merging this or suggesting updates. In the meantime, I don't think there is anything that should stop you from publishing your own fork and working with us to get your changes upstreamed, so I don't think you have to feel like this really blocks adoption.

src/types.ts
```diff
@@ -2303,74 +2303,53 @@ export class ZodFunction<
...
     const params = { errorMap: ctx.params.errorMap };
     const fn = data;
 
     if (this._def.returns instanceof ZodPromise) {
       return OK(async (...args: any[]) => {
-        const parsedArgs = handleArgs(
-          args,
-          await this._def.args.spa(args, params)
-        );
+        const error = new ZodError([]);
```
scotttrinh, line 2325: One thing I can share my experience on is not initializing an empty error. We did a lot of work to remove these allocations for performance reasons earlier in the year. Should be as simple as moving this into the catch blocks below.

scotttrinh (changes requested):

src/types.ts
colinhacks: Very true though I think in this case it's equivalent to the current implementation. Both `handleArgs` and `handleReturns` create new ZodErrors currently.

rubber-duck-software: Hi @scotttrinh! I appreciate your response :). I'm not a pro at how node works internally, but strangely the stack trace is dramatically different depending on where the error is initialized. 

Stack Trace for Error Initialized in catch block 

(node:27924) UnhandledPromiseRejectionWarning: ZodError: [
  {
    "code": "invalid_return_type",
    "returnTypeError": {
      "issues": [
        {
          "code": "invalid_type",
          "expected": "string",
          "received": "number",
          "path": [],
          "message": "Expected string, received number"
        }
      ],
      "name": "ZodError"
    },
    "path": [],
    "message": "Invalid function return type"
  }
]
    at new ZodError (/Users/USERNAME/zod/src/ZodError.ts:134:5)
    at /Users/USERNAME/zod/src/types.ts:2335:27


Stack Trace for Error Initialized outside of catch block: 
```
(node:27945) UnhandledPromiseRejectionWarning: ZodError: [
  {
    "code": "invalid_return_type",
    "returnTypeError": {
      "issues": [
        {
          "code": "invalid_type",
          "expected": "string",
          "received": "number",
          "path": [],
          "message": "Expected string, received number"
        }
      ],
      "name": "ZodError"
    },
    "path": [],
    "message": "Invalid function return type"
  }
]
    at new ZodError (/Users/USERNAME/zod/src/ZodError.ts:134:5)
    at ZodFunction.<anonymous> (/Users/USERNAME/zod/src/types.ts:2325:23)
    at step (/Users/USERNAME/zod/src/types.ts:59:23)
    at Object.next (/Users/USERNAME/zod/src/types.ts:40:53)
    at /Users/USERNAME/zod/src/types.ts:34:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/USERNAME/zod/src/types.ts:30:12)
    at Object.numToString (/Users/USERNAME/zod/src/types.ts:1761:24)
    at callIt (/Users/USERNAME/zod/src/playground.ts:12:3)
    at callIt2 (/Users/USERNAME/zod/src/playground.ts:16:3)
```

colinhacks:
Sorry @rubber-duck-software, didn't mean to leave this so long. 

The reason I haven't merged it is that I don't really know what's happening. Is the goal just to shorten the number of stack frames that are internal to Zod? Is the problem that the stack trace is getting truncated before you can see which LOC in your codebase threw the error? 

In my testing, your version only decreased the number of Zod-internal stack frames by one. Could you provide a code snippet + associated stack traces where this isn't the case? It's possible the complexity of the input/returnType schemas are relevant here (I was using simple schemas in my tests).

Finally - if the stack trace is getting truncated before they're useful, I recommend increasing the stack trace limit. The default is 10 which is why you may not be getting as much detail as you need to properly debug. You can just execute `Error.stackTraceLimit = 100` anywhere in your application to change this globally in your codebase.

colinhacks:
@rubber-duck-software Also since you're a sponsor, feel free to DM me on Twitter @colinhacks whenever you need speedy attention on an issue/PR 👍

rubber-duck-software:
This could be an environment issue I suppose. I have tested this on both node v14.15.4 and node v12.20.2 with the same result. While my initial assessment was that it was a stack trace limit issue, I later found that increasing the stack trace limit did not change the number of frames associated with the error. On my local machine (as well as on our dev/prod servers), it seems that node could not fully trace the origin of the error. 

So the issue is not that the stack trace is dominated by zod internals as I initially supposed, rather the issue is that node cannot produce a stack trace which escapes Zod internals. It somehow loses the full stack trace, so that the entirety of the stack trace is something like 

```
at new ZodError (/Users/USERNAME/zod/src/ZodError.ts:134:5)
at handleResult (/Users/USERNAME/zod/src/types.ts:76:19)
at ZodString.<anonymous> (/Users/USERNAME/zod/src/types.ts:149:12)
at step (/Users/USERNAME/zod/src/types.ts:59:23)
at Object.next (/Users/USERNAME/zod/src/types.ts:40:53)
at fulfilled (/Users/USERNAME/zod/src/types.ts:31:58)
```

Here node traced the error to where it was constructed, but could not trace further up. I would expect the original caller to be included in the stack somewhere, but it is not.

Here is my `playground.ts` file to reproduce the issue

```
import { z } from "./index";

export const numToString = z
  .function()
  .args(z.number())
  .returns(z.string().promise())
  .strictImplement(async (arg1) => {
    return arg1 as any;
  });

function callIt() {
  numToString(3);
}

function callIt2() {
  callIt();
}

function callIt3() {
  callIt2();
}

callIt3();
```

Note: on my branch I know get a stack trace like this for the same `playground.ts`

```
at new ZodError (/Users/USERNAME/zod/src/ZodError.ts:134:5)
at ZodFunction.<anonymous> (/Users/USERNAME/zod/src/types.ts:2325:23)
at step (/Users/USERNAME/zod/src/types.ts:59:23)
at Object.next (/Users/USERNAME/zod/src/types.ts:40:53)
at /Users/USERNAME/zod/src/types.ts:34:71
at new Promise (<anonymous>)
at __awaiter (/Users/USERNAME/zod/src/types.ts:30:12)
at Object.numToString (/Users/USERNAME/zod/src/types.ts:1761:24)
at callIt (/Users/USERNAME/zod/src/playground.ts:12:3)
at callIt2 (/Users/USERNAME/zod/src/playground.ts:16:3)
```

src/types.ts
scotttrinh: I wonder if this is related to returning a promise directly vs `return await`. https://www.even.li/posts/2020-06-04-async-stack-traces/ has some detail here.

scotttrinh: Eh, that didn't seem to make any difference. Going to poke at it some more.

colinhacks:
Ah okay, I'm seeing this behavior with your playground. Thanks for the explanation! Merged in zod@3.7.3.

Issue #311, opened 2021-02-04
Incorrect error for unions with literals


aliksend:
**Description**
I use `zod@2.0.0-beta.30`
I want to use zod to parse structures like
```typescript
type T = { val: 'a', a?: string } | { val: 'b', b?: number }
```
but I get incorrect error when I tries to parse structure with incorrect field's type

**Demo**
```typescript
import * as z from 'zod'

const Schema = z.union([
  z.object({
    val: z.literal('a'),
    a: z.ostring()
  }),
  z.object({
    val: z.literal('b'),
    b: z.onumber()
  })
])

console.log(Schema.parse({
  val: 'a',
  a: 123         
}))
```

**Expected error**
```
Error: [
  {
    "code": "invalid_type",
    "expected": "string",
    "received": "number",
    "path": [
      "a"
    ],
    "message": "Expected string, received number"
  }
]
```

**Actual error**
```
Error: [
  {
    "code": "invalid_literal_value",
    "expected": "b",
    "path": [
      "val"
    ],
    "message": "Input must be \"b\""
  }
]
```

OzoTek:
I'm having the same issues, without literals. Basically any union types containing objects seems to yield wrong error messages.

Tried with `zod@1.11.1` and `zod@next`.
It seems like it builds up errors as it tries to parse each type and is not able to choose the most relevant one.

Here's another example:

```typescript
import * as zod from zod;

const schema = zod.union([
    zod.object({
        a: zod.string(),
    }),
    zod.object({
        b: zod.boolean()
    })
]);

schema.parse({ b: 'test' });
```

colinhacks:
Zod isn't able to guess which schema in the union is most "relevant". This just isn't possible in the general case. It may seem obvious to you that the `val` literals are being used as the discriminator for the union, but how would Zod be able to figure that out? Consider the general case — what if there were multiple keys associated with literal values, etc. This logic gets stupidly complicated really fast and just makes it harder for people to understand what's going on.

The real mystery here is why you received a "invalid_literal_value" error. You should have received an "invalid_union" that contains the errors thrown by each schema within the union. @aliksend when I run the code you provided this is the error I get: (this is on Zod 3 so there will be some small discrepancies but the top-level error should be `invalid_union`.

```ts
Error: [
  {
    "code": "invalid_union",
    "unionErrors": [
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "number",
            "path": [
              "a"
            ],
[...]

PR #578, opened 2021-08-09
Add benchmarks for object unions

scotttrinh:
Local run:

System stats

```
[snip other benchmarks]
z.union: double: valid: a x 2,285,600 ops/sec ±4.88% (84 runs sampled)
z.union: double: valid: b x 558,052 ops/sec ±5.98% (80 runs sampled)
z.union: double: invalid: null x 18,932 ops/sec ±12.05% (62 runs sampled)
z.union: double: invalid: wrong shape x 16,755 ops/sec ±4.62% (74 runs sampled)
z.union: many: valid: a x 2,090,784 ops/sec ±7.77% (81 runs sampled)
z.union: many: valid: c x 342,439 ops/sec ±6.93% (80 runs sampled)
z.union: many: invalid: null x 12,029 ops/sec ±1.76% (86 runs sampled)
z.union: many: invalid: wrong shape x 11,738 ops/sec ±1.84% (91 runs sampled)
```


scotttrinh:
Interesting that getting a miss on the first member of the union is so expensive, and seems to get more expensive the further "away" the correct member is from the first one.

scotttrinh:
I think we can/should(/I might take a swing at) do something similar to this suggestion: https://github.com/colinhacks/zod/issues/311#issuecomment-818266883

It'd be nice to be able to tell Zod which field was the discriminant so we didn't have to run the full schema parsing on each schema at runtime.

PR #623, opened 2021-08-31
Added multipleOf number validation (evenly divisible).

StefanTerdell:
Came across [multipleOf in the JSON schema docs](https://json-schema.org/understanding-json-schema/reference/numeric.html#multiples) and thought it'd be nice to have in Zod. 

Easily implemented with a simple refinement, but then I can't convert it to a JSON schema and will loose sleep. :(

Jokes aside, it's a pretty basic feature and may be useful for things like `<input type="number" step=x />` in form generators etc.


StefanTerdell:
Easily implemented with a simple refinement, but then I can't convert it to a JSON schema and will loose sleep. :(

src/types.ts
```diff
@@ -736,6 +746,21 @@ export class ZodNumber extends ZodType<number, ZodNumberDef> {
...
+        ...this._def.checks,
+        {
+          kind: "multipleOf",
+          value: value,
+          message: errorUtil.toString(message),
+        },
+      ],
+    });
+
+  mod = this.multipleOf;
```
scotttrinh, line 762: Not a blocker:

I'm not sure this alias is helpful or clear. It reads a bit like it is transforming the value rather than validating it.

scotttrinh (changes requested):
One non-blocking bit of feedback, but also, could you reset the permissions back to `100644` on all of these updated files?

src/types.ts
StefanTerdell: Hmm yep, that's fair. Perhaps `.step(x)` to coincide with dom input api?

scotttrinh: Sure! I personally thing `multipleOf` is perfectly clear and that aliases are noisy/distracting, but that's 100% my personal opinion so if it seems more user-friendly to have something like `step`, that seems better than `mod` to me.

StefanTerdell: Ok, changed name of alias to step for now since I personally like having shorthands around, but really wouldn't mind if you decided to remove it before release.

scotttrinh (approved):
🙏

Issue #621, opened 2021-08-30
Add .partialBy or .makeOptional


rubber-duck-software:
A common use case for objects is making some subset of its fields optional. Currently, the syntax for this is

```js

const mySchema = z.object({
    foo: z.string(),
    bar: z.number(),
    xyz: z.boolean()
})

const myNextSchema = mySchema.omit({
    foo: true,
    bar: true
}).augment({
    foo: mySchema.shape.foo,
    bar: mySchema.shape.bar
})

// {
//      foo?: string | undefined,
//      bar?: number | undefined,
//      xyz: boolean
// }

```

I propose a new method on ZodObject which would allow for doing something like
```js
const mySchema = z.object({
    foo: z.string(),
    bar: z.number(),
    xyz: z.boolean()
})

const myNextSchema = mySchema.partialBy({
    foo: true,
    bar: true
})
// {
//      foo?: string | undefined,
//      bar?: number | undefined,
//      xyz: boolean
// }
```

colinhacks:
A modified version of this has dropped in Zod 3.9: https://github.com/colinhacks/zod/pull/663

PR #624, opened 2021-08-31
Add partialBy to ZodObject

rubber-duck-software:
GH Issue #621


scotttrinh:
@rubber-duck-software 

Implementation looks solid! Would you be interested in adding some tests similar to [the suite for pick and omit?](https://github.com/colinhacks/zod/blob/master/src/__tests__/pickomit.test.ts) If not, we can add them on our end when integrating this PR.

scotttrinh (approved):
Awesome, looks great!

colinhacks:
A good idea and a good implementation. That said, I think this should be an overload of `.partial`. Could you submit another PR? Then I'll publish this as a part of zod@3.9.

colinhacks:
@rubber-duck-software Scratch that, I just did this myself. This should land in zod@3.9 later today.

jlmcabral:
Hello @colinhacks ,

When will this be officially released?

PR #640, opened 2021-09-10
fix typos

frankie303:


src/types.ts
```diff
@@ -2902,7 +2902,7 @@ export class ZodEffects<
   constructor(def: ZodEffectsDef<T>) {
     super(def);
     // if (def.schema instanceof ZodEffects) {
-    //   throw new Error(ZodFirstPartyTypeKind.ZodEffectscannot be nested.");
+    //   throw new Error(ZodFirstPartyTypeKind.ZodEffects cannot be nested.");
```
scotttrinh, line 2905: Can you revert this one? It's actually invalid even with the typo fixed, so no reason to add more git history to this line.

scotttrinh (changes requested):
Thanks for taking the time to PR this! One small request.

src/types.ts
frankie303: Sure I can revert it but just curious, what is the side effect of adding one more line to git history in this case?

scotttrinh: Not a huge deal, just makes doing a `git blame` on this chunk of commented out code have a random typo fix commit in the middle of it. Definitely not an actual problem 😅

colinhacks: Oops sorry Scott, saw this review immediately after merging. This whole comment is getting deleted in v3.9 so doesn't really matter.

PR #642, opened 2021-09-10
Fix react-native/metro bundler warning

adbl:
When using Zod in a react-native app, we see the following warning when running the bundler:
```
warn Package zod has been ignored because it contains invalid configuration. Reason: Package subpath './package.json' is not defined by "exports" in <path>/node_modules/zod/package.json
```

The problem is described in https://github.com/react-native-community/cli/issues/1168 and that is where it should be properly fixed. In the meantime (that issue has been open for some time), the warning can be fixed in affected packages by adding an export for `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
   "module": "./lib/index.mjs",
   "exports": {
     "require": "./lib/index.js",
-    "import": "./lib/index.mjs"
+    "import": "./lib/index.mjs",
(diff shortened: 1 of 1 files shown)


adbl:
The fix wasn't correct, closing for now.

adbl:
This seems to work, but I'm not really sure what it means to but the previous exports inside `"."` ... 🤷‍♂️

Issue #636, opened 2021-09-07
Error "Intersection results could not be merged" on array fields in zod 3.8.2


azhiv:
Consider the following example:
```ts
const arraySchema = zod.object({
  attachments: zod.array(zod.object({}))
});

console.log(
  arraySchema.and(arraySchema).safeParse({
    attachments: []
  })
);
```
The result is a `ZodError`:
```ts
ZodError: [
  {
    "code": "invalid_intersection_types",
    "path": [],
    "message": "Intersection results could not be merged"
  }
]
```
At the same time a very similar object schema is parsed successfully:
```ts
const objectSchema = zod.object({
  attachment: zod.object({})
});

console.log(
  objectSchema.and(objectSchema).safeParse({
    attachment: {}
  })
);
```
[Here](https://codesandbox.io/s/heuristic-khayyam-62tpo?file=/src/index.ts:719-871) is a sandbox with zod 3.
[Here](https://codesandbox.io/s/wonderful-smoke-6osu7?file=/src/index.ts) is a sandbox with zod 2 which gives success for both cases.

mmahalwy:
running into this as well

azhiv:
@mmahalwy This issue prevents us from migrating to ver. 3, so the workaround for us is to stay on ver. 2 for now.
However, I assume that the fix should be significantly simple. I guess the issue is located in `mergeValues` https://github.com/colinhacks/zod/blob/0b7a9837b64cd8af5580424fe09e03d98958d917/src/types.ts#L2051
There's a check that the two values equal, or their types are `ZodParsedType.object`. But is case of arrays the type is probably different (`ZodParsedType.array`?).

azhiv:
@colinhacks Would you mind taking a look at the above PR?

mmahalwy:
oh wow, thanks @azhiv

PR #676, opened 2021-09-29
Support array type during intersection

azhiv:
The change is motivated by https://github.com/colinhacks/zod/issues/636.
It should be possible to intersect schemas that consist of common fields, even
in the common fields are defined via `z.array`. Prior to the current change it
was only possible to intersect primitives/objects.


mmahalwy:
@colinhacks any chance to take a 👁️

colinhacks:
Great PR, merged in 3.9.8 👍

PR #681, opened 2021-10-01
feat: trim method for ZodString

### LINKED ISSUE #653: Feature request: First class string trimming support?
Edit: I've since changed my mind, but leaving open to allow discussion.

I suppose this feature request is also a question about how the maintainers view the library in terms of its usage in form validation.

A very common use case when dealing with user input in forms is you need to ignore/strip whitespace. `"  "` or `"michael@example.com "` isn't valid input in some cases, and using `.min(1)` or `.email()` respectively will get what might be seen by some as the wrong result. In those cases you want to parse or validate the input after stripping the whitespace, and also transform the input to strip it out altogether in the ouput.

This question was raised in https://github.com/colinhacks/zod/issues/614, and `preprocess` was suggeted. The solution using `preprocess` presumably goes a follows:

```ts
z.preprocess((arg) => typeof arg === "string" ? arg.trim() : arg, z.string().min(1)).parse(" "); // throws
z.preprocess((arg) => typeof arg === "string" ? arg.trim() : arg, z.email()).parse("michael@example.com "); // ouput: "michael@example.com"
```

Now this seems quite verbose and you are having to validate the `arg` value ahead of your processing step (kind of the purpose of zod in the first place - to validate that something is a string etc...), and other libraries such as `yup` have builtin trim methods that will transform the input ahead of validation as a builtin operation:

```ts
Yup.string().trim().min(1).validateSync(" ") // ValidationError "this must be at least 1 characters"
```

My sense is `zod` wants there to be a strong seperation between transformation and parsing. Is there space for zod having better support for what appears to be a pretty common use case? Something along the lines of:

```ts
// transforms after checking that input is a string
z.string().trim().min(1).parse(" ") // throws
z.string().trim().email().parse("michael@example.com ") // ouput: "michael@example.com"

// or does not change output, and force you to add a transform if you want the parsed ouput to have whitespace stripped
z.string().trim().min(1).transform(str => str.trim()).parse(" ") // throws
z.string().trim().email().transform(str => str.trim()).parse("michael@example.com ") // ouput: "michael@example.com"
```

The latter case is still a little noisy and confusing, but at least I'm dealing with a string by this point instead of an `unknown` because it has already passed parsing as a string.

Does this fit into `zod`'s vision (I'm guessing that it might not)?

Djaler:
resolves #653


DannyFeliz:
@colinhacks what do you think about this PR?

DannyFeliz:
ping

vladshcherbin:
bump

DannyFeliz:
ping

SrBrahma:
Up. Found https://github.com/neuledge/computed-types and thought it is cool to have trim() integrated.

SrBrahma:
Many thanks, @colinhacks! I have many places to add it on my backend!

PR #752, opened 2021-11-06
Records should be partial so we're not required to use all keys

jeremybull:


colinhacks:
This isn't really how Record works in TypeScript:

```ts
const arg: Record<"foo" | "bar", string> = {
  foo: "some string",
};
// Property 'bar' is missing in type '{ foo: string; }' but required in type 'Record<"foo" | "bar", string>'.ts(2741)
```

Perhaps a better idea is to add a `.partial` method to ZodRecord akin to the same method on ZodObject. I'll add this to my list, or a PR is welcome.

Fangmingdu:
I’m also in need of this change. It would be great if we can get this in. Thanks!

PR #758, opened 2021-11-11
Remove duplicated block on custom error messages

5minpause:
Almost the same block on custom error messages can be found in lines 336–345.

```js
const name = z.string({
  required_error: "Name is required",
  invalid_type_error: "Name must be a string",
});
```

Leaves the question whether the properties in that object can be both => 
`required`/`required_error` and `invalid`/`invalid_type_error`

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,15 +312,6 @@ tuna.value; // "tuna"
 
 ## Strings
 
-You can customize certain errors when creating a string schema.
-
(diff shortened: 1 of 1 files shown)


colinhacks:
Wow good catch. `invalid` and `required` aren't actually supported at all...only `required_error` and `invalid_type_error`.

Issue #764, opened 2021-11-16
`description` not presented in the `ProcessedCreateParams`


equt:
According to the type sig, I'm expecting to retrieve my description set in the `RawCreateParams` previously.

```typescript
import { z } from 'zod'

const foo = z.string({ description: 'Foo' })
console.log(foo.description) // undefined in console
```

Then I check the code and find out that the processing function seems to [have no way passing the description data out](
https://github.com/colinhacks/zod/blob/72fa277a34970e542565c6d5630e7f451ff1e343/src/types.ts?_pjax=%23js-repo-pjax-container%3Afirst-of-type%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%3Afirst-of-type%2C%20%5Bdata-pjax-container%5D%3Afirst-of-type#L86-L103).

Is this a bug or I just miss something? Since I'm not really familiar with the codebase, just want to be sure before a PR.

equt:
My bad, after a thorough check, it turns out to be caused by another framework's cache mechanism :)

equt:
It turns out that I'm testing on my forked version, so there is really an issue. (sigh)

PR #767, opened 2021-11-17
fix: Add `description` to `ProcessedCreateParams`

equt:
c.c. #764


colinhacks:
Yep, seems to be an oversight. Thanks!

Issue #772, opened 2021-11-19
Intersection with date field


dirgapeter:
When schema with date field is used in intersection, parse fail with 
```json
{
          "code": "invalid_intersection_types",
          "path": [],
          "message": "Intersection results could not be merged"
}
```

example:
```javascript
it('intersection with date', () => {
    const schema = z.object({
      date: z.date(),
    });

    const result = schema.and(schema).safeParse({
      date: new Date(),
    });

    expect(result.success).toBe(true);
});
```

PR #773, opened 2021-11-19
fix: intersection with date fields

dirgapeter:
resolves [#772](https://github.com/colinhacks/zod/issues/772)


colinhacks:
Love it, thanks!

PR #777, opened 2021-11-22
Don't include received value in enum/literal error messages (fixes #461)

### LINKED ISSUE #461: Literal and enum validation errors contain input data
Normally `ZodError` does not contain the actual value that was invalid. IMHO this is good - errors are often logged, and there are many situations where you don't want to log the value - the value could be sensitive (credit card number, personally identifiable information, password/secret token of some kind), or it might be unsafe to log (might be several gigabytes, contain newlines, NUL bytes, ANSI terminal escape sequences, HTML, ... - some of which might cause problems when viewing logs later). 

However, for `z.literal` and `z.enum`, the message does contain the input data:

````
> z.literal('Foo').parse('Bar')
Uncaught ZodError: [
  {
    "code": "invalid_type",
    "expected": "Foo",
    "received": "Bar",
    "path": [],
    "message": "Expected Foo, received Bar"
  }
]
<snip>

> z.enum(['Foo','Bar']).parse('Bad')
Uncaught ZodError: [
  {
    "code": "invalid_enum_value",
    "options": [
      "Foo",
      "Bar"
    ],
    "path": [],
    "message": "Invalid enum value. Expected 'Foo' | 'Bar', received 'Bad'"
  }
]
<snip>
```

I would suggest aligning these two with the other errors (it looks like these are the only two exceptions), so that logged error messages can never contain the invalid input data itself.

pasieronen:
See #461 

Also the error produced by `ZodLiteral` did not actually match the `ZodInvalidTypeIssue` interface or `ERROR_HANDLING.md` documentation (`expected` and `received` were not actually `ZodParsedType`s)

### DIFF
--- a/deno/lib/ZodError.ts
+++ b/deno/lib/ZodError.ts
@@ -281,9 +281,7 @@ export const defaultErrorMap = (
     case ZodIssueCode.invalid_enum_value:
       message = `Invalid enum value. Expected ${issue.options
         .map((val) => (typeof val === "string" ? `'${val}'` : val))
-        .join(" | ")}, received ${
-        typeof _ctx.data === "string" ? `'${_ctx.data}'` : _ctx.data
-      }`;
+        .join(" | ")}`;
       break;
     case ZodIssueCode.invalid_arguments:
       message = `Invalid function arguments`;
--- a/deno/lib/__tests__/error.test.ts
+++ b/deno/lib/__tests__/error.test.ts
@@ -398,6 +398,19 @@ test("strict error message", () => {
   }
 });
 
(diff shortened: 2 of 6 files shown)


scotttrinh (approved):
I support this stance in general, and this PR in particular looks good to me! I'm not sure what we consider our "compatibility" responsibility with regards to message strings, so I'd love to here from @colinhacks before merging.

colinhacks:
LGTM - Changes to error language don't merit a major version bump imo so let's merge

I'll publish a minor release with this and some other changes this weekend.

PR #784, opened 2021-11-24
Handle singular and plural verbiage

bsell93:
Hello! Would love to handle singular/plural grammar in the error messages. Open to suggestions & other changes. I thought I'd at least put this out there to see if there are any thoughts on this.

### DIFF
--- a/src/ZodError.ts
+++ b/src/ZodError.ts
@@ -302,11 +302,11 @@ export const defaultErrorMap = (
       if (issue.type === "array")
         message = `Should have ${issue.inclusive ? `at least` : `more than`} ${
           issue.minimum
-        } items`;
+        } item${issue.minimum === 1 ? '' : 's'}`;
       else if (issue.type === "string")
         message = `Should be ${issue.inclusive ? `at least` : `over`} ${
           issue.minimum
-        } characters`;
+        } character${issue.minimum === 1 ? '' : 's'} long`;
       else if (issue.type === "number")
         message = `Value should be greater than ${
           issue.inclusive ? `or equal to ` : ``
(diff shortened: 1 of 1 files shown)


scotttrinh (approved):
@bsell93 

Thanks for bringing this up! User facing copy is always such an 🙃 thing in error messaging, so improving that is always a welcome update. There is probably something more consistent that we can do, but I think what you have here is an improvement already, and Colin has stated that copy changes are not "breaking", so I think this is safe to merge. Thanks!

src/ZodError.ts
```diff
@@ -317,11 +317,11 @@ export const defaultErrorMap = (
       if (issue.type === "array")
         message = `Should have ${issue.inclusive ? `at most` : `less than`} ${
           issue.maximum
-        } items`;
+        } item${issue.maximum === 1 ? '' : 's'}`;
       else if (issue.type === "string")
         message = `Should be ${issue.inclusive ? `at most` : `under`} ${
           issue.maximum
-        } characters long`;
+        } character${issue.minimum === 1 ? '' : 's'} long`;
```
scotttrinh, line 324:
```suggestion
        } character${issue.maximum === 1 ? '' : 's'} long`;
```

src/ZodError.ts
```diff
@@ -302,11 +302,11 @@ export const defaultErrorMap = (
       if (issue.type === "array")
         message = `Should have ${issue.inclusive ? `at least` : `more than`} ${
           issue.minimum
-        } items`;
+        } item${issue.minimum === 1 ? '' : 's'}`;
       else if (issue.type === "string")
         message = `Should be ${issue.inclusive ? `at least` : `over`} ${
           issue.minimum
-        } characters`;
+        } character${issue.minimum === 1 ? '' : 's'} long`;
```
scotttrinh, line 309: did you mean to add ` long` here? If so, we'll need to update the failing test.

scotttrinh (changes requested):

colinhacks:
Thanks for this PR @bsell93. It motivated me to go a slightly different direction with these error messages that I think is more explicit: https://github.com/colinhacks/zod/commit/2021a5e07c3e0a06a33b4c44441fc9863ae51ae4 

Closing for now.

bsell93:
Glad to bring inspiration! 😁

PR #799, opened 2021-11-29
Update types.ts

JacobWeisenburger:
- Removed some duplication:
```ts
{ success: true; data: Output } | { success: false; error: ZodError<Input> }
```
```ts
SafeParseReturnType<Input, Output>
```

- Added some better support for [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).
```ts
export type SafeParseSuccess<Output> = { success: true; data: Output; error: undefined };
```
```ts
export type SafeParseError<Input> = { success: false; data: undefined; error: ZodError<Input> };
```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -60,9 +60,7 @@ export interface ZodTypeDef {
 const handleResult = <Input, Output>(
   ctx: ParseContext,
   result: SyncParseReturnType<Output>
-):
-  | { success: true; data: Output }
-  | { success: false; error: ZodError<Input> } => {
+): SafeParseReturnType<Input, Output> => {
   if (isValid(result)) {
     return { success: true, data: result.value };
   } else {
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -102,8 +100,8 @@ function processCreateParams(params: RawCreateParams): ProcessedCreateParams {
   return { errorMap: customMap };
 }
 
-export type SafeParseSuccess<Output> = { success: true; data: Output };
-export type SafeParseError<Input> = { success: false; error: ZodError<Input> };
+export type SafeParseSuccess<Output> = { success: true; data: Output; error: undefined };
+export type SafeParseError<Input> = { success: false; data: undefined; error: ZodError<Input> };
```
scotttrinh, line 104: Maybe this is a stylistic quirk, but I find `: undefined` to be misleading. It's not defined as `undefined`, it's just not present, which maybe is just a semantic quibble, but I think that's an important distinction. Unless we want to actually pass `undefined` for those keys. I guess it's not strictly semantic-only since it is different at runtime as well:

```ts
const a = { a: undefined };
"a" in a; // true
```

As it is, type narrowing should give you what you want (via `if`/`else` or ternary, etc), even if it doesn't take advantage of optional chaining.

JacobWeisenburger: Sorry, I'm new to pull requests, and I couldn't figure out how to add changes to this one, so I made a new one. Hopefully, you or someone else can let me know how to make changes to a pull request the right way so I know for next time. :D

[Update types.ts #799 Continued #809](https://github.com/colinhacks/zod/pull/809)

PR #809, opened 2021-12-03
Update types.ts #799 Continued

JacobWeisenburger:
Sorry, I'm new to pull requests, and I couldn't figure out how to add changes to my last one, so I made a new one. Hopefully, someone can let me know how to make changes to a pull request the right way so I know for next time. :D

Previous PR: [Update types.ts #799](https://github.com/colinhacks/zod/pull/799)

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -60,9 +60,7 @@ export interface ZodTypeDef {
 const handleResult = <Input, Output>(
   ctx: ParseContext,
   result: SyncParseReturnType<Output>
-):
-  | { success: true; data: Output }
-  | { success: false; error: ZodError<Input> } => {
+): SafeParseResult<Input, Output> => {
   if (isValid(result)) {
     return { success: true, data: result.value };
   } else {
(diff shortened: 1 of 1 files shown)


scotttrinh:
@JacobWeisenburger 

I can definitely help you figure out how to add changes to your PR. The overall idea is to just add commits to your local branch, and then push those commits to _your_ branch on _your_ fork on GitHub, and it will automatically show up here as an update.

However, I don't think we want to make the change you're suggesting here. We specifically want a discriminated union here to make it plain what the valid states are. I know that breaks the conditional chaining that you're hoping to get, but I think I speak for the rest of the maintainers when I say that we're fine with that trade-off and it gives you better runtime guarantees in your code.

I'll leave this PR open for a bit so you can mess around with other strategies to get the ergonomics you're looking for, but I don't believe there is a straightforward way to get optional chaining outside of using conditionals to narrow if we keep the discriminated union (and, to reiterate, we'd like to keep that!)

JacobWeisenburger:
ok 😞

Thanks for your patience and the extra explanation.

Issue #702, opened 2021-10-11
`ZodNumber.step` does not use the same logic as number input `step`


scotttrinh:
This implementation fails with cases like the following:

`23.9 % 0.01 = 0.009999999999998082`

_(the client enters a price, that must have at max 2 decimal places, = step(0.01))_

https://stackoverflow.com/questions/7037926/javascript-how-to-tell-if-one-number-is-a-multiple-of-another

https://stackoverflow.com/questions/3966484/why-does-modulus-operator-return-fractional-number-in-javascript

@StefanTerdell @colinhacks @scotttrinh 

For my case, it would be good a validation like max/minDecimals(), but maybe you can come up with better solutions.

_Originally posted by @SrBrahma in https://github.com/colinhacks/zod/issues/623#issuecomment-939585373_

JacobWeisenburger:
Has this issue been resolved? I see that https://github.com/colinhacks/zod/pull/623 was merged, but not sure if that fixed this.

PR #816, opened 2021-12-07
fix: multiple of decimals

mmahalwy:
Consider this:
```ts
const percentage = z.number().min(0).max(1).step(0.001);
```

Previously, `percentage.parse(0.5)` would fail because the Modulus operator would return non-zero
```
> 0.5 % 0.001
0.0009999999999999896
```

Instead, this PR adds the logic to check if the division is a whole number. Ideally, `Number.isInteger` would be used but it's banned via eslint rule

### DIFF
--- a/deno/lib/__tests__/number.test.ts
+++ b/deno/lib/__tests__/number.test.ts
@@ -11,6 +11,7 @@ const lteFive = z.number().lte(5);
 const intNum = z.number().int();
 const multipleOfFive = z.number().multipleOf(5);
 const stepSixPointFour = z.number().step(6.4);
+const percentage = z.number().min(0).max(1).step(0.001);
 
 test("passing validations", () => {
   gtFive.parse(6);
@@ -20,6 +21,7 @@ test("passing validations", () => {
   intNum.parse(4);
   multipleOfFive.parse(15);
   stepSixPointFour.parse(12.8);
+  percentage.parse(0.5);
 });
 
 test("failing validations", () => {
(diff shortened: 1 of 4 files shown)


mmahalwy:
bump :)

mmahalwy:
@scotttrinh this is blocked us - can you take a look?

scotttrinh:
@mmahalwy 

Sorry for the delay. Honestly, I'm a little worried about this since it's kind of breaking the existing behavior even if it's more "correct" and aligned with the original intent. I'd feel more comfortable if @colinhacks gave the go-ahead on this before merging.

In the meantime, you're free to publish a fork while we work it out, I believe our license is rather permissive in that regards. You could also use a custom refine function until this is resolved to unblock your team, but I'll let you make that decision for yourself.

mmahalwy:
@colinhacks bumping this up :)

colinhacks:
This solution feels pretty hacky in comparison to the solution proposed in #702. There are also still some bugs in this implementation such as:

```ts
z.number().step(0.0001).parse(3.01);
// fails
```

I [just implemented](https://github.com/colinhacks/zod/commit/eeabcdd0d4011fc08368d20091bfffd98fc4a7e0) the "float-safe modulus" solution which should solve this problem once and for all.  Mo, does this address your use cases?

mmahalwy:
@colinhacks Works for me! Yeah I felt my implementation was hacky especially that I couldn't use Number.isInteger

PR #827, opened 2021-12-15
Fix broken twitter link

ajcwebdev:
The Twitter link was using Markdown inside of HTML and thus was broken because it's being centered in a `<p>` tag. I changed it to just a regular ol' `<a>` tag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 
 </p>
 <p align="center">
-by [@colinhacks](https://twitter.com/colinhacks)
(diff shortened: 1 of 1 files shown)


scotttrinh:
@ajcwebdev The link seems to work fine in the README as-is? This doesn't seem like a big deal, but could you maybe provide a screenshot of what you're seeing?

scotttrinh:
Oh, I was actually looking at the wrong one. 🤦

Issue #831, opened 2021-12-18
Mapping readonly array to Zod union  


cefn:
Hi and thanks for Zod. It's been a joy to use. 

Being able to use the following would be even nicer, but it fails for a variety of hard-to-fix reasons...

```typescript
const items = ["a","b","c"] as const;
const unionSchema = z.union((items.map((item) => z.literal(item)));
```

I often use an `as const` array or object to simultaneously define types (guiding the compiler) AND sequences (used at runtime). 

For example in the toy case below, the `PRIORITIES` list defines the type of a `Priority` string, and can also be iterated to construct a dropdown menu for example. This is a common emerging pattern in typescript codebases. It can ensure that variants are defined once, and that none are missed in mapped types either at compile-time or runtime.

Defining a zod schema from an `as const` array hits a dead end for me. Perhaps it's impossible. Perhaps I am missing how to do this properly. If it's this hard, some kind of utility type could be needed.

# Problems

There are an interlocking set of problems...
* Zod not accepting `Readonly` definitions meaning the use of `as const` causes arguments to be rejected.
* the required arity of a union (at least two) meaning arrays of unknown length are not accepted - normally resolved by using readonly arrays with a known length
* a [weakness of typescript](https://github.com/microsoft/TypeScript/issues/44309), which can't keep the tuple nature when using e.g. map() 

# Solution: Utility Function

If a utility function was provided, I imagine overloaded signatures as follows...

```typescript
type ZodLiteralTuple<T extends readonly Primitive[]> = {
  readonly [K in keyof T]: ZodLiteral<T[K]>;
};

function createUnionSchema<T extends readonly []>(values: T): ZodNever;
function createUnionSchema<T extends readonly [Primitive]>(
  values: T
): ZodLiteral<T[0]>;
function createUnionSchema<
  T extends readonly [Primitive, Primitive, ...Primitive[]]
>(values: T): z.ZodUnion<ZodLiteralTuple<T>>;
```

Here, ZodLiteralTuple is a type to use as a type assertion after `Array.map`. It preserves tuple arity and projects each concrete literal found in T into a corresponding ZodUnionOptions datastructure. 

![image](https://user-images.githubusercontent.com/159819/146686024-e853dbc6-be11-4426-8f1b-6577b4ec9fcf.png)

## CASES 

* `createUnionSchema([])` should resolve to `z.never()`
* `createUnionSchema(["single"])` should return `z.literal("single")`
[...]

PR #832, opened 2021-12-18
Add Readonly to ZodUnionOptions

cefn:
Addresses issues where `as const` needs to be used in definitions to allow further type inference.

Currently `as const` values are prevented from being used since the existing signature requires writeable arrays and tuples 

Using `as const` to declare actual readonly tuples instead of populating writeable arrays enables strong type inference and ensures arity is detectable for calls to union constructors (which need a tuple of at least 2 members).

Fixes https://github.com/colinhacks/zod/issues/831

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1774,9 +1774,9 @@ export type AnyZodObject = ZodObject<any, any, any>;
 //////////                    //////////
 ////////////////////////////////////////
 ////////////////////////////////////////
-type ZodUnionOptions = [ZodTypeAny, ...ZodTypeAny[]];
+type ZodUnionOptions = Readonly<[ZodTypeAny, ...ZodTypeAny[]]>;
 export interface ZodUnionDef<
-  T extends ZodUnionOptions = [ZodTypeAny, ZodTypeAny, ...ZodTypeAny[]]
+  T extends ZodUnionOptions = Readonly<[ZodTypeAny, ZodTypeAny, ...ZodTypeAny[]]>
 > extends ZodTypeDef {
   options: T;
   typeName: ZodFirstPartyTypeKind.ZodUnion;
@@ -1864,7 +1864,7 @@ export class ZodUnion<T extends ZodUnionOptions> extends ZodType<
     return this._def.options;
   }
 
-  static create = <T extends [ZodTypeAny, ZodTypeAny, ...ZodTypeAny[]]>(
+  static create = <T extends Readonly<[ZodTypeAny, ZodTypeAny, ...ZodTypeAny[]]>>(
     types: T,
     params?: RawCreateParams
   ): ZodUnion<T> => {


cefn:
This is less of a blocker for me now, since I fixed it by remapping my types to be (unnecessarily) writeable in https://github.com/colinhacks/zod/issues/831#issuecomment-997444465  In general `readonly` is a better signature but I'll just leave this here.

colinhacks:
Great stuff, I agree this is the best solution. And yeah, `z.enum` is for string literals only. I don't think it's worth introducing something akin to `createUnionSchema` into core...creating unions of non-string primitive literals is pretty obscure use case IMO. But I'm glad you posted your solution for future reference.

cefn:
Hey, thanks also for looking into `createUnionSchema`! 

I was pleased with how the implementation fell in place in the end and its generality to any const literal array. 

It did occur to me that it could be an alternative implementation of `z.enum` (in that it handles a superset of cases), but I don't have a good handle on the consequences in terms of e.g. potential branching complexity in the typescript compilation process or other issues that my implementation might introduce across the zod core as a whole.

As you say, there's a reference implementation now for those who need it. Glad of your expertise and continued maintenance!

Aaronius:
My scenario is I have a list of "exemption payee codes" (from a United States W-9 form), which can be one of 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, or 13. Depending on the "tax classification", only certain exemption payee codes are valid. So, I'd like to be able to do something this:

```
const exemptPayeeCode = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13] as const;

// Variants based on tax classification:
const individualExemptPayeeCodes: readonly ExemptPayeeCode[] = [1, 2, 3, 4, 6, 7, 11, 12] as const;
const llcExemptPayeeCodes: readonly ExemptPayeeCode[] = [1, 2, 3, 4, 6, 7, 11, 12] as const;
const cCorpExemptPayeeCodes: readonly ExemptPayeeCode[] = [1, 2, 3, 4, 5, 6, 7, 9, 11, 12] as const;
const sCorpExemptPayeeCodes: readonly ExemptPayeeCode[] = [1, 2, 3, 4, 5, 6, 7, 9, 11, 12] as const;
const partnershipExemptPayeeCodes: readonly ExemptPayeeCode[] = [1, 2, 3, 4, 6, 7, 11, 12] as const;
const trustEstateExemptPayeeCodes: readonly ExemptPayeeCode[] = [
  1, 2, 3, 4, 6, 7, 8, 10, 11, 12, 13,
] as const;
const otherExemptPayeeCodes: readonly ExemptPayeeCode[] = [
  1, 2, 3, 4, 6, 7, 8, 10, 11, 12, 13,
] as const;
```

And build validators for each of the variants. Right now, with Zod only, it's a bit of a pain.

@cefn's solution gets me closer, but I have to make adaptations because if I do something like this:

```
createUnionSchema(individualExemptPayeeCodes)
```

TypeScript complains:

image

If I take off the `readonly ExemptPayeeCode[]` type from my constant arrays, it works, but I'd rather not because I want to ensure I'm not fat-fingering one of the payee codes in my array.

marcebdev:
Will tack on that I'd like to see this, I'm sure it's not the most common form of validation but having to validate against a set of say numeric literals certainly is a use case, at least with numbers.

I would imagine as of today most people are putting the numbers as strings (and transforming) to get around this but I'm surprised a validation library itself isn't able to validate against a list of codes itself.

KNedelec:
I agree, unions of integer literals are not an obscure use case to validate integer codes. Would love to have it included in core

cefn:
> Will tack on that I'd like to see this, I'm sure it's not the most common form of validation but having to validate against a set of say numeric literals certainly is a use case, at least with numbers.
> 
> I would imagine as of today most people are putting the numbers as strings (and transforming) to get around this but I'm surprised a validation library itself isn't able to validate against a list of codes itself.

For those arriving at this thread, an example implementation which is general to const arrays of literals (not limited to strings) and which might be considered to merge as a core feature is at... https://github.com/colinhacks/zod/issues/831#issuecomment-997444465

PR #837, opened 2021-12-26
Further translate untranslated English parts

geenva:
In this pull request, I have translated multiple headers and contents that have not been translated fully into Chinese.


colinhacks:
Thanks!!

PR #838, opened 2021-12-27
feat: default `unknownKeys` strategy can be overridden by parse param

pepakriz:
Hi,

It'd be nice to have an option for forcing the default `unknownKeys` strategy on all objects that have not explicitly defined their own unknownKeys behavior. In real life, I sometimes need to have just one schema for one complex object, but once with `strict` strategy (mostly in endpoint input validations to avoid mistakes) and once with `passthrough` (for response validation from 3rd party APIs). With the current Zod version, I don't see any simple solution for my use case without big code duplication. So I'd like to know your opinion on my PR that perfectly solves my problem.

Huge thanks for your awesome library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -728,17 +728,22 @@ const deepPartialUser = user.deepPartial();
 
 #### Unrecognized keys
 
-By default Zod objects schemas strip out unrecognized keys during parsing.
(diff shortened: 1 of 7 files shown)


colinhacks:
I'm broadly opposed to parse params for stuff like this. Schemas should be fully self-contained. Seems like you're use case is entirely covered with something like this, no?

```ts
const baseObject = z.object({ ... });

baseObject.strict().parse(...);
baseObject.passthrough().parse(...);
```

pepakriz:
Your code example shows the modification of one object layer. I had in mind deep nested object structures. What is your recommendation how to achieve something like `baseObject.deepStrict().parse(...);`?

In my project as a workaround I've written a small function helper for deep object transformation, but it's ugly as hell :) so I'm looking for a better solution.

PR #840, opened 2021-12-28
Fix instanceof map and set

NumminorihSF:
There are some cases then zod is running in es5 env.

In such cases zod throws an error like "Map is not defined".
This commit adds `typeof X` check before all `instanceof X`. It allows skip `instanceof` checks in case of missed `X`.

There is one more issue in case of `#safeParse()` with missed `Map`.
To prevent this issue `ParseContext#typeCache` becomes nullable. So, when something calls `#safeParse()`, zod checks if `Map` present and creates `#typeCache` in this case. If `Map` is missed, zod sets `#typeCache` to `undefined` and check runs slower.


colinhacks:
Looks fantastic, good solution!

PR #856, opened 2022-01-15
Add types to extract error results from schema

### LINKED ISSUE #855: Helper type to infer errors from flattened `safeParse` error result.
When using `flatten()` on error the result from `safeParse()`, it doesn't currently return type-safe `fieldErrors`, as it is typed with index signatures.

The idea is to provide an additional helper type similar to `z.infer`, but which provides a type-safe flattened error result type.

My current solution looks like this:

```ts
type inferSafeParseErrors<T extends z.ZodType<any, any, any>, U = string> = {
  formErrors: U[];
  fieldErrors: {
    [P in keyof z.infer<T>]?: U[];
  };
};
```

The name of the type is subject to change.

A use case for this is when doing form validation, where you want to have type safety and autocompletion around error types from validated fields. [See an example of this with form validation in Remix](https://gist.github.com/fnky/7fe414b402baf5c9ba8f7ddecfdd263e).

fnky:
- Adds two types `z.inferFlattenedErrors` and alias type `z.inferFormErrors` to [infer error result types for `flatten()` from a schema](https://github.com/colinhacks/zod/pull/856/files#diff-ae3ae59764197d1e97d44cb8d123d849b078e60d66aa6d8332851e7b51fb24abR377).
- Adds additional overloading for `flatten` to [support type-safety with the new types](https://github.com/colinhacks/zod/pull/856/files#diff-ae3ae59764197d1e97d44cb8d123d849b078e60d66aa6d8332851e7b51fb24abR413-R417).

Fixes #855

### DIFF
--- a/src/ZodError.ts
+++ b/src/ZodError.ts
@@ -217,12 +217,24 @@ export class ZodError<T = any> extends Error {
     this.issues = [...this.issues, ...subs];
   };
 
-  flatten = <U = string>(
+  flatten(
+    mapper?: (issue: ZodIssue) => string
+  ): {
+    formErrors: string[];
+    fieldErrors: { [k: string]: string[] };
+  };
+  flatten<U>(
+    mapper?: (issue: ZodIssue) => U
+  ): {
+    formErrors: U[];
+    fieldErrors: { [k: string]: U[] };
+  };
+  flatten<U = string>(
     mapper: (issue: ZodIssue) => U = (issue: ZodIssue) => issue.message as any
   ): {
     formErrors: U[];
     fieldErrors: { [k: string]: U[] };
-  } => {
+  } {
     const fieldErrors: any = {};
     const formErrors: U[] = [];
     for (const sub of this.issues) {
@@ -234,7 +246,7 @@ export class ZodError<T = any> extends Error {
       }
     }
     return { formErrors, fieldErrors };
-  };
+  }
 
   get formErrors() {
     return this.flatten();
--- a/src/types.ts
+++ b/src/types.ts
@@ -49,7 +49,23 @@ export type ZodTypeAny = ZodType<any, any, any>;
 export type TypeOf<T extends ZodType<any, any, any>> = T["_output"];
 export type input<T extends ZodType<any, any, any>> = T["_input"];
 export type output<T extends ZodType<any, any, any>> = T["_output"];
-export type { TypeOf as infer };
+export type TypeOfFlattenedError<
+  T extends ZodType<any, any, any>,
+  U = string
+> = {
+  formErrors: U[];
+  fieldErrors: {
+    [P in keyof TypeOf<T>]?: U[];
+  };
+};
+export type TypeOfFormErrors<
+  T extends ZodType<any, any, any>
+> = TypeOfFlattenedError<T>;
+export type {
+  TypeOf as infer,
+  TypeOfFlattenedError as inferFlattenedErrors,
+  TypeOfFormErrors as inferFormErrors,
+};
 
 export type CustomErrorParams = Partial<util.Omit<ZodCustomIssue, "code">>;
 export interface ZodTypeDef {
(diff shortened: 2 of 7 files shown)


src/types.ts
```diff
@@ -49,7 +49,23 @@ export type ZodTypeAny = ZodType<any, any, any>;
...
+export type TypeOfFlattenedError<
+  T extends ZodType<any, any, any>,
+  U = string
+> = {
+  formErrors: U[];
+  fieldErrors: {
+    [P in keyof TypeOf<T>]?: U[];
+  };
+};
+export type TypeOfFormErrors<
```
fnky, line 61: This could be a single `z.inferFormErrors` type that just has the signature of `z.inferFlattenedError`, but I wanted to avoid confusion between results of `flatten()` vs `formErrors` getter.

What would you prefer? A single type, or is the convince alias type useful?

kentcdodds:
This would be quite helpful! I would like to use this in the example project for integrating Remix with zod.

src/ZodError.ts
```diff
@@ -217,12 +217,24 @@ export class ZodError<T = any> extends Error {
...
+    formErrors: U[];
+    fieldErrors: { [k: string]: U[] };
+  };
+  flatten<U = string>(
     mapper: (issue: ZodIssue) => U = (issue: ZodIssue) => issue.message as any
   ): {
     formErrors: U[];
     fieldErrors: { [k: string]: U[] };
-  } => {
+  } {
```
scotttrinh, line 237: I so rarely use this method syntax: are there any gotchas or considerations for changing from an arrow function bound to a property?

src/types.ts
scotttrinh: I personally appreciate the alias here since it actually gives a pretty nice hint at what is actually happening.

scotttrinh (commented):
Are there any type-level tests that we can add, maybe some [`AssertEquals`](https://github.com/colinhacks/zod/blob/73a9a628e8e3f512908b5f2b4116a87197f43407/src/helpers/util.ts#L2-L6) to the [errors test suite](https://github.com/colinhacks/zod/blob/master/src/__tests__/error.test.ts)?

src/ZodError.ts
fnky: The only difference is this case, is that it will technically allow user to change `this` context with `bind`, `call` and `apply`. But I'd say that is an edge-case.

fnky:
> Are there any type-level tests that we can add, maybe some [`AssertEquals`](https://github.com/colinhacks/zod/blob/73a9a628e8e3f512908b5f2b4116a87197f43407/src/helpers/util.ts#L2-L6) to the [errors test suite](https://github.com/colinhacks/zod/blob/master/src/__tests__/error.test.ts)?

I was thinking the same. There's a few libraries to do so; [dtslint](https://github.com/microsoft/dtslint) and [tsd](https://github.com/SamVerschueren/tsd) are the two most prominent. The latter has some [differences with dtslint](https://github.com/SamVerschueren/tsd/issues/10).

scotttrinh:
> I was thinking the same.

I just meant literally using the existing `AssertEqual` type in a test case in the errors test suite. We have a few examples around: https://github.com/colinhacks/zod/search?q=AssertEqual

fnky:
> https://github.com/colinhacks/zod/search?q=AssertEqual

Oh, sorry I totally misunderstood. I'll add some tests for this :)

fnky:
@scotttrinh I added tests for type inference and type assertion with `flatten()` as well as additional checks.

I don't know if  `all-errors.test.ts` is the right place as `safeparse.test.ts` also seemed like a good location for this, but let me know :)

src/ZodError.ts
fnky: 💯 . You're welcome. Agreed, it's always those nuances that can be hard to catch.

scotttrinh (approved):
Great tests, great implementation, thanks for your contribution!

kentcdodds:
Wonderful! Thank you!

kentcdodds:
When can we expect a release of this?

Issue #854, opened 2022-01-13
UUID regex only validate Variant 1 UUIDs


narrowei:
Hey everyone,

When I tried to use the UUID function to validate this UUID: **b3ce60f8-e8b9-40f5-1150-172ede56ff74** it returns invalid UUID, but it's a validate UUID. After digging into the source code, I found the reason.

Here is the regex we are currently using (v3.11.6) for validate uuid: 
``` javascript
const uuidRegex = /^([a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}|00000000-0000-0000-0000-000000000000)$/i;
```
However the fourth part of the regex **[89ab][a-f0-9]{3}** only validate for Variant 1 UUIDs. 
[The definition of variant from Wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier)

Not sure why the code limits the valid UUID to Variant 1 type?

JacobWeisenburger:
I see that this was merged. Can I close this?

PR #860, opened 2022-01-17
change UUID regex to support different UUID variants

narrowei:
PR for issue #854 - UUID regex only validate Variant 1 UUIDs

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -400,7 +400,7 @@ export interface ZodStringDef extends ZodTypeDef {
 }
 
 const cuidRegex = /^c[^\s-]{8,}$/i;
-const uuidRegex = /^([a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}|00000000-0000-0000-0000-000000000000)$/i;
+const uuidRegex = /^([a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[a-f0-9]{4}-[a-f0-9]{12}|00000000-0000-0000-0000-000000000000)$/i;
 // from https://stackoverflow.com/a/46181/1550155
 // old version: too slow, didn't support unicode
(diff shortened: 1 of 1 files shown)


scotttrinh:
@narrowei 

Do you mind adding a few tests for the various UUID versions to the string tests?

https://github.com/colinhacks/zod/blob/73a9a628e8e3f512908b5f2b4116a87197f43407/src/__tests__/string.test.ts#L80-L89

narrowei:
@scotttrinh  No problem! will add unit tests for different various

narrowei:
@scotttrinh have added tests to cover variants 0 and 2. Other files were automatically changed by the pre-commit lint fix.

scotttrinh:
Thanks for the test cases, this is looking good! I'll spend some time in the next day or two to get it ready for merging. Not sure what the release schedule is going to be, but I don't think there is anything blocking this from going out in the next release. Thanks for getting this together!

> Other files were automatically changed by the pre-commit lint fix.

I think this might be due to some issue with prettier picking up your own configuration or using some global config? I can't see why it would've changed code that has ostensibly been sitting in the upstream branch all this time and wasn't changed. We can probably sort this out during the merge, but if you have time to remove any of the extraneous changes sooner than we do, feel free to push another commit.

narrowei:
Hey @scotttrinh, thanks for reviewing my code and quick response! I have removed extraneous changes and pulled the code from the master branch.

src/__tests__/string.test.ts
```diff
@@ -81,6 +81,8 @@ test("uuid", () => {
   const uuid = z.string().uuid("custom error");
   uuid.parse("9491d710-3185-4e06-bea0-6a2f275345e0");
   uuid.parse("00000000-0000-0000-0000-000000000000");
+  uuid.parse("b3ce60f8-e8b9-40f5-1150-172ede56ff74"); // Variant 0 - RFC 4122: Reserved, NCS backward compatibility
+  uuid.parse("92e76bf9-28b3-4730-cd7f-cb6bc51f8c09"); // Variant 2 - RFC 4122: Reserved, Microsoft Corporation backward compatibility
```
scotttrinh, line 85: Forgive my ignorance (I typically only use v4 UUIDs), but do we care about "variants" here or "versions" as specified in RFC 4122? I think we want to be able to validate v1 - v5 UUIDs regardless of the "variant", but I'm not entirely sure I understand what a variant is supposed to be.

scotttrinh: Looking a little deeper, I think I just misunderstood your initial requirement here, so if it is indeed the ability to use these variants, and the new RegExp properly handles any _version_ I think this is fine, I just want to get some clarity for myself.

narrowei: No worries, most of the UUIDs nowadays are variant 1 however some old systems still generate variant 0 or 2 UUIDs, but according to RFC 4122, they are all valid UUIDs. I only changed the fourth part of the regex which is for variant, it won't affect the version validation (v1 to v5) on the third part of the regex.

narrowei:
Hey @scotttrinh, any luck about the merge?

scotttrinh (approved):
Sorry for the delay in merging, this looks good to me!

PR #864, opened 2022-01-18
fix: publish package with `src`

### LINKED ISSUE #828: Failed to parse source map
Using zod: 3.11.6

We are having warning in `index.mjs`
```
WARNING in ./node_modules/zod/lib/index.mjs
Module Warning (from ./node_modules/source-map-loader/dist/cjs.js):
Failed to parse source map from 'C:\repos\Git\WorkZone\WZ-Teams\Source\tabs\node_modules\zod\src\types.ts' file: Error: ENOENT: no such file or directory, open 'C:\repos\Git\WorkZone\WZ-Teams\Source\tabs\node_modules\zod\src\types.ts'
```

apparently the referenced `src\types.ts` is not shipped with package, leading to such error.

Luke-zhang-04:
- fixes #828

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "./package.json": "./package.json"
   },
   "files": [
-    "/lib"
+    "/lib",
+    "/src"
   ],
   "repository": {
     "type": "git",


JohanMouchet (approved):

tpetrina:
Hi, any chance of this going in release?

colinhacks:
For the moment I just removed source maps. I don't like the idea of shipping TS files in the bundle - it can break other builds and causes bloat for everyone else (which is the whole problem minimization + source maps are intended to combat). Other major projects (`io-ts`) also don't bother with source maps, and I'm starting to think it's more trouble than it's worth.

Luke-zhang-04:
If removing source maps is the desired solution I can close this PR. As you said, source maps really don't have much benefit in this case.

Source maps may be beneficial for debugging, and if that's something we want then we should ship the source so the source maps actually work. I fail to see how it would break any builds, since the built JS files will still be there, and I wouldn't really say it's bloat since it's just a couple of files, especially if you don't include the test and benchmark folders.

PR #872, opened 2022-01-19
Return invalid_union on dirty results

### LINKED ISSUE #866: Wrong error when using a union with strict objects
I have two object types with one key each (with different names), and I want to parse an object and verify that it has one and only one of those keys. To do that, I created a union schema with two strict objects:

```ts
z.union([
  z.object({ a: z.number() }).strict(),
  z.object({ b: z.number() }).strict()
])
```

If I parse `{ a: 1 }` or `{ b: 1 }` it validates successfully as expected, and if I parse `{ c: 1 }` it gives an `invalid_union` error with the issue from both union elements in `unionErrors` as expected.

However if I parse `{ a: 1, b: 1 }` I don't get an `invalid_union` error. I instead get an `unrecognized_keys` error, i.e. the same error you would get if you parsed this directly with `z.object({ a: z.number() }).strict()`. I also see that in other to change this error message, I have to add an `errorMap` to the object schema, rather than the union schema. Since none of the object schemas match the input, I would expect to get an `invalid_union` here as well.

This makes it hard to get a useful error message for this use case.

It seems like what's happening is that when checking if any of the union schemas is valid, it doesn't consider them to be strict, then selects the first since it matches when it's not strict, and then applies strict afterwards?

trygveaa:
When one or more of the results in the union is dirty the union would
previously just return that result. However this can make the error
message strange and inconvenient to override in some cases, drops the
rest of the dirty results if there are multiple, and hides the fact that
it's a union.

Therefore, it's changed to return an invalid_union issue with all the
dirty results. When there are some dirty results, the aborted results
are not included, since they are not relevant. When there are no dirty
results, all the results are included as before.

An example of an error message that's strange and inconvenient to
override is if you have a union of two strict objects, e.g.:

```ts
z.union([
    z.object({ a: z.number() }).strict(),
    z.object({ b: z.number() }).strict()
])
```

If the input is `{ a: 1, b: 1 }`, it would previously return the result
from the first object, which would be code unrecognized_keys with keys
`b`. If you wanted to override the error message, you would have to do
that in the first object, it wouldn't suffice to override the message in
union.

With this patch, the result would be code invalid_union with the two
unrecognized_keys in the unionErrors array. Since the error is now
always invalid_union it suffices to override the error message in union
for both this case when both keys are provided, and for the case when
none of the union schemas match.

Fixes #866

The PR also includes one more commit:

> Mark union as dirty when the results are dirty
> 
> When one or more of the results of the schemas in the union are dirty, I
> think it makes sense to mark the union itself as dirty too.

### DIFF
--- a/deno/lib/__tests__/unions.test.ts
+++ b/deno/lib/__tests__/unions.test.ts
@@ -102,6 +102,74 @@ test("return all dirty results over aborted", () => {
   }
 });
 
+test("don't continue parsing on aborted result", () => {
+  const result = z
+    .union([z.number(), z.boolean()])
+    .refine(() => false)
+    .safeParse("a");
+  expect(result.success).toEqual(false);
+  if (!result.success) {
+    expect(result.error.issues).toEqual([
+      {
+        code: "invalid_union",
+        message: "Invalid input",
+        path: [],
+        unionErrors: [
+          new ZodError([
+            {
+              code: "invalid_type",
+              expected: "number",
+              received: "string",
+              path: [],
+              message: "Expected number, received string",
+            },
+          ]),
+          new ZodError([
+            {
+              code: "invalid_type",
+              expected: "boolean",
+              received: "string",
+              path: [],
+              message: "Expected boolean, received string",
+            },
+          ]),
+        ],
+      },
+    ]);
+  }
+});
+
+test("continue parsing on dirty result", () => {
+  const result = z
+    .union([z.number(), z.string().refine(() => false)])
+    .refine(() => false)
+    .safeParse("a");
+  expect(result.success).toEqual(false);
+  if (!result.success) {
+    expect(result.error.issues).toEqual([
+      {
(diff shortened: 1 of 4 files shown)


trygveaa:
Uh, please don't close it. Could someone review it please? I can resolve the conflicts that has occurred if someone responds.

trygveaa:
Still waiting for feedback.

trygveaa:
...

brupxxxlgroup:
Why is it closed?

trygveaa:
It never got any review or feedback and there's a bot closing PRs without any activity. I gave up trying to keep it open in the end.

IMO having a bot that automatically closes PRs that haven't been reviewed is pretty hostile.

trygveaa:
I see the open PRs now that are older than 3 months haven't been closed though, so maybe the bot was removed. Unfortunately, I don't have permission to reopen the PR.

PR #881, opened 2022-01-24
added z.date documentation to README.md. fixes #880

### LINKED ISSUE #880: z.date documentation needed
### Discussed in https://github.com/colinhacks/zod/discussions/879

<sup>Originally posted by **blocka** January 24, 2022</sup>
Very surprised that I could not find anything on this.

If I have a date schema `const schema = z.date()`  and I pass it a date string `schema.parse('2022-01-12T00:00:00.000Z')`
I would expect to get back a JS Date object. Instead I get an exception that I've passed a string instead of a date. ...</div>

There really should be some documentation in the README about how z.date works.

JacobWeisenburger:


JacobWeisenburger:
This would fix #880

Issue #604, opened 2021-08-23
Feature: String starts/ends with 


alii:
I was writing a validator for a hex color code and thought that being able to check that it started with a `#` would be a nice addition.

E.g.: `z.string().length(7).startsWith('#')`

DASPRiD:
You can already do that:

```ts
z.string().length(7).regex(/^#/);
```

agustif:
An updated version if you need supporting 3 and 8 digit hex too

  `color: z.string().min(4).max(9).regex(/^#/),`
This allows for
#FFF (White, simplified, it just becomes #FFFFFF)
and #00000080 (Black, with 80% transparency)

in case helps anyone looking for this

PR #893, opened 2022-01-28
Hex validator for Zod

CaptainObi:
This is just a simple hex validator for Zod. The idea was floated in #604 although this isn't exactly what the feature requested. If this isn't needed that's fine, I will write something that directly implements the feature request. Sorry for the crazy diff, I believe that's the prettier linter at work.


scotttrinh:
> If this isn't needed that's fine, I will write something that directly implements the feature request

Thanks for the PR @CaptainObi ! As you noted, I believe the change that @colinhacks was advocating for there was the `startsWith` method. I think having a "is CSS style RGB hex string" might be a little _too_ specific to warrant inclusion into the core library, but I tend to be pretty conservative when it comes to the core.

> Sorry for the crazy diff, I believe that's the prettier linter at work.

Yeah, I believe it's probably your editor doing some auto formatting with prettier, so if it's possible to prevent auto-formatting on save and update the PR, that would be helpful. Otherwise we're likely to just open a separate PR with your changes implemented.

CaptainObi:
It seems that the formatting that is going on is due to the husky hooks. How can I bypass those?

scotttrinh:
> It seems that the formatting that is going on is due to the husky hooks. How can I bypass those?

I _think_ it's likely picking up some kind of configuration of yours somewhere, which is why I suspect an editor issue rather than a husky precommit hook. You can bypass any hooks when committing with a [`--no-verify`](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---no-verify) flag to see if that helps.

agustif:
This one would be nice to get! Let me know if I can help with anything about shipping this PR

Issue #885, opened 2022-01-25
No way to validate NaN


cyberia-ng:
- `z.number()` does not accept NaN (arguably correct behaviour)
- `z.union([z.number(), z.literal(NaN)])` does not accept NaN (presumably because under the hood `.literal` uses `===`)

I saw somebody in #512 mention `z.nan()` but this does not seem to exist (yet?)

My current workaround:
```ts
export const numberWithNaN = z.custom<number>(
    (val) => typeof val === 'number',
    (val) => ({ message: `Expected number (including NaN), received ${inspect(val)}` }),
);
```

Proposal: add `z.nan()`, or make `z.literal` use `Object.is` for equality under the hood.

noyan-alimov:
Hi @cyberia-ng

I created a PR where I added `z.nan()` (here)[https://github.com/colinhacks/zod/pull/898]

linkb15:
any updates on this? PR got conflicts too

PR #898, opened 2022-01-30
NaN type

noyan-alimov:
added NaN type


noyan-alimov:
This PR addresses [this issue](https://github.com/colinhacks/zod/issues/885)

Issue #792, opened 2021-11-26
What is a good strategy to get a nice (nicer) error for invalid discriminated unions


ottokruse:
First of all: thanks for creating `zod` it is an awesome library.

Consider the following schema (contrived examples of course), and test. When you run this the error message is really long and hard for humans to grok. Note the schema has a discriminated union (the field `type` on each object is the discriminating field):

```typescript
import { z } from "zod";

const typeA = z.object({
    type: z.literal("A"),
    prop1: z.string() 
})

const typeB = z.object({
    type: z.literal("B"),
    prop2: z.string() 
})

const typeC = z.object({
    type: z.literal("C"),
    prop3: z.string() 
})

const typeD = z.object({
    type: z.literal("D"),
    prop4: z.string() 
})

const schema = z.object({ mytypes: z.union([typeA, typeB, typeC, typeD]).array() });

schema.parse({
    mytypes: [ 
        { type: "A", prop1: "foo" },
        { type: "D", bar: "BAR" } // misses prop4
    ]
});

```

<details>
  <summary>The long and hard to read error message</summary>
  
```
ZodError: [
  {
    "code": "invalid_union",
    "unionErrors": [
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "A",
            "received": "D",
            "path": [
              "mytypes",
              1,
              "type"
            ],
            "message": "Expected A, received D"
          },
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "undefined",
            "path": [
              "mytypes",
              1,
              "prop1"
            ],
            "message": "Required"
          }
        ],
        "name": "ZodError"
      },
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "B",
            "received": "D",
            "path": [
              "mytypes",
              1,
              "type"
            ],
            "message": "Expected B, received D"
          },
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "undefined",
            "path": [
              "mytypes",
              1,
              "prop2"
            ],
            "message": "Required"
          }
        ],
        "name": "ZodError"
      },
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "C",
            "received": "D",
            "path": [
              "mytypes",
              1,
              "type"
            ],
            "message": "Expected C, received D"
          },
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "undefined",
            "path": [
              "mytypes",
              1,
              "prop3"
            ],
            "message": "Required"
[...]

PR #899, opened 2022-01-30
Discriminated union

### LINKED ISSUE #894: Add `discriminatedUnion` schema type
Our current implementation naively goes through and tests each member in a union and reports back all errors. This is both noisy and slower for cases where the union is a discriminated union.

Adding this issue to discuss how best to tackle adding a new schema type for discriminated unions.

### Async?

How do we support an `async` version of the schema?

### Requiring schemas to be objects with the discriminate as a key

I _think_ there should be a way to do this, but I think it needs to be absolutely type-level constrained that you can't add a schema to the union that wouldn't contain the discriminant, so it needs to exclude non-objects and objects who's `shape` does not include the discriminant.

alexxander:
Resolves #894.

I implemented the `z.discriminatedUnion(discriminator, options)` method and aliased it by `z.union(discriminator, options)`. I also added tests and benchmarks.

Here are the results of the benchmarks to demonstrate the speed improvements introduced by the new `z.discriminatedUnion` method:
```
z.union: double: valid: a x 310,690 ops/sec ±3.68% (98 runs sampled)
z.union: double: valid: b x 317,585 ops/sec ±0.13% (99 runs sampled)
z.union: double: invalid: null x 52,452 ops/sec ±0.69% (94 runs sampled)
z.union: double: invalid: wrong shape x 46,605 ops/sec ±0.43% (94 runs sampled)
z.union: many: valid: a x 151,210 ops/sec ±3.75% (95 runs sampled)
z.union: many: valid: c x 155,035 ops/sec ±0.21% (97 runs sampled)
z.union: many: invalid: null x 31,971 ops/sec ±0.53% (93 runs sampled)
z.union: many: invalid: wrong shape x 28,777 ops/sec ±0.42% (95 runs sampled)

z.discriminatedUnion: double: valid: a x 857,014 ops/sec ±3.61% (99 runs sampled)
z.discriminatedUnion: double: valid: b x 876,301 ops/sec ±0.15% (95 runs sampled)
z.discriminatedUnion: double: invalid: null x 130,574 ops/sec ±0.55% (98 runs sampled)
z.discriminatedUnion: double: invalid: wrong shape x 124,104 ops/sec ±3.36% (98 runs sampled)
z.discriminatedUnion: many: valid: a x 859,193 ops/sec ±0.28% (94 runs sampled)
z.discriminatedUnion: many: valid: c x 858,599 ops/sec ±0.30% (94 runs sampled)
z.discriminatedUnion: many: invalid: null x 124,720 ops/sec ±0.36% (96 runs sampled)
z.discriminatedUnion: many: invalid: wrong shape x 117,588 ops/sec ±3.44% (93 runs sampled)

z.string: empty string x 3,263,033 ops/sec ±1.31% (94 runs sampled)
z.string: short string x 3,254,089 ops/sec ±0.82% (93 runs sampled)
z.string: long string x 3,249,335 ops/sec ±0.73% (93 runs sampled)
z.string: invalid: null x 138,359 ops/sec ±0.58% (95 runs sampled)
z.string: manual parser: long x 883,218,417 ops/sec ±0.31% (92 runs sampled)
z.object: empty: valid x 2,477,603 ops/sec ±0.18% (97 runs sampled)
z.object: empty: valid: extra keys x 2,291,072 ops/sec ±0.12% (95 runs sampled)
z.object: empty: invalid: null x 142,172 ops/sec ±0.54% (97 runs sampled)
z.object: short: valid x 1,139,297 ops/sec ±3.54% (99 runs sampled)
z.object: short: valid: extra keys x 1,086,536 ops/sec ±0.30% (99 runs sampled)
z.object: short: invalid: null x 141,700 ops/sec ±0.56% (96 runs sampled)
z.object: long: valid x 481,833 ops/sec ±3.78% (91 runs sampled)
z.object: long: valid: extra keys x 469,050 ops/sec ±0.24% (100 runs sampled)
z.object: long: invalid: null x 145,378 ops/sec ±0.57% (97 runs sampled)
Done in 165.07s.
```


CantelopePeel:
Thanks @alexxander for this PR. The team I work with would have an immediate use for this as the current errors for unions can leave something to be desired. I am not sure what the current approval process for this project is but I would be happy to do an initial review before @colinhacks.

EdsterG:
@alexxander this PR is a game changer and very useful. Thank you for the good work. While testing I ran into a case where the error was incorrect if the value of the discriminator didn't match anything in the union. I would expect the error to be something along the lines of: "Received invalid value 'IncorrectType' for discriminator 'type'. Expecting value to be one of [ValidType1, ValidType2, ...]".

alexxander:
Hi @EdsterG, firstly, thanks for the compliments on my work, I am glad you find this useful.

However, I did try to handle that case you mentioned and I am failing to reproduce the incorrect error. Though it might be that something slipped through my fingers. I actually even created a new zod issue type for this particular purpose. See the test case bellow:

```ts
test("invalid discriminator value", () => {
    expect.assertions(1);
    try {
      z.discriminatedUnion("type", [
        z.object({ type: z.literal("a"), a: z.string() }),
        z.object({ type: z.literal("b"), b: z.string() }),
      ]).parse({ type: "x", a: "abc" });
    } catch (e: any) {
      expect(JSON.parse(e.message)).toEqual([
        {
          code: z.ZodIssueCode.invalid_union_discriminator,
          expectedOneOf: ["a", "b"],
          received: "x",
          message:
            "Invalid discriminator value. Expected one of: a, b. Received x.",
          path: ["type"],
        },
      ]);
    }
  });
```

I tried to pass other values in the `type` property - such as `null`, an object, `undefined`. I even tried leaving the property out. And in all the cases it seemed to work. Is there anything I am missing?

alexxander:
Hi @CantelopePeel, before this PR, I also implemented a self-standing function to handle discriminated unions with zod. I guess you might be interested in that as well: https://github.com/colinhacks/zod/issues/792#issuecomment-1023251415.

EdsterG:
@alexxander yes that was my mistake. I was getting a very confusing error message and accidentally thought it was related to incorrect discriminator value. Your code correctly reports when an invalid discriminator is passed in. On that same note, there seems to be a previous issue to remove input values from the error messages. Might make sense to follow the same pattern and avoid having the input value as part of the error message. https://github.com/colinhacks/zod/issues/461

scotttrinh:
> On that same note, there seems to be a previous issue to remove input values from the error messages. Might make sense to follow the same pattern and avoid having the input value as part of the error message

Yeah, it has been our practice to _not_ leak input or schema data in error messaging to avoid logging sensitive or personally-identifying information to logs/etc. Applications _should_ have the context at the call-site to report that sort of data themselves if they find it useful.

src/types.ts
```diff
@@ -1893,18 +1893,179 @@ export class ZodUnion<T extends ZodUnionOptions> extends ZodType<
     return this._def.options;
   }
 
-  static create = <
-    T extends Readonly<[ZodTypeAny, ZodTypeAny, ...ZodTypeAny[]]>
+  static create<
```
scotttrinh, line 1896: I personally find overloading like this to be a bit of a headache when trying to understand, maintain, and debug things. Additionally, I think it's a bit of a big change to introduce a `ZodType` constructor that actually might return one of two actual `ZodType`s.

Unless @colinhacks or the other maintainers disagree with me on this, I think we can leave out the aliasing for now and (maybe) introduce it later. At the very least, you could cherrypick this change into a separate PR and we can discuss the merits of it without holding up the overall progress on `discriminatedUnion`. Does that seem reasonable?

scotttrinh (changes requested):
Fantastic work here! Tests and implementation look great, documentation is well-done, and the style is consistent with what we have in general. 👏 

One small request to move the aliasing to `union` to a separate PR so we don't block merging this while we discuss the merits of the alias.

src/ZodError.ts
```diff
@@ -290,6 +298,11 @@ export const defaultErrorMap = (
     case ZodIssueCode.invalid_union:
       message = `Invalid input`;
       break;
+    case ZodIssueCode.invalid_union_discriminator:
+      message = `Invalid discriminator value. Expected one of: ${issue.expectedOneOf.join(
+        ", "
+      )}. Received ${issue.received}.`;
```
scotttrinh, line 304: As mentioned, I think we want to avoid revealing the input data here. Following the existing example from `invalid_enum_value`:

```suggestion
      message = `Invalid discriminator value. Expected one of: ${issue.expectedOneOf.join(
        ", "
      )}`;
```

alexxander: I removed it from the error message and the issue object. I also updated the tests.

alexxander: And as you mentioned `invalid_enum_value`, I made the `invalid_discriminator_value` issue even more similar to it.

src/types.ts
alexxander: removed

alexxander:
Previously, the only supported discriminator value type was `string`. In the last commit, I added support for other primitive types supported by the `z.literal` function. The drawback is that I needed to use the `Map` data type. Should this be a problem, I can implement it using a separate object for each type:
```js
const map = {
  'string': {},
  'number': {},
  'bigint': {},
  'boolean': {},
  'object': {},
  'undefined': {}
};
map[typeof discriminatorValue][discriminatorValue] = schema;
```

alexxander:
Also, there are the failing _deno_ tests. I guess I have to modify the tests to be compatible with both _jest_ and _deno_? Any hints @scotttrinh?

How about using something like https://github.com/gcaptn/tincan? (Though this might be for a different issue.)

src/helpers/typeAliases.ts
```diff
@@ -0,0 +1,2 @@
+export type Primitive = string | number | bigint | boolean | null | undefined;
+export type Scalars = Primitive | Primitive[];
```
EdsterG, line 2: Is `Scalars` actually use in this PR?

scotttrinh:
> The drawback is that I needed to use the Map data type.

Glad you called that out. Hmm. We don't have a strong platform support story, but if it's easy enough to implement in the nested way you've outlined then that's probably the safest bet here without wanting a major version bump and a line-in-the-sand policy about platform support.

> Also, there are the failing deno tests. I guess I have to modify the tests to be compatible with both jest and deno?

Yeah, we are trying to be friendly to the Deno community as much as is feasible. I can take a look at the failing tests, but I have 0 experience with Deno so won't be too much specific help.

src/helpers/typeAliases.ts
alexxander: Nope, and I do not actually know whether it is even used anywhere in the repo, is it?

I moved it to the typeAliases file, because it is very closely related to the Primitive type (which I needed to move). Also it was defined right under it.

deno/lib/__tests__/discriminatedUnions.test.ts
```diff
@@ -0,0 +1,145 @@
+import { expect } from "https://deno.land/x/expect@v0.2.6/mod.ts";
+const test = Deno.test;
+
+import * as z from "../index.ts";
+
+describe("discriminated union", () => {
```
EdsterG, line 6: Looking at the other deno tests, none of them are wrapped in a `describe` block, removing it here should fix the deno tests.

EdsterG:
Would it be possible to make this support lazy schemas? Otherwise it's not possible to use with discriminated recursive schemas.

EdsterG:
@alexxander that would be super helpful, I know there's also a `z.late.object` methods that's implemented but undocumented. It's referenced here https://github.com/colinhacks/tozod but I'm not familiar with the inner workings of Zod so am having trouble getting it to work.

PabloDinella:
Just would like to say I'm also looking forward to this PR, more human errors will be a very good improvement.

scotttrinh:
> I know there's also a z.late.object methods that's implemented but undocumented.

I believe that is a 2.0 construct. As far as I can tell in the source code, there is not `late` class.

I'm actually wondering if we should put off supporting `lazy` here? @alexxander if you dig in a bit and run into any roadblocks, I think we should consider shipping what you have here and breaking off the further work into a separate PR.

deno/lib/__tests__/discriminatedUnions.test.ts
alexxander: Yeah, that was one thing... But I also had to get rid of the other jest functions. I simplified the tests so that they use only the test wrappers and the `expect(...).toEqual(...)` construct.

alexxander:
> Would it be possible to make this support lazy schemas?
@EdsterG I tried some simple tests with `z.lazy`, and it seemed to work well. Did you find any particular issues?

EdsterG:
@alexxander here's a toy example where I get type errors:
```
import {z} from 'zod';

type Category1 = {
  type: 'Category1';
  subcategories1: Array<Category>;
};

type Category2 = {
  type: 'Category2';
  subcategories2: Array<Category>;
};

const Category1Schema: z.ZodSchema<Category1> = z.object({
  type: z.literal('Category1'),
  subcategories1: z.array(z.lazy(() => CategorySchema)),
});

const Category2Schema: z.ZodSchema<Category2> = z.object({
  type: z.literal('Category2'),
  subcategories2: z.array(z.lazy(() => CategorySchema)),
});

const CategorySchema = z.discriminatedUnion('type', [Category1Schema, Category2Schema]);
type Category = z.infer<typeof CategorySchema>;
```
I get the following type error:
```
error TS2740: Type 'ZodType<Category1, ZodTypeDef, Category1>' is missing the following properties from type 'ZodObject<{ type: ZodLiteral<Primitive>; } & ZodRawShape, any, any, { [x: string]: any; [x: number]: any; }, { [x: string]: any; [x: number]: any; }>': _shape, _unknownKeys, _catchall, _cached, and 16 more.

const CategorySchema = z.discriminatedUnion('type', [Category1Schema, Category2Schema]);
```

alexxander:
@EdsterG This fails because `z.discriminatedUnion` expects instances of `ZodObject`. You, however, typed `Category1Schema` and `Category2Schema`  as `ZodSchema`, which is an alias of `ZodType`.

Nonetheless, it is still possible to achieve what you want by slightly reorganizing the types.

```ts
type Category1 = {
  type: "Category1";
  subcategories1: Array<Category>;
};

type Category2 = {
  type: "Category2";
  subcategories2: Array<Category>;
};

type Category = Category1 | Category2;

const Category1Schema = z.object({
  type: z.literal("Category1"),
  subcategories1: z.array(z.lazy(() => CategorySchema)),
});

const Category2Schema = z.object({
  type: z.literal("Category2"),
  subcategories2: z.array(z.lazy(() => CategorySchema)),
});

const CategorySchema: z.ZodSchema<Category> = z.discriminatedUnion("type", [
  Category1Schema,
  Category2Schema,
]);
```
This actually even matches the example from the docs. Therefore, I do not really think this is an issue. Any views on this @scotttrinh?

EdsterG:
Ah, I see. The idea was that many types of categories would be defined through out the lifetime of the project so I wanted to avoid maintaining two sources of truth:
```
type Category = Category1 | Category2;
```
and
```
const CategorySchema: z.ZodSchema<Category> = z.discriminatedUnion("type", [
  Category1Schema,
  Category2Schema,
]);
```

However, I would agree that this change should go into a separate PR. In my opinion, this PR should be merged as is because its impact is too great to wait any longer.

alexxander:
> ```
> import {z} from 'zod';
> 
> type Category1 = {
>   type: 'Category1';
>   subcategories1: Array<Category>;
> };
> 
> type Category2 = {
>   type: 'Category2';
>   subcategories2: Array<Category>;
> };
> 
> const Category1Schema: z.ZodSchema<Category1> = z.object({
>   type: z.literal('Category1'),
>   subcategories1: z.array(z.lazy(() => CategorySchema)),
> });
> 
> const Category2Schema: z.ZodSchema<Category2> = z.object({
>   type: z.literal('Category2'),
>   subcategories2: z.array(z.lazy(() => CategorySchema)),
> });
> 
> const CategorySchema = z.discriminatedUnion('type', [Category1Schema, Category2Schema]);
> type Category = z.infer<typeof CategorySchema>;
> ```
> 
> I get the following type error:
> 
> ```

@EdsterG, actually, even if `z.discriminatedUnion` accepted anything object-like, typescript would still throw recursion-related errors with your solution quoted above. The thing is that `Category` is inferred from `CategorySchema`, which refers to `Category1Schema` of type `Category1`. However, `Category1` refers to `Category`.

alexxander:
@scotttrinh I added the simplified implementation of the map data structure as it was outlined in one of my previous comments (`util.SimpleMap`). It implements `get(key)`, `set(key, value)`, `has(key)`, `keys()` and the `size` property. Though I am not completely happy about the additional complexity it introduces (I mean, `Map` is well documented in the standards, while this one is custom).

scotttrinh:
@alexxander 

Looking at https://caniuse.com/mdn-javascript_builtins_map I'm personally fine with IE11+ here, so `Map` still seems like a reasonable choice. We use `const`/`let` which has even worse support than `Map` (TIL!).

@EdsterG 

RE: The type errors related to `ZodSchema`/`ZodType`

We shouldn't do too much work to support explicit types like the examples given here. Zod takes a pretty opinionated and strong stance that it should be the source of truth for types and does a best-effort job of supporting defining the types in TypeScript first. The big place of friction with this is external types that you do not control, so it's definitely ideal for it to work well if possible, but we should not spend too much development effort trying to get it to work. If `z.infer`/`z.input`/`z.output` is able to infer the correct type, that's our primary concern.

alexxander:
@scotttrinh So I should revert it to the state in which `Map` is used, right?
Is there anything else that has to be done (or that has to be changed) before this can be merged?

src/types.ts
```diff
@@ -1907,6 +1905,150 @@ export class ZodUnion<T extends ZodUnionOptions> extends ZodType<
...
+  >(
+    discriminator: Discriminator,
+    types: Types,
+    params?: RawCreateParams
+  ): ZodDiscriminatedUnion<Discriminator, DiscriminatorValue, Types[number]> {
+    // Get all the valid discriminator values
+    const options: util.SimpleMap<
+      DiscriminatorValue,
+      Types[number]
+    > = new util.SimpleMap();
```
scotttrinh, line 2021: Let's revert this to using a native `Map` here which has good support (IE11+).

scotttrinh (changes requested):

scotttrinh:
> @scotttrinh So I should revert it to the state in which Map is used, right?
> Is there anything else that has to be done (or that has to be changed) before this can be merged?

Just went back through the code, and it looks very good to me, thank you for all of your hard work here! Once you get it back to using a `Map`, do you mind just running the benchmarks again and updating the description if they're different at all from the first time you ran it?

src/types.ts
alexxander: Reverted

alexxander:
@scotttrinh I reverted it. So now it uses `Map` again. I also ran the benchmarks and updated the pull request description.

scotttrinh (approved):
Amazing work, and thanks so much for the valuable contribution and patience @alexxander ! 🎉

flybayer:
Whoa, amazing! This is just what we needed today. But looks like this isn't published to npm yet? Any way we can get a beta published?

src/types.ts
```diff
@@ -1907,6 +1905,147 @@ export class ZodUnion<T extends ZodUnionOptions> extends ZodType<
...
 
+/////////////////////////////////////////////////////
+/////////////////////////////////////////////////////
+//////////                                 //////////
+//////////      ZodDiscriminatedUnion      //////////
+//////////                                 //////////
+/////////////////////////////////////////////////////
+/////////////////////////////////////////////////////
+
+type ZodDiscriminatedUnionOption<
```
luixo, line 1916: I believe this type should be exported, as all of the types in the package are.
Should I open a PR?
@alexxander

PR #999, opened 2022-03-06
Add URN support

stoopkidddd:
[URNs](https://en.wikipedia.org/wiki/Uniform_Resource_Name) are still used in a lot of places despite being sort of deprecated for URIs. One place that comes to mind is AWS.

I added support for a `z.string().urn()` that can be check for validity of any URN.

A basic example:

```ts
const urn = z.string().urn();
urn.parse('urn:isbn:0451450523');
```

I also added custom parser support, so that you can create schemas for the URNs themselves, instead of them all just being a "string". Example code below:

```ts
import { createUrnUtil } from 'urn-lib';
const arnParser = createUrnUtil('arn', {
  components: [ // protocol is automatically added (protocol = urn or arn or whatever)
    'partition',
    'service',
    'region',
    'account-id',
    'resource', // if more:separations:exist after this, they are handled properly
  ],
  separator: ':',
  allowEmpty: true, // arn does stuff like arn:::s3 and stuff
});

const str = 'arn:aws:autoscaling:us-east-1:123456789012:scalingPolicy:c7a27f55-d35e-4153-b044-8ca9155fc467:autoScalingGroupName/my-test-asg1:policyName/my-scaleout-policy';

const arn = z.string().urn('custom error', arnParser);

arn.parse(str);
```

I pulled in [urn-lib](https://github.com/cmawhorter/urn-lib) to do this. Not sure if it is going to be okay with the maintainers to have an outside dependency. A regex could sort of suffice for the most basic of URNs, but I could not find one that covers all scenarios, and it also doesn't give you the ability to create custom parsers for your own URN namespace definitions.


stoopkidddd:
Going to close this PR as zod does not want to introduce additional code and increase the package size. For anybody who might find this, an alternate approach is to use superRefine(). An example is below

```typescript
import { createUrnUtil } from 'urn-lib';
const arnParser = createUrnUtil('arn', {
  components: [ // protocol is automatically added (protocol = urn or arn or whatever)
    'partition',
    'service',
    'region',
    'account-id',
    'resource', // if more:separations:exist after this, they are handled properly
  ],
  separator: ':',
  allowEmpty: true, // arn does stuff like arn:::s3 and stuff
});

const UrnValidator: (errMessage: string) => z.SuperRefinement<string> =
  (errMessage) => (val, ctx) => {
    const parsedUrn = arnParser.parse(val);
    const validationErrors = !parsedUrn
      ? "Invalid URN"
      : arnParser.validate(parsedUrn);
    ctx.addIssue({
      code: z.ZodIssueCode.custom,
      message: errMessage || validationErrors?.toString(),
    });
  };

const schema = z.string().superRefine(UrnValidator("Invalid URN"));
```

PR #1002, opened 2022-03-08
Add ZodDiscriminatedUnion to ZodFirstPartySchemaTypes

jacobmischka:
Is there a reason why `ZodDiscriminatedUnion`, which is a `ZodFirstPartyTypeKind`, isn't in the `ZodFirstPartySchemaTypes` union type? It seems like an oversight?


colinhacks:
Indeed. Thanks for the PR!

PR #1003, opened 2022-03-09
Add react-zorm to "Form integrations" section

esamattis:
It would be really cool have this mentioned here 🙏


colinhacks:
Happy to add! Great work with `react-zorm`, it's phenomenal 🤙

Issue #814, opened 2021-12-07
Add file type


jgradzki:
Hi, as stated in contributing documentation, Im creating this issue to propose implementing "ZodFile" type. 

I have already implemented this in my project (I had to overwrite some types etc), and I could make PR with this.
In my current implementation, I have checks for size(min/max) and mime type, I'm open to suggestions to add something more.

JacobWeisenburger:
PRs are always welcome. Please explain more about this use case and provide some code examples show how it could be used. Is this something that would only work with node/deno or could this work for the browser too?

andresgutgon:
I'm also interested in solving this issue 👍

andresgutgon:
I would love an API like this for the `file` Zod type:

``` javascript
const avatar = z.file().type(["jpg", "png"]).size(1048576) // 1MB in bytes
```

I'm trying to do my own `ZodType` like `ZodString` is done for example.

But I'm not sure if there is another way. The way I'm trying looks super complex. Also Typescript is barking me all the time becasue `ZodType` only allow [core types](https://github.com/colinhacks/zod/blob/master/src/helpers/parseUtil.ts#L10-L31)

As far as I know a file to be validated need to stream their content in chunks so you can check if the size is not bigger than expected in `z.file().size(XXX)`. I was thinking that the input for this file would be an object with 3 keys:

``` javascript
const avatar = z.file().type(["jpg", "png"]).size(1048576) // 1MB in bytes

const result = await avatar.safeParseAsync({
  filestream: ReadableStream, // with this internally we create a NodeFIle
  filename: "my-avatar.jpg", // Necessary for creating the file
  mimeType: "image/jpg" // This is the data that is validated agains what developer wants.
})

The output would be a [File](https://developer.mozilla.org/en-US/docs/Web/API/File) (or NodeFile equivalent).
```

jgradzki:
@andresgutgon I'm working on it now, I should make PR very soon. I made min/max instead of `size`, just like it is made in NumberType

andresgutgon:
Very cool! But what would be the flow with this validation?

Should I have already the file created in my server in disk? And then check for `mimetype` and `size`?

What I was think was different `input` and `output` and inside the `FileType` the actual `file` is created and returned. 
This process would be always `async`.

I'm thinking in this because my use case is validate a request on a node server. 

1. I want to validate all my `fields` in the request
2. I want to validate all my `files` in the request.

[...]

PR #1017, opened 2022-03-14
feat: implement File type

jgradzki:
#814

```js
z.file().type(["image/png", "image/jpeg"]).min(2000).max(4000);
z.file().type("image/png");
```

For Node.js:
```js
import { File } from "@web-std/file";

z.file({}, File);
```

Getters:

```js
z.file().min(1).minSize;
z.file().max(5).maxSize;
z.file().type("image/png").allowedTypes
```


src/types.ts
```diff
@@ -3488,6 +3488,122 @@ export class ZodNaN extends ZodType<number, ZodNaNDef> {
...
+    return this._addCheck({
+      kind: "type",
+      value: accept,
+      ...errorUtil.errToObj(message),
+    });
+  }
+
+  static create = (): ZodFile => {
+    if (typeof File === "undefined") {
+      throw new Error("File not supported in current environment");
```
andresgutgon, line 3597: We could use something like [@web-std/file](https://github.com/web-std/io/tree/main/packages/file)

jgradzki:
@JacobWeisenburger :point_up:

src/types.ts
```diff
@@ -3488,6 +3488,141 @@ export class ZodNaN extends ZodType<number, ZodNaNDef> {
...
 
+/////////////////////////////////////////
+/////////////////////////////////////////
+//////////                     //////////
+//////////      ZodFile         //////////
+//////////                     //////////
+/////////////////////////////////////////
+/////////////////////////////////////////
+
+// const FileConstructor = typeof File === "undefined" ? WebStdFile : File;
```
andresgutgon, line 3499: 🔥  : )

jgradzki: :( fixed

src/types.ts
```diff
@@ -3488,6 +3488,139 @@ export class ZodNaN extends ZodType<number, ZodNaNDef> {
...
+  type(accept: string | string[], message?: errorUtil.ErrMessage) {
+    return this._addCheck({
+      kind: "type",
+      value: accept,
+      ...errorUtil.errToObj(message),
+    });
+  }
+
+  static create = <T extends Function = typeof File>(
+    fileConstructor?: T
```
andresgutgon, line 3606: If we want to support custom errors I think you need to pass [RawCreateParams](https://github.com/colinhacks/zod/blob/efc37cb94d8d21161529462e4ad3109078ba119c/src/types.ts#L644) like in `ZodString` right?

andresgutgon: I'm talking about this:
image

andresgutgon: I think is the zod way to allow translated errors

jgradzki: Ok, so no problem? :D 
I have removed some extra `"` in "File not supported ..." error message, also changed separator in `issue.expected.join` from `,` to ` or `

jgradzki: nvm, i see what you talking about, will update

jgradzki: i wonder which parameter should be first, for now file constructor is second

deno/lib/types.ts
```diff
@@ -3488,6 +3488,141 @@ export class ZodNaN extends ZodType<number, ZodNaNDef> {
...
+    return this._addCheck({
+      kind: "type",
+      value: accept,
+      ...errorUtil.errToObj(message),
+    });
+  }
+
+  static create = <T extends Function = typeof File>(
+    params?: RawCreateParams,
+    fileConstructor?: T
```
andresgutgon, line 3607: I would switch the order here. It's more common to set the constructor than the params I think. Feel free to ignore, just an opinion

jgradzki: maybe, I will wait for one more opinion maybe and then we will decide

jgradzki:
> It would be nice to have a getter like [string isEmail](https://github.com/colinhacks/zod/blob/26b94211eecb6c1db772e1d3235aa1edee419db7/src/types.ts#L606-L608) to get the expected `types`, `min` and `max`.
> 
> Would be something like this:
> 
> ```js
> const myFile = z.file().min(2).max(39).type(["application/pdf", "image/png"])
> 
> // The getters
> myFile.type // ["application/pdf", "image/png"]]
> myFile.min // 2
> myFile.max // 39
> ```
> 
> I would need this to use in [a code like this one](https://gist.github.com/andresgutgon/36e01c91acb7a0c2dc7509b2b046a6cc)

added, had to use different names ;)

deno/lib/types.ts
```diff
@@ -3488,6 +3488,175 @@ export class ZodNaN extends ZodType<number, ZodNaNDef> {
...
+  typeName: ZodFirstPartyTypeKind.ZodFile;
+  checks: ZodFileCheck[];
+}
+
+export class ZodFile<FileConstructor extends Function> extends ZodType<
+  FileConstructor,
+  ZodFileDef
+> {
+  constructor(def: ZodFileDef, private readonly fileConstructor: Function) {
+    super(def);
```
andresgutgon, line 3514: Can you add here the `description` field? Usually this is not used. But they way I'm trying to use it is helpful to be able to see if current validator is a `ZodFile`

``` javascript
constructor(def: ZodFileDef, private readonly fileConstructor: Function) {
  super(def);
  def.description = "ZodFile";
}
```

andresgutgon: Ideally this should be `typeName` but i think the getter is not exposed in the base class `ZodType` right?

andresgutgon:
### Proposal: Change input from a `File` class to a File description
I don't see a way to make this validation work on the backend :/

The problem:
> You don't know the size of the file until you write in disk

And you don't want to write in disk the full file submitted by a user. Imagine is a HUGE file

Ideally this validation would have as input a file descriptor with the attributes. But not necessary to be an instance of any `File` class. This way is flexible enough.

Ex.:
```javascript
const myFile = z.file().max(3394499)

myFile.safeParse({ size: 50000, type: "image/jpg" }) // type FileInput = { size: number, type: string }
```
What do you think? As output can be a `File`. In Node the people of Remix [did his own file](https://github.com/remix-run/remix/blob/main/packages/remix-node/fetch.ts#L70) because [web File](https://developer.mozilla.org/en-US/docs/Web/API/File) I think doesn't have a way of defining the `filepath`

andresgutgon:
Ok I think current implementation is flexible enough. For my use case I don't want to build a real `File` object while doing validation. But this code allow me to pass a class that looks like a `File` 

``` javascript
export class FileShape { constructor(public type: string, public size: number) {}}

// Then I can use normal:
const myFile = z.file({}, FileShape).max(34033).type(["images/jpeg"])
```

The reason for not wanting to use a real `File` implementation is because I don't know the size of the file in advance on the server. So I can't build a real file and pass to `zod`. 
What I do is to stream the contents on the file + write to disk those contents and measure if they are less than the `.max` attribute on the zod file validation.

When I know that the size of the file submitted is too big I set that `size` in my `ShapeFile` like this
``` javascript
const file = new FileShape("image/jpeg", 0)
const myValidation = z.file({}, FileShape).max(34033).type(["images/jpeg"])
// ...
// Do some work with Node streams + write to disk + measure bytes
// If I detect the the file is too big set that size in my FileShape like this:
file.setSize(23232323) // Too big
myValidation.parse(file) // Now that I know the actual file size I want to get the errors.
```

Do you thing is a reasonable way of using this or is fragile?

I like your PR 👍

deno/lib/ZodError.ts
```diff
@@ -356,6 +364,11 @@ export const defaultErrorMap = (
     case ZodIssueCode.not_multiple_of:
       message = `Number must be a multiple of ${issue.multipleOf}`;
       break;
+    case ZodIssueCode.invalid_file_type:
+      message = `The file type should be ${issue.expected.join(
+        " or "
+      )}. Received: ${issue.received}`;
+      break;
```
andresgutgon, line 371: Missing here messages for `too_big` and `too_small`. Without it for `file` the error goes to the generic:
image

andresgutgon: I can do this changes if you give me push permission to your branch. If not is fine I'll wait :)

deno/lib/types.ts
```diff
@@ -3488,6 +3488,175 @@ export class ZodNaN extends ZodType<number, ZodNaNDef> {
...
+  | { kind: "min"; value: number; message?: string }
+  | { kind: "type"; value: string | string[]; message?: string };
+
+export interface ZodFileDef extends ZodTypeDef {
+  typeName: ZodFirstPartyTypeKind.ZodFile;
+  checks: ZodFileCheck[];
+}
+
+export class ZodFile<FileConstructor extends Function> extends ZodType<
+  FileConstructor,
```
andresgutgon, line 3510: Would be possible to support an Array of `File`s?

The use case is if people use [multiple=true](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-multiple) in an `<input type="file" multiple />`

This way these upload can be processed in the server and use this validation like this:

```
const myRequest = z.object({
  gallery: z.file({}, File).array()
})
```

But for this to work we need to manage the case when the input is an array. 

I would play with it in my dev environment with your fork : )

andresgutgon: Forget it. This works:
``` javascript
const myRequest = z.object({
  gallery: z.array(z.file({}, File))
})
```

Nothing to do here : )

andresgutgon:
@jgradzki could you take a look to my comments? I would really love to have this in "zod" ❤️  

@colinhacks what's your opinion here?

andresgutgon:
I'm being thinking on the approach we're taking here of requiring a constructor. It has some issues. 

The first issue is that you need to pass a constructor on Non-web environments or [simulate a web file API in Node](https://www.npmjs.com/package/@web-std/file). This is annoying and I think we can avoid it.

The second issue is that what you see after parsing your `ZodFile` is `typeof File`
image

This is a problem because then Typescript doesn't know what properties should have the returned `File | FileLike` class.

## My proposal
My proposal is to define a minimum common interface that `ZodFile` should require and any class that respect this interface can be returned. 

``` javascript
interface IFile {
  name: string;
  size: number;
  type: string;
}
```

If we return an Interface then Typescript know what is that. Look how nice : )

image


Also here is clear what you have
image

### Implementation
I'm still working on it. That interface is missing for example `filename` which is basic in a node environment. Also we should add all [official web attributes and methods according to the spec](https://developer.mozilla.org/en-US/docs/Web/API/File)

But not check that passed object is an `instanceof File`. Instead just check that it has what we need for this `ZodFile` which is `size` and `type`

``` javascript
    const hasSize = !!input.data?.size;
    const hasType = !!input.data?.type;
    if (!hasSize || !hasType) {
      const issue = makeIssue({ //... set the issue as you do })
    }
```

What do you think?

andresgutgon:
Please take a look to [THIS SANDBOX](https://codesandbox.io/s/pensive-frost-txd20s?file=/src/fileInterface.ts) I put together both apprach `fileInterface.ts` and `fileConstructor.ts`

README.md
```diff
@@ -433,24 +434,58 @@ const isActive = z.boolean({
...
+>   if (typeof arg == "string" || arg instanceof Date) return new Date(arg);
+> }, z.date());
+> type DateSchema = z.infer<typeof dateSchema>;
+> // type DateSchema = Date
+>
+> dateSchema.safeParse(new Date("1/12/22")); // success: true
+> dateSchema.safeParse("2022-01-12T00:00:00.000Z"); // success: true
+> ```
+
+## Files [browser only]
```
andresgutgon, line 458: Why only browser? At this point a lot of people is using Zod in browser and in Node. Why not a solution valid for both environments?

In my case it looks something like this:

``` javascript
const schema = z.object({
  name: z.string(),
  avatar: z.file() 
})

const requestHandler = (request, response) => {
  const result = parser.parse({ request, schema })

  if (!result.success) {
     // Handle error
  } 
  
  // At this point is a file in NodeJS. That can be a file in filesystem o an object in S3
  console.log("DATA", result.data.avatar)
}
```

All the implementation on how to create that file is inside the parser. But Zod is there to provide a nice validation system. Would be super nice if instead of `file` type be limited to a web `File` object would be a contract that can satisfy both web and Node.

I'm not thinking on adding any Node dependency here. Just have a file validator that accept a contract instead of a fixed `File` class. In the end the validator just want the file to respond to `size` and `type` properties right?

README.md
```diff
@@ -433,24 +434,58 @@ const isActive = z.boolean({
...
 
 ```ts
-const dateSchema = z.preprocess((arg) => {
-  if (typeof arg == "string" || arg instanceof Date) return new Date(arg);
-}, z.date());
-type DateSchema = z.infer<typeof dateSchema>;
-// type DateSchema = Date
+z.date().min(1); // no empty files
+z.date().max(10e6); // max 10MB
+z.date().types(["image/png", "image/jpeg"]);
```
jgradzki, line 476: z.file I guess :D

README.md
jgradzki: I started thinking that we cannot satisfy browser and node.js in one type. My intentions were strictly to validate for File instance. I think there should be completely different type for node.js/deno and I think you should make your own in your project as handling files on server can be done in few different ways I guess.

andresgutgon: > I started thinking that we cannot satisfy browser and node.js in one type. My intentions were strictly to validate for File instance. I think there should be completely different type for node.js/deno and I think you should make your own in your project as handling files on server can be done in few different ways I guess.

That's the thing : ) 

1. In browser a `File` is checked for `size` and `type`
2. In Node/Deno a `Something` is checked for `size` and `type` 

The only job of this validator is to check if the `Thing` passed respond to `type` and `size`

And then in the return part could be something like this:

``` javascript
interface IFile {
  public name: string
  public size: number
  public filepath: string
}
```

I like about this proposal that is flexible enough to serve browser and Node. 

## Cons
In browser you would have to wrap your `File` with an `IFile` like interface
``` javascript
class WebFile implements IFile {
  filepath: string = ""
  constructor(file: File) {
     this.name = file.name
     this.size = file.size
  }
}
```

andresgutgon: Totally fine with implementing [my own file for Node](https://github.com/andresgutgon/remix-storage/blob/main/packages/bodyparser/src/zod-addons/internal/file.ts#L104) 😄 

But I think Zod is awesome and people is already using it in the server. `file` should handle both platforms somehow. But as I said totally happy if this doesn't happen inside Zod

README.md
jgradzki: @colinhacks what next :)

RicardoValero95:
What happened to this?

dennemark:
any chance this could be added?

PR #1021, opened 2022-03-17
Precache, avoid cache-checking boolean values, and avoid unnecessary spreads

tmcw:
Combination of two things that I could find that caused significant performance bumps for my benchmark:

- This optimizes the cache for literals. true, false, null, and undefined are always going to be equal to each other: we might as well fill the cache with them and when they're returned, avoid the `cacheAndReturn` logic.
- In hot code, using `...` is causing performance drag because, in the current setup, zod's distributed files are using TypeScript helpers. This I'll file another PR for: my suspicion is that a very large portion of zod's performance problems are caused by that.

### DIFF
--- a/deno/lib/helpers/parseUtil.ts
+++ b/deno/lib/helpers/parseUtil.ts
@@ -50,7 +50,7 @@ export const getParsedType = (
 
   switch (t) {
     case "undefined":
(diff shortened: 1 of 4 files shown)


tmcw:
I think removing the typeCache is a better solution than this one for primitives, and switching to es2018 is a better solution than avoiding spread syntax (for now).

PR #1026, opened 2022-03-17
Fast unions

tmcw:
Okay, one more for fun I guess.

Right now Zod's unions work by evaluating all options and returning the first valid result, the first dirty result, or an error. So, the performance of a union check is strictly related to the possible options.

This changes the (sync) union mechanism to instead succeed fast. If the first option works, then it returns immediately. In the success case, this means that `z.union` can be faster than evaluating all the options. That's basically it. Also, it's a little more stingy with how it uses memory, by retaining less of results.


scotttrinh:
This looks great! I'll take a better look tomorrow and try to think of possible unintended implications.

PR #1048, opened 2022-03-25
Make literal error message explicit

### LINKED ISSUE #1045: More expressive error messsage possible for "Expected string, received string"
Thanks for zod, I love it!

[Example:](https://runkit.com/gregoor/zod-string-literal-error)
```js
const { z } = require("zod");
const Special = z.literal("wat");
console.log(Special.safeParse("nah").error.issues[0].message);
```
which outputs `Expected string, received string`

Wouldn't `Expected "wat", received "nah"` be more preferable? I suppose there is some nuance and edge cases where this isn't feasible?

Gregoor:
Fixes #1045

Initially I added a separate test, but then I noticed that there was a pre-existing test covering this that just needed to be updated, so I just left it at that.

### DIFF
--- a/deno/lib/__tests__/error.test.ts
+++ b/deno/lib/__tests__/error.test.ts
@@ -419,7 +419,7 @@ test("literal default error message", () => {
   } catch (err) {
     const zerr: z.ZodError = err as any;
     expect(zerr.issues.length).toEqual(1);
-    expect(zerr.issues[0].message).toEqual("Expected string, received string");
+    expect(zerr.issues[0].message).toEqual('Expected "Tuna", received "Trout"');
   }
 });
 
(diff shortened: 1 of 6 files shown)


colinhacks:
Thanks for the great PR. Ultimately I decided it wasn't worth trying to abuse the `invalid_type` issue here, so it makes more sense to introduce a new `invalid_literal` issue type. Zod also has a policy of not putting any user input into the error messages for security reasons. I implemented this in https://github.com/colinhacks/zod/pull/1052 so I'm gonna close this for now.

Gregoor:
Makes sense, thanks for the info, I can see how not having this in is a better security default. That said, for cases where it matters less like mine, I was thinking of turning `errorIncludesInputData` once and if #851 lands. Would you be open to a PR that builds on that to show literal values if that flag is set?

PR #1054, opened 2022-03-27
fix(readme): Use Correct Test Link

FlorianWendelborn:
Also added a filter for the master branch, so only relevant tests are shown


colinhacks:
whoops, copy/paste fail..

Issue #696, opened 2021-10-09
Throwing errors in preprocess


hampuskraft:
Hi there,

I'm trying to implement string coercion using `z.preprocess`, but there seems to be no way to indicate coercion errors:
```ts
export const Str = z.preprocess((value) => {
  switch (typeof value) {
    case 'bigint':
    case 'boolean':
    case 'number':
      return value.toString();

    case 'object':
      if (value == null) {
        return value;
      }
      if (value instanceof Date) {
        return value.toISOString();
      }
      throw new Error('Could not interpret value as string');

    case 'string':
      return value.trim();

    case 'undefined':
      return value;

    default:
      throw new Error('Could not interpret value as string');
  }
}, z.string());

Str.parse([]); // Error: Could not interpret value as string (rather than a ZodError)
```

I tried to create verbose workarounds but none ended up working. Thanks!

Kind regards,
Hampus Kraft.

scotttrinh:
The few times I've played around with `preprocess`, I actually use zod schemas with the "looser" version of what I expect. For instance, if you're using `preprocess` to get from a serialized query-parameter style `string`-heavy object to a parsed object, something like:

```ts
// Query parameter schema
const loose = z.object({
  aBoolean: z.string().transform(Boolean),
  aNumber: z.string().transform(Number),
});

// Domain schema
const query = z.object({
  aBoolean: z.boolean(),
  aNumber: z.number(),
});

const fromQueryParams = z.preprocess(
  loose.parse,
  query
);
```

I have to admit that I haven't used `preprocess` in a serious manner yet, so there might be others who have come up with something more generic like what you we originally attempting.

selimb:
> Perhaps you should be able to add issues inside `preprocess` similar to `superRefine`. I'll look into this. It would save you from having to call `coerceString` twice.

+1 for adding issues in `preprocess`. @colinhacks Would you be willing to accept a PR on this?

IMO this would make it much more straightforward to build a reusable set of "lenient preprocessors" on top of zod (#804 #1019), while keeping *good error messages*.

I currently work around it with something like this:
```typescript
function maybeParseNumber(v: unknown): unknown | number {
  if (typeof v !== "string") {
    return v;
  }
  if (!v) {
    return v; // Otherwise "" is parsed as 0
  }
  const ret = Number(v);
  return isNaN(ret) ? v : ret;
}

// Preprocessors for stringly-typed data.
const zs = {
[...]

PR #1056, opened 2022-03-29
feat: allow ctx.addIssue from transform

morgs32:
Closes out part of #696 . Except only on transforms!

```
z
    .string()
    .transform((data, ctx) => {
      const i = strs.indexOf(data)
      if (i === -1) {
        ctx.addIssue({
          code: 'custom',
          message: `${data} is not one of our allowed strings`,
        })
      }
      return data.length
    })
    .parse("asdf");
```
Will throw. `ctx` works just like in superRefine()

### DIFF
--- a/src/__tests__/transformer.test.ts
+++ b/src/__tests__/transformer.test.ts
@@ -30,7 +30,7 @@ test("transform ctx.addIssue", () => {
       return data.length
     })
     .parse("asdf");
-  }).toThrow();
+  }).toThrow('asdf is not one of our allowed strings');
 });
 
 test("basic transformations", () => {
--- a/src/__tests__/transformer.test.ts
+++ b/src/__tests__/transformer.test.ts
@@ -10,6 +10,29 @@ const stringToNumber = z.string().transform((arg) => parseFloat(arg));
 //   .transform((n) => String(n));
 const asyncNumberToString = z.number().transform(async (n) => String(n));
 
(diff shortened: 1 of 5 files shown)


src/__tests__/transformer.test.ts
```diff
@@ -10,6 +10,29 @@ const stringToNumber = z.string().transform((arg) => parseFloat(arg));
...
+      if (i === -1) {
+        ctx.addIssue({
+          code: 'custom',
+          message: `${data} is not one of our allowed strings`,
+        })
+      }
+      return data.length
+    })
+    .parse("asdf");
+  }).toThrow();
```
FlorianWendelborn, line 33: I think it would be better if this test could check for the kind of error, ideally the custom code/message.

Otherwise, this test would also pass if `ctx` is `undefined` as it would throw a JS error

morgs32: Good call. Checking the whole error message now.

FlorianWendelborn:
Well, I'm in favor of this. I don't have the power to merge anything here though, just a `zod` user who likes to contribute a bit with suggestions

FlorianWendelborn (approved):

morgs32:
insertgif.com

colinhacks:
Sorry for the delay on this. Fantastic PR. Landed in 3.15.0.

PR #1057, opened 2022-03-30
Faster extra keys detection

tmcw:
This increases object parsing performance by about 8-10% by skipping a call to Object.keys and instead using `in`. I know it's odd, but this seemingly insignificant few lines of code really seems to be a hotspot.

Benchmark results: https://observablehq.com/d/8978a77c49a9c6ef - there's a lot of test jitter in mainly unrelated benchmarks - no reason to think this'll change string parsing performance.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1558,8 +1558,12 @@ export class ZodObject<
     const { status, ctx } = this._processInputParams(input);
 
     const { shape, keys: shapeKeys } = this._getCached();
-    const dataKeys = util.objectKeys(ctx.data);
-    const extraKeys = dataKeys.filter((k) => !shapeKeys.includes(k));
+    const extraKeys: string[] = [];
+    for (const key in ctx.data) {
+      if (!shapeKeys.includes(key)) {
+        extraKeys.push(key);
+      }
+    }
 
     const pairs: {
       key: ParseReturnType<any>;
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -1558,8 +1558,12 @@ export class ZodObject<
     const { status, ctx } = this._processInputParams(input);
 
     const { shape, keys: shapeKeys } = this._getCached();
-    const dataKeys = util.objectKeys(ctx.data);
-    const extraKeys = dataKeys.filter((k) => !shapeKeys.includes(k));
+    const extraKeys: string[] = [];
+    for (const key in ctx.data) {
+      if (!shapeKeys.includes(key)) {
+        extraKeys.push(key);
+      }
+    }
 
     const pairs: {
       key: ParseReturnType<any>;


deno/lib/benchmarks/competitive.ts
```diff
@@ -0,0 +1,85 @@
+import Benchmark from "benchmark";
+import mz from "myzod";
```
scotttrinh, line 2: I expect to see an update to `package.json` that adds `myzod` to `devDependencies`, no?

tmcw: ugh, this was included by accident - see #1059 for that issue. i’ll remove it when i’m back at a keyboard

tmcw: Removed the stray file.

FlorianWendelborn (approved):
I ran a test locally with my real-world schema:

Before:

```
Benchmark #1: yarn run start
  Time (mean ± σ):     42.640 s ±  1.392 s    [User: 50.160 s, System: 4.063 s]
  Range (min … max):   40.499 s … 44.512 s    10 runs
```

After:

```
Benchmark #1: yarn run start
  Time (mean ± σ):     38.663 s ±  0.267 s    [User: 47.281 s, System: 3.767 s]
  Range (min … max):   38.375 s … 39.139 s    10 runs
```

This was measured with `hyperfine` at 10 runs (without using the system in the meantime). Since the worst of the after runs was still better than the best of the before runs I’d say this pretty much lines up with the PR’s claims of ~10% improved performance.

Code LGTM as well, but perhaps I’m missing something.

src/types.ts
```diff
@@ -1558,8 +1558,12 @@ export class ZodObject<
     const { status, ctx } = this._processInputParams(input);
 
     const { shape, keys: shapeKeys } = this._getCached();
-    const dataKeys = util.objectKeys(ctx.data);
-    const extraKeys = dataKeys.filter((k) => !shapeKeys.includes(k));
+    const extraKeys: string[] = [];
+    for (const key in ctx.data) {
+      if (!shapeKeys.includes(key)) {
```
FlorianWendelborn, line 1563: Would it make a difference to use a `Set` for the shape keys? Should be faster than `Array.prototype.includes`

tmcw: I tried and that and it was hard to produce any consistent improvement. I suspect that it's similar perf for smallish objects but faster for objects with lots of keys? I'll try adding a very wide object to the realworld bench to see.

scotttrinh: Yeah, I suspect homgeneous array include calls are pretty optimized, but Set certainly makes sense semantically. Not a blocker here!

scotttrinh (approved):

src/types.ts
FlorianWendelborn: I’ve found this benchmark for this exact comparison. Maybe it helps with a future PR for this hotspot: <https://jsbench.me/irkhdxnoqa/2>

PR #1067, opened 2022-04-04
Add UMD build

rinsuki:
Add UMD build to zod.

## Example

with UMD build, you can use zod on web browser without bundler/ESM support, like this:

```html
<script src="https://some.cdn.example/npm/zod/lib/index.umd.js"></script>
<script>
// @ts-check
const { z } = Zod

z.object({hello: z.literal("world")})
</script>
```

## Background

I want to use zod in UserScript, but I have some issues:

- Zod is pretty big to bundle it to UserScript (>2000 lines if you wont minify it)
  - some hosting sites are restricting file size, like [Greasy Fork](https://greasyfork.org/en/help/code-rules)
- If bundle libraries to UserScript, makes hard to:
  - read/audit source codes (because lines of source code will increase)
  - determine this source code comes from which version of libraries

so I added a UMD build to Zod, which allows UserScript to use `@require` to load zod, like `// @require  https://some.cdn.like.jsdelivr.example/npm/zod@x.x.x/lib/index.umd.js`.


package.json
```diff
@@ -49,11 +49,12 @@
     "check": "yarn lint:check && yarn prettier:check",
     "fix": "yarn lint:fix && yarn prettier:fix",
     "clean": "rm -rf lib/* deno/lib/*",
-    "build": "yarn run clean && npm run build:cjs && npm run build:esm && npm run build:deno",
+    "build": "yarn run clean && npm run build:cjs && npm run build:esm && npm run build:deno && yarn run build:tweak-dts",
     "build:deno": "node ./deno/build.mjs",
     "build:esm": "rollup --config rollup.config.js",
     "build:cjs": "tsc --p tsconfig.cjs.json",
     "build:types": "tsc --p tsconfig.types.json",
+    "build:tweak-dts": "echo 'export as namespace Zod;' >> lib/index.d.ts",
```
scotttrinh, line 57: Is there something we can do in the source to get this same effect without having to patch the artifact?

rinsuki: I find way to do it without patch to generated dts! can you review it?

scotttrinh: Definitely better! Can you give me a little context as to why we need to export a namespace for the UMD target? Is that because it adds a global binding to the library?

rinsuki: > Is that because it adds a global binding to the library?

Yes!

FYI: If project uses module (like `import ...`), global binding will be disabled (tsc says `'Zod' refers to a UMD global, but the current file is a module. Consider adding an import instead.ts(2686)`)

colinhacks:
@rinsuki I'd rather the name of the global namespace be `z` as this is the recommended way to consume Zod's API. Is there a reason this wouldn't work?

rinsuki:
1. `z` are too short and might be overwritten by other scripts (some dumbs script writes temporary variables to global)
2. if we uses `z` for UMD name, and if users `.ts` file doesn't using import, `z` will be automatically available on global. then users may forget to import, and causes runtime error. it would be good to force user to write `const { z } = Zod`, i think.

PR #1083, opened 2022-04-14
docs: infer `Literal` from the schema in JSON type example

jdbruijn:
I was using this great recommendation for JSON type and saw this small improvement to reduce duplication.


scotttrinh (approved):
Seems sensible, thanks!

PR #1084, opened 2022-04-14
Add parent ctx to refinement

michiim:
make refine more powerful by offering parent (data) ctx to validation function

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -46,6 +46,7 @@ import {
 export type RefinementCtx = {
   addIssue: (arg: IssueData) => void;
   path: (string | number)[];
+  parent: ParseContext;
 };
 export type ZodRawShape = { [k: string]: ZodTypeAny };
 export type ZodTypeAny = ZodType<any, any, any>;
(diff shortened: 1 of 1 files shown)


colinhacks:
I think this is a bit overkill - I'd rather provide specific fields from the parent context instead of dumping the whole thing. I consider most of `ParseContext` an "internal API" and I'd rather not leak the full complexity of it to the user. Any particular piece of `ctx` you'd like to include?

michiim:
For a complex validation of nested objects I need the parent data (up to top level) from the context.

colinhacks:
Why not just add the refinement higher up in your schema hierarchy so it has access to all the data it needs? A bit more detail about your use case would be useful.

michiim:
very simplified scenario:

A customer object with a list of addresses. Depending on the customer type, the address must have a filled "company" field or not. The "company" field must be filled in if the customer type is "company".

{
  type: 'COMPANY' (or 'PRIVATE')
  number: 1234
  addresses: [
    {
      company: '' <= error
      street: 'street 1
      postalCode: '1234
      city: 'City
    }
  ]
}

I know I could make the refinement on top level, but I don't really find that attractive. I automatically create the zod schema from my ORM configuration and thus have property level definitions, which makes much more sense to me.  I want to have a refinement on the property "company" because it makes the automatic generation of the zod schema much easier and more logical for me.

PR #1113, opened 2022-05-03
fix: add type-predicate function through the `is`-method.

merisbahti:
Is it possible for is to add an `is`-method, which basically is the type: `is(data: unknown): data is Output`?

This feature is called [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)

### DIFF
--- a/deno/lib/__tests__/is.test.ts
+++ b/deno/lib/__tests__/is.test.ts
@@ -1,0 +1,23 @@
+import { expect } from "https://deno.land/x/expect@v0.2.6/mod.ts";
+const test = Deno.test;
+
+import { z } from "../index.ts";
+
+test("test is ", () => {
+  const schema = z.object({ key1: z.string(), key2: z.number() });
+  type ExpectedType = { key1: string; key2: number };
+
+  const value = { key1: "a", key2: 5 } as unknown;
+
+  expect.assertions(1);
+  if (schema.is(value)) {
+    const typeCheckedValue: ExpectedType = value;
+    expect(typeCheckedValue).toBe(value);
+  } else {
+    //@ts-expect-error
+    const typeCheckedValue: ExpectedType = value;
+    // So that the "unused" error goes away
+    expect(typeCheckedValue).toBe(value);
+    expect(false).toBe(true);
+  }
+});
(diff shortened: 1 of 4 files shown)


lichstam:
i don't see a reason for this...

joking :), gj!

scotttrinh:
It might seem surprising that we don't already have this, right? 😂 We don't provide type predicates because schemas can transform their inputs. It's one of the reasons we refer to Zod schemas as parsers.

It's not impossible to write a helper that will give you a type predicate for the input type but we've purposefully avoided having one in Zod itself since it can be a footgun with `preprocess`.

merisbahti:
Very good, thank you! :) 

I'll save it here, so that if anybody searches for the keywords "type predicate" I'll find this explained here :)

THEtheChad:
Forgive me, @scotttrinh, but I don't understand your explanation.

Yes, you can mutate the data using `.transform` and `.preprocess`, but in either case, the resulting object's output type is known.

```ts
const MyType = z.preprocess(
  (val) => String(val),
  z.string().transform((v) => Number(v))
);
  
 z.infer<typeof MyType> // number
```

I use Zod to strongly type all of my API endpoints. The issue I run into a lot is when the response type has a union. I need a type predicate to differentiate which version of the union I received. This is the ideal scenario for this type of method.

```ts
const User =  z.object({
    first_name: z.string(),
    last_name: z.string()
})

const Staff = User.extend({
    group: z.string(),
    isAdmin: z.boolean()
})

// this type of scenario
const Response = z.object({
    id: z.string(),
    user: z.union([ User, Staff ])
})

response.user.group // can't be accessed here

if( Staff.isType(response.user) ){
    response.user.group // no errors
}
```

THEtheChad:
I was following up on this in my original thread and it occurred to me what the real issue is.

```ts
const MyType = z.preprocess(
  (val) => String(val),
  z.string().transform((v) => Number(v))
);
 
// this is what we're asserting
z.infer<typeof MyType> // number

const v = '2'
if( MyType.isType( v ) ){
    // this executes because v passes the type predicate test
}
```
Here, we're passing in a string that IS parsable by `MyType` but is not the type of the resulting output.

I know that, in my use case, I'm asking for a type predicate that allows me to safely access values of things that ARE parsable my type. So really, it might be a semantic issue with the name of the method vs the actual utility of the type predicate itself. Maybe something along the lines of `.isParsable`, which would allow for type narrowing and better convey what the method is actually doing?

Issue #1121, opened 2022-05-08
Zod resolved type change for Opaque type between v3.14.4 -> v3.15.1


bluepnume:
I have the following code:

```typescript
type Opaque<Type, Token = unknown> = Type & { tag: Token }; 

enum FACE {
    QUEEN = 'QUEEN',
    KING = 'KING'
}

enum SUIT {
    HEARTS = 'HEARTS',
    DIAMONDS = 'DIAMONDS'
}

type FaceCard<Face extends FACE> = Opaque<`${ Face }_${ SUIT }`, 'FaceCard'>;

type Hand = z.output<
    z.ZodObject<{
        card : z.ZodType<FaceCard<FACE.QUEEN>, z.ZodTypeDef, string>,
    }>
>;
```

In `zod@3.14.4`, `Hand` resolved to:

```typescript
type Hand = {
    card: FaceCard<FACE.QUEEN>;
}
```

But now in `zoid@3.15.1`, `Hand` is resolving to:

```typescript
type Hand = {
    card: (FaceCard<FACE.QUEEN> | undefined) & FaceCard<FACE.QUEEN>;
}
```

Why did the second type get more complex here, when the first one is correct?

(By the way -- **incredible** library. Thank you for making this, seriously.)

bentefay:
Hi @bluepnume. 

Sorry about this. I changed the definition of `z.ZodObject` in my PR (https://github.com/colinhacks/zod/issues/1121) that was recently merged. It's behaviour was unchanged in all the cases I tested, but it looks like you found a broken case. I've fixed the type to work with your example: 

https://github.com/colinhacks/zod/pull/1122

As a side node, although the type looks ugly and we should definitely fix the problem, I believe the types `{ property: (T | undefined) & T }` and `{ property: T }` are technically the same, so hopefully you shouldn't be seeing any issues with using the type.

bluepnume:
Thanks @bentefay! Appreciate the fast turnaround!

No huge rush to merge, since I can just lock to `3.14.4` for now. But since you asked, the issue that this causes for me is demonstrated with the following code:

```typescript
type QueenCard = Hand['card'];

type CardValue = {
    [ FACE.QUEEN ] : 12,
    [ FACE.KING ] : 13,
};

type CardToValue<Face extends FACE> = Face extends keyof CardValue
    ? CardValue[Face]
    : never;

const cardToValue = <Face extends FACE>(card : FaceCard<Face>) : CardToValue<Face> => {
[...]

PR #1122, opened 2022-05-09
Restore existing type inference behaviour of z.ZodObject

bentefay:
Fixes the issue documented here: https://github.com/colinhacks/zod/issues/1121

I'm pretty sure my PR introduced this issue: https://github.com/colinhacks/zod/pull/1117

In almost all cases `z.flatten<{ property: (T | undefined) & T }>` flattens to `{ property: T }`. However, this does not appear to be the case for some very specific types like this one:

```TypeScript
type Opaque<Type, Token = unknown> = Type & { tag: Token };

  enum FACE {
    QUEEN = "QUEEN",
    KING = "KING",
  }

  enum SUIT {
    HEARTS = "HEARTS",
    DIAMONDS = "DIAMONDS",
  }

  type FaceCard<Face extends FACE> = Opaque<`${Face}_${SUIT}`, "FaceCard">;

  type Hand = z.objectUtil.flatten<
    z.ZodObject<{
      card: z.ZodType<FaceCard<FACE.QUEEN>, z.ZodTypeDef, string>;
    }>["_output"]
  >;
  ```

In this case `Hand` is flattening to:
```TypeScript
type Hand = {
    card: (FaceCard<FACE.QUEEN> | undefined) & FaceCard<FACE.QUEEN>;
}
```

Fortunately, I now better understand the problem I was trying to fix in my PR, and it is actually possible to write `z.ZodObject` in a way that preserves both behaviours. 

It also has the benefit of being easier to understand:

```TypeScript
  type optionalKeys<T extends object> = {
    [k in keyof T]: undefined extends T[k] ? k : never;
  }[keyof T];

  type requiredKeys<T extends object> = {
    [k in keyof T]: undefined extends T[k] ? never : k;
  }[keyof T];

  export type addQuestionMarks<T extends object> = Partial<
    Pick<T, optionalKeys<T>>
  > &
    Pick<T, requiredKeys<T>>;
```

Unfortunately I couldn't see a way to write a test for this, as:

```TypeScript
type Hand = {
    card: (FaceCard<FACE.QUEEN> | undefined) & FaceCard<FACE.QUEEN>;
}
```

and 

```TypeScript
type Hand = {
    card: FaceCard<FACE.QUEEN>
}
```

are equal as far as `util.AssertEqual` is concerned.

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -1363,13 +1363,18 @@ export namespace objectUtil {
     [k in Exclude<keyof U, keyof V>]: U[k];
   } & V;
 
-  export type requiredKeys<T extends object> = {
+  type optionalKeys<T extends object> = {
+    [k in keyof T]: undefined extends T[k] ? k : never;
+  }[keyof T];
+
+  type requiredKeys<T extends object> = {
     [k in keyof T]: undefined extends T[k] ? never : k;
   }[keyof T];
 
-  export type addQuestionMarks<T extends object> = {
-    [k in keyof T]?: T[k];
-  } & { [k in requiredKeys<T>]: T[k] };
+  export type addQuestionMarks<T extends object> = Partial<
+    Pick<T, optionalKeys<T>>
+  > &
+    Pick<T, requiredKeys<T>>;
 
   export type identity<T> = T;
   export type flatten<T extends object> = identity<{ [k in keyof T]: T[k] }>;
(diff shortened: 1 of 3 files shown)


colinhacks:
Sounds good to me. That test may be possible if we switched to using something like https://github.com/dsherret/conditional-type-checks which is a little smarter than the current `util.AssertEqual` trick, but I've been procrastinating on that. That test isn't vital at the moment.

PR #1123, opened 2022-05-09
Added expected and received properties to enum and native enum

### LINKED ISSUE #1042: Read received value in invalid_enum_value issue
In `issue.code === 'invalid_type'` I can access `issue.received`. 
In `issue.code === 'invalid_enum_value'` this property does not exist. 

Is there any way to understand what was received and why the validation failed?

ElianCordoba:
Fixes #1042

Small note, I typed the `received` property in `ZodInvalidEnumValueIssue` as `any` since we are returning what was inputed originally, instead of it's `ParsedType`. I think this is right but will wait for confirmation :)

Also added an small util to reduce the code duplication around quoting and joining arrays of strings.

TL;DR;

Given
```ts
enum Fish {
  Tuna = "Tuna",
  Trout = "Trout"
}

z.nativeEnum(Fish).parse("Salmon");
```

```diff
- "Invalid enum value. Expected 'Tuna' | 'Trout'"
+ "Invalid enum value. Expected 'Tuna' | 'Trout' received 'Salmon'"
```

### DIFF
--- a/deno/lib/ZodError.ts
+++ b/deno/lib/ZodError.ts
@@ -68,6 +68,8 @@ export interface ZodInvalidUnionDiscriminatorIssue extends ZodIssueBase {
 }
 
 export interface ZodInvalidEnumValueIssue extends ZodIssueBase {
+  expected: ZodParsedType;
+  received: any;
   code: typeof ZodIssueCode.invalid_enum_value;
   options: (string | number)[];
 }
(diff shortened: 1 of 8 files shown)


ElianCordoba:
What about deno? Should I run the `build.mjs` file?

colinhacks:
Thanks for this. It's a tricky problem. Zod has a policy against logging schema input to the console, but arguably the invalid enum value itself is a type in this situation. I decided the root of the problem is that `ZodInvalidEnumValueIssue` should be limited to the scenario where a string/number is provided as input. Everything else should be a `ZodInvalidTypeIssue`. I made this change directly to your PR. This is now the behavior:

```ts
z.enum(["Tuna", "Trout"]).parse(12);
// ZodIssueCode.invalid_type.invalid_type
// "Expected 'Tuna' | 'Trout', received number"

z.enum(["Tuna", "Trout"]).parse("Salmon");
// ZodIssueCode.invalid_enum_value
// "Invalid enum value. Expected 'Tuna' | 'Trout', received 'Salmon'"
```

PR #1129, opened 2022-05-11
Fix ctx.addIssue in transform to work correctly with parseAsync

bentefay:
A recent PR added `ctx.addIssue` to `transform`: https://github.com/colinhacks/zod/pull/1056

This is fantastic! 

There appears to be a small bug - calling `ctx.addIssue` in `transform` does not work correctly when using `parseAsync`. This results in a confusing behaviour where `parseAsync` and `parse` return different results.

### Reproduction

```TypeScript
z
  .string()
  .transform((val, ctx) => {
    const parsed = parseInt(val);
    if (isNaN(parsed)) {
      ctx.addIssue({
        code: z.ZodIssueCode.custom,
        message: "Not a number",
      });
    }
    return parsed;
  })
  .parse("Not a number");
````

__Expected__: Zod error
__Actual__: Success

I've added a test to confirm it is now working correctly. I also took the chance to update the README to reflect this new feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1604,7 +1604,28 @@ const stringToNumber = z.string().transform((val) => myString.length);
 stringToNumber.parse("string"); // => 6
 ```
 
-> ⚠️ Transform functions must not throw. Make sure to use refinements before the transform to make sure the input can be parsed by the transform.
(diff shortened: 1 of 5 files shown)


colinhacks (approved):

colinhacks:
Looks solid, and good catch!

PR #1131, opened 2022-05-12
Instanceof issue fatal

### LINKED ISSUE #1001: New `refine` behaviour issues with `z.custom` (and `z.instanceof`)
Because `z.custom` returns a `ZodType<T>`. successive refinements believe to operate on a `T` when they're actually receiving the untouched input instead.

### Example
```ts
import { z } from 'zod';

class Foo {
  isValid(): boolean {
    return true;
  }
}

const FooSchema = z.instanceof(Foo)
  .refine((val) => {
    // `val` is typed as `Foo` because `z.instanceof(Foo)` is a `ZodType<Foo>`
    // but we will be operating on the raw input instead causing a runtime error.
    return val.isValid();  
  });

FooSchema.parse(undefined); // TypeError: cannot read properties of undefined (reading 'isValid')
```

### Possible Solution
I think a viable non-breaking solution would be to have `z.custom` rely on `superRefine` instead of `refine` and create an issue with `fatal: true` when the validation fails.

Stouffi:
closes #1001 

Here is a codesandbox to illustrates the problem encountered: https://codesandbox.io/s/youthful-sunset-6jn810?file=/src/index.test.ts

An issue with instanceof schema should be fatal in order to prevent running refinement on a value which might be incompatible with the provided instance type.

This PR introduces an additional optional parameter in the `custom` signature to specify wether the issue should be fatal.

```ts
declare const custom: <T>(
  check?: (data: unknown) => any,
  params?: Parameters<ZodTypeAny["refine"]>[1],
  fatal?: boolean
) => ZodType<T>
```

However it would be better to specify it as an object to specify more options in the future.

```ts
interface CustomOptions {
  fatal: boolean
}

declare const custom: <T>(
  check?: (data: unknown) => any,
  params?: Parameters<ZodTypeAny["refine"]>[1],
  opts?: Partial<CustomOptions>
) => ZodType<T>
```

Let me know if this PR needs improvements to meet the guidelines.


colinhacks:
Immaculate PR, thanks!

Dezzj (commented):
Veiwed

Dezzj (commented):
Veiwed

PR #1134, opened 2022-05-14
feat(doc): fix a wrong example at README

dhythm:
# Overview
I found that the example code for `parseAsync` is wrong.
`.refine` should be passed when the callback returns `true`. Thus, I would like to send a PR 🙇‍♂️


colinhacks:
Thanks!!

Issue #213, opened 2020-11-04
Feature Request: Allow generating an object with a full set of defaults that may not validate


abierbaum:
I am looking to use zod as a way to validate user configuration files for a command line utility.  It looks like a great fit compared to what we are doing now, but I am having trouble with one aspect.

The workflow we use is:

- Allocate an initial set of defaults in a nested object  (note: this isn't fully filled in so won't validate the schema)
- Pass to a user provided configuration function (ex. app.config.js) that extends the configuration
- Fill in final settings based upon combination of user and defaults
- Validate final config with schema.  (looking for extra properties, validation checks, etc)

My problem is that in this model I don't see a way to use zod to create the initial defaults object.  I know I could create it as a plain old javascript object separate from the schema, but I was hoping to take advantage of the .default() methods within the schema to have the schema be self describing without having a separate large defaults object.

So my hope was to make something like this work:

```javascript
const ConfigSchema = z.object({
   /** The name to use. */
   name: z.string().default('blah'),

   /** Another nested part. */
   nested: z
      .object({
         value: z.number(),
         val2: z
            .string()
            .nullable()
            .refine((v) => v && v.length === 5, 'Must be length 5.'),
      })
      .default({
         value: 20,
         val2: null,
      }),
});

type ConfigType = z.infer<typeof ConfigSchema>;

describe('zod', () => {
   it('validates a good value', () => {});

   it('can provide defaults to override', () => {
      // We can build a set of default values that don't validate yet
      const default_results = ConfigSchema.safeParse({});  // <--- Need something here to just get defaults
      const cfg = default_results.data;

      expect(cfg.nested.val2).toEqual(null);

      // then: user modifies it with config overrides
      cfg.nested.val2 = '12345';

      // and: now it can validate
      const results = ConfigSchema.safeParse(cfg);
   });
});
```

Is anything like this possible or practical as a feature in zod?  I think it is effectively a SafeParse that doesn't do any checks, but instead just sets the default values on the object passed in.

vriad:
I'm not clear on what you're asking for exactly. Can you provide some simpler examples of the behavior you want?

abierbaum:
Let me try simplifying the example above:

```
const ConfigSchema = z.object({
   name: z.string().default('UserName'),
   image: z.string().nullable().default(null).refine(s => s != null && x.length > 5')
});

type ConfigType = z.infer<typeof ConfigSchema>;

// Build a base set of defaults to pass off to user config module to update/add required configuration
[...]

PR #1146, opened 2022-05-17
Add nullableInput

ArnaudBarre:
Hello and thanks for this lib that has completely changed my way of doing client and server side validations.

One pattern that I'm missing is the ability to change the input type while keeping the output type strict. This PR includes my most needed one: Adding nullable type on the input.

## Why? 

I'm using schema["_input"] type to strongly type the initial values of my form, and some values are required but can't have a meaningful default (like choosing an entity from a list) and are set to null.

## Current solution

```ts
export const addNullInput = <A, B, C>(value: ZodType<A, B, C>) =>
  value
    .nullable()
    .refine((v) => v !== null, "Required")
    .transform((v) => v!);
```

The problem is it creates some runtime and type-checking overhead that could be simplified with a support from the lib. 

## Related

https://github.com/colinhacks/zod/issues/213

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -364,6 +364,7 @@ export abstract class ZodType<
     this.superRefine = this.superRefine.bind(this);
     this.optional = this.optional.bind(this);
     this.nullable = this.nullable.bind(this);
+    this.nullableInput = this.nullableInput.bind(this);
     this.nullish = this.nullish.bind(this);
     this.array = this.array.bind(this);
     this.promise = this.promise.bind(this);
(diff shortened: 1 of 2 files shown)


colinhacks:
Interesting use case. It recently because possible to add issues inside transforms, similar to `superRefine`, which unlocks a cleaner approach to this problem:

```ts
  const noNull = <T>(val: T | null, ctx: z.RefinementCtx) => {
    if (val === null) {
      ctx.addIssue({
        code: z.ZodIssueCode.invalid_type,
        expected: z.ZodParsedType.string,
        received: z.ZodParsedType.null,
        message: "Required",
      });
    }
    return val!;
  };
  const schema = z.number().nullable().transform(noNull);
```

You can define `noNull` somewhere and import it wherever it's needed. Because it's generic, it can be used on any nullable schema and the input/output types will be what you expect.

PR #1149, opened 2022-05-18
Added test for enums with defaults

ElianCordoba:
I wasn't sure about how defaults worked with enums so I played around a bit. I thought they may be worth adding to the test suite :)


colinhacks:
Thanks! Can you change `test.only` to `test`?

colinhacks:
Perfect thanks!

PR #1150, opened 2022-05-19
Add README for deno

omar-dulaimi:


colinhacks:
I'd rather not have a full duplicate of the README here. It's 99.9% redundant and will get out of sync quickly. Can you modify this PR to add Deno instructions to the main README instead? Then just copy the primary README to `deno/lib` inside `deno:build`.

PR #1168, opened 2022-05-25
Update README for z.record key type schema

zzyyxxww:
This documents the behaviour when passing two arguments to `z.record`.  
I updated the adjacent sections to what I added to make it consistent.

This is open to suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1075,7 +1075,7 @@ If you want to validate both the keys and the values, use
 `z.record(keyType, valueType)`:
 
 ```ts
-const NoEmptyKeysSchema = z.record(z.string().nonempty(), z.number());
+const NoEmptyKeysSchema = z.record(z.string().min(1), z.number());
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1069,9 +1069,22 @@ userStore["77d2586b-9e8e-4ecf-8b21-ea7e0530eadd"] = {
 }; // TypeError
 ```
 
+### Record key type
+
+If you want to validate both the keys and the values, use
+`z.record(keyType, valueType)`:
+
+```ts
+const NoEmptyKeysSchema = z.record(z.string().nonempty(), z.number());
```
scotttrinh, line 1078: `nonempty` is actually deprecated in favor of `min(1)`.

```suggestion
const NoEmptyKeysSchema = z.record(z.string().min(1), z.number());
```

scotttrinh (changes requested):

PR #1175, opened 2022-05-29
Don't use invalid_type_error for `undefined` fields.

### LINKED ISSUE #1174: Don't let `invalid_type_error` replace the default `required_error`
The default error when parsing `undefined` for a required value is "Required". When a value overrides its type error with `invalid_type_error`, that message also also overrides the default message for undefined values. I would expect that only `required_error` would override that message.

```
z.number().parse(undefined);
// Throws error with message "Required"

z.number({ invalid_type_error: 'Wrong type!' }).parse(undefined);
// Throws error with message "Wrong type!"
// Expected behavior: Throws error with message "Required"
```

kbongort:
Fixes #1174

### DIFF
--- a/README.md
+++ b/README.md
@@ -1076,8 +1076,8 @@ If you want to validate both the keys and the values, use
 
 ```ts
 const NoEmptyKeysSchema = z.record(z.string().min(1), z.number());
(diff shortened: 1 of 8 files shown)


kbongort:
@colinhacks Hi, any chance you could give this a look? Thank you!

PR #1197, opened 2022-06-08
add example of the `.superRefine` `path` property

JonParton:
The ability to target particular fields with the `ctx.addIssue` function when using `.superRefine` wasn't very clear in the docs. 

This adds an explanation and aims to: 

- Show that the `path` property is available for this purpose. 
- Show that the path property is relative to the zod property being `superRefine`'d not the schema root.
- Give a relevant example of it's use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1641,7 +1641,7 @@ const updateUserForm = z.object({
 
 ```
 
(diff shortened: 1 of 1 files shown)


JonParton:
When neatening up this pull request, I did notice that `.refine` does have an explanation of it's `path` property but `.superRefine` does not... I think this is where I picked up and inferred its use the first time I learnt how but missed it when I came back to the docs today. 

In light of this we may not need to repeat some of this knowledge again but just reference to it? However I think one thing that's not hugely clear is the relative nature of the path property (Which is very powerful!)... So may need an addition? 

Noticed this is also explored a bit more in some commented out docs below refine!?

Let me know thoughts on what to do but it did catch me out again today and I had to do some playing with code to figure it out!

JonParton:
Bump...

Issue #1075, opened 2022-04-07
Allow more types in discriminated unions


dbeckwith:
It seems like `z.discriminatedUnion` currently has quite narrow requirements for what you can use it with. Here are some example use cases that seem sound to me (in the sense that the discriminator should still be able to uniquely determine the union member) but currently aren't allowed:

```typescript
const A = z.object({ type: z.literal('a') });
const B = z.object({ type: z.literal('b') });
const C = z.object({ type: z.literal('c') });
const D = z.object({ type: z.literal('d').optional() });
const E = z.object({ type: z.enum(['e', 'ee', 'eee']) });

const AorB = z.discriminatedUnion('type', [A, B]);

// using another discriminated union with the same discriminator as a union member
const AorBorC = z.discriminatedUnion('type', [AorB, C]);

// the discriminator is optional in one union member
const CorD = z.discriminatedUnion('type', [C, D]);

// the discriminator is an enum
const CorE = z.discriminatedUnion('type', [C, E]);
```

Any chance these cases can be supported? It would great if the requirement for discriminated union members could be widened to something like "any type that has the discriminator as a field and the type of the discriminator is a subtype of `string | null | undefined` and is mutually exclusive between the discriminated union members", but supporting at least the cases I mentioned above would still be much appreciated.

scotttrinh:
Yeah, all of those cases seem reasonable. Furthermore, `true`/`false`/`Number` should be valid discriminator types. Basically, it would be good to support any value that TypeScript itself supports, and I think with `Map` we can get the right behavior. PRs welcome!

vlovich:
Another one is `intersection`:
```
z.discriminatedUnion('method', [
  z.intersection(
    z.object({ 'method': 'get' }),
    getSchema,
  ),
  z.intersection(
    z.object({ 'method': 'put' }),
    putSchema,
  ),
])
```

vlovich:
That one can almost be solved by using `.merge` instead of `.intersection` but that's not always possible. `.union`/`.or` also poses a challenge.

watsonsong:
It seems zod not support null and undefined discriminants.
For example I want to define the structure: https://github.com/microsoft/TypeScript/pull/27631
`type Result<T> = { error?: undefined, value: T } | { error: Error };`

I test this code but report error:
```
type Result = { error?: undefined, value: string } | { error: Error };
const ResultSchema = z.discriminatedUnion('error', [
    z.object({error: z.string().min(1)}),
[...]

Issue #1015, opened 2022-03-13
Support multiple discriminated unions on same object


Aaronius:
First of all, thank you for Zod! I'm just starting to use it on a new project. I have a case come up quite frequently that I don't know how to resolve using current Zod functionality. I've asked a similar question [here](https://github.com/colinhacks/zod/discussions/1006) and haven't received a response, so I'm logging an issue considering functionality may need to be added to Zod.

I believe I want multiple discriminated unions on the same object. Here's some code that illustrates what I'm trying to accomplish:

```js
const schema = z.object({
    sku: z.string()
  }).merge(z.discriminatedUnion("product", [
    z.object({
      product: z.literal("shoes"),
      price: z.number(),
      material: z.union([z.literal("leather"), z.literal("suede")])
    }),
    z.object({
      product: z.literal("shirt"),
      price: z.string()
    })
 ]).merge(z.discriminatedUnion("color", [
    z.object({
       color: z.literal("red"),
       size: z.enum(["s", "m"])
    }),
    z.object({
      color: z.literal("blue"),
      size: z.enum(["m", "l"])
    })
  ]);
```

In other words, the object will (1) always have a sku, (2) always have `product` and `price` (and sometime `material`), and (3) always have `color` and `size`.  The combinations would be restricted by the discriminated unions.

This doesn't work, though, because ZodDiscriminatedUnion doesn't have a merge property and ZodDiscriminatedUnion can't be merged into ZodObject.

I'm not particular about the implementation, so long as it's simple (it's a common use case) and preferably does not require `refine` or `superRefine` (though I'd still like to understand how solve it with `refine` or `superRefine` in the meantime). Thanks!

FlorianWendelborn:
Not a fix on zod’s end, but it’s often a good idea to split this up into sub-objects like this:

```ts
const schema = z.object({
	sku: z.string(),
	product: z.discriminatedUnion('type', [
		z.object({
			type: z.literal('shoes'),
			price: z.number(),
			material: z.union([z.literal('leather'), z.literal('suede')]),
		}),
		z.object({
			type: z.literal('shirt'),
			price: z.string(),
		}),
	]),
	color: z.discriminatedUnion('type', [
		z.object({
			type: z.literal('red'),
			size: z.enum(['s', 'm']),
		}),
		z.object({
			type: z.literal('blue'),
			size: z.enum(['m', 'l']),
		}),
	]),
})
```

[...]

PR #1213, opened 2022-06-19
Zod discriminated union discriminator value accepts zod enum

tianhuil:
Zod discriminated union discriminator value accepts zod enum:

```ts
const item = z
  .discriminatedUnion("type", [
    z.object({ type: z.literal("a"), a: z.string() }),
    z.object({ type: z.enum(["b", "c"]), b: z.string() }),
  ])
  .parse({ type: "b", b: "abc" });
```

Updates test, README.md documentation, and some minor changes from lint / prettier.

Requested in https://github.com/colinhacks/zod/issues/1075 and https://github.com/colinhacks/zod/issues/1015.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1617,6 +1617,12 @@ export function deepPartialify<T extends ZodTypeAny>(
       ...schema._def,
       shape: () => newShape,
     }) as any;
+  } else if (schema instanceof ZodUnion) {
+    type Options = [ZodTypeAny, ZodTypeAny, ...ZodTypeAny[]];
+    const options = schema._def.options as Options;
+    return ZodUnion.create(
+      options.map((option) => deepPartialify(option)) as Options
+    ) as any;
   } else if (schema instanceof ZodArray) {
     return ZodArray.create(deepPartialify(schema.element)) as any;
   } else if (schema instanceof ZodOptional) {
--- a/src/types.ts
+++ b/src/types.ts
@@ -1603,7 +1603,9 @@ export type SomeZodObject = ZodObject<
   any
 >;
 
-function deepPartialify(schema: ZodTypeAny): any {
+export function deepPartialify<T extends ZodTypeAny>(
+  schema: T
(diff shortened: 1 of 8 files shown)


tianhuil:
@colinhacks -- any thoughts?

src/types.ts
```diff
@@ -2232,11 +2267,22 @@ export class ZodDiscriminatedUnion<
...
+          options.set(schema.value, type);
+          numDiscriminatorValues += 1;
+        } else if (schema instanceof ZodEnum) {
+          schema.options.forEach((value: DiscriminatorValue & string) =>
+            options.set(value, type)
+          );
+          numDiscriminatorValues += schema.options.length;
+        } else {
+          throw new Error("Invalid types provided to ZodDiscriminatedUnion");
+        }
```
scotttrinh, line 2285: Do we need this runtime check and error here?

src/types.ts
```diff
@@ -1603,7 +1603,9 @@ export type SomeZodObject = ZodObject<
   any
 >;
 
-function deepPartialify(schema: ZodTypeAny): any {
+export function deepPartialify<T extends ZodTypeAny>(
```
scotttrinh, line 1606: I think it would be a better to split off this update to `deepPartialify` since it's not necessary for the rest of the work here.

scotttrinh (changes requested):

maxArturo:
Hello @tianhuil @scotttrinh ! I'm wondering if this is still being worked on? It seems that the PR's feature [already works](https://github.com/colinhacks/zod/issues/1075#issuecomment-1324937116). Just asking because I ran into this while implementing https://github.com/colinhacks/zod/pull/1589.

tianhuil:
@scotttrinh / @maxArturo: I'm happy to resubmit with just the changes concerning `deepPartialify()` to add union and discriminated union, if that is still of interest.  Sorry, I stopped paying attention once the issue became stale but happy to re-engage.

Issue #1085, opened 2022-04-15
Support for schemas that is an enum of the keys of an object?


bravely:
Essentially what I'm looking for would be something akin to this:

```typescript
const Inputs = Z.object({
  email: Z.string().email(),
  password: Z.string().min(8),
  username: Z.string().min(3),
  id: Z.string().uuid(),
  base: Z.unknown().optional(),
})
const InputFields: Z.ZodEnum<['email', 'password', 'username', 'id', 'base']> = Z.keyof(Inputs)
```

Is any form of the above currently possible, and if not, would anything bar its creation?

scotttrinh:
Hmmm.. Nothing comes to mind already. One tricky thing here is that you could have a schema like `z.object({}).passthrough()` which should make the `keyof` type `never`, which might be a little tricky.

I don't think there is anything _impossible_ here, and PRs are welcome!

ecyrbe:
Hello, i'm also in need for this feature. As i need to use an enum to validate user input based only on a provided schema for [zodios](https://github.com/ecyrbe/zodios) library.
Could you point to me where to start or check if i want to propose a Pull Request for this .

PR #1216, opened 2022-06-20
feat(enum): return enum from object keys

ecyrbe:
This should close this issue: #1085 

usage
```typescript
const result: Z.ZodEnum< [ 'a', 'b' ] > = z.object({ a: z.string(), b: z.string().optional() }).enum();
```  
  
it allows to extract the keys of a zob object schema as a zod enum.

Tell me if it's ok.


ecyrbe:
For information, i use an AI that corrects automatically markdown text and that fixes table of content.  
  
If you want me to revert the fixes it did to zod documention, please telle me.

colinhacks:
That conversion from a union to a tuple is some black magic. Love it.

ecyrbe:
@colinhacks yeah, i became a typescript vodoo master recently. Don't ask me what it cost me 🤣

betafcc:
Really cool trick @ecyrbe, discovered half of it to make [tuple permutations](https://gist.github.com/betafcc/f9084dd9f42ffea59ad6f2ff366b864f) but didn't occur to me to use the function overload trick to get ordering, nice.

Do beware of how [TS treats key order differently than JS engines](https://2ality.com/2015/10/property-traversal-order-es6.html#traversing-the-own-keys-of-an-object) though:

```ts
z.object({ a: z.string(), '1': z.string() }).keyof()._def.values
// TS says it's ['a', '1']
// JS runtime it's ['1', 'a']
```

Seems to not be an issue for internal use with zod enum, but thought pointing this out can be of some value since the test cases only have non numeral strings

PR #1219, opened 2022-06-23
docs: fix typo

jackton1:


Svish:
If I read it correctly, I think the text is actually correct as it is, and the correction makes it incorrect?

The type is `NaN`, so although a bit confusing, it "must be _not_ a number", I think 🤔

jackton1:
Oh, Interesting I see the word play now.

PR #1222, opened 2022-06-24
Added native support for min and max date validations

### LINKED ISSUE #1089: Request: ZodDate min and max
I'm happy to give this a shot myself if you think it's worthwhile to add!

I get that this can currently be achieved with a refinement:

```js
z.date().refine((d) => d >= new Date("2020-01-01") && d <= new Date("2030-01-01"))
```

... but it seems like a reasonable top-level concept to me:

```js
z.date().min(new Date("2020-01-01")).max(new Date("2030-01-01"))
```

Balastrong:
This PR fixes #1089.

`ZodDate` now can natively validate for `min` and `max` allowed dates.

### DIFF
--- a/deno/lib/ZodError.ts
+++ b/deno/lib/ZodError.ts
@@ -97,14 +97,14 @@ export interface ZodTooSmallIssue extends ZodIssueBase {
   code: typeof ZodIssueCode.too_small;
   minimum: number;
   inclusive: boolean;
-  type: "array" | "string" | "number" | "set";
+  type: "array" | "string" | "number" | "set" | "date";
 }
 
 export interface ZodTooBigIssue extends ZodIssueBase {
   code: typeof ZodIssueCode.too_big;
   maximum: number;
   inclusive: boolean;
-  type: "array" | "string" | "number" | "set";
+  type: "array" | "string" | "number" | "set" | "date";
 }
 
 export interface ZodInvalidIntersectionTypesIssue extends ZodIssueBase {
@@ -360,6 +360,10 @@ export const defaultErrorMap = (
         message = `Number must be greater than ${
           issue.inclusive ? `or equal to ` : ``
         }${issue.minimum}`;
+      else if (issue.type === "date")
+        message = `Date must be greater than or equal to ${new Date(
+          issue.minimum
+        )}`;
       else message = "Invalid input";
       break;
     case ZodIssueCode.too_big:
@@ -375,6 +379,10 @@ export const defaultErrorMap = (
         message = `Number must be less than ${
           issue.inclusive ? `or equal to ` : ``
         }${issue.maximum}`;
+      else if (issue.type === "date")
+        message = `Date must be smaller than or equal to ${new Date(
+          issue.maximum
+        )}`;
       else message = "Invalid input";
       break;
     case ZodIssueCode.custom:
--- a/deno/lib/ZodError.ts
+++ b/deno/lib/ZodError.ts
@@ -361,9 +361,9 @@ export const defaultErrorMap = (
           issue.inclusive ? `or equal to ` : ``
         }${issue.minimum}`;
       else if (issue.type === "date")
-        message = `Date must be greater than or equal to ${new Date(
-          issue.minimum
-        )}`;
+        message = `Date must be greater than ${
+          issue.inclusive ? `or equal to ` : ``
+        }${new Date(issue.minimum)}`;
       else message = "Invalid input";
       break;
     case ZodIssueCode.too_big:
@@ -380,9 +380,9 @@ export const defaultErrorMap = (
           issue.inclusive ? `or equal to ` : ``
         }${issue.maximum}`;
       else if (issue.type === "date")
-        message = `Date must be smaller than or equal to ${new Date(
-          issue.maximum
-        )}`;
+        message = `Date must be smaller than ${
+          issue.inclusive ? `or equal to ` : ``
+        }${new Date(issue.maximum)}`;
       else message = "Invalid input";
       break;
     case ZodIssueCode.custom:
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -1015,9 +1015,9 @@ export interface ZodDateDef extends ZodTypeDef {
 export class ZodDate extends ZodType<Date, ZodDateDef> {
   _parse(input: ParseInput): ParseReturnType<this["_output"]> {
     const parsedType = this._getType(input);
-    const ctx: undefined | ParseContext = this._getOrReturnCtx(input);
 
     if (parsedType !== ZodParsedType.date) {
+      const ctx = this._getOrReturnCtx(input);
       addIssueToContext(ctx, {
         code: ZodIssueCode.invalid_type,
         expected: ZodParsedType.date,
@@ -1027,6 +1027,7 @@ export class ZodDate extends ZodType<Date, ZodDateDef> {
     }
 
     if (isNaN(input.data.getTime())) {
+      const ctx = this._getOrReturnCtx(input);
       addIssueToContext(ctx, {
         code: ZodIssueCode.invalid_date,
       });
@@ -1034,10 +1035,12 @@ export class ZodDate extends ZodType<Date, ZodDateDef> {
     }
 
     const status = new ParseStatus();
+    let ctx: undefined | ParseContext = undefined;
 
     for (const check of this._def.checks) {
       if (check.kind === "min") {
         if (input.data.getTime() < check.value) {
+          ctx = this._getOrReturnCtx(input, ctx);
           addIssueToContext(ctx, {
             code: ZodIssueCode.too_small,
             message: check.message,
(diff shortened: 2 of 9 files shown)


deno/lib/ZodError.ts
```diff
@@ -360,6 +360,10 @@ export const defaultErrorMap = (
         message = `Number must be greater than ${
           issue.inclusive ? `or equal to ` : ``
         }${issue.minimum}`;
+      else if (issue.type === "date")
+        message = `Date must be greater than or equal to ${new Date(
```
Svish, line 364: Should use issue.inclusive to decide whether the "or equal to" should be added, as is done for other types.

deno/lib/ZodError.ts
```diff
@@ -375,6 +379,10 @@ export const defaultErrorMap = (
         message = `Number must be less than ${
           issue.inclusive ? `or equal to ` : ``
         }${issue.maximum}`;
+      else if (issue.type === "date")
+        message = `Date must be smaller than or equal to ${new Date(
```
Svish, line 383: Should use issue.inclusive to decide whether the "or equal to" should be added, as is done for other types.

deno/lib/types.ts
```diff
@@ -1004,38 +1004,91 @@ export class ZodBoolean extends ZodType<boolean, ZodBooleanDef> {
...
         code: ZodIssueCode.invalid_date,
       });
       return INVALID;
     }
 
+    const status = new ParseStatus();
+
+    for (const check of this._def.checks) {
+      if (check.kind === "min") {
+        if (input.data.getTime() < check.value) {
```
Svish, line 1040: In your error message, you hard-coded `or equal to`, but here you use `<`. For dates, it would (in my opinion) make more sense if `min` and `max` were inclusive. So I would switch this to `<=` and add `inclusive: true` to the issue generated if it's not valid.

deno/lib/types.ts
```diff
@@ -1004,38 +1004,91 @@ export class ZodBoolean extends ZodType<boolean, ZodBooleanDef> {
...
+          addIssueToContext(ctx, {
+            code: ZodIssueCode.invalid_date,
+            message: check.message,
+          });
+          status.dirty();
+        }
+      } else if (check.kind === "max") {
+        if (input.data.getTime() > check.value) {
```
Svish, line 1048: In your error message, you hard-coded `or equal to`, but here you use `>`. For dates, it would (in my opinion) make more sense if `min` and `max` were inclusive. So I would switch this to `>=` and add `inclusive: true` to the issue generated if it's not valid.

src/ZodError.ts
```diff
@@ -360,6 +360,10 @@ export const defaultErrorMap = (
         message = `Number must be greater than ${
           issue.inclusive ? `or equal to ` : ``
         }${issue.minimum}`;
+      else if (issue.type === "date")
+        message = `Date must be greater than or equal to ${new Date(
```
Svish, line 364: Should use issue.inclusive to decide whether the "or equal to" should be added, as is done for other types.

src/ZodError.ts
```diff
@@ -375,6 +379,10 @@ export const defaultErrorMap = (
         message = `Number must be less than ${
           issue.inclusive ? `or equal to ` : ``
         }${issue.maximum}`;
+      else if (issue.type === "date")
+        message = `Date must be smaller than or equal to ${new Date(
```
Svish, line 383: Should use issue.inclusive to decide whether the "or equal to" should be added, as is done for other types.

src/types.ts
```diff
@@ -1004,38 +1004,91 @@ export class ZodBoolean extends ZodType<boolean, ZodBooleanDef> {
...
         code: ZodIssueCode.invalid_date,
       });
       return INVALID;
     }
 
+    const status = new ParseStatus();
+
+    for (const check of this._def.checks) {
+      if (check.kind === "min") {
+        if (input.data.getTime() < check.value) {
```
Svish, line 1040: In your error message, you hard-coded `or equal to`, but here you use `<`. For dates, it would (in my opinion) make more sense if `min` and `max` were inclusive. So I would switch this to `<=` and add `inclusive: true` to the issue generated if it's not valid.

src/types.ts
```diff
@@ -1004,38 +1004,91 @@ export class ZodBoolean extends ZodType<boolean, ZodBooleanDef> {
...
+          addIssueToContext(ctx, {
+            code: ZodIssueCode.invalid_date,
+            message: check.message,
+          });
+          status.dirty();
+        }
+      } else if (check.kind === "max") {
+        if (input.data.getTime() > check.value) {
```
Svish, line 1048: In your error message, you hard-coded `or equal to`, but here you use `>`. For dates, it would (in my opinion) make more sense if `min` and `max` were inclusive. So I would switch this to `>=` and add `inclusive: true` to the issue generated if it's not valid.

Svish (changes requested):
Added some comments that I hope can be changed, to make it work a bit more consistently with other checks and errors.

* `min` and `max` should be inclusive (to match other `min` and `max` checks)
* issue generated should include `inclusive: true`
* `inclusive` should be used in error map to check whether `or equal to` should be used or not. Even though the current `min` and `max` check (should) be hard-coded to be inclusive, I can easily create my own non-inclusive issue from a refinement or transform, so it's good if the default error map takes that into account, like e.g. the one for `number` does.

Balastrong:
Thank you @Svish for the review! I'll be really happy to conform to the library standards :)

I'm a bit confused, the checks are already inclusive by default.
`if (input.data.getTime() > check.value) { throw error }` already means that if the submitted value equals the max, it is considered valid.

That's also why the hardcoded message - it's inclusive by default.

I get the feedback of adding the `inclusive` field to show the error message as it might bet used by something else in the future, but the min/max check looks correct to me.

I'll push asap the new changes :)

Svish:
> Thank you @Svish for the review! I'll be really happy to conform to the library standards :)
> 
> I'm a bit confused, the checks are already inclusive by default. `if (input.data.getTime() > check.value) { throw error }` already means that if the submitted value equals the max, it is considered valid.
> 
> That's also why the hardcoded message - it's inclusive by default.
> 
> I get the feedback of adding the `inclusive` field to show the error message as it might bet used by something else in the future, but the min/max check looks correct to me.
> 
> I'll push asap the new changes :)

Oh! Sorry, I must be read the checks backwards. I think I just expected to see an `<=` when you're message read "or equal". My bad! But yeah, more _inclusive_ error messages at least is great 😁

src/types.ts
```diff
@@ -1004,38 +1004,97 @@ export class ZodBoolean extends ZodType<boolean, ZodBooleanDef> {
...
+  | { kind: "max"; value: number; message?: string };
 export interface ZodDateDef extends ZodTypeDef {
+  checks: ZodDateCheck[];
   typeName: ZodFirstPartyTypeKind.ZodDate;
 }
 
 export class ZodDate extends ZodType<Date, ZodDateDef> {
   _parse(input: ParseInput): ParseReturnType<this["_output"]> {
     const parsedType = this._getType(input);
+    const ctx: undefined | ParseContext = this._getOrReturnCtx(input);
```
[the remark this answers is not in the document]
Balastrong: Sure things! I didn't know the implication, now I took as example `ZodString` and pushed, thanks :)

src/__tests__/date.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const minCheck = z.date().min(benchmarkDate);
+const maxCheck = z.date().max(benchmarkDate);
+
+test("passing validations", () => {
+  minCheck.parse(new Date(benchmarkDate));
+  minCheck.parse(new Date(benchmarkDate.getTime() + 1));
+
+  maxCheck.parse(new Date(benchmarkDate));
+  maxCheck.parse(new Date(benchmarkDate.getTime() - 1));
```
Svish, line 16: I would feel safer if you added an equality check as well. Both these should pass, if the check is inclusive:

```ts
minCheck.parse(new Date(benchmarkDate.getTime())

maxCheck.parse(new Date(benchmarkDate.getTime())
```

Svish: No, wait, sorry, keep reading things wrong. That's what the previous line without the `.getTime()` does 🤦‍♂️

Nevermind me!

Svish:
Unless it already is possible and I've just missed it in the PR, could you please you add getters for the minimum and maximum date, so that it's possible to get those values "back out" of the schema? Equivalent to how `ZodString` has `minLength` and `maxLength`, so that you for example can do `z.string().min(1).minLength` and get `1`. Could be called `get minDate()` and `get maxDate()` for instance.

These getters makes it possible to set the correct max and min values of an input component from a schema directly, rather than having to pass those in manually. 

Might be possible to just copy them from `ZodString`?
```ts
  get minDate() {
    let min: Date | null = null;
    this._def.checks.map((ch) => {
      if (ch.kind === "min") {
        if (min === null || ch.value > min) {
          min = ch.value;
        }
      }
    });
    return min;
  }

  get maxDate() {
    let max: Date | null = null;
    this._def.checks.map((ch) => {
      if (ch.kind === "max") {
        if (max === null || ch.value < max) {
          max = ch.value;
        }
      }
    });
    return max;
  }
```

Balastrong:
That's indeed a valid suggestion, I'll add them :)

colinhacks:
An immaculate PR — thanks!! 🙌

Issue #1223, opened 2022-06-24
Request: Introduce a convert method, to allow for transforms that may fail


Duncan3142:
When attempting a transform, it's possible that the transform may fail. In this case, it's necessary to return a typecast dummy value.

I think having a 'convert' method on ZodType, which could return a success or failure value, would make for nicer DX.

The convert method would essentially combine a refine and transform, where the refinement failure would be fatal but the transform is necessary to determine the refinement.

The following serves as an example:

```ts
const IdToUser = z
  .string()
  .uuid()
  .convert(async (id) => {
    try {
      const user = await getUserById(id);
      return z.success(user);
    } catch (e) {
      return z.failure([
        {
          code: "custom",
          message: "Unable to fetch user",
        },
      ]);
    }
  });
```

See https://github.com/colinhacks/zod/pull/1224 for a potential implementation

kibertoad:
there is a pr waiting for verdict for this

Duncan3142:
This issue appears to be stale, so I've created a new issue describing an alternate approach

https://github.com/colinhacks/zod/issues/1390

PR https://github.com/colinhacks/zod/pull/1389 is a potential implementation for the above

PR #1224, opened 2022-06-24
Add 'convert' effect to ZodType

Duncan3142:
Add a convert method to ZodType, to allow for transforms that may fail

Closes https://github.com/colinhacks/zod/issues/1223

Hopefully the examples in the README make the use case clear


colinhacks:
Definitely an interesting solution. I believe all of this can already by accomplished with `ctx.addIssue` inside a transform, no? [I'm just realizing this wasn't documented, but it landed in 3.15. https://github.com/colinhacks/zod/pull/1056](https://github.com/colinhacks/zod#validating-during-transform)

Duncan3142:
This PR appears to be stale, so I've submitted an alternative solution

https://github.com/colinhacks/zod/pull/1389

PR #1235, opened 2022-06-29
Add string.startsWith and string.endsWith

grumpyoldman-io:
This adds a startsWith and endsWith helper on the string like discussed here https://github.com/colinhacks/zod/issues/604

small aside, the playground.ts is missing, which is mentioned in the contrib

### DIFF
--- a/README.md
+++ b/README.md
@@ -485,6 +485,8 @@ z.string().length(5, { message: "Must be exactly 5 characters long" });
 z.string().email({ message: "Invalid email address" });
 z.string().url({ message: "Invalid url" });
 z.string().uuid({ message: "Invalid UUID" });
+z.string().startsWith("https://", { message: "Must provide secure URL" });
+z.string().endsWith(".com", { message: "Only .com domains allowed" });
 ```
 
 ## Numbers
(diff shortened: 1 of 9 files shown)


grumpyoldman-io:
CC @alii

alii:
ooh awesome :D thank you!

AustinShelby:
Transforming the parameter users give to the `startsWith` function into a regex doesn't work in certain edge cases.

Consider the following scenario;

```ts
const x = z.string().startsWith("[a-g]");
x.parse("[a-g]");
```

By first look this should work as the string `[a-g]` starts with `[a-g]`.

The problem is that when you transform `[a-g]` into a regex using your implementation it checks if the string starts with any character from `a` to `g`, which it doesn't as it starts with the character `[`.

I suggest using the native `String` functions [startsWith](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) and [endsWith](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith).

PR #1236, opened 2022-06-29
Adds remix-domains to the README under Ecosystem

gustavoguichard:
We'd like to add remix-domains in the README (under ecosystem) as our library is only possible due to the existence of Zod <3 and we believe it improves a lot the DX of Remix.

LMK if this PR is welcome or should I make any changes to it.


colinhacks:
An extremely welcome PR of course :)

Issue #1208, opened 2022-06-15
Allow setting a label and pass it through to error map


Svish:
Don't remember the exact rule, but I believe there is a WCAG requirement that error messages should refer to the name of the field it's about. I.e. `Field is required` is less good than for example `"First name" is required`.

Would be great if we could add a label to our schema, which would be used in the default error map, _and_ be made available in custom error map functions.

```ts
const mySchema = z.object({
  firstName: z.string({ label: 'First name' }),
  dateOfBirth: z.date({ label: 'Date of birth' }),
})
```

I see there's already a `description`, but it's not documented anywhere, so not sure what that is for. If it's actually for a label, then I guess this would be solved simply by exposing that `description` in the error map somehow. Maybe it could come as part of the `ErrorMapCtx` object?

<img src="https://user-images.githubusercontent.com/142162/173804787-a9987eb7-946d-489b-b6b4-4d09679858ff.png" alt="screenshot" width="350"/>

WazzaJB:
I had a stab at this, not sure it's necessarily right or in the right place though given I have attached the info to `issue` as opposed to context...

Find myself in need of this functionality too 👍 

![Screenshot 2022-07-04 at 15 27 33](https://user-images.githubusercontent.com/2930904/177174599-649cfc57-efe3-4880-b91b-bf7b570398cc.png)

Notice the addition of "type" and "description" alongside how I have interpolated the description into the error message, the gist below shows how I achieved it. Pretty nasty implementation just to show in principle how error messages can more closely resemble how Yup handles them.

https://gist.github.com/WazzaJB/2eca3df3369c91317edaad3ed96862a6

zynth17:
any roadmap to implement this?
in `yup` they have label method to override the field name.
this comes in handy when validation an object, so we can rename each field with human readable text not object name.

PR #1241, opened 2022-07-04
Extended issue data to include type and description

WazzaJB:
In line with issue #1208 I required visibility of validator type and description within my error maps to create user friendly errors, this PR adds that support.

I have not yet added documentation or tests but will do so providing I am on the right lines and this is something we want to move forward with. Thanks for Zod!

### DIFF
--- a/deno/lib/ZodError.ts
+++ b/deno/lib/ZodError.ts
@@ -35,8 +35,84 @@ export const ZodIssueCode = util.arrayToEnum([
 
 export type ZodIssueCode = keyof typeof ZodIssueCode;
 
+export const ZodIssueType = util.arrayToEnum([
+  "string",
+  "number",
+  "nan",
+  "bigint",
+  "boolean",
+  "date",
+  "undefined",
+  "null",
+  "any",
+  "unknown",
+  "never",
+  "void",
+  "array",
+  "object",
+  "union",
+  "discriminated_union",
+  "intersection",
+  "tuple",
+  "record",
+  "map",
+  "set",
+  "function",
+  "lazy",
+  "literal",
+  "enum",
+  "effects",
+  "native_enum",
+  "optional",
+  "nullable",
+  "default",
+  "promise",
+  "custom",
+]);
+
+export type ZodIssueType = keyof typeof ZodIssueType;
+
+/**
+ * Map ZodIssueCode into pretty ZodIssueType
+ */
+export enum ZodIssueTypeMap {
+  ZodString = "string",
+  ZodNumber = "number",
+  ZodNaN = "nan",
+  ZodBigInt = "bigint",
+  ZodBoolean = "boolean",
+  ZodDate = "date",
+  ZodUndefined = "undefined",
+  ZodNull = "null",
+  ZodAny = "any",
+  ZodUnknown = "unknown",
+  ZodNever = "never",
+  ZodVoid = "void",
+  ZodArray = "array",
+  ZodObject = "object",
+  ZodUnion = "union",
+  ZodDiscriminatedUnion = "discriminated_union",
+  ZodIntersection = "intersection",
+  ZodTuple = "tuple",
+  ZodRecord = "record",
+  ZodMap = "map",
+  ZodSet = "set",
+  ZodFunction = "function",
+  ZodLazy = "lazy",
+  ZodLiteral = "literal",
+  ZodEnum = "enum",
+  ZodEffects = "effects",
+  ZodNativeEnum = "native_enum",
+  ZodOptional = "optional",
+  ZodNullable = "nullable",
+  ZodDefault = "default",
+  ZodPromise = "promise",
+}
+
 export type ZodIssueBase = {
   path: (string | number)[];
+  type?: ZodIssueType;
+  description?: string;
   message?: string;
 };
 
@@ -97,14 +173,12 @@ export interface ZodTooSmallIssue extends ZodIssueBase {
   code: typeof ZodIssueCode.too_small;
   minimum: number;
   inclusive: boolean;
-  type: "array" | "string" | "number" | "set";
 }
 
 export interface ZodTooBigIssue extends ZodIssueBase {
   code: typeof ZodIssueCode.too_big;
   maximum: number;
   inclusive: boolean;
-  type: "array" | "string" | "number" | "set";
 }
 
 export interface ZodInvalidIntersectionTypesIssue extends ZodIssueBase {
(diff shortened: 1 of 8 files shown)


Svish:
I believe `type` already is included in most issues? It's just sometimes it's called something else, like for the `invalid_type` issue where it's called `expected` instead.

Svish:
Yeah, I'm not sure what would be best here. I don't even require the `type` myself really, I just want a way to label stuff.

For my issue #1208, it feels like the cleanest solution would be to have all schemas add a `label` option, in addition to `description`, and then have both of those be available to the error map and possible to get from a schema. That would cover both being able to include the label in error messages, _and_ adding labels (`<label>` or `labelled-by`) and descriptions (`described-by` attribute) to stuff based upon a schema.

WazzaJB:
To be honest, the existing type would probably be enough to cover my needs too. However, it means you have to conditionally check the issue code in order to satisfy TS that `type` exists in line with the union. This PR goes some way to solve that by making `type` optional on the base issue type.

As an example, the "too_small" and "too_big" messages do not differ based on type (I believe) which can lead to you having a string which is too short throwing a message like "String is too **small**" as opposed to "String is too **short**".

My gist shows how you could create a locale definition similar to Yup and differ your messages based on type:

https://gist.github.com/WazzaJB/2eca3df3369c91317edaad3ed96862a6#file-playground-ts-L4

Svish:
I already check for type within each code, so that's not an issue for me. My current error map looks like this, and I find it fairly clean actually:

```ts
setErrorMap((issue, ctx) => ({ message: zodErrorMap(issue, ctx) }));

function zodErrorMap(...[issue, ctx]: Parameters<ZodErrorMap>): string {
  /* eslint-disable @typescript-eslint/switch-exhaustiveness-check */
  switch (issue.code) {
    case ZodIssueCode.invalid_type:
      switch (issue.expected) {
        case 'number':
          return 'må være et gyldig tall';

        case 'integer':
          return 'kan ikke ha desimaler';

        case 'date':
          return 'må være en gyldig dato';

        case 'array':
          switch (issue.received) {
            case 'null':
            case 'undefined':
              return 'må fylles ut';
            default:
              return 'må være en liste';
          }

        case 'object':
          switch (issue.received) {
            case 'null':
            case 'undefined':
              return 'må fylles ut';
            default:
              return 'må være et objekt';
          }

        case 'string':
          return 'må fylles ut';

        case 'boolean':
          switch (issue.received) {
            case 'null':
            case 'undefined':
              return 'må fylles ut';
            default:
              return 'må være en boolsk verdi';
          }
      }

      // TODO: Replace this mess with `case 'enum':` when possible [torber]
      // @see https://github.com/colinhacks/zod/issues/1217
      if (/^(?:'(\w+)'(?: \| )?)+$/.test(issue.expected)) {
        return 'må være valgt';
      }

      break;

    case ZodIssueCode.too_small:
      switch (issue.type) {
        case 'number':
          return [
            'må være større enn',
            issue.inclusive ? 'eller lik' : '',
            issue.minimum,
          ].join(' ');

        case 'string':
          return issue.minimum === 1 && issue.inclusive
            ? 'må fylles ut'
            : [
                'må være',
                issue.inclusive ? 'minst' : 'lengre enn',
                issue.minimum,
                'tegn',
              ].join(' ');

        case 'array':
          return issue.minimum === 1 && issue.inclusive
            ? 'må fylles ut'
            : [
                'må ha',
                issue.inclusive ? 'minst' : 'flere enn',
                issue.minimum,
              ].join(' ');
      }
      break;

    case ZodIssueCode.too_big:
      switch (issue.type) {
        case 'number':
          return [
            'må være mindre enn',
            issue.inclusive ? 'eller lik' : '',
            issue.maximum,
          ].join(' ');

        case 'string':
          return issue.inclusive
            ? `kan ikke være mer enn ${issue.maximum} tegn`
            : `må være kortere enn ${issue.maximum} tegn`;
      }
      break;

    case ZodIssueCode.invalid_literal:
      return `må være "${String(issue.expected)}"`;

    // ... continued
  }

  // Fallback
  console.warn(`Missing error translation in ${zodErrorMap.name} function`, {
    issue,
    ctx,
  });
  return ctx.defaultError;
}
```

colinhacks:
Hi folks, interesting discussion here. 

I seems like adding a `label` property and passing it into error maps is an unnecessary layer of indirection. Same for `description`, which I only grudgingly added to support various projects that were trying to do code-generation from Zod schemas. 

Error messages can be customized in this way with existing parameters:

```ts
z.string({
  required_error: "First name is required",
  invalid_type_error: "First name is not valid"
});
```

You can build your own abstractions on top of Zod if you find this too verbose.

```ts
const labeledString = (label: string) => z.string({
  required_error: `${label} is required`,
  invalid_type_error: `${label} is not valid`
})
```

I'll leave this open for more discussion for a while but currently I'm not convinced this presents a better alternative to the current APIs. Zod represents types, not forms, so introducing concepts of "labels" and the like strikes me as too domain-specific.

Svish:
@colinhacks 

> I seems like adding a `label` property and passing it into error maps is an unnecessary layer of indirection. Same for `description`, which I only grudgingly added to support various projects that were trying to do code-generation from Zod schemas.
> 
> Error messages can be customized in this way with existing parameters:
> 
> ```ts
> z.string({
>   required_error: "First name is required",
>   invalid_type_error: "First name is not valid"
> });
> ```
> 
> You can build your own abstractions on top of Zod if you find this too verbose.
> 
> ```ts
> const labeledString = (label: string) => z.string({
>   required_error: `${label} is required`,
>   invalid_type_error: `${label} is not valid`
> })
> ```
> 
> I'll leave this open for more discussion for a while but currently I'm not convinced this presents a better alternative to the current APIs. Zod represents types, not forms, so introducing concepts of "labels" and the like strikes me as too domain-specific.

Although I understand the wish to keep Zod focused, this basically removes us being able to use the `error_map`.

Error messages _can_ be customized with existing parameters, but that would require us to customize _all_ error messages, for _every_ rule, _everywhere_. Especially with teams, it's difficult to remember doing this everywhere, and also to be consistent  everywhere. Not to mention the mess it would be if we needed to reword one of those later.

Building abstractions on top of Zod would also be quite a mess, in my opinion, especially since you'd also need to override all the error messages for other rules, like `min` and `max`. Not even sure how we'd do that, without things becoming super messy.

Zod is such a nice library, and the `error_map` has been very nice and clean for creating a easy to maintain, single source, of translated error messages. The only thing it misses, is the ability to include a label in the error message, so that the error messages ca be more readable and clear to users, and accessibility compliant.

Really hope you can reconsider this issue. A label (or description, or whatever you want to call it) that can be attached to a rule, and be available via a getter and to the error map, would make Zod basically perfect.

WazzaJB:
@Svish has summed up my thoughts better than I could have too.

I think keeping Zod focussed is definitely sensible, but as it stands it's not really flexible enough (I don't think) to set nice global error rules. The ecosystem around Zod is super impressive and I feel that some flexibility here would create lots of opportunity.

What if we took a step back to something like a `meta` property which was available in the error mapper? It would provide tonnes of flexibility and not carry any opinions or preferential use? Having this `meta` interface overridable in some way would mean Zod dependants could enforce typings on some meta keys?

With this we wouldn't be restricted to just a label/description and this might mean there is less maintenance overhead? Ie use meta for whatever you want and it's your job as a consumer not to break it?

Apologies if I am not conveying my thoughts well or they are too far from Zod's mission.

I'm going to have a go at wrapping Zod as you have suggested and see if I can arrive at a similar result regardless.

WazzaJB:
@colinhacks would it be fair to me to say that based on the signature of "RawCreateParams" I can only set the invalid type and required error strings? In order to replace invalid_regex or min/max errors for example with labels, I would be unable to?

WazzaJB:
Well, I got something working... https://gist.github.com/WazzaJB/5700b07c6bcf0c5847a2a50051f68310

Just use "label" instead of Zod and set the error map there. I am sure there are downsides to this approach but not quite sure what.

tonyxiao:
@colinhacks is code generation an important use case for Zod? We're doing a lot of the same stuff and thinking of storing JSON.stringified string inside description field. 🙈

PR #1248, opened 2022-07-06
ZodString.minLength defaults to null to be consistent with maxLength

### LINKED ISSUE #1246: Getter maxLength in ZodString has incorrect type (null)
**Snippet:**

```typescript
const s = z.string().max(10);
type L = typeof s.maxLength;
```

**Expected:**

```typescript
type L = number | null        // or ideally just number
```

**Actual:**

```typescript
type L = null
```

**Setup:**
- zod v3.17.3
- typescript v4.7.4

**Context:**

This is the current implementation of `maxLength`.

```typescript
  get maxLength() {
    let max: number | null = null;
    this._def.checks.map((ch) => {
      if (ch.kind === "max") {
        if (max === null || ch.value < max) {
          max = ch.value;
        }
      }
    });
    return max;        // typescript considers max to be of type null here
  }
```

The issue here is that Typescript ignores callbacks when it comes to type narrowing (assumes that they won't have any side effects). See https://github.com/microsoft/TypeScript/issues/35593 and https://github.com/microsoft/TypeScript/issues/9998. Since the only other operation on the variable `max` is its assignment to `null`, Typescript will narrow down its type to just `null`.

**Potential Fixes:**

- Add an explicit return type annotation of `number | null`.
- Replace `.map` with a for-of loop.

Balastrong:
This PR fixes #1246.

`minLength` and `maxLength` are inconsistent in `ZodString`, this PR makes the behaviour more consistent by returning `null` if no max is defined (not changed) and if no min is defined (instead of `-Infinity`).

This PR also improves the syntax by removing the incorrect usage of `Array.map` that had side effects & no return).

A couple tests are also added to make sure it works as intended.

### DIFF
--- a/deno/lib/__tests__/string.test.ts
+++ b/deno/lib/__tests__/string.test.ts
@@ -166,9 +166,11 @@ test("checks getters", () => {
 test("min max getters", () => {
   expect(z.string().min(5).minLength).toEqual(5);
   expect(z.string().min(5).min(10).minLength).toEqual(10);
+  expect(z.string().minLength).toEqual(-Infinity);
 
   expect(z.string().max(5).maxLength).toEqual(5);
   expect(z.string().max(5).max(1).maxLength).toEqual(1);
+  expect(z.string().maxLength).toEqual(Infinity);
 });
 
 test("trim", () => {
--- a/deno/lib/__tests__/string.test.ts
+++ b/deno/lib/__tests__/string.test.ts
@@ -166,11 +166,11 @@ test("checks getters", () => {
 test("min max getters", () => {
   expect(z.string().min(5).minLength).toEqual(5);
   expect(z.string().min(5).min(10).minLength).toEqual(10);
-  expect(z.string().minLength).toEqual(-Infinity);
+  expect(z.string().minLength).toEqual(null);
 
   expect(z.string().max(5).maxLength).toEqual(5);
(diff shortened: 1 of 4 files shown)


ealmansi-uc:
Looking at other types in the library, it seems to be that the most consistent API would be to return `number | null` for both min and max, rather than returning `number`. Some examples below:

```typescript
// ZodNumber
    get minValue(): number | null;
    get maxValue(): number | null;
```

```typescript
// ZodArrayDef
    minLength: {
        value: number;
        message?: string;
    } | null;
    maxLength: {
        value: number;
        message?: string;
    } | null;
```

```typescript
// ZodSetDef
    minSize: {
        value: number;
        message?: string;
    } | null;
    maxSize: {
        value: number;
        message?: string;
    } | null;
```

Therefore, I would suggest this instead:

```typescript
// ZodString
    get minLength(): number | null;
    get maxLength(): number | null;
```

Balastrong:
Hey @ealmansi-uc, I guess you're right. I have updated the PR to reflect the same logic used in `ZodNumber` and use `number | null`. Should look better now. Thank you!

ealmansi-uc:
@Balastrong LGTM now, although I'm not a contributor to this repo 😅  you'll probably want to update the PR's title
keep in mind this is a breaking change, I'm not sure how conservative this project is around breaking changes

colinhacks:
Thanks, good PR!

Issue #1240, opened 2022-07-03
how to express number string


UNDERCOVERj:
our team handle i64 into string，so how should we validate this number string type

AustinShelby:
What kind of validations are you planning on doing? One thing you can do is first transforming the string into a [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) so you can do simple comparisons.

UNDERCOVERj:
I just want to validate a string which is like numeric type。In class-validator, it is expressed as IsNumberString

UNDERCOVERj:
>

UNDERCOVERj:
> I just want to validate a string which is like numeric type。In class-validator, it is expressed as IsNumberString

AustinShelby:
If by `i64` you mean the Rust primitive type https://doc.rust-lang.org/std/primitive.i64.html then you can either use `refine` to write a custom validator function to check if the string is a `BigInt`.

```ts
const numberString = z.string().refine((value) => {
  try {
    BigInt(value);
    return true;
  } catch (e) {
    return false;
  }
});
```

or use a regex to check if the string only contains numbers

```ts
const numberString = z.string().regex(/^[0-9]*$/)
```

UNDERCOVERj:
> If by `i64` you mean the Rust primitive type https://doc.rust-lang.org/std/primitive.i64.html then you can either use `refine` to write a custom validator function to check if the string is a `BigInt`.
> 
> ```ts
> const numberString = z.string().refine((value) => {
>   try {
>     BigInt(value);
>     return true;
>   } catch (e) {
>     return false;
>   }
> });
> ```
> 
> or use a regex to check if the string only contains numbers
> 
> ```ts
> const numberString = z.string().regex(/^[0-9]*$/)
> ```

ok，thank you very much。Can I add numberString as a method in z？

icecreamparlor:
Not solved yet.
Fixed with PR https://github.com/colinhacks/zod/pull/1570

PR #1266, opened 2022-07-18
feat: support numberic string

UNDERCOVERj:
I want to express number string。just use `z.string().numeric()` to replace `z.string().regex(/^[0-9]*$/)`

https://github.com/colinhacks/zod/issues/1240


RobinTail:
what about negative and float numbers?

UNDERCOVERj:
> what about negative and float numbers?

I think you are right，I think there should have a way for users to provide custom validators for ZodString

PR #1273, opened 2022-07-20
Fix endsWith default message

carlosriveroib:
The endsWith default message is "Invalid input: must start with" and should be "Invalid input: must end with"


colinhacks:
Oops 🤦‍♂️

PR #1280, opened 2022-07-22
Refactor type checks

colinhacks:


ecyrbe:
conditional type checks lib seems hacky and over complicated.
can't you just use this :  
```typescript
export type IfEquals<T, U, Y = unknown, N = never> = (<G>() => G extends T
  ? 1
  : 2) extends <G>() => G extends U ? 1 : 2
  ? Y
  : N;
```
i use it for zodios and it seems to pass all my checks.  
i made an article about it [here on dev.to](https://dev.to/ecyrbe/how-to-unit-test-your-typescript-utility-types-3cnm)

colinhacks:
Sure, sounds good! That's some dark type magic.

Issue #1257, opened 2022-07-12
Incorrect validation with `optional` and `default` properties


leoavelino7:
## The problem 

I noticed that when setting the `optional` property before the `default` the validation does not have the same behavior.
Look at the example below:

```tsx

const mySchema1 = z.object({
  name: z.string(),
  surname: z.string().min(1).max(120).optional().default("")
});

const mySchema2 = z.object({
  name: z.string(),
  surname: z.string().min(1).max(120).default("").optional()
});

const data = {
  name: "Leo"
};

const result1 = mySchema1.safeParse(data);
/**
 * Result:
 * {
 *  "success": false,
 *  "error": {
 *    "issues": [
 *      {
 *        "code": "too_small",
 *        "minimum": 1,
 *        "type": "string",
 *        "inclusive": true,
 *        "message": "String must contain at least 1 character(s)",
 *        "path": ["surname"]
 *      }
 *    ],
 *    "name": "ZodError"
 *  }
 * }
 */

const result2 = mySchema2.safeParse(data);
/**
 * Result:
 * { "success": true, "data": { "name": "Leo" } }
 */
```
I expected `result1` to have the same behavior as `result2`.

### Possible fix
Change type `default(def: util.noUndefined<Input>): ZodDefault<this>;` to also accept `undefined`

```tsx

const mySchema1 = z.object({
  name: z.string(),
  surname: z.string().min(1).max(120).optional().default(undefined)
});

```

### **Packages and versions:**
"zod": 3.17.3
"typescript": "4.5.5"

colinhacks:
Behaving as intended. Order matters here. Think of each method call as adding another layer to the onion. When you call `parse`, the layers are unwrapped.

First schema:
- initial:  `undefined`
- `.default("")`: `undefined` becomes `""`
- `.optional()`: has no effect because the value is defined
- `.min(1)`: throws because "" has length zero

Second schema
- initial:  `undefined`
- `.optional()`: returns `undefined` as final value

General rule, there's no reason to use both `optional` and `default` at the same time. Calling `default` makes the schema optional implicitly

cj:
@colinhacks I just run into this issue with ZInfer, it does not make those fields optional:

![image](https://user-images.githubusercontent.com/1819/180666384-ab8c11ad-046e-41f4-a0bb-5758b832d200.png)

PR #1285, opened 2022-07-24
Fix ZodDefault not being optional for infer

cj:
This fixes ZodDefault not being optional for infer https://github.com/colinhacks/zod/issues/1257#issuecomment-1193396806


colinhacks:
Nope, this is very intentional. A schema with a default returns `undefined`.

cj:
@colinhacks thank you for the clarification!

Issue #1287, opened 2022-07-25
Feature request: add external params for custom refine and transform


roblabat:
In my use case I would like to validate some fields of an object in front of an other dynamic object. I didn't found a way to do that with the current implementation . I think the best way to do that would be to add custom parameters to the parse functions that would be accessible by the callback of transform and refine functions (for exemple with a ctx property):

```ts
const foo = z.object({
  id: z.string(),
  data: z.unknown()
});

consr bar = z.object({
  fooRef: z.string().refine((data, ctx) => ctx.customParams?.include(data))
});

const a = z.parse({id: 'a', data: 'a data'});
const b = z.parse({id: 'b', data: 'b data'});

const ids = [a,b].map(obj => obj.id);

const obj1 = z.parse({fooRef: 'b'}, ids);  // successfully returns {ref: 'b'}
const obj2 = z.parse({fooRef: 'c'}, ids);  // throw a  parsing error
```

This exemple could be extended with transform to directly parse obj1 as {foo: {data: 'b data'}} for exemple.

PR #1288, opened 2022-07-25
[#1287] add params to transform and refine

roblabat:
This PR solve the issue #1287 by adding the ability to add a params object as optional argument to parse functions, this params argument will be available in transform and refine callbacks to perform more customizable data checks and transformation.

I will extend this PR with a few doc when I'll get some time.

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -53,47 +53,6 @@ export type TypeOf<T extends ZodType<any, any, any>> = T["_output"];
 export type input<T extends ZodType<any, any, any>> = T["_input"];
 export type output<T extends ZodType<any, any, any>> = T["_output"];
 export type { TypeOf as infer };
-export { ZodEffects as ZodTransformer };
-export { ZodType as Schema, ZodType as ZodSchema };
-export {
-  anyType as any,
-  arrayType as array,
-  bigIntType as bigint,
-  booleanType as boolean,
-  dateType as date,
-  discriminatedUnionType as discriminatedUnion,
-  effectsType as effect,
-  enumType as enum,
-  functionType as function,
-  instanceOfType as instanceof,
-  intersectionType as intersection,
-  lazyType as lazy,
-  literalType as literal,
-  mapType as map,
-  nanType as nan,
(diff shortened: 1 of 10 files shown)


deno/lib/types.ts
```diff
@@ -45,13 +45,55 @@ import {
 export type RefinementCtx = {
   addIssue: (arg: IssueData) => void;
   path: (string | number)[];
+  params?: any;
```
UNDERCOVERj, line 48: this should be Input ?

roblabat: No this is a parameter object added by the PR to the parse method, the idea of this parameter is to customization to refinement based on data available at runtime during the parsing. I typed it as any as typing it would mean to add a param generic type to ZodType and propagate it across different implementations of ZodType, it would be great but I didn't add time to do that yet. It would be a subject for a future PR.

colinhacks:
This is a good PR but I'm generally opposed to contextual validations. Zod schemas should be self-contained, and schema that rely on params break encapsulation. https://github.com/colinhacks/zod/issues/487#issuecomment-860231900

roblabat:
@colinhacks 
Ok, I'm not sure of the benefits of keeping shema self-contained no mater what as for certain use cases it would lead into generating the schema in the data validation process. Which is, I think, a bad pattern as it would cause performance issues as you would re-declare a new schema object each time you want to perform a validation losing all the benefits of declaring a reusable self-contained schema.
I think it's a trade off between strict encapsulation and performance and even if I won't encourage the pattern I think it's not the role of the library to make this choice for the users. But a warning about encapsulation breaking of this pattern could be nice in the the documentation of the functionality.

roblabat:
I'm closing this PR after @colinhacks arguments and example of workaround mentioned in https://github.com/colinhacks/zod/issues/487#issuecomment-860231900

I'm not 100% sure about the argument of strict encapsulation but I understand it and acknowledge its relevance.
I would also argu that strict typings for that functionnality would be a big work that make this PR not mature enough for merge. 

Feel free to take back this code if you have interest on it and want to go further on this subject.

Issue #1176, opened 2022-05-30
Allow specific properties to be `.required`


SrBrahma:
I have a mySchema with ~10 properties.

In my API, I use `mySchema.partial()` as all props are optional, BUT one.

I wish I could simply `mySchema.partial().required({ myRequired: true })`, as partial allows.

It would be a better solution than .partial({ 9 props }) or merging with the mySchema.shape.myRequired.

ivosabev:
+1 fo this, it is very common scenario. 

For example I have a `orderSchema` and I have an `updateOrder()` function that takes `orderSchema.partial()`, but I want the `orderSchema.shape.id` property to be required, so I would like to do `orderSchema.partial().required({id:true})` instead of the hack I am currently using `orderSchema.partial().omit({id:true}).extend({id: orderIdSchema})`

Antman261:
+1 -- would like this so zod can mirror the TypeScript `Required` utility type https://www.typescriptlang.org/docs/handbook/utility-types.html#requiredtype

YannBcf:
+1, I also needed this feature, thanks for your workaround @ivosabev

SrBrahma:
Just implemented it in #1315.

Can you please take a look, @colinhacks?

SrBrahma:
Up

SrBrahma:
Merged in #1315

PR #1315, opened 2022-08-05
Add `mask` parameter to `.required` method

SrBrahma:
This solves the #1176 by adding the `mask` parameter to the `.required()` method.

* I have used as base the `.partial(mask)` method.
* I have also added the `.required` to the README (it didn't exist before) with its new mask explanation.
* Tests added and renamed the `with mask` test to `partial with mask` as we now have the `required with mask`.
* Deno support


SrBrahma:
Up

scotttrinh (approved):
Looks good to me! I'll check benchmarks to ensure we're not accidentally slowing anything down, too.

scotttrinh:
No noticeable changes in the benchmarks for the existing functionality, so I'd say this is ready to merge! Thanks @SrBrahma for the contribution.

PR #1321, opened 2022-08-07
Add `class` getter to the return of `z.instanceof`

patr0nus:


colinhacks:
This is a well made PR, but I'm not convinced it's worth introducing a new subclass of `ZodType` just to add the `class` getter. Is there a compelling argument in favor? Keep in mind that I'm in a continual struggle to keep the bundle size of Zod low - it's at 10KB currently and I get complaints with some regularity.

PR #1342, opened 2022-08-17
feat: add `decode` and `safeDecode` methods to check type at build time instead of runtime

julienfouilhe:
This adds the following methods:

- `decode`
- `safeDecode`
- `decodeAsync`
- `safeDecodeAsync`

Those methods are extremely similar to their `parse` counterparts, the differences are:
- They expect a `Input` as an argument instead of `unknown`.

Using zod as such is therefore possible:

```ts
const schema = z.object({ k:  z.number().min(1) });

const valid = { k: 12 };
const invalidWrongType = "value";
const invalidValue = { k: 0 };
const invalidUnknown = { k: 12 } as unknown;

schema.decode(valid) // success
schema.decode(invalidWrongType) // Typescript error
schema.decode(invalidValue) // Runtime error
schema.decode(invalidUnknown) // Typescript error, should use .parse instead
```

Checks, refiners, transformers are still run, it just delegates type checking to typescript?

The main use case is when you want to apply business rules to multiple fields that can return an error, but primitives types have already been checked some other way, for instance:

- Schema-first gRPC/graphql servers with generated stub files that check messages/queries already.
- Databases accessed using some kind of ORM that already checks primitives.

For instance, I have a gRPC server and I want to transform the message `{ email: string; phoneNumber: string | null }` into `{ email: EmailAddress; phoneNumber: PhoneNumber | null }`

Without something like zod, it can be painful:
```ts
try {
  const email = EmailAddress.parse(request.email);
  const phoneNumber = request.phoneNumber ? PhoneNumber.parse(request.phoneNumber) : null;
  // rest of your code
} catch (e) {
  // handle error
}
``` 

It's even more painful when your `parse` methods are safe and do not throw.

```ts
  const email = EmailAddress.parse(request.email);
  const phoneNumber = request.phoneNumber ? PhoneNumber.parse(request.phoneNumber) : null;
  
  if (email instanceof Error || phoneNumber instanceof Error) {
    // handle error
  }
``` 

With zod and the `decode` methods, I can ensure that my zod schema matches my gRPC message, which results in cleaner and more reusable/composable pieces of code:

```ts
const schema = z.object({ 
  email: z.string().refine(/* ... */).transform(/* ... */),
  phoneNumer: z.string().refine(/* ... */).transform(/* ... */).nullable(), // see the typo here in the key?
})

const parsed = schema.safeDecode(request) // Typescript error, key phoneNumer is expected but received phoneNumber
if (!parsed.success) {
  // handle error
}
```

Some more info in this discussion https://github.com/colinhacks/zod/discussions/1335

I'm not too sure about:
- Should we keep validating the type at runtime to prevent unexpected errors when the type of the input is `any`?
- Are these changes enough, it seems like some structural changes could be good to reduce the number of paths.


README.md
```diff
@@ -1535,6 +1539,65 @@ For convenience, this has been aliased to `.spa`:
 await stringSchema.spa("billie");
 ```
 
+### `.check`
```
scotttrinh, line 1542: Yeah, I think maybe we can iterate on the name to make it more obvious what it does, let me spend a few minutes spitballing and brainstorming:

- `parseInput`/`safeParseInput`: We're still parsing rather than validating, so maybe this makes that more clear and makes it more obvious that the argument it expects is of type `Input`?
- `decode`/`safeDecode`: Maybe this is closer to a codec kind of thing where you have some defined input type that you're decoding from?
- `construct`/`safeConstruct`: Maybe these are "constructors" of this type? Maybe too much of an overloaded term?
- `make`/`safeMake`: Taking a page from the OCaml/ReScript world, most constructors there are named `make`.

src/types.ts
```diff
@@ -265,6 +329,8 @@ export abstract class ZodType<
 
   /** Alias of safeParseAsync */
   spa = this.safeParseAsync;
+  /** Alias of safeCheckAsync */
+  sca = this.safeParseAsync;
```
scotttrinh, line 333:
```suggestion
  sca = this.safeCheckAsync;
```
—?

scotttrinh (changes requested):

README.md
julienfouilhe: I like decode, I feel like it's the name that describes best the fact that transformations are applied?

colinhacks:
Sorry, I'm not really seeing the appeal of this. I get that it's somewhat duplicative, but I recommend using regular `.parse` here? It's weird to provide a method that skips some checks (type validations) but not others (refinements). This distinction is hard to explain to users who aren't familiar with Zod internals. 

I suppose I'm not clear either on whether your concern here is performance or DX/API. It seems like you want to skip validations but apply transforms? You may be better suited just defining your transform logic externally to Zod in this case. I don't think this use case is popular enough to merit a major addition to the API.

julienfouilhe:
@colinhacks My only usecase is that `parse` expects an `unknown` object and sometimes it doesn't make sense. Sometimes you have a schema like this one: `const dateTimeSchema = z.string().refine(externalDateTimeValidationFunction)` and you want to check that something you already know is a `string` is valid. But `dateTimeSchema.parse(3)` throws a runtime error, when it could have been caught by typescript at build time if the `decode` method, added in this PR, existed.

Example of where it can be useful: You have two APIs, one REST API with a `body: unknown` that you `parse` using a zod schema. The other API is a gRPC API that gives you a `message: { id: string, prop1: string | null }` that you `decode` using the same zod schema. `Advantage`: If the type of `message` is not compatible with the zod schema input, there will be a typescript error, no risk this will be sent in production.

The removal of runtime validations checks can be reverted, I just did it because it felt like it wasn't necessary.

Issue #1292, opened 2022-07-26
Addition of `.keyof()` breaks assignability for `ZodObjects`


itsgiacoliketaco:
Hi there,

#1216 appears to break assignability of `ZodObjects` in a subtle way.  For example, the following worked before Zod 3.17.9, but fails in the latest Zod (3.17.10).  I'm on TypeScript version 4.7.4.

```ts
declare const superset: { foo: string } & z.ZodObject<{
  prop1: z.ZodString;
  prop2: z.ZodNumber;
}>; // The `{ foo: string }` intersection is necessary for this to fail for some reason.

const subset: z.ZodObject<{ prop1: z.ZodString }> = superset;
//    ^^^^^^ Error here
```

This produces the following type error:

```
Type '{ foo: string; } & ZodObject<{ prop1: ZodString; prop2: ZodNumber; }, "strip", ZodTypeAny, { prop1: string; prop2: number; }, { prop1: string; prop2: number; }>' is not assignable to type 'ZodObject<{ prop1: ZodString; }, "strip", ZodTypeAny, { prop1: string; }, { prop1: string; }>'.
  The types returned by 'keyof()._parse(...)' are incompatible between these types.
    Type 'ParseReturnType<"prop1" | "prop2">' is not assignable to type 'ParseReturnType<"prop1">'.
      Type 'OK<"prop1" | "prop2">' is not assignable to type 'ParseReturnType<"prop1">'.
        Type 'OK<"prop1" | "prop2">' is not assignable to type 'OK<"prop1">'.
          Type '"prop1" | "prop2"' is not assignable to type '"prop1"'.
            Type '"prop2"' is not assignable to type '"prop1"'.ts(2322)
```

The problem arises because `z.ZodEnum<["prop1", "prop2"]>` is not assignable to `z.ZodEnum<["prop1"]>`.  This makes sense!  However, I would argue that `superset` _should_ be assignable to `subset`, because that aligns with TypeScript's structural typing.  In particular, in TypeScript,
```ts
interface Superset {
  prop1: string;
  prop2: number;
}
```
is assignable to
```ts
interface Subset {
  prop1: string
}
```

[...]

PR #1346, opened 2022-08-19
Replace `ZodObject.keyof()` with `z.keyof()`

itsgiacoliketaco:
Minimal alternative to #1347.

This is a breaking change, which I introduced to revert the accidental breaking change described in #1292.

I have replaced the `ZodObject.keyof()` method with a `z.keyof()` helper function.  They work similarly:

```ts
const schema = z.object({ a: z.string(), b: z.number() });
// BEFORE:
schema.keyof(); // ZodEnum<["a", "b"]>
// AFTER:
z.keyof(schema) // ZodEnum<["a", "b"]>
```

I also added a test case for the problem described in #1292.  Removing `ZodObject.keyof()` is required to fix the assignability problem.

I will also propose an alternative PR that expands `z.keyof()` to work with other schemas, not just `ZodObject` (#1347)

cc @ecyrbe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@
   - [.promise](#promise)
   - [.or](#or)
   - [.and](#and)
+- [Keyof](#zkeyof)
 - [Guides and concepts](#guides-and-concepts)
   - [Type inference](#type-inference)
   - [Writing generic functions](#writing-generic-functions)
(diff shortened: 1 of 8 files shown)


ecyrbe:
cc @colinhacks

itsgiacoliketaco:
This PR is complete, just awaiting review.  I don't think that's the same as stale.

PR #1347, opened 2022-08-19
Replace `ZodObject.keyof()` with comprehensive `z.keyof()`

itsgiacoliketaco:
Expanded alternative to #1346.

This is a breaking change, which I introduced to revert the accidental breaking change described in #1292.

I have replaced the `ZodObject.keyof()` method with a `z.keyof()` helper function.  Unlike in #1346, this `z.keyof()` helper function works for any object-like schema, such as intersections of objects.  I think this is an improvement to `ZodObject.keyof()`, as it behaves more like TypeScript's `keyof` operator.

```ts
const schema = z.object({ a: z.string(), b: z.number() });
// BEFORE:
schema.keyof(); // ZodEnum<["a", "b"]>
// AFTER:
z.keyof(schema) // ZodEnum<["a", "b"]>
// ALSO WORKS AFTER:
const intersection = z.intersection(
  z.object({ a: z.string() }),
  z.object({ b: z.string() }),
)
z.keyof(intersection) // ZodEnum<["a", "b"]>
```
See the tests for all the supported cases.

I also added a test case for the problem described in #1292.  Removing `ZodObject.keyof()` is required to fix the assignability problem.

cc @ecyrbe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@
   - [.promise](#promise)
   - [.or](#or)
   - [.and](#and)
+- [Keyof](#zkeyof)
 - [Guides and concepts](#guides-and-concepts)
   - [Type inference](#type-inference)
   - [Writing generic functions](#writing-generic-functions)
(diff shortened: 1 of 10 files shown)


ecyrbe:
cc @colinhacks

PR #1355, opened 2022-08-24
Update readme with strictNullChecks

eber404:
I was facing some issues to access property `error` as result of `safeParse` method, then I realize that `strictNullChecks` should be true on `tsconfig.json` as well. Can we add it on readme file?

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,7 +342,7 @@ There are a growing number of tools that are built atop or support Zod natively!
 ### Requirements
 
 - TypeScript 4.1+!
-- You must enable `strict` mode in your `tsconfig.json`. This is a best practice for all TypeScript projects.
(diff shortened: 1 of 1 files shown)


scotttrinh:
`strict` includes `strictNullChecks` so what's likely happening in your case is that some base config is specifically setting `strictNullChecks: false`? Closing for now, but if you can demonstrate that this is a problem with a minimal repro CodeSandbox or something, feel free to reopen!

scotttrinh:
Relevant part of the docs: https://www.typescriptlang.org/tsconfig#strictNullChecks

> Default:
> `true` if [strict](https://www.typescriptlang.org/tsconfig#strict),,`false` otherwise.

PR #1377, opened 2022-09-04
docs: remove dead link from TOC of README.md

erikhofer:
There is a link to `.enum()` under "Objects" which does not exist (anymore?).


colinhacks:
Fixed, thanks!

PR #1386, opened 2022-09-06
add uniq to ZodArray

roblabat:
This PR add a uniq validation to ZodArray to validate that the array doesn't contains any duplicates.

ps: Currently this validation doesn't check for deep duplications but only simple ones.


src/types.ts
```diff
@@ -1423,6 +1424,26 @@ export class ZodArray<
...
+            if (a === b) {
+              addIssueToContext(ctx, {
+                code: ZodIssueCode.duplicate_element,
+              });
+              status.dirty();
+              return;
+            }
+          }
+        }
+      })();
```
Balastrong, line 1444: I'm not sure why you used an IIFE with a double for loop here.
I think a really easy and quick approach would be using `Set` in something like:

```suggestion
if (new Set(ctx.data).size < ctx.data.length) {
  addIssueToContext(ctx, {
    code: ZodIssueCode.duplicate_element,
  });
  status.dirty();
}
```

Balastrong (changes requested):

src/types.ts
roblabat: I used an IIFE with double loop for execution complexity optimisation, this way I'm able to stop the processing as soon as I find a duplicates avoiding to check the rest of the array if it's not necessary. Using A Set would mean to check duplicates for all elements of the array then check the lenght of the Set to get the result. The code would be simpler but the execution would be worse.

ArthurKa: Did you test performance of double loop over a Set so you say it's more optimized?

ArthurKa: By the way in case of `unique array` schema may be prefer to use `z.set` instead of array?

ArthurKa:
By the way in case of `unique array` schema may be prefer to use `z.set` instead of array?

roblabat:
@ArthurKa, One of the main use of Zod is to validate external data input as JSON message (from http call or Mongo collection for example). Those kind of messages doesn't supports Sets as Sets or in memory objects and aren't par of the JSON specification. To validate arrays with uniq items comming from JSON inputs you can't use z.set()

ArthurKa:
Make sense. What about `.refine`?

```ts
const schema = z.number().array().refine(e => new Set(e).size === e.length, {
  message: 'Array should have unique elements',
});
```

src/types.ts
roblabat: Effectively I didn't tested it before committing it. I just made a few tests and they show that the double loop as heavily better performance on arrays with duplicates but worst performance on arrays without duplicates. I found an other implementation with object and simple loop, here is my test code: 
``` ts
import _ from 'lodash';

function checkDuplicateDoubleLoop(list: any[]) {
  return (() => {
    for (let i = 0; i < list.length; i++) {
      const a = list[i];

      for (let j = 0; j < i; j++) {
        if (a === list[j]) {
          return false;
        }
      }
    }
    return true;
  })();
}

function checkDuplicateSet(list: any[]) {
  return new Set(list).size >= list.length;
}

function checkDuplicateObject(list: any[]) {
  return (() => {
    const obj: any = {};
    for (const elem of list) {
      if (obj[elem]) {
        return false;
      } else {
        obj[elem] = true;
      }
    }
    return true;
  })();
}

function checkDeepDuplicateDoubleLoop(list: any[]) {
  return (() => {
    for (let i = 0; i < list.length; i++) {
      const a = list[i];

      for (let j = 0; j < i; j++) {
        if (_.isEqual(a, list[j])) {
          return false;
        }
      }
    }
    return true;
  })();
}

function checkDeepDuplicateSet(list: any[]) {
  return new Set(list).size >= list.length;
}

(() => {
  const list = [];

  for (let i = 0; i < 10000000; i++) {
    list.push(i);
  }

  // commented because too long

  // console.time('noDupDoubleLoop');
  // console.log('noDupDoubleLoop: ', checkDuplicateDoubleLoop(list));
  // console.timeEnd('noDupDoubleLoop');

  console.log('noDupDoubleLoop: Too long');

  console.time('noDupSet');
  console.log('noDupSet: ', checkDuplicateSet(list));
  console.timeEnd('noDupSet');

  console.time('noDupObject');
  console.log('noDupObject: ', checkDuplicateObject(list));
  console.timeEnd('noDupObject');

  list.unshift(0);

  console.time('dupDoubleLoop');
  console.log('dupDoubleLoop: ', checkDuplicateDoubleLoop(list));
  console.timeEnd('dupDoubleLoop');

  console.time('dupSet');
  console.log('dupSet: ', checkDuplicateSet(list));
  console.timeEnd('dupSet');

  console.time('dupObject');
  console.log('dupObject: ', checkDuplicateObject(list));
  console.timeEnd('dupObject');
})();

(() => {
  const list = [];

  for (let i = 0; i < 10000; i++) {
    list.push({ i });
  }

  console.time('noDeepDupDoubleLoop');
  console.log('noDeepDupDoubleLoop: ', checkDeepDuplicateDoubleLoop(list));
  console.timeEnd('noDeepDupDoubleLoop');

  console.time('noDeepDupSet');
  console.log('noDeepDupSet: ', checkDuplicateSet(list));
  console.timeEnd('noDeepDupSet');

  console.time('noDeepDupObject');
  console.log('noDeepDupObject: ', checkDuplicateObject(list));
  console.timeEnd('noDeepDupObject');

  list.unshift({ i: 0 });

  console.time('deepDupDoubleLoop');
  console.log('deepDupDoubleLoop: ', checkDeepDuplicateDoubleLoop(list));
  console.timeEnd('deepDupDoubleLoop');

  console.time('deepDupSet');
  console.log('deepDupSet: ', checkDuplicateSet(list));
  console.timeEnd('deepDupSet');

  console.time('deepDupObject');
  console.log('deepDupObject: ', checkDuplicateObject(list));
  console.timeEnd('deepDupObject');
})();
```

and results : 

``` bash
noDupDoubleLoop: Too long
noDupSet:  true
noDupSet: 2.184s
noDupObject:  true
noDupObject: 742.271ms
dupDoubleLoop:  false
dupDoubleLoop: 0.232ms
dupSet:  false
dupSet: 2.179s
dupObject:  false
dupObject: 0.099ms
noDeepDupDoubleLoop:  true
noDeepDupDoubleLoop: 31.074s
noDeepDupSet:  true
noDeepDupSet: 1.935ms
noDeepDupObject:  false // expect true
noDeepDupObject: 0.114ms
deepDupDoubleLoop:  false
deepDupDoubleLoop: 0.396ms
deepDupSet:  true // expect false
deepDupSet: 1.172ms
deepDupObject:  false
deepDupObject: 0.222ms
```

I will update my PR to implement the object solution.
You can also notice that I tested Set and object solution for deep object unicity check and for that check only the Double loop solution is working. I will so implement the deepUniq function with the double loop solution (I made the deep check using lodash for going faster for testing but for final implementation I will make it in an hand made function to avoid adding lodash in Zod dependencies just for that utility.

PS: trying to go further in testing I also noticed that on big arrays with more than 10 000 000 items you can reach Set maximun size limit. The Set solution would then be not suitable for a production reliable librairy like Zod  (even if I do not recommend using the uniq check on such arrays because of processing time expect for exceptional use cases)

roblabat:
Refine would definitely work and that's what I'm using currently. But I think this uniq check is a quite common use case and a great functionality to add to Zod. It would definitely be cleaner to used than using big refine on all arrays that need this checks.
By the way quite all checks of Zod (like min, max, length, elements on arrays) could be implemented with refine, but we add then to the library to make it cover more use cases.

Balastrong:
We can discuss the implementation but I also agree having a `.uniq` shortcut is good :D

deno/lib/helpers/util.ts
```diff
@@ -70,6 +70,24 @@ export namespace util {
...
+
+    if (typeof a === "object" && typeof b === "object") {
+      const aKeys = objectKeys(a);
+      const bKeys = objectKeys(b);
+
+      if (aKeys.length !== bKeys.length) {
+        return false;
+      }
+
+      return aKeys.every((key) => isEqual(a[key], b[key]));
```
webNeat, line 86: This check doesn't guarantee that objects `a` and `b` are equal. For example:
```
isEqual({a: 1, b: undefined}, {a: 1, c: 1}) //=> true
```

roblabat: I didn't saw that possibility...
I added a check for existance of the property in b object with hasOwnProperty in a new commit. @webNeat Do you see other cases where this check would not be enough to check equality between a and b objects ?

webNeat: The only other case I can see with normal values is the case of `NaN`
```
const a = NaN
console.log(a === a) //=> false
```

roblabat: I think we can leave the case of NaN Like it is, for objects with NaN values in it will return false but as NaN === NaN returns false too this behavior is more consistant than returning true (I think returning a different result than the === operator for non object values would be inconsistant with the goal that function which is to provider an === operator with a consistant behavior of === check on values of objects instead of their memory adress).

So if you don't have any more problematic cases I mark this thread as resolved.

webNeat: I understand your point, if the goal is the provide an operator that behaves exactly the same as `===` then I agree with you.
if in the other hand, the goal is to use this to remove duplicated values, then two instances of `NaN` should be considered the same value, this is what the native `Set` does:
```ts
new Set([NaN, 1, NaN]) //=> Set(2) { NaN, 1 }
```
I have also checked the implementation of `eq` method from lodash and it seems to use the same logic of the `Set`: https://github.com/lodash/lodash/blob/master/test/eq.test.js#L9
```ts
_.eq(NaN, NaN) //=> true
```
What do you think?

roblabat: I'm not sure of that, the symbolic of NaN is to express a non numeric value that you tried to use as a number. But this is used no matter what non numeric value you have then it make sense that NaN != NaN as if you try to test 2 * x === 3 * x except for x = 0 you should always have false. If NaN === NaN you would here have 2 * x === 3 * x returning true for x = 'a' for exemple. So for me it makes sense returning false on isEqual(NaN, NaN) even it makes the array having multiple NaN. 

But I also understand your point of vue and honestly I can't say which one is the best point of vue.
Maybe @colinhacks as the owner of Zod could choose which behavior makes the more sense for him ?

roblabat: I would add that for me this point is not really a crucial concern has using Zod  with typescript strict type checking, you should never see your code generate NaN values and so this choice should quite never have any consequence.

psirenny:
Is it possible to add a comparator function? The use case I have in mind is to check a _subset_ of duplicate properties. For example:

```ts
const schema = z.string().array().uniq((a, b) => a.id === b.id);

schema.parse([
  { "id": "1", "firstName": "Alice"  },
  { "id": "2", "firstName": "Bob" }, // Error: duplicate `id` property
]);
```

roblabat:
@psirenny, Following your suggestion I added two functions on ZodArray to my PR uniqBy and uniqWith that respectively take a mapping function and a comparator function for custom unicity checks.

calebwilson706:
Any chance of getting this released?

README.md
```diff
@@ -1057,6 +1057,120 @@ z.string().array().length(5); // must contain 5 items exactly
 
 Unlike `.nonempty()` these methods do not change the inferred type.
 
+### `.uniq`
```
marcus13371337, line 1060: Do you see any benefits of using a shortened name for `unique`? Instead of `.unique()` more clear/less distraction for developers in my opinion!

roblabat: I where just conforming to Lodash wording that use the key .uniq as I more often see the key .uniq in libs and because I'm more used to this keyword.
But that's not really a concern for me, it's more a question of personal preference I think so I will not try argue on this point and simply align the PR if necessary.

igalklebanov: I think the name is confusing. `uniq` in lodash transforms the input, while the suggested method here checks and "returns" the input as-is.

psirenny: Is the function name the last remaining piece to solve? These other tools/specifications use some variation on the word **unique**:
- ajv: `uniqueItemProperties`
- joi: `unique`
- json-schema: `uniqueItems`

Perhaps in the future, array coercion might remove duplicates. If that's the case, any mutability implied by the word **unique** might actually be desirable? Anyway, just my 2¢.

The word **distinct** also seems to be a good fit here.

Calyhre:
Hello, could we put this back on the burner? A `uniq` method would be great as a short hand but also to represent [uniqueness in JSON schema](https://json-schema.org/understanding-json-schema/reference/array#uniqueItems).

roblabat:
If we could have a validation from @colinhacks about the naming of the function that where making debate (personaly I don't care about this naming) and the interest of adding this. Then I can make a new PR addapting this both to zod V3 and V4.

Calyhre:
BTW, if you end up here wondering how to represent `uniqueItems` in JSON schema, you can do it like so:

```ts
const schema = z.object({
  items: z.array(z.enum(["a", "b"])).meta({
    uniqueItems: true,
  }),
})

z.toJSONSchema(schema)
```

Will give you

```json
{
  "type": "object",
  "properties": {
    "items": {
      "uniqueItems": true,
      "type": "array",
      "items": {
        "type": "string",
        "enum": [
          "a",
          "b"
        ]
      }
    }
  },
  "required": [
    "items"
  ],
  "additionalProperties": false
}
```

PR #1389, opened 2022-09-07
Introduce ZodTransformError type, to allow for transforms that fail

Duncan3142:
This PR introduces a new Error type, `ZodTransformError`, which accepts an array of issues in it's constructor.

Closes https://github.com/colinhacks/zod/issues/1390

When thrown inside a transform, this error class will be caught by Zod, and the issues it contains treated a fatal and added to the parse result.

The rational for this PR is to allow for transforms that may fail, such as the following example.

```ts
import { ZodTransformError } from "zod";

const user = z.string().transform(async (id) => {
  try {
    const user = await getUserById(id);
    return user;
  } catch (e) {
    throw new ZodTransformError([{
      code: "custom",
      message: "User not found",
    }]);
  }
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1768,6 +1768,23 @@ const Strings = z.string().transform((val, ctx) => {
 });
 ```
 
+When attempting a transform that fails, and there is no value to return, the transform may throw an error of type `ZodTransformError`.
+The `ZodTransformError` constructor accepts an array of issues, all of which are treated as fatal. When a transform throws a `ZodTransformError`, no subsequent transforms or refinments are attempted..
+
+```ts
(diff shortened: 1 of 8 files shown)


colinhacks:
Thanks for this PR! This same scenario can now be addressed with `ctx.addIssue` inside the `.transform()` method:

```ts
const user = z.string().transform(async (id, ctx) => {
  try {
    await "asdf";
    throw new Error();
  } catch (e) {
    ctx.addIssue({
      code: z.ZodIssueCode.custom,
      message: "User not found :(",
    });
  }
});

```

Closing for now, let me know if the new solution doesn't cover your use cases.

PR #1393, opened 2022-09-08
Optimize object strip case

tmcw:
Refs #205

This skips the check for extra keys in the default case of `strip` being the object parsing strategy, because we don't use the extra keys.

This yields a pretty modest performance bump - about 5% in the realworld case in my testing.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1669,9 +1669,17 @@ export class ZodObject<
 
     const { shape, keys: shapeKeys } = this._getCached();
     const extraKeys: string[] = [];
-    for (const key in ctx.data) {
-      if (!shapeKeys.includes(key)) {
-        extraKeys.push(key);
+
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -1669,9 +1669,17 @@ export class ZodObject<
 
     const { shape, keys: shapeKeys } = this._getCached();
     const extraKeys: string[] = [];
-    for (const key in ctx.data) {
-      if (!shapeKeys.includes(key)) {
-        extraKeys.push(key);
+
+    if (
+      !(
```
FlorianWendelborn, line 1674: Couldn't this `!(A && B)` be simplified to `!B || !A` so it can short-circuit?

I'd also guess that putting B first is faster as it's a simple comparison, but not sure on that

PR #1415, opened 2022-09-18
Chain schemas

Duncan3142:
Allow schemas to be chained, passing the output of one as the input to another.
The output of the first schema is type-checked to ensure compatibility with the input of the second schema.

```ts
const createUserDTOSchema = z
  .object({
    name: z.string(),
    email: z.string(),
    active: z.boolean(),
  })
  .transform((input) => {
    return {
      id: uuid(),
      ...input,
    };
  });

const updateUserDTOSchema = z.object({
  id: z.string(),
  name: z.string(),
  email: z.string(),
  active: z.boolean(),
});

const userSchema = z.object({
  id: z.string().trim().uuid(),
  name: z.string().trim().min(1),
  email: z
    .string()
    .trim()
    .email()
    .refine((val) => allowedDomains.includes(emailDomain(val))),
  active: z.boolean(),
});

const createUserSchema = createUserDTOSchema.chain(userSchema);
const updateUserSchema = updateUserDTOSchema.chain(userSchema);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1822,14 +1822,16 @@ const updateUserDTOSchema = z.object({
   active: z.boolean(),
 });
 
-const userSchema = z
-  .object({
-    id: z.string().trim().uuid(),
-    name: z.string().trim().min(1),
-    email: z.string().trim().email(),
-    active: z.boolean(),
-  })
-  .refine((val) => allowedDomains.includes(emailDomain(val.email)));
+const userSchema = z.object({
+  id: z.string().trim().uuid(),
+  name: z.string().trim().min(1),
+  email: z
+    .string()
+    .trim()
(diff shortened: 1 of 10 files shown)


jamesarosen:
For anyone who comes across this issue, this feature exists as `.pipe` and `z.pipeline`:

```ts
// with .pipe:
const createUserSchema = createUserDTOSchema.pipe(userSchema);
const updateUserSchema = updateUserDTOSchema.pipe(userSchema);

// or with z.pipeline:
const createUserSchema = z.pipeline(createUserDTOSchema, userSchema);
const updateUserSchema = z.pipeline(updateUserDTOSchema, userSchema);
```

PR #1417, opened 2022-09-20
Introduce data hydration

ArthurKa:
Implement data hydration.

For example, when you need to validate any input data and get `string[]` on output, without hydration you can only get validation error or set default value (if parse `undefined`), but with hydration if data validation fails it fallbacks on what you pass to `.hydrate` method.

```ts
const schema = z.string().hydrate(/* '' by default */).array().hydrate([]);

schema.parse(undefined); // => []
schema.parse({ a: 123 }); // => []
schema.parse([]); // => []
schema.parse(['asd']); // => ['asd']
schema.parse([1]); // => ['']
schema.parse([1, 'a', 2]); // => ['', 'a', '']
```

So on output you always have `string[]`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@
   - [.superRefine](#superRefine)
   - [.transform](#transform)
   - [.default](#default)
+  - [.hydrate](#hydrate)
   - [.optional](#optional)
   - [.nullable](#nullable)
   - [.nullish](#nullish)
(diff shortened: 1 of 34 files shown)


colinhacks:
Great idea and great PR. Ultimately this was implemented in #1537 as `.catch` and just landed in Zod 3.20. Thanks for the amazing work here and apologies for the delay in reviewing this.

PR #1435, opened 2022-09-27
Allows email address domains to be IP addresses

Tyler-Murphy:
Sorry I didn't make an issue first. Figured it's a small change, so I'd just offer a solution whether it's acceptable or not.

References:
- https://learn.microsoft.com/en-us/archive/blogs/testing123/email-address-test-cases
- https://github.com/validatorjs/validator.js/issues/800

This doesn't make sure the IP address is valid. It only makes sure that it's not obviously invalid.

### DIFF
--- a/deno/lib/__tests__/string.test.ts
+++ b/deno/lib/__tests__/string.test.ts
@@ -55,6 +55,15 @@ test("more email validations", () => {
   }
 });
 
+test("email with IP address domain validations", () => {
+  const email = z.string().email();
+  email.parse("mojojojo@123.123.123.123");
+  email.parse("mojojojo@[123.123.123.123]");
+  expect(() => email.parse("mojojojo@[123.123.123]")).toThrow()
+  expect(() => email.parse("mojojojo@[123.123.123.1234]")).toThrow()
+  expect(() => email.parse("mojojojo@[123.123.123.]")).toThrow()
+})
+
 test("url validations", () => {
   const url = z.string().url();
   try {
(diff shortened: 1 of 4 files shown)


kibertoad:
can you benchmark perf impact of this change on plain emails?

kibertoad:
I think validator.js approach of making this optional generally makes more sense, especially if this makes the hot path regex slower

Tyler-Murphy:
Sure, added to `benchmarks/string.ts` like this:

```ts
const emailSchema = z.string().email()

...

  .add("email", () => {
    emailSchema.parse(`someone@example.com`)
  })
```

Without IP support: 7,224,694 ops/sec ±0.59% (96 runs sampled)
With IP support: 7,175,461 ops/sec ±0.62% (92 runs sampled)

I also tried with 1000 random email addresses from https://www.randomlists.com/email-addresses?qty=1000 where each one was checked in a `for .. of` loop.

Without IP support: 7,459 ops/sec ±0.39% (97 runs sampled)
With IP support: 7,584 ops/sec ±0.56% (97 runs sampled)

And finally, I replaced two of the 1000 random email addresses with email addresses that use IP addresses, still using the regex with IP support: 7,663 ops/sec ±0.41% (97 runs sampled)

The speed is so similar that I don't think it's worth making it optional.

Tyler-Murphy:
Leave open

Tyler-Murphy:
Not stale

Tyler-Murphy:
not stale

colinhacks:
Thanks for this (now very old) PR! The final word on email regexes landed here: https://github.com/colinhacks/zod/pull/2157 Further discussion can happen in new issues/PRs. 👍

PR #1438, opened 2022-09-29
Update documentation to include `strictNullChecks` as required

artokun:
This PR is intended to help other engineers avoid the same mistake I made by having `strict` set to true but not `strictNullChecks` which led to very frustrating / confusing functionality when dealing with static typechecking

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,7 @@ There are a growing number of tools that are built atop or support Zod natively!
 
 - TypeScript 4.1+!
 - You must enable `strict` mode in your `tsconfig.json`. This is a best practice for all TypeScript projects.
+- Make sure that `strictNullChecks` mode is also enabled, otherwise static typechecking will return all properties as optional!
 
   ```ts
   // tsconfig.json
@@ -351,7 +352,8 @@ There are a growing number of tools that are built atop or support Zod natively!
     // ...
     "compilerOptions": {
       // ...
-      "strict": true
+      "strict": true,
+      "strictNullChecks: true,
     }
   }
   ```


scotttrinh:
FWIW, `strict` implies `strictNullChecks`, so I think our stance is still "enable `strict`" since we rely on the most strict behavior possible. The fact that you can opt out of certain parts of the strict-mode is understandable but unfortunate, and this is a speed bump that people run into, so I'm 👍 for this change for practicality,

README.md
```diff
@@ -344,14 +344,16 @@ There are a growing number of tools that are built atop or support Zod natively!
...
 
   ```ts
   // tsconfig.json
   {
     // ...
     "compilerOptions": {
       // ...
-      "strict": true
+      "strict": true,
+      "strictNullChecks: true,
```
scotttrinh, line 356: I don't think this line should be part of our setup instructions. It's only the case that you shouldn't set it to `false`, rather than that you need to set it to `true`.

README.md
```diff
@@ -344,14 +344,16 @@ There are a growing number of tools that are built atop or support Zod natively!
 
 - TypeScript 4.1+!
 - You must enable `strict` mode in your `tsconfig.json`. This is a best practice for all TypeScript projects.
+- Make sure that `strictNullChecks` mode is also enabled, otherwise static typechecking will return all properties as optional!
```
scotttrinh, line 347: I think we should highlight the "override" behavior here. Maybe something like:

```suggestion
- Make sure that `strictNullChecks` mode has not been disabled explicitly by setting it to `false` elsewhere, otherwise TypeScript will infer all properties as optional. If you're extending a base config, you may need to set this to be explicitly `true`.
```

PR #1448, opened 2022-10-02
Add `ZodSymbol`

santosmarco-caribou:
# Overview

This PR adds support for the `ZodSymbol` type, which parses the primitive JavaScript type `symbol`.

`ZodSymbol.create()`/`z.symbol()` can be called with or without a defined symbol as the first argument.

If passed a **defined** symbol, inferring will reflect `typeof [name of const that defines the symbol]`

Example:

```ts
const definedSymbol = Symbol("symbol");
z.symbol(definedSymbol); // => infers `typeof definedSymbol`
```

If called without a symbol as the first argument, then it will infer `symbol` and validate _any_ symbol.

```ts
z.symbol(); // => parses any symbol and infers `symbol`
```

### Preventing an impossible validation

 **Important:** For the compiler to correctly reflect the type of a `unique symbol` it [**must** be defined in a const declaration or as a readonly static property.](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#unique-symbol)
 
 So, `z.symbol(Symbol())` would pass type validation and return `ZodSymbol<symbol>` but we know that it would fail _every_ time because no value would ever be assignable to this symbol (since there is no way to access it).

 To prevent people from calling `z.symbol()` with a symbol that's not a `unique symbol` (e.g., `symbol(Symbol("unassigned"))` accidentally and cause `zod` to **always** error out, we fail on compile time:

 ```ts
 <S extends symbol>(
   symbol: S extends symbol ? (symbol extends S ? /* symbol is not unique */ never : /* OK */ S) : never,
   params?: RawCreateParams
): S extends symbol ? (symbol extends S ? never : ZodSymbol<S>) : never
 ```

## Summary

```ts
z.symbol(); // => this one accepts any symbol and infers `symbol`

z.symbol(Symbol("unassigned")); // => this one will fail on compile time

const mySymbol = Symbol();
z.symbol(mySymbol); // => this one will infer `typeof mySymbol` and validate only symbols that are equal to `mySymbol`
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -428,6 +428,7 @@ z.number();
 z.bigint();
 z.boolean();
 z.date();
+z.symbol();
 
 // empty types
 z.undefined();
@@ -449,7 +450,11 @@ z.never();
 ```ts
 const tuna = z.literal("tuna");
 const twelve = z.literal(12);
+const twobig = z.literal(2n); // bigint literal
 const tru = z.literal(true);
+
+const terrificSymbol = Symbol("terrific");
+const terrific = z.literal(terrificSymbol);
 
 // retrieve literal value
 tuna.value; // "tuna"
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -428,6 +428,7 @@ z.number();
 z.bigint();
 z.boolean();
 z.date();
+z.symbol();
 
 // empty types
 z.undefined();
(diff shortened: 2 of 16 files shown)


colinhacks:
Great PR, thanks for your work on this and sorry for the long delay.

I propose a slightly different approach, namely that we use `z.literal()` for declaring a validator for a specific schema. This is more analogous to how `z.string()` and `z.literal("tuna")` works for strings. I pushed a commit to reflect this.

Ultimately I didn't want to clutter up the implementation of ZodLiteral too much, so unfortunately I ended up dropping your clever solution for preventing `z.symbol(Symbol('whatever'))`! I figure anyone bothering with literal symbol schemas probably knows better than to create an anonymous symbol inside their schema definition...but we'll see.

Thanks again!

Issue #1370, opened 2022-09-01
Feature request: `.deepStrict`


liammcdermott:
Like `.deepPartial` is to `.partial` it would be beneficial to have a `.deepStrict` function that ensures every level in an object, array, or tuple hierarchy is `.strict`.

For example, this code does _not_ generate a ZodError:
```
const person = z
  .object({
    name: z.string(),
    meta: z.object({
      born: z.string(),
    }),
  })
  .strict();

person.parse({
  name: "bob dylan",
  meta: {
    born: "a place",
    extraKey: 61,
  },
});
```

With `.deepStrict` it would work like this:
```
const person = z
  .object({
    name: z.string(),
    meta: z.object({
      born: z.string(),
    }),
  })
  .deepStrict();

person.parse({
  name: "bob dylan",
  meta: {
    born: "a place",
    extraKey: 61,
  },
});
// => throws ZodError
```

colinhacks:
This is very hard to do in the static domain. Currently `.deepPartial` is very limited and error-prone for similar reasons. I'll leave this open but I recommend using `z.strictObject` if you want a shorthand for declaring lots of strict objects without needing a method.

liammcdermott:
Thank you for considering this Colin, that's fair enough. Thank you for the reference to `z.strictObject`, I didn't see that, much nicer than calling `.strict()` on every object.

renatoargh:
When using Joi there is kind of a "global" option you can enable so that the `strict` feature will be applied recursively across all your objects by default. For sure it is a challenge to make such configuration reflect statically in the types but maybe it is a really good improvement to Zod to introduce such feature in runtime only at least. 

Not sure about the implementation details but from the developer perspective it would be something like this:

```typescript
schema.safeParse({
  deepStrict: true,
})
```

elmeister:
Would be nice to have such `deepStrict: true` option to cover cases, when extra keys are not welcomed. 
Not only it would allow to define more `clear` schemas - without the need to add `.strict()` everywhere, but also support cases requested in #1449, when we need to use the same shared schemas in both ways - as strict and "normal" ones.

StefanTerdell:
Having the behaviour of the schema change based on the parsing seems like a mistake to me. There would be no way to account for it in conversion / translation and it could act inconsistently in different parts of the code with no compiler errors to show for it.

philip-peterson:
Another option would be something like:

`z.strict().object({ ... })`

[...]

PR #1455, opened 2022-10-03
Add `deepStrict` flag to `parse()` (#1370)

elmeister:
Straightforward attempt to implement feature requested in #1370 
```javascript
schema.safeParse({
  deepStrict: true,
})
```

### DIFF
--- a/deno/lib/__tests__/parser.test.ts
+++ b/deno/lib/__tests__/parser.test.ts
@@ -46,3 +46,24 @@ test("invalid enum value", () => {
 test("parsing unknown", () => {
   z.string().parse("Red" as unknown);
 });
+
+test("parsing with deepStrict", () => {
+  const deepSchema = z.object({
+    nestedObject: z.object({
+      knownNestedKey: z.number(),
+    })
+  });
+
+  const deepObject = {
+    nestedObject: {
+      knownNestedKey: 1,
+      unknownNestedKey: 1,
+    }
+  }
+
+  deepSchema.parse(deepObject);
+
+  expect(() => deepSchema.parse(deepObject, {
+    deepStrict: true,
+  })).toThrow();
+});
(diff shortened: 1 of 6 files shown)


StefanTerdell:
Having the behaviour of the schema change based on the parsing seems like a mistake to me. There would be no way to account for it in conversion / translation and it could act inconsistently in different parts of the code with no compiler errors to show for it.

paescuj:
> Having the behaviour of the schema change based on the parsing seems like a mistake to me. There would be no way to account for it in conversion / translation and it could act inconsistently in different parts of the code with no compiler errors to show for it.

I agree, but having a `deepStrict` method would be great!

elmeister:
Well, we already have [deepPartialify](https://github.com/colinhacks/zod/blob/e8248f7230781206dbebddb9b39845a5be49e790/src/types.ts#L1613-L1638)
So we could introduce something similar for `deepStrictify` like this
```typescript
function deepStrictify(schema: ZodTypeAny): any {
  if (schema instanceof ZodObject) {
    const newShape: any = {};

    for (const key in schema.shape) {
      const fieldSchema = schema.shape[key];
      newShape[key] = ZodOptional.create(deepStrictify(fieldSchema));
    }
    return new ZodObject({
      ...schema._def,
      shape: () => newShape,
      unknownKeys: "strict",
    }) as any;
  } else if (schema instanceof ZodArray) {
    return ZodArray.create(deepStrictify(schema.element));
  } else if (schema instanceof ZodTuple) {
    return ZodTuple.create(
      schema.items.map((item: any) => deepStrictify(item))
    );
  }
  return schema;
}
```
I can raise another PR with `deepStrict()` method

elmeister:
Closed in favor of [the new PR](https://github.com/colinhacks/zod/pull/1467)

PR #1462, opened 2022-10-05
Add ZodObject partial and required benchmarks

scotttrinh:
Add benchmark for ZodObject partial and required. Also adds a GitHub Actions workflow for running benchmarks against current `master` and PR `HEAD` so we can compare when looking at PRs.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,41 @@
+name: benchmark
+
+on:
+  push:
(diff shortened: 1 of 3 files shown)


scotttrinh:
Benchmarks take a looooong time, so I didn't want to merge this without a 👍 from you @colinhacks. Seems handy, but we could also just use a single 1x1 matrix instead of the 1x2 to reduce CI minute "spend".

PR #1464, opened 2022-10-05
Forward path param for safeParseAsync method to _parse method

### LINKED ISSUE #1451: schema.safeParseAsync method is ignoring the path param
The `safeParseAsync` method on a zod schema is ignoring the provided `path` param

Here's the output when using `safeParse`:
```ts
const schema = z.object({
	debug: z.boolean(),
});

const result = schema.safeParse({}, { path: ["body"] });
console.log(result.success ? result.data : result.error.issues[0]?.path); // ["body", "debug"]
```

and `safeParseAsync`:
```ts
const schema = z.object({
	debug: z.boolean(),
});

const result = await schema.safeParseAsync({}, { path: ["body"] });
console.log(result.success ? result.data : result.error.issues[0]?.path); // ["debug"]
```

Zod version: 3.18.0

T3NED:
**Problem**
The `path` option for `safeParseAsync` is not forwarded to the `_parse` method

**Solution**
Swap the empty array argument which is passed into the `safeParseAsync` method to the defined `path` in the current context

Fixes #1451

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -255,7 +255,7 @@ export abstract class ZodType<
       parsedType: getParsedType(data),
     };
 
-    const maybeAsyncResult = this._parse({ data, path: [], parent: ctx });
+    const maybeAsyncResult = this._parse({ data, path: ctx.path, parent: ctx });
     const result = await (isAsync(maybeAsyncResult)
       ? maybeAsyncResult
       : Promise.resolve(maybeAsyncResult));
--- a/src/types.ts
+++ b/src/types.ts
@@ -255,7 +255,7 @@ export abstract class ZodType<
       parsedType: getParsedType(data),
     };
 
-    const maybeAsyncResult = this._parse({ data, path: [], parent: ctx });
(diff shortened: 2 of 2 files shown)


colinhacks:
Hm, the fact that `path` can be passed to `safeParseAsync` actually isn't intentional. I'm inclined to remove it actually. What's your use case? This would just concatenate the value of `path` to the start of every issue in the ZodError - not very useful.

T3NED:
> Hm, the fact that `path` can be passed to `safeParseAsync` actually isn't intentional. I'm inclined to remove it actually. What's your use case? This would just concatenate the value of `path` to the start of every issue in the ZodError - not very useful.

API validation. I'm using NestJS and have a validation pipe which uses the provided schema to validate either the `body`, `query` or `params` of a request. I want to prefix the issue `path` with the request property name. For convenience, I pass in the given property name that being validated into `path`. Otherwise, I'd have to do something like:

```ts
const schema = z.object({
    [requestPropertyName]: validationSchema
});
```

colinhacks:
This has been fixed in another PR - landing soon.

PR #1467, opened 2022-10-06
Introduce `.deepStrict()` method

elmeister:
Introduce `.deepStrict()` method, requested in #1370 
Previous related discussion can be found [in this PR](https://github.com/colinhacks/zod/pull/1455)

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1610,18 +1610,26 @@ export type SomeZodObject = ZodObject<
   any
 >;
 
+const getModifiedShape = (
+  deepModify: (schema: ZodTypeAny) => any,
+  schema: AnyZodObject
+) => {
+  const newShape: Record<string, ZodTypeAny> = {};
+
+  for (const key in schema.shape) {
+    const fieldSchema = schema.shape[key];
+    newShape[key] = ZodOptional.create(deepModify(fieldSchema));
+  }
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -1886,6 +1910,10 @@ export class ZodObject<
     return deepPartialify(this) as any;
   }
 
+  deepStrict(): AnyZodObject {
```
colinhacks, line 1913: Sorry, this doesn't fly. The difficulty of typing the result of this method is why it doesn't exist. The schema hierarchy is always fully typed, all the way down, no exceptions. If people want a more "declarative" way to make their object schemas strict, you can use `z.strictObject` instead of `z.object`.

elmeister: Thanks for the feedback, two questions please. 

1st one is your opinion about approach on the same subject, proposed in https://github.com/colinhacks/zod/pull/1455
```typescript
schema.safeParse({
  deepStrict: true,
})
```
This could solve any issues with types and coverage, and allowed us to treat data as TypeScript does - without allowing optional keys.

2nd question - if you think that 1st case not an option - should I try to update all `strip` to `strict` in types of objects, returned by `deepStrict()`, or there are more deep concerns I'm not aware of?

PR #1476, opened 2022-10-09
feat: Add lazy to root Zod class

santosmarco-caribou:
# Overview

Very tiny PR to add the `ZodLazy` type as a convenience method in the root `ZodType` class.

Also noticed the `brand` convenience method wasn't bound in the constructor, so I went ahead and fixed this too.

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -368,6 +368,8 @@ export abstract class ZodType<
     this.and = this.and.bind(this);
     this.transform = this.transform.bind(this);
     this.default = this.default.bind(this);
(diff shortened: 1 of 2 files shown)


colinhacks:
Appreciate this PR, though ultimately `.lazy()` isn't really needed as a method. The only reason it exists is so Zod can represent recursive schemas, since those require a type cast during declaration: https://github.com/colinhacks/zod#recursive-types. If a schema has already been declared, there's no reason I can think of to wrap it in `ZodLazy`. Gonna close this unless someone comes in with a use case.

Issue #1400, opened 2022-09-11
Feature request. Enable or disable .strict() by argument


max-mykhailenko:
For the staging environment, I want to provide additional protocol validation with `strict` parsing, but it can break something on production, so I see the next solution
`.strict(process.env.NODE_ENV === 'development')`
and it gives
* `.strict()` — strict mode enabled
* `.strict(true)` — strict mode enabled
* `.strict(false)` — strict mode disabled

Is it hard to implement?

max-mykhailenko:
Please not stale this issue :) I hope that PR will be merged soon. @brunocrosier thank you for participating.

JacobWeisenburger:
Is this what you are looking for?
https://github.com/colinhacks/zod/pull/1484#issuecomment-1284640538

max-mykhailenko:
@JacobWeisenburger hi
```js
const rawSchema = z.object({ ... });
const schema = process.env.NODE_ENV === 'development' ? rawSchema : rawSchema.strict()
```

I have a lot of different schemas and don't want to create add condition before each.

JacobWeisenburger:
Does this work for you?

```ts
function makeSchemaStrict<ObjSchema extends z.ZodObject<z.ZodRawShape>> ( objSchema: ObjSchema ) {
    if ( process.env.NODE_ENV === 'production' ) return objSchema
    return objSchema.strict()
}
```

PR #1484, opened 2022-10-12
Enable or disable `.strict()` by argument

brunocrosier:
Closes https://github.com/colinhacks/zod/issues/1400

### DIFF
--- a/deno/lib/__tests__/object.test.ts
+++ b/deno/lib/__tests__/object.test.ts
@@ -103,6 +103,16 @@ test("strict", () => {
   expect(val.success).toEqual(false);
 });
 
+test("strict with enabled = true", () => {
+  const val = z.object({ points: z.number() }).strict(undefined, true).safeParse(data);
+  expect(val.success).toEqual(false);
+});
+
+test("strict with enabled = false", () => {
+  const val = z.object({ points: z.number() }).strict(undefined, false).parse(data);
+  expect(val).toEqual({ points: 2314 });
+});
+
 test("catchall inference", () => {
   const o1 = z
     .object({
--- a/deno/lib/__tests__/object.test.ts
+++ b/deno/lib/__tests__/object.test.ts
@@ -104,12 +104,12 @@ test("strict", () => {
 });
 
 test("strict with enabled = true", () => {
-  const val = z.object({ points: z.number() }).strict(undefined, true).safeParse(data);
+  const val = z.object({ points: z.number() }).strict({ enabled: true }).safeParse(data);
   expect(val.success).toEqual(false);
 });
 
 test("strict with enabled = false", () => {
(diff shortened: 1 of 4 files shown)


colinhacks:
Interesting idea, but I can't merge this as-is. The return type signature can be inconsistent with actual behavior:

```ts
  strict(
    opts?: errorUtil.ErrMessage & { enabled?: boolean }
  ): ZodObject<T, "strict", Catchall> {}
```

Note the `"strict"` in the type signature. This is technically fixable but this isn't what I'd recommend for your use case. If you need some sort of conditional switch like this, I recommend creating the two possible schemas and switching between them with the ternary operator. It's a language-level approach that doesn't require additional complication in the Zod API.

```ts
const rawSchema = z.object({ ... });
const schema = process.env.NODE_ENV === 'development' ? rawSchema : rawSchema.strict()
```

brunocrosier:
> Interesting idea, but I can't merge this as-is. The return type signature can be inconsistent with actual behavior:
> 
> ```ts
>   strict(
>     opts?: errorUtil.ErrMessage & { enabled?: boolean }
>   ): ZodObject<T, "strict", Catchall> {}
> ```
> 
> Note the `"strict"` in the type signature. This is technically fixable with generics but this isn't what I'd recommend for your use case. If you need some sort of conditional switch like this, I recommend creating the two possible schemas and switching between them with the ternary operator. It's a language-level approach that doesn't require additional complication in the Zod API.
> 
> ```ts
> const rawSchema = z.object({ ... });
> const schema = process.env.NODE_ENV === 'development' ? rawSchema : rawSchema.strict()
> ```

I see what you're saying. Would you be happy to accept a solution which uses generics to correctly show either `"strict"` or `"strip"` using generics?

brunocrosier:
@colinhacks have updated this PR to use generics for the return type, and simplify the parameter: https://github.com/colinhacks/zod/pull/1484/commits/4127e113fa07ec55b344669b5d7d261304d84fa9

max-mykhailenko:
@colinhacks may I ask you to review this PR?

colinhacks:
My proposed solution with the ternary operator is what I would recommended. This API doesn't strike me as very "Zoddy". In general Zod prefers separate explicit methods over single methods that accept configuration options. There are a dozen methods that could conceivably support the same `enabled` flag so it's weird that only this one method supports it. That said I also don't think adding the `enabled` flag to all those methods is worth the squeeze. And given that this can be achieved easily enough using language-level constructs, it's a no from me. Closing for now, but thanks for your work on this, Bruno  🙌

max-mykhailenko:
@colinhacks do you recommend to have something like
```
function makeSchemaStrict<T extends z.ZodTypeAny>(schema: T) {
  if(process.env.NODE_ENV === 'production'){
    return schema
  }
  return schema.strict();
}
```

Issue #126, opened 2020-08-25
Support for date, time and date-time format strings


maneetgoyal:
Currently, Zod is lacking support for validating date, time and date-time stamps. As a result, users may need to implement custom validation via [`.refine`](https://github.com/vriad/zod#custom-validation). But the use case for date/time/date-time is very common, so if Zod ships it out-of-the-box like it does for `uuid`, `email`, etc, it would be helpful.

- Stemmed from https://github.com/vriad/zod/issues/30#issuecomment-679383286
- AJV's in-built regexes are [here](https://github.com/ajv-validator/ajv/blob/master/lib/compile/formats.js#L37) for reference.

lazharichir:
Agreed! Although date and time validation is a tricky issue as different developers expect different data (ISO? YYYY-MM-DD? 12h Time? With milliseconds? Timezone notation?).

maneetgoyal:
- Now that `zod.string().regex(...)` is available since `v1.11`, it has become easier to implement custom date/time/date-time patterns. 
- Some out-of-the-box support for date/time/etc. can still be useful though. Using `zod.string().regex()`, users can modify the default behavior if it doesn't meet their expectations. 
- The default behavior I was proposing is from AJV which is a hugely popular data validation library. I am assuming they'd have brought in many improvements to their regex over time making it increasingly robust. But yes, there can be some unhandled cases as suggested above in https://github.com/vriad/zod/issues/126#issuecomment-680782741 and in https://github.com/vriad/zod/issues/30#issuecomment-673133452.

maneetgoyal:
If it helps, [here's](https://github.com/korzio/djv/blob/master/lib/utils/formats.js#L16) how `djv` is handling date-time strings. Their approach looks a bit similar to the idea that emerged in https://github.com/vriad/zod/issues/120 (i.e. using instantiation as a way to validate).

However, this approach may come with its own [drawbacks](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#:~:text=Description,method%20and%20the%20Date%20object.).

grreeenn:
I use the `.refine()` method for these cases, works like a charm

marlonbernardes:
Hey @vriad are you still open to having AJV date/time/date-time regexes as built-ins? If so I'll open a PR sometime this week.

vriad:
## RFC for datetime methods

[...]

PR #1488, opened 2022-10-13
Add `.utc()` string type

samchungy:
Adds `.utc()` from: https://github.com/colinhacks/zod/issues/126#issuecomment-708589382.

I think milliseconds being required makes sense but I'm happy to be corrected. I'm also happy to rename this matcher to be more specific to being a matcher for the [`.toISOString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) format which I think is fairly used and easy to access in Javascript.

`.dateIsoString()` could also work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -489,6 +489,7 @@ z.string().email();
 z.string().url();
 z.string().uuid();
 z.string().cuid();
+z.string().utc(); // Date.prototype.toISOString() format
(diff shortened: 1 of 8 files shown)


samchungy:
Closing to raise a slightly different PR to unlink from the existing issue

PR #1492, opened 2022-10-14
Removed BRAND from ZodBrand Input definition

Xetera:
A type's brand should only be accessible in its output aka `z.infer` or `z.output` but it's also carried over to its input. This makes it difficult to create a typesafe parse function as the code doesn't compile

For example:
```ts
const Age = z.number().min(18).brand<"Age">()

type AgeInput = z.input<typeof Age>
```
an input for `Age.parse(x)` aka `AgeInput` should just be `number` but it's `number & BRAND<"Age">` which doesn't make sense

[Working typescript playground example](https://www.typescriptlang.org/play?ssl=3&ssc=46&pln=3&pc=1#code/JYWwDg9gTgLgBAbzgLzgXzgMyhEcDkyEAJvgFBkDGEAdgM7wCCA5gKZwC8KAdDQK4gARqygAKAJTcQwGqICMADkmCoAQxrEAPACIWrbQD4JFTHxqUYwWnBx8YI0TLB2AXHH5CR4xGThwwqlB0rI40zjDiZGgmZhZWNP6BwaKqbG7I3E52mjAAnmCsEJhwegbeCFFAA)

### DIFF
--- a/deno/lib/__tests__/branded.test.ts
+++ b/deno/lib/__tests__/branded.test.ts
@@ -47,6 +47,17 @@ test("branded types", () => {
     true
   );
 
+  // keeping brands out of input types
+  const age = z.number().brand<"age">();
+
+  type Age = z.infer<typeof age>;
+  type AgeInput = z.input<typeof age>;
+
+  util.assertEqual<AgeInput, Age>(false);
+
+  util.assertEqual<number, AgeInput>(true);
+  util.assertEqual<number & z.BRAND<"age">, Age>(true);
+
   // @ts-expect-error
   doStuff({ name: "hello there!" });
 });
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -3795,11 +3795,7 @@ export type BRAND<T extends string | number | symbol> = {
 export class ZodBranded<
   T extends ZodTypeAny,
   B extends string | number | symbol
-> extends ZodType<
-  T["_output"] & BRAND<B>,
-  ZodBrandedDef<T>,
-  T["_input"] & BRAND<B>
-> {
+> extends ZodType<T["_output"] & BRAND<B>, ZodBrandedDef<T>, T["_input"]> {
   _parse(input: ParseInput): ParseReturnType<any> {
     const { ctx } = this._processInputParams(input);
     const data = ctx.data;
--- a/src/__tests__/branded.test.ts
+++ b/src/__tests__/branded.test.ts
@@ -46,6 +46,17 @@ test("branded types", () => {
     true
   );
 
+  // keeping brands out of input types
+  const age = z.number().brand<"age">();
+
+  type Age = z.infer<typeof age>;
+  type AgeInput = z.input<typeof age>;
+
+  util.assertEqual<AgeInput, Age>(false);
+
+  util.assertEqual<number, AgeInput>(true);
+  util.assertEqual<number & z.BRAND<"age">, Age>(true);
+
   // @ts-expect-error
   doStuff({ name: "hello there!" });
 });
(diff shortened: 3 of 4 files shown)


colinhacks:
Great call. Thanks!

KATT:
I want the opposite behavior of this 😢

KATT:
opened #3380 where we could have both 🙃

colinhacks:
The inferred input type is supposed to be the *set of inputs that will pass validation*. So actually, this inferred input type for this schema:

```ts
z.number().brand<"USD">();
```

should be

```
number | (number & BRAND<"USD">)
```

Zod should definitely allow a "pre-branded" value, but it also "brands" any plain `number` values that are passed in. I can't tell if this pleases everyone or no one...but I do think it's the most correct behavior.

KATT:
I'll add some context on what I'm doing.

I'm building a query string builder for filtering data where one of the parts is a date range

A date range is serializable to the URL and transformable into a real date range.

Some examples of _input_ values:

- this_week
- past_week
- 2023-04-01,2024-04-30

for the last one, I want to _brand_ it as `string & X`, otherwise all of them will be `string` in the union, which is annoying

in my case I don't care about the brand on the output at all (but it doesn't hurt it either)

### Code

```tsx
import { Temporal } from '@js-temporal/polyfill';
import { z } from 'zod';

type DateRange = [Temporal.PlainDate, Temporal.PlainDate];
function getWeekDateRange(date: Temporal.PlainDate): DateRange {
  const start = date.subtract({ days: date.dayOfWeek });

  const end = start.add({ days: 6 });

  return [start, end];
}

export const anyDateRange = z.literal('any').transform(() => null);

export const thisWeekDateRange = z
  .literal('this_week')
  .transform((): DateRange => {
    const now = Temporal.Now.plainDate('iso8601');

    return getWeekDateRange(now);
  });

export const customDateRange = z
  .string()
  .regex(/^\d{4}-\d{2}-\d{2},\d{4}-\d{2}-\d{2}$/)
  .transform((value): DateRange => {
    const parts = value.split(',') as [string, string];
    const start = Temporal.PlainDate.from(parts[0]);
    const end = Temporal.PlainDate.from(parts[1]);

    return [start, end];
  })
  .brand('custom_date_range');

export const filters = z.object({
  dateRange: z
    .union([
      //
      anyDateRange,
      thisWeekDateRange,
      customDateRange, // <- try commenting out this line
    ])
    .default('any'),
});

type Input = z.input<typeof filters>;
type Output = z.output<typeof filters>;

type DateRangeInput = Input['dateRange'];
//     ^?

```

colinhacks:
Even if the brand was applied to the input type, it's still pretty un-ergonomic for the end user to cast a string to the branded type:

```ts
type customDateRange = z.output<typeof customDateRange>;
const value = "2021-01-01,2021-01-02" as unknown as customDateRange;
```

How about this?

```ts
const dateRangeParser = z
  .string()
  .regex(/^\d{4}-\d{2}-\d{2},\d{4}-\d{2}-\d{2}$/)
  .transform((value): DateRange => {
    const parts = value.split(',') as [string, string];
    const start = Temporal.PlainDate.from(parts[0]);
    const end = Temporal.PlainDate.from(parts[1]);

    return [start, end];
  })
  .brand('custom_date_range');

const customDataRange = z
  .custom<`${number}-${number}-${number},${number}-${number}-${number}`>()
  .pipe(dateRangeParser);

type customDataRange = z.input<typeof customDataRange2>; 
// `${number}-${number}-${number},${number}-${number}-${number}`
```

KATT:
That's awesome. Thanks ♥

PR #1494, opened 2022-10-15
add `.utc()`/`.iso8601()` string formats

samchungy:
Adds `.utc()` and `.iso8601()` from https://github.com/colinhacks/zod/issues/126.

Through this PR I learnt that dates can be very inconsistent but still valid 🤯 eg. you can give any number of ms which to me is a little silly. Imagine having `2022-07-08T00:00:00.000Z` alongside `2022-07-08T00:00:00.0000000000000000Z` in your system.

So I've tried to come up with a solution which is flexible but can be customised to be more strict if users wish.

`ms: true` and `ms: false` control whether users want to check for `ms` or not with the default being to accept both formats. I've also added the ability to specify the ms digit precision with `msLength` since technically any number of milliseconds is considered a "valid" timestamp 🙄 .

```ts
// date formats
z.string().utc();
z.string().utc({ ms: true }); // ms only
z.string().utc({ ms: true, msLength: 3 }); // ms only, 3 digit precision
z.string().utc({ ms: false }); // no ms only

z.string().iso8601(); // allows offset
z.string().iso8601({ ms: true }); // ms only
z.string().iso8601({ ms: true, msLength: 5 }); // ms only, 5 digit precision
z.string().iso8601({ ms: false }); // no ms only
```

Lastly I've added getters for both of these which take the same args so users can find ones which they have specifically created.

```ts
// date formats
z.string().utc().isUTC(); // true
z.string().utc({ ms: true }).isUTC({ ms: true }); // true
z.string().utc({ ms: true, msLength: 3 }.isUTC({ ms: true, msLength: 3 }) // true
z.string().utc({ ms: false}).isUTC({ ms: false }); // true
z.string().utc().isUTC({ any: true }); // matches all the above
```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -462,13 +462,9 @@ type ZodStringCheck =
   | { kind: "regex"; regex: RegExp; message?: string }
   | { kind: "trim"; message?: string }
   | {
-      kind: "utc";
-      options?: StringDateOptions;
-      message?: string;
-    }
-  | {
-      kind: "iso8601";
-      options?: StringDateOptions;
+      kind: "datetime";
+      offset: boolean;
+      precision: number | null;
       message?: string;
     };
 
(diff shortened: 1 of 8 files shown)


src/types.ts
```diff
@@ -460,7 +460,8 @@ type ZodStringCheck =
   | { kind: "startsWith"; value: string; message?: string }
   | { kind: "endsWith"; value: string; message?: string }
   | { kind: "regex"; regex: RegExp; message?: string }
-  | { kind: "trim"; message?: string };
+  | { kind: "trim"; message?: string }
+  | { kind: "utc"; options?: { milliseconds?: Boolean }; message?: string };
```
samchungy, line 464: Went with an options object here for better visual clarity.

`.utc(true)`/`.utc(false)` did not read as well to me.

src/types.ts
```diff
@@ -599,6 +600,28 @@ export class ZodString extends ZodType<string, ZodStringDef> {
           });
           status.dirty();
         }
+      } else if (check.kind === "utc") {
+        const parsedDate = new Date(input.data);
+
+        const isDate = !isNaN(parsedDate.getTime());
+        const isMsString = isDate && parsedDate.toISOString() === input.data;
```
colinhacks, line 607: Interesting approach but I think this is better implemented with regexes for the "ISO with ms" and "ISO without ms".

colinhacks (changes requested):

colinhacks:
This looks great, thanks! Any interest in adding the `.iso8601()` method to this PR too? 😛

samchungy:
> This looks great, thanks! Any interest in adding the `.iso8601()` method to this PR too (which allows offsets)? 😛
> 
> I like this PR more than the other.

Only because you asked nicely 😛

samchungy:
Question is - it looks like the regex I copy pasted supports an unlimited number of ms. Should we cap it at 3 or should we provide an option which allows someone to configure it. Eg. By default only accept 3 ms characters

colinhacks (approved):

colinhacks:
Sorry, after all this I had a change of heart and went a pretty different direction. Take a look and let me know if you have any thoughts.

samchungy:
> Sorry, after all this I had a change of heart and went a pretty different direction. I just pushed a couple commits to this PR to reflect my latest thinking on this API.
> 
> ```ts
> z.string().datetime();  // millis supported but optional, offsets not allowed
> z.string().datetime({ offset: true });  // millis supported but optional, offsets allowed
> z.string().datetime({ precision: 3 });  // millis required to 3 decimal places, offsets not allowed
> z.string().datetime({ precision: 0 });  // millis not allowed, offsets not allowed
> z.string().datetime({ precision: 3, offset: true });  // millis required to 3 decimal places, offsets allowed
> ```
> 
> The main driver of this approach is that the result of `new Date().toISOString()` should pass validation against `z.string().datetime()` out of the box, so millis should be supported without additional configuration.

Oooo I do like this approach. A few configs which are merged together like precision and offset. I like it. Let me know if you want me do anything :)

colinhacks:
Awesome. Thanks for your work on this Sam! Hard to believe it's been two years since I first wrote out that mini-RFC for `.utc()/.iso8601()` 🫠 Feels great to finally merge something here.

devj3ns:
Awesome addition to zod 😍

@colinhacks when will this feature be released on npm?

mauriciabad:
@colinhacks any news about the released on npm?

samchungy:
> @colinhacks any news about the release on npm?

If you're super keen jump on the beta [3.20.0-beta.0](https://www.npmjs.com/package/zod/v/3.20.0-beta.0). Release notes are [here](
https://github.com/colinhacks/zod/releases/tag/v3.20.0).

Otherwise, we wait! 😄 Please remember Colin is just a solo person working on this.

samchungy:
Nevermind it's officially in https://github.com/colinhacks/zod/releases/tag/v3.20 :)

PR #1502, opened 2022-10-17
branded type as normal argument

KATT:
I sometimes use zod in regular `.js`-files and would like to leverage branded types there.

This adds an overload to make these two the same:

```ts
z.string().brand<"X">()
z.string().brand("X")
```

Can clean up tests and recheck docs if you like the idea.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1950,8 +1950,7 @@ z.intersection(z.object({ name: z.string() }), z.object({ age: z.number() }));
 
 ### `.brand`
 
-`.brand<T>() => ZodBranded<this, B>`, or
-`.brand<T>(brand: T) => ZodBranded<this, B>`
+`.brand<T>() => ZodBranded<this, B>`
 
 TypeScript's type system is structural, which means that any two types that are structurally equivalent are considered the same.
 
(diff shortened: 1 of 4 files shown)


scotttrinh:
I can't think of a reason to block this, but branding specifically is of little use in places where there is not a type checker. Is this to help with the pseudo checking that LSP provides through TypeScript or something?

colinhacks:
I don't like this idea, since it would seem to imply that we do something with the runtime value passed in. Branding is a static-only thing and I think the API should reflect that.

KATT:
For further context, I have an `env.js` file because it allows me to use it in my `next.config.js` without transpilation. Typing works fine even if I import a JS-file everywhere I want to use it.

Example:

- https://github.com/trpc/trpc/blob/next/examples/next-prisma-starter/next.config.js#L1-L3
- https://github.com/trpc/trpc/blob/next/examples/next-prisma-starter/src/server/env.js


The workaround is to do some babel-hack to make my `next.config.js` into TS. At the end of the day, this is because Next.js doesn't allow my config file to be TS out-of-the-box & this was the least painful way I've encountered to deal with this.

Issue #1471, opened 2022-10-07
intersection of records differs from typescript


bug-brain:
```ts
const type = z.intersection(
  z.object({ a: z.null() }),
  z.record(z.string().refine((s): s is "b" => s === "b"), z.undefined()),
)
const input: z.infer<typeof type> = { a: null, b: undefined }
type.parse(input)
```
should work according to the types but throws
```
ZodError: [
  {
    "code": "custom",
    "message": "Invalid input",
    "path": [
      "a"
    ]
  },
  {
    "code": "invalid_type",
    "expected": "undefined",
    "received": "null",
    "path": [
      "a"
    ],
    "message": "Expected undefined, received null"
  }
]
```
A key is not required to match the index signature but if it does its value should match the intersection of value types.

scotttrinh:
Confirmed that this is broken, and I agree that we should match TypeScript's behavior here. PRs welcome!

captain-yossarian:
Will pick it up.

ochan12:
Hey! I can do this. Is the desired behaviour that key/value pairs are stripped off from the object if the key parsing is not valid?

colinhacks:
Yep I agree, though some people will be annoyed - the same set that are annoyed that `z.object` also strips unknown keys by default. We'll need to add `.strict` and `.passthrough` to ZodRecord as well so there's still a way to opt into the old behavior.

ochan12:
@colinhacks I created a PR without the strict and the passthrough #1514 . Will it be necessary to add it?

chuck-flowers:
I believe this issue has been prematurely closed. The PR was also auto-closed without any feedback. Is this being addressed in a different issue?

PR #1514, opened 2022-10-24
intersection of records differs from typescript

ochan12:
Values are only parsed if the Record key is valid, otherwise the key/value pair is stripped off. 

Related issue: #1471


ochan12:
Hey @colinhacks , do you think you could take a look at this request?

Issue #1227, opened 2022-06-26
New ZodDefaultOnMismatch class for providing a default value when there is a type mistmatch


seancrowe:
## Summary
A new class ZodDefaultOnMismatch which will replace the data with the default value when the provided value is a mismatch in type with the expected value. When data is undefined, ZodDefaultOnMismatch acts like ZodDefault.

```javascript
const deleteAllFiles = z.boolean().defaultOnMismatch(false).parse("monkeys")  // deleteAllFiles = false
```

Below, I layout two real world use cases. Whether or not this gets accepted, we will need this functionality, and I am hoping that we don't have to work off a forked version of zod.

Below, I also layout three implementations. All are tested and ready to be merged. I just need @colinhacks to tell me which is a sounder implementation, so I know which branch to make a pull request. Also, if the naming should be changed, I am up for that as well.

The implementation I suggest is the first one: https://github.com/seancrowe/zod/commit/40579cacf411f37a0458dd1abf85b8aba679fb81

So, if you don't feel like reading through all these implementations, the first one is the simplest and logical choice at the expense of JavaScript users requiring discipline.

### Use Case 1
Application is pulling data from a JSON file where both the data's type is not guaranteed to be correct or even defined. Our application wants to use zod to verify the parsed JSON, but we don't want to fail if there is a type mismatch, but instead default to a safe value.

### Use Case 2
Our application uses an FP technique of passing data around as objects. This data is passed to a function, the function will copy the data, modify the data, and potentially add more data. Due to JavaScript being dynamic typed and TypeScript not having sound typing, the types defined by TypeScript are not guaranteed 100% at compile time. This scenario means that as we pass these data objects around, we cannot be guaranteed 100% that the data object is exactly as we expect.

To defend against this, a type check must be made everytime. If there is a type mismatch, we don't want to cause an error, but instead default to a safe value with the correct type. Since these objects can be complex or contain extra data our current function does not care about, we need make sure the input and output is untouched accept for the specific properties we are interested in.

## ZodDefaultOnMismatch
Unfortunately, Zod does not have built way to support the two above use cases. Maybe it can be done with preprocess, but for our use case, that defeats the purpose of zod if every object property needs a handwritten type check anyway. 

Instead, all zod needs is something like ZodDefault but that will use the default value not only when there is undefined but when there is a mismatch in types. Thus ZodDefaultOnMismatch.

[...]

PR #1537, opened 2022-11-02
#1227 Feature default on mismatch

seancrowe:
# Feature
Below, I layout two real world use cases. Unfortunately, Zod does not have built way to support the two use cases. Maybe it can be done with preprocess, but for our use case, that defeats the purpose of zod if every object property needs a handwritten type check anyway.

Instead, all zod needs is something like ZodDefault but that will use the default value not only when there is undefined but when there is a mismatch in types. Thus ZodDefaultOnMismatch.

A new class ZodDefaultOnMismatch which will replace the data with the default value when the provided value is a mismatch in type with the expected value. When data is undefined, ZodDefaultOnMismatch acts like ZodDefault.

```typescript
const deleteAllFiles = z.boolean().defaultOnMismatch(false).parse("monkeys")  // deleteAllFiles = false
```

# Use Case

## Use Case 1

Application is pulling data from a JSON file where both the data's type is not guaranteed to be correct or even defined. Our application wants to use zod to verify the parsed JSON, but we don't want to fail if there is a type mismatch, but instead default to a safe value.

## Use Case 2

Our application uses an FP technique of passing data around as objects. This data is passed to a function, the function will copy the data, modify the data, and potentially add more data. Due to JavaScript being dynamic typed and TypeScript not having sound typing, the types defined by TypeScript are not guaranteed 100% at compile time. This scenario means that as we pass these data objects around, we cannot be guaranteed 100% that the data object is exactly as we expect.

To defend against this, a type check must be made everytime. If there is a type mismatch, we don't want to cause an error, but instead default to a safe value with the correct type. Since these objects can be complex or contain extra data our current function does not care about, we need make sure the input and output is untouched accept for the specific properties we are interested in.

# Notes
At first, I named this ZonDefaultAlways, but that implied it would always default to a value no matter what. ZodDefaultOnMismatch, while lengthy, is descriptive in what it does. However, I am not set on ZodDefaultOnMismatch. Maybe there is a better name?

See issue #1227 for alternative implementations.


colinhacks:
Hi! I really like this but it's not quite the API or behavior I had in mind. I've updated the PR to reflect a general-purpose `.catch` method that will detect any parsing failure (not just `invalid_type`) and return a default "catch value".

seancrowe:
@colinhacks Sorry for the late reply. I tried to start a convo #1227, as I assume you would have some opinions. However, we are all busy... so no worries. I was too busy to reply to the changes. Of course they make sense to the overall project.

Happy this idea got merged. I will need to make a few changes for my production software to get align with these changes.

PR #1542, opened 2022-11-05
fix #1046 `.required()` doesn't remove optional flag from the result of `.nullish()`.

### LINKED ISSUE #1046: `.required()` doesn't remove optional flag from the result of `.nullish()`
The (undocumented) function `.required` is not working properly for properties defined with `.nullish()`:

```js
const base = z.object({
  required: z.string(),
  nullable: z.string().nullable(),
  optional: z.string().optional(),
  nullableOptional: z.string().nullable().optional(),
  nullish: z.string().nullish(),
})

const req = base.required()
```

The resulting inferred type of `req` is:
```ts
{
    required: string;
    nullable: string | null;
    optional: string;
    nullableOptional: string | null;
    nullish?: string | null | undefined;
}
```

As you can see, all fields that were defined as `.optional()` have properly been made non-optional except for the one created using `.nullish()`

This seems to arise from the fact that `.nullish()` is defined to return a `ZodNullable<ZodOptional<T>>` and `.required` simply returns the outermost `ZodOptional` without doing any type recursion, and as such it cannot see the optional in that type.

igalklebanov:
Hey :wave:

Love zod, we use it in production.. they used to call me the "zod guy" cause I'm the one who introduced it to the codebase. lol.

First time contributing here, sorry If I didn't do things properly.

fixes #1046.

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -779,7 +779,7 @@ nullableString.parse(null); // => null
 Or use the `.nullable()` method.
 
 ```ts
-const E = z.string().nullable(); // equivalent to D
+const E = z.string().nullable(); // equivalent to nullableString
 type E = z.infer<typeof E>; // string | null
 ```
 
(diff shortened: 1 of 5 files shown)


colinhacks:
Thanks! Everything looks great :)

PR #1555, opened 2022-11-12
add `fatal` to `ZodIssue`.

### LINKED ISSUE #1406: `fatal` missing from issue-type
From [my question asked on Discord](https://discord.com/channels/893487829802418277/893488038477434881/1018893492342226994):

When adding an issue (e.g. via `ctx.addIssue` in `superRefine`), you can set whether the issue is `fatal` or not:

```ts
    ctx.addIssue({
      code: 'custom',
      fatal: true, // Allowed
      params: {
        foo: 'bar',
      },
      path: [],
      message: 'Some message',
    });
```
But when manually creating a new `ZodError` with a matching issue, Typescript complains that `fatal` does not exist:
```
    const err = new z.ZodError([
      {
        code: 'custom',
        fatal: true, // Not allowed, with typescript-error below
        params: {
          foo: 'bar',
        },
        path: [],
        message: 'Some message',
      },
    ]);
```
> Object literal may only specify known properties, and 'fatal' does not exist in type 'ZodCustomIssue & { message: string; }'

As mentioned in the reply to my question on Discord, there seems to be an oversight here. Is `fatal` is missing on `ZodCustomIssue` or `ZodIssueBase`, or is it "correct" that it doesn't exist there? I'm guessing not, since the `fatal` field does come back out on an issue when a fatal issue is created. Typewise, it seems to only be added to `IssueData`, which is the type `ctx.addIssue` accepts?

---

The reason why I ran into this, is because I have written a few "custom schemas" for often used "types", for re-use throughout my code-base, and to make sure they're done correctly, I've written some unit-tests for them:

```ts
    it('invalid should generate expected issue', () => {
      expect(AccountNumberSchema.safeParse('1234 56 78907')).toEqual({
        success: false,
        error: new z.ZodError([
          {
            code: 'custom',
            fatal: true, // Gives TS error, as mentioned
            params: {
              code: 'invalid_type',
              expected: 'account_number',
            },
            path: [],
            message: 'must be a valid account number',
          },
        ]),
      });
    });
```

igalklebanov:
Hey 👋 

closes #1406.

### DIFF
--- a/src/ZodError.ts
+++ b/src/ZodError.ts
@@ -147,7 +147,10 @@ export type ZodIssueOptionalMessage =
   | ZodNotMultipleOfIssue
   | ZodCustomIssue;
 
-export type ZodIssue = ZodIssueOptionalMessage & { message: string };
+export type ZodIssue = ZodIssueOptionalMessage & {
+  fatal?: boolean;
+  message: string;
+};
 
 export const quotelessJson = (obj: any) => {
   const json = JSON.stringify(obj, null, 2);
@@ -296,7 +299,6 @@ export type IssueData = stripPath<ZodIssueOptionalMessage> & {
   path?: (string | number)[];
   fatal?: boolean;
 };
-export type MakeErrorData = IssueData;
 
 export type ErrorMapCtx = {
   defaultError: string;
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -779,7 +779,7 @@ nullableString.parse(null); // => null
 Or use the `.nullable()` method.
 
 ```ts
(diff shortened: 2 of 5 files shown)


src/ZodError.ts
```diff
@@ -296,7 +299,6 @@ export type IssueData = stripPath<ZodIssueOptionalMessage> & {
   path?: (string | number)[];
   fatal?: boolean;
 };
-export type MakeErrorData = IssueData;
```
igalklebanov, line 299: This one's unused btw.. 😕

colinhacks:
Sounds good to me!

Issue #1473, opened 2022-10-07
zod.preprocess causes compilation error starting from 3.19.1


azhiv:
After the recent release v3.19.1, the following code started to fail [compiling](https://codesandbox.io/s/beautiful-khayyam-kx97bw?file=/src/index.ts:0-282):
```ts
import * as zod from "zod";

const schema = zod.object({
  failureReasons: zod.preprocess(
    rel => [rel],
    zod.array(zod.object({})),
  )
});

export class Editor<
  S extends zod.ZodType<zod.TypeOf<typeof schema>>
    = typeof schema                    // <-- the error pops here
> {
  constructor(public schema: S) {}
}

```
This is the error from sandbox:
```
Type 'ZodObject<{ failureReasons: ZodEffects<ZodArray<ZodObject<{}, "strip", ZodTypeAny, {}, {}>, "many">, {}[], unknown>; }, "strip", ZodTypeAny, { ...; }, { ...; }>' does not satisfy the constraint 'ZodType<{ failureReasons?: {}[]; }, ZodTypeDef, { failureReasons?: {}[]; }>'.
  The types of '_input.failureReasons' are incompatible between these types.
    Type '{}' is missing the following properties from type '{}[]': length, pop, push, concat, and 26 more.ts(2344)
```
@colinhacks The error is not reproduced with zod v3.19.0, I guess the root cause is this [commit](https://github.com/colinhacks/zod/commit/2893e1659875944198aff28a3ba404858f834ed8).

PS When I replicate the same setup in VSCode locally, the error is slightly different:
```
Type 'ZodObject<{ failureReasons: ZodEffects<ZodObject<{}, "strip", ZodTypeAny, {}, {}>, {}, unknown>; }, "strip", ZodTypeAny, { failureReasons: {}; }, { ...; }>' does not satisfy the constraint 'ZodType<{ failureReasons: {}; }, ZodTypeDef, { failureReasons: {}; }>'.
  The types of '_input.failureReasons' are incompatible between these types.
    Type 'unknown' is not assignable to type '{}'.ts(2344)
```
I couldn't locate the difference, but the general behaviour is still reflecting the error, so I post the issue as is.

ecyrbe:
hello, 
i think issues are not just related to `preprocess`. The new ZodType definition change made my typescript compiler to become extremly slow and bug on many generics with the typescript error `definition is possibly infinite`.
I had to rolled back to v3.19.0

mlev:
[...]

Issue #1458, opened 2022-10-03
Any way to avoid casting preprocess return type, normally returns unknown


abacaj:
When using preprocess the return type is "unknown":

schema:
```
export const Size = zod.object({
  width: z.preprocess((input) => {
    if (isNil(input)) return undefined;

    return parseInt(input as string, 10);
  }, z.number()),
});
```

type:
```
zod.ZodEffects<zod.ZodNumber, number, unknown>
```

casting it works:
```
export const Size = zod.object({
  width: z.preprocess((input) => {
    if (isNil(input)) return undefined;
    return parseInt(input as string, 10);
  }, z.number()) as zod.ZodEffects<zod.ZodNumber, number, number>
});
```

I did create a helper function:
```
function prepocessNumber() {
  return z.preprocess((input) => {
    try {
      return parseInt(input as string, 10);
    } catch (ex) {
      return input;
    }
  }, z.number()) as zod.ZodEffects<zod.ZodNumber, number, number>
}

export const Size = zod.object({
  width: prepocessNumber()
});
```

which gives me the expected return type (number).

Isn't clear to me if there is an alternative but maybe this is correct?

azhiv:
@abacaj Can you please check if this is relevant for v3.19.0? My guess is that the recent update 3.19.1 contains an "enhancement" that breaks the existing behaviour. I may have submitted a duplicated issue though, let's see what the dev team responds.

Xetera:
Can confirm this is fixed by pinning the version to 3.19.0. Thought it was a limitation of zod, glad to see it's just a regression

denvaki:
sorry for opening this, but I still experience unknown type. I tried author example and w/o casting it has `zod.ZodEffects<zod.ZodNumber, number, unknown>`.
Package version is 3.19.1

Royal-lobster:
I am facing the same issue, for parsing the json

```
z.preprocess(
    (v) => (typeof v === "string" ? JSON.parse(v as string) : v),
    z.array(jsonSchema),
)
```

I am expecting it to patch jsonSchema type and throw error if it does not on runtime. is there any other way to achieve this without type cast

PR #1557, opened 2022-11-13
Revert "Make unknown input type for preprocess"

azhiv:
This reverts commit 2893e1659875944198aff28a3ba404858f834ed8.
Make the input type generic again. The original change breaks existing
behaviour, its motivation is vague.

This PR aims to fix a couple of issues raised since 3.19.1 was released: https://github.com/colinhacks/zod/issues/1473 and https://github.com/colinhacks/zod/issues/1458.

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -3591,7 +3591,7 @@ export class ZodEffects<
     preprocess: (arg: unknown) => unknown,
     schema: I,
     params?: RawCreateParams
-  ): ZodEffects<I, I["_output"], unknown> => {
+  ): ZodEffects<I, I["_output"]> => {
     return new ZodEffects({
       schema,
       effect: { type: "preprocess", transform: preprocess },
--- a/src/__tests__/transformer.test.ts
+++ b/src/__tests__/transformer.test.ts
@@ -201,7 +201,7 @@ test("preprocess", () => {
 
   const value = schema.parse("asdf");
   expect(value).toEqual(["asdf"]);
-  util.assertEqual<typeof schema["_input"], unknown>(true);
+  util.assertEqual<typeof schema["_input"], string[]>(true);
 });
 
 test("async preprocess", async () => {
--- a/src/types.ts
+++ b/src/types.ts
@@ -3591,7 +3591,7 @@ export class ZodEffects<
     preprocess: (arg: unknown) => unknown,
     schema: I,
     params?: RawCreateParams
-  ): ZodEffects<I, I["_output"], unknown> => {
+  ): ZodEffects<I, I["_output"]> => {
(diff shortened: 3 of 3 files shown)


colinhacks:
The commit this reverts only sets the `Input` type to unknown, not the output type. This is working as intended:

```ts
export const Size = z.object({
  width: z.preprocess((input) => {
    if (input === null) return undefined;

    return parseInt(input as string, 10);
  }, z.number()),
});

const arg = Size.parse({}).width; // this is `number`
```

Are you seeing something different?

azhiv:
@colinhacks I'm not sure how exactly the other folks encounter the issue, but in our case it arises because `typeof schema` is no longer assignable to `z.ZodType<z.TypeOf<typeof schema>>` (when schema contains a `preprocess` directive).
To provide you with a bit more context, consider the following piece of [code](https://codesandbox.io/s/beautiful-khayyam-kx97bw?file=/src/index.ts:0-282):
```ts
import * as zod from "zod";

const schema = zod.object({
  failureReasons: zod.preprocess(
    rel => [rel],
    zod.array(zod.object({})),
  )
});

export class Editor<
  S extends zod.ZodType<zod.TypeOf<typeof schema>>
    = typeof schema
> {
  constructor(public schema: S) {}
}
```
Having declared this base configuration, we may want to extend the schema and declare another class with an updated schema and editor class:
```ts
const descriptionSchema = schema.and(
  z.object({
    description: z.string(),
  })
)

class DescriptionEditor extends Editor<typeof descriptionSchema> {}
```
This is convenient because the validation schema and Editor infrastructure are reused. But starting from v3.19.1 we are no longer allowed to do that. This PR is aimed to address this problem.

PS I don't insist on reverting your commit, we could probably come up with a solution that satisfies both sides.

colinhacks:
I see. This would be an issue if you're assigning to `z.ZodType<z.TypeOf<typeof schema>>` (which should rarely be necessary). The reason is that `ZodType` actually accepts 3 generic arguments:

```ts
export abstract class ZodType<
  Output = any,
  Input = Output
> {}
```

As you can see the `Input` defaults to the value of `Output` which causes the error when trying to assign a `ZodPreprocess` instance. Try this instead:

```ts
const arg: ZodType<number, ZodTypeDef, unknown> = z.preprocess(...)
```

azhiv:
@colinhacks I'm not sure if this is what you meant, but explicitly specifying generic parameters for ZodType solves the original problem:
```ts
export class Editor<
  S extends zod.ZodType<zod.TypeOf<typeof schema>, zod.ZodTypeDef, unknown>
    = typeof schema
> {
  constructor(public schema: S) {}
}
```
Thank you!

Issue #1563, opened 2022-11-16
`pick`, `omit`, `require` & `partial` should use arrays instead of objects


janpaepke:
I was wondering what the rationale behind `pick`, `omit`, `require` and `partial` expecting objects is?

Each value is optional and can only be true, so using
```ts
.pick(['foo', 'bar'])
```
feels more sensible than
```ts
.pick({ foo: true, bar: true })
```

There is another reason than convenience:  
Currently you can (accidentally) add properties, which do not exist on the original object.  
With an array you could ensure type safety.  

The signature would look like this
```ts
pick<Mask extends Array<keyof T>>(mask: Mask): ZodObject<Pick<T, Extract<keyof T, keyof Mask>>, UnknownKeys, Catchall>;
```

If you agree I could offer to implement this.
Either in a backwards compatible way (also accepting objects, using overloads) or as a breaking change.

finaloop-igal:
> Currently you can (accidentally) add properties, which do not exist on the original object.
With an array you could ensure type safety.

This is solvable, without going the "array way". Have recently opened a PR that solves this in Kysely.

igalklebanov:
> There is another reason than convenience:
Currently you can (accidentally) add properties, which do not exist on the original object.
With an array you could ensure type safety.

Check out #1564, it's fixable.

janpaepke:
Hi Igal,
thank you for your response and contribution for making unknown properties fail.
I am still wondering though about the rationale.
Why would we be using an Object here? An Array makes way more sense to me.

BenWaNH:
I agree with @janpaepke 
We'll should always set to true
And it will be more easier to use, when receiving an array of pick props that we'll can set in pick method directly instead parse in object.

janpaepke:
*updated title and description, to include `require` and `partial`*

Again: I am willing to add a PR for this, but want to understand the reason for using an object in the first place.

janpaepke:
@colinhacks May I ask for your insight?

capaj:
@janpaepke Prisma uses the same object notation for selecting fields. 
IMHO it is easier to type correctly in typescript. I don't mind the extra verbosity, but you're right that doing `pick({field: false})` doesn't make sense so using a Boolean might be a bit overkill

colinhacks:
Too much of a footgun. Many many people will inevitably try this:

```ts
const keys = ['a', 'b'];

const schema = z.object({
  a: z.string(),
  b: z.string(),
  c: z.string()
[...]

PR #1564, opened 2022-11-16
deny unexpected keys @ `ZodObject`'s `.omit(mask)`,`.pick(mask)`,`.required(mask)` & `.partial(mask)` at compile time.

igalklebanov:
Hey 👋 

`ZodObject` methods that receive a mask as input currently accept unexpected keys (keys that do not exist in object's shape) without any typescript errors. This can be safer.

Issue is also mentioned @ #1563.

This PR limits `omit()`/`pick()`/`required()`/`partial()` mask object to `ZodObject`'s shape, introducing a welcome type-safety breaking change to `zod`.

It does so without affecting autocompletion.


deno/lib/__tests__/pickomit.test.ts
```diff
@@ -85,6 +85,7 @@ test("pick a nonexistent key", () => {
 
   const pickedSchema = schema.pick({
     a: true,
+    // @ts-expect-error
     doesntExist: true,
```
igalklebanov, line 89: this now throws a ts error, which is proof that the change did work. But I believe we should still do some typescript testing like `tsd`..

colinhacks:
Love it, thanks!!!

igalklebanov:
Love the helper name change!

PR #1574, opened 2022-11-19
Remove datetime from README

carlgieringer:
It isn't implemented (#126)

There is recent activity in #126; let me know if this is the wrong approach. I think the changes contemplated recently in #126 will add `.utc` and would require removing `.datetime` anyways.

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,7 +501,6 @@ z.string().regex(regex);
 z.string().startsWith(string);
 z.string().endsWith(string);
 z.string().trim(); // trim whitespace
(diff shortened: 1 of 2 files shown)


samchungy:
> It isn't implemented (#126)
> 
> There is recent activity in #126; let me know if this is the wrong approach. I think the changes contemplated recently in #126 will add `.utc` and would require removing `.datetime` anyways.

Negative. We went with adding datetime. It'll be added in the next release.

rschick:
Any plans to release to npm?

carlgieringer:
Oh, sorry I missed it!

PR #1599, opened 2022-11-25
docs: likely a minor typo

JakeBruner:
_i.e._ means _that is_,
whereas
_e.g._ means _for example_

I'm assuming `.optional()` is just an example here


colinhacks:
Good catch 😛

PR #1604, opened 2022-11-25
feat: Add abort early (fatal) option to refine (#1603)

### LINKED ISSUE #1603: Add an option to .refine() to abort early
As mentioned in https://github.com/colinhacks/zod#abort-early, chained refinements are all executed, which may not be desirable in all cases. That is why `superRefine` has a way of _aborting early_ - preventing further refinements from being executed.
I propose adding this functionality to `refine` too.

DanielBreiner:
Closes #1603

`CustomErrorParams` is not used anywhere else besides `refine`, so it should be fine to add the `fatal` field directly there. The object gets then passed to `ctx.addIssue`.

### DIFF
--- a/deno/lib/__tests__/refine.test.ts
+++ b/deno/lib/__tests__/refine.test.ts
@@ -217,3 +217,25 @@ test("fatal superRefine", () => {
   expect(result.success).toEqual(false);
   if (!result.success) expect(result.error.issues.length).toEqual(1);
 });
+
+test("fatal refine", () => {
+  const Strings = z
+    .string()
+    .refine((val) => val === "", {
+      message: "foo",
+      fatal: true,
+    })
+    .superRefine((val, ctx) => {
+      if (val !== " ") {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          message: "bar",
+        });
+      }
+    });
+
+  const result = Strings.safeParse("");
+
+  expect(result.success).toEqual(false);
+  if (!result.success) expect(result.error.issues.length).toEqual(1);
+});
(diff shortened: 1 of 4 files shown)


deno/lib/__tests__/refine.test.ts
```diff
@@ -217,3 +217,25 @@ test("fatal superRefine", () => {
...
 });
+
+test("fatal refine", () => {
+  const Strings = z
+    .string()
+    .refine((val) => val === "", {
+      message: "foo",
+      fatal: true,
+    })
+    .superRefine((val, ctx) => {
```
maxArturo, line 228: @DanielBreiner Would it make sense to add also a test that flexes the `fatal` option stopping processing before the next `refine`/`superRefine`? 

As it is, you're testing the second `superRefine`'s message, but we're not testing that the `fatal` option actually stops processing. Something like your original issue use case would work great IMO!

```ts
  const testSchema = z
    .object({ test: z.literal(true) })
    .nullable()
    .refine((val) => val !== null, {
      message: "foo",
      fatal: true, // errors without this 
    })
    .superRefine((val, ctx) => {
      if ((val as any).test) {
        console.log("actually made it here");
        ctx.addIssue({
          code: z.ZodIssueCode.custom,
          message: "bar",
        });
      }
    });

  const result = testSchema.safeParse(null);
```

maxArturo (changes requested):
Thanks @DanielBreiner ! Just one comment to flex the working `fatal` option.

deno/lib/__tests__/refine.test.ts
DanielBreiner: Good idea. Added.

colinhacks:
I like this a lot and will hopefully land it in 3.21. 3.20 is already feature locked but 3.21 will follow swiftly.

mwarger:
Thanks for this!  Really looking forward to this PR to clean up some superrefines.  Do you have an ETA on 3.21?

DanielBreiner:
Any update on when we may see this merged? Or the underlying issue addressed in a different way?

PR #1619, opened 2022-11-30
Fix: dynamic zod.object() schema returned from function body with generic type, done right

dilame:
I made a small, but very important fix of `objectUtil.addQuestionMarks` type. 
TBH i have no idea why it works, but from my logical point of view this looks right. 
It should close a bunch of issues, so i will not enumerate it here.
With this fix we are able to write something like this, that was not possible before:

```ts
import { z } from 'zod';

function anyZodRefinement<T>(t: T): z.ZodType<T> {
  return z.any().refine(() => t);
}

function dynamicZodObject<T>(t: T) {
  return z.object({
    arbitrary: anyZodRefinement(t),
  });
}

function doSomethingUseful<T>(t: T) {
  const schema = dynamicZodObject(t);
  const parsed = schema.parse({ arbitrary: t });
  // In the previous implementation this fiend have not been existed
  return parsed.arbitrary;
  // ^? returns type <T> here!!!
}

const result = doSomethingUseful('test' as const)
// ^? 'test' literal type!!!
```

And it will work as expected!

Please, don't ignore this PR, i have spent the whole sleepless night and day figuring out how to fix it 😄

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -1663,14 +1663,15 @@ export namespace objectUtil {
     [k in keyof T]: undefined extends T[k] ? k : never;
   }[keyof T];
 
-  type requiredKeys<T extends object> = {
+  type filterRequiredOnly<T extends object> = {
     [k in keyof T]: undefined extends T[k] ? never : k;
-  }[keyof T];
+  };
 
-  export type addQuestionMarks<T extends object> = Partial<
-    Pick<T, optionalKeys<T>>
+  export type addQuestionMarks<T extends object> = Pick<
+    T,
+    keyof filterRequiredOnly<T>
   > &
-    Pick<T, requiredKeys<T>>;
+    Partial<Pick<T, optionalKeys<T>>>;
 
   export type identity<T> = T;
   export type flatten<T extends object> = identity<{ [k in keyof T]: T[k] }>;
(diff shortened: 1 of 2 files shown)


dilame:
nvm, this is a bad solution. I have not found a better one yet though

dilame:
Asked TypeScript team to help :)

https://github.com/microsoft/TypeScript/issues/51709#issuecomment-1333611156

dilame:
By the way, for those who look for a workaround – you can manually implement `ZodObject` type like this.

```
export type ZodObjectIdentity<T extends z.ZodRawShape, Side extends '_input' | '_output'> = {
  [K in keyof T]: T[K][Side];
};

export function zodObjectIdentity<T extends z.ZodRawShape>(
  t: T,
): z.ZodObject<T, 'strip', z.ZodTypeAny, ZodObjectIdentity<T, '_output'>, ZodObjectIdentity<T, '_input'>> {
  // @ts-ignore
  return z.object(t);
}
```

I know it's not perfect, but it works. 
The whole bug narrows down to `objectUtil.addQuestionMarks` type – it's the place where the generics get lost. Fortunately zod allows to manually redefine output type. But will be overwritten and lost again if you can any `zod.object()` method such as `.strict()`, `.merge()`, `.pick()` etc..

@colinhacks is there any chance we can add extra API to allow more verbose object description (exactly like `io-ts` in your comparison) that will preserve the generics?

PR #1620, opened 2022-11-30
Fix issue #1611

### LINKED ISSUE #1611: Create exact error message for z.array().length().
Given this schema:

`const testSchema = z.array(z.number()).length(4);`

For arrays with less than 4 elements, it returns the message: 
>Should have at least 4 items

For arrays with more than 4 elements, it returns the message: 
>Should have at most 4 items

I wish instead for it to return a message explaining that it should be precisely 4 items:
>  Should have exactly 4 items

I know I could achieve that with a custom refine validation, but it would be nice to have it baked in.

Of course, if you find this suggestion valid, I could make a PR for it.

john-schmitz:
closes #1611 

I quickly drafted a working solution so we could iterate and discuss its design.

I also added the same proposed behavior to strings as well, since both strings and arrays share the `.length` function.

Looking further into the codebase, I think it would be ideal to create a new check (something like "exact_legth") and a ZodIssueCode so we can share the same locale behavior between Arrays and Strings.

Whats your take? How should I integrate those changes into Zod's current codebase?

### DIFF
--- a/deno/lib/ZodError.ts
+++ b/deno/lib/ZodError.ts
@@ -106,6 +106,7 @@ export interface ZodTooSmallIssue extends ZodIssueBase {
   code: typeof ZodIssueCode.too_small;
   minimum: number;
   inclusive: boolean;
+  exact: boolean;
   type: "array" | "string" | "number" | "set" | "date";
 }
 
@@ -113,6 +114,7 @@ export interface ZodTooBigIssue extends ZodIssueBase {
   code: typeof ZodIssueCode.too_big;
   maximum: number;
   inclusive: boolean;
+  exact: boolean;
   type: "array" | "string" | "number" | "set" | "date";
 }
 
--- a/deno/lib/__tests__/validations.test.ts
+++ b/deno/lib/__tests__/validations.test.ts
@@ -42,6 +42,24 @@ test("array length", async () => {
   }
 });
 
+test("string length", async () => {
+  try {
+    await z.string().length(4).parseAsync("asd");
+  } catch (err) {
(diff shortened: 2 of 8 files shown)


maxArturo:
> Whats your take? How should I integrate those changes into Zod's current codebase?

IMO this seems very sensible - especially since there's no API changes and seems in line with the current way of organizing error messages. Just my $0.02 USD. Thanks for the pr @john-schmitz !

john-schmitz:
Glad to help. The impact Zod had on my workflow is immense. 
Always good to give back.

santosmarco-caribou:
Hmm... I'm not sure if I like this solution. (but correct me if I'm wrong)

You are setting the default message directly on the method, which will promote it to the highest level of priority when generating the final message.

Say, for example, I've decided to define my error map on the `.parse()` call:

```ts
const myArraySchema = z.array(z.string()).length(2).parse(["foo"], {
  errorMap: (issue) => {
    if (issue.code === "too_small") {
      return { message: "Hey yo, the array must have a length of 2." };
    }
    return { message: "Nope" };
  },
});
```

Since I'm not setting the `"Hey yo, the array must have a length of 2."` message in the `.length(2)` call, it will be overwritten by your default message because messages set directly on the check (`'length'`) call have the highest priority (and I'm not setting none, but you are).

Default messages should **always** have the lowest priority.

Can you please add a test of the above code and see the result?

colinhacks:
As @santosmarco-caribou says, the original approach does break Zod's error message hierarchy. I pushed a commit that uses a different approach, adding an optional `exact` flag (alongside `inclusive`) that we can use to customize the default error message here. Achieves the goal in a backwards compatible way.

PR #1652, opened 2022-12-08
feat: Add `extract`/`exclude` methods to `ZodEnum`

santosmarco-caribou:
This PR adds two methods to `ZodEnum`:

1. `extract`
2. `exclude`

Both work exactly like the TS utility types of the same name.

You call `.extract()` with values that exist in the enum, and it returns a new one with these values only. The same thing for `.exclude()` but it returns a new enum _without_ the values you passed in.

## Example

Screenshot 2022-12-08 at 5 52 30 AM

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -1261,7 +1261,7 @@ type MyUnion =
   | { status: "failed"; error: Error };
 ```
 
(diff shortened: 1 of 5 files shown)


colinhacks:
I dig it! Sorry I couldn't fit this into 3.20 - had to stop tinkering and just cut a release. The naming is of course perfect. This'll get merged soon.

PR #1666, opened 2022-12-12
docs: add `zod-i18n-map` to README ecosystem links

aiji42:
I am developing a library to translate zod default error messages into multiple languages. [`zod-i18n-map`](https://github.com/aiji42/zod-i18n)
It is useful and I would like to add it to the ecosystem list.


colinhacks:
This is excellent! We'll have to chat about what Zod can do to support i18n better out of the box.

aiji42:
Thank you very much. If there is anything I can do to help, I will cooperate.

naquiroz:
Could there be a way to use this without dependning on i18next?

PR #1670, opened 2022-12-12
Updated data hydration

ArthurKa:
Implement data hydration.

For example, when you need to validate any input data and get `string[]` on output, without hydration you can only get validation error or set default value (if parse `undefined`), but with hydration if data validation fails it fallbacks on what you pass to `.hydrate` method.

```ts
const schema = z.string().hydrate(/* '' by default */).array().hydrate([]);

schema.parse(undefined); // => []
schema.parse({ a: 123 }); // => []
schema.parse([]); // => []
schema.parse(['asd']); // => ['asd']
schema.parse([1]); // => ['']
schema.parse([1, 'a', 2]); // => ['', 'a', '']
```

So on output you always have `string[]`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@
   - [.transform](#transform)
   - [.default](#default)
   - [.catch](#catch)
+  - [.hydrate](#hydrate)
   - [.optional](#optional)
   - [.nullable](#nullable)
   - [.nullish](#nullish)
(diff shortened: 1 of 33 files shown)


igalklebanov (commented):
This is pretty confusing now that `.catch` is available.

JacobWeisenburger:
Is this what you are looking for?

```ts
function hydrate<Schema extends z.ZodTypeAny> ( schema: Schema, hydratedValue: any ) {
    return z.any().transform( value => {
        const result = schema.safeParse( value )
        if ( result.success ) return result.data
        return hydratedValue
    } )
}

const schema = hydrate( hydrate( z.string(), '' ).array(), [] )

const tests = [
    { value: undefined, expected: [] },
    { value: { a: 123 }, expected: [] },
    { value: [], expected: [] },
    { value: [ 'asd' ], expected: [ 'asd' ] },
    { value: [ 1 ], expected: [ '' ] },
    { value: [ 1, 'a', 2 ], expected: [ '', 'a', '' ] },
]

tests.forEach( test => {
    const result = schema.safeParse( test.value )
    if ( result.success ) {
        const passed = isEqual( result.data, test.expected )
        if ( passed ) {
            console.log( test.value, '=>', result.data )
        }
    }
} )
```

ArthurKa:
This one is pretty good, but nope. No chaining, no default parameters, no type check of hydrated parameter according to schema.

JacobWeisenburger:
> No chaining, no default parameters, no type check of hydrated parameter according to schema

What do you mean by this?

Issue #1628, opened 2022-12-03
Field becomes optional when value is any or unknown


jasonkuhrt:
```ts
const s = z.object({ x: z.unknown() })
type s = z.infer<typeof s>
// s has { x?: unknown } but should be { x: unknown }
```

colinhacks:
I'd say this is actually behaving as intended. Currently Zod can't really differentiate between `{arg: string | undefined}` and `{arg?: string | undefined}`. When you pass a key schema `S` s.t. `undefined extends infer<typeof S>`, Zod will add the question mark. In this case `undefined` extends `unknown` (because everything does), hence the question mark. This also agrees with runtime behavior:

```ts
const schema = z.object({ x: z.unknown() });
schema.parse({}); // works
```

jasonkuhrt:
How does one write `anything but something must be passed in and it cannot be undefined`?

colinhacks:
> How does one write `anything but something must be passed in and it cannot be undefined`?

This is the best I've got:

```ts
const schema = z.object({
    field: z.custom((x) => x !== undefined),
  });
```

colinhacks:
> It would be really nice if we could define "any object" as a property (and it is a required property when it does schema validation).

Cool idea! Open to this as a PR. Just setting the "default shape" to `{}` should work, though you'll end up with an empty object every time unless you use `.passthrough()`.

```ts
z.object().passthrough()
```

ranmocy:
If you guys like me need to make sure the return type is compatible with existing interfaces, then you could do something like this:

```typescript
object({
  field: custom<MyType>((val) => val !== undefined && val !== null),
})
// Or
object({
  field: object({}).passthrough().refine<MyType>((val): val is MyType => val !== undefined && val !== null),
})
```

lostpebble:
> Cool idea! Open to this as a PR. Just setting the "default shape" to `{}` should work, though you'll end up with an empty object every time unless you use `.passthrough()`.

I've just added a PR which pretty much does exactly that (makes the shape an empty object and set the object to `"passthrough"`. Its added as the type `z.anyObject()`.

igalklebanov:
> I've just added a PR which pretty much does exactly that (makes the shape an empty object and set the object to "passthrough". Its added as the type z.anyObject().

Isn't that the purpose of [`z.record`](https://github.com/colinhacks/zod#records)?

Possible today:

```ts
z.record(z.unknown()) // don't care about key type.
```

Proposal:

```ts
[...]

PR #1675, opened 2022-12-12
feat: Add the anyObject type

lostpebble:
Fixes https://github.com/colinhacks/zod/issues/1628 .

This allows the use of `z.anyObject()` for validation situations where a property is required to be an object but we don't care to validate its exact shape / properties. Other types are easy to define as indiscriminate versions of themselves, except for objects. And using  `z.any()` causes the property to be optional (as noticed in https://github.com/colinhacks/zod/issues/1628).

### DIFF
--- a/deno/lib/__tests__/object.test.ts
+++ b/deno/lib/__tests__/object.test.ts
@@ -4,6 +4,7 @@ const test = Deno.test;
 
 import { util } from "../helpers/util.ts";
 import * as z from "../index.ts";
+import { SafeParseSuccess } from "../index.ts";
 
 const Test = z.object({
   f1: z.number(),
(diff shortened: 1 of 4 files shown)


niba:
Hey guys, what is the status of this one? We started using zod in our company and this is a big issue for us.

JacobWeisenburger:
I'm not sure that we need `z.anyObject` because we have `z.record(z.any())`. It seems to work the way that you have described. Is there anything I'm missing?

```ts
const anyObjectSchema = z.record( z.any() )
type AnyObject = z.infer<typeof anyObjectSchema>
// type AnyObject = {
//     [ x: string ]: any
// }

console.log( anyObjectSchema.safeParse( { a: 1, b: 2 } ).success ) // true
console.log( anyObjectSchema.safeParse( {} ).success ) // true
console.log( anyObjectSchema.safeParse( [] ).success ) // false
console.log( anyObjectSchema.safeParse( 'foo' ).success ) // false
console.log( anyObjectSchema.safeParse( 42 ).success ) // false
console.log( anyObjectSchema.safeParse( null ).success ) // false
console.log( anyObjectSchema.safeParse( undefined ).success ) // false
```

lostpebble:
This PR was made in response to this comment: https://github.com/colinhacks/zod/issues/1628#issuecomment-1338138675

Also, the inferred type from `z.record( z.any())` is `{ [ x: string ]: any }`. This is not exactly the TypeScript `object` type (which I've made `z.anyObject()` output to), although they would generally be interchangeable.

I also think that since we have `z.object()` - having a similarly named `z.anyObject()` makes more sense than having to find and understand the ability of `z.record( z.any())` to do the same. I've never made use of the `z.record()` type, and while it does work for this situation it feels like a bit of a hack rather than a natural type definition for this situation.

jasonkuhrt:
There are situations when a type error arises because a type index is expected but not present. I cannot off hand reproduce the situation though. So I'm not sure to what extent @JacobWeisenburger's point stands--about the need or not for a differentiation. I would expect however that a differentiation is needed under advanced cases, type logic, multiple times derived types, etc.

JacobWeisenburger:
Either of these seem to do what you are asking for. So I'm still not really understanding what `z.anyObject()` is supposed to do that other schemas don't.

```ts
const emptyObjSchema = z.object( {} )
console.log( emptyObjSchema.safeParse( {} ).success ) // true
console.log( emptyObjSchema.safeParse( { foo: 'foo' } ).success ) // true
console.log( emptyObjSchema.safeParse( [] ).success ) // false
console.log( emptyObjSchema.safeParse( () => { } ).success ) // false
console.log( emptyObjSchema.safeParse( 'foo' ).success ) // false
console.log( emptyObjSchema.safeParse( 42 ).success ) // false
console.log( emptyObjSchema.safeParse( null ).success ) // false
console.log( emptyObjSchema.safeParse( undefined ).success ) // false

const recordSchema = z.record( z.any() )
console.log( recordSchema.safeParse( {} ).success ) // true
console.log( recordSchema.safeParse( { foo: 'foo' } ).success ) // true
console.log( recordSchema.safeParse( [] ).success ) // false
console.log( recordSchema.safeParse( () => { } ).success ) // false
console.log( recordSchema.safeParse( 'foo' ).success ) // false
console.log( recordSchema.safeParse( 42 ).success ) // false
console.log( recordSchema.safeParse( null ).success ) // false
console.log( recordSchema.safeParse( undefined ).success ) // false
```

lostpebble:
> Either of these seem to do what you are asking for. So I'm still not really understanding what `z.anyObject()` is supposed to do that other schemas don't.

Its about the returned type when inferring the type from zod. We still want the type to be `object` instead of `any` or `Record<string, any>` or whatnot. Its not an issue of validation per se- its about how we use zod and TypeScript together to define our interfaces.

PR #1678, opened 2022-12-12
Issue 1677 Fixing ZodString::isDatetime.

### LINKED ISSUE #1677: 3.20: ZodString::isDatetime — Cannot read properties of undefined
Reproduction

```ts
const schema = z.string().datetime();
const { isDatetime } = schema;
isDatetime(); // TypeError: Cannot read properties of undefined (reading '_def')
```

does not affect `isEmail`.

Most likely reason is that `isDatetime` described as method, not getter prop like `isEmail`.
It's unbound.

```ts
// source
    isDatetime() {
        return !!this._def.checks.find((ch) => ch.kind === "datetime");
    }
    get isEmail() {
        return !!this._def.checks.find((ch) => ch.kind === "email");
    }
```

RobinTail:
Closes #1677

Making `isDatetime` consistent with other getters of `ZodString`, fixing its binding.

### DIFF
--- a/deno/lib/__tests__/string.test.ts
+++ b/deno/lib/__tests__/string.test.ts
@@ -192,16 +192,19 @@ test("trim", () => {
 
 test("datetime", () => {
   const a = z.string().datetime({});
-  expect(a.isDatetime()).toEqual(true);
+  expect(a.isDatetime).toEqual(true);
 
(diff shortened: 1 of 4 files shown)


RobinTail:
@colinhacks && @samchungy , please review

colinhacks:
Good catch! 😅

PR #1680, opened 2022-12-12
add more test cases for `z.coerce`.

igalklebanov:
Hey 👋 

I always forget these js weird spots, and wanted to gain more confidence with the new method.

Btw, some cases throw native errors and not `z.ZodError`s, should we catch and handle it the "zod way"?


colinhacks:
Thanks! Love this!

PR #1688, opened 2022-12-13
feat: add index signature inference when using `passthrough` on ZodObjects

santosmarco-caribou:
# Overview

As we know, the `passthrough` option allows any key to, as the name suggests, pass through the parser.

This PR adds a `{ [k: string] any }` index signature on the Zod Objects marked as `passthrough`. (because that's what happens in reality, right?)

If you declare a `catchall`, though, the `catchall` has priority over it (as says the docs and the code) so the inference caused by the `passthrough` option goes away.

Screenshot 2022-12-13 at 2 26 50 PM


deno/lib/README.md
```diff
@@ -361,6 +361,7 @@ There are a growing number of tools that are built atop or support Zod natively!
 - [`zodix`](https://github.com/rileytomasek/zodix): Zod utilities for FormData and URLSearchParams in Remix loaders and actions.
 - [`formik-validator-zod`](https://github.com/glazy/formik-validator-zod): Formik-compliant validator library that simplifies using Zod with Formik.
 - [`zod-i18n-map`](https://github.com/aiji42/zod-i18n): Useful for translating Zod error messages.
+- [`@modular-forms/solid`](https://github.com/fabian-hiller/modular-forms): Modular form library for SolidJS that supports Zod for validation.
```
santosmarco-caribou, line 364: I have no idea why this keeps getting added here automatically.

PR #1699, opened 2022-12-15
docs: fix typo

zetaraku:
The original snippet created `stringSchema1` and `stringSchema2` variables but forgot to use the correct names after that.

This PR first fixes the names and realized that the example can be fixed by only using one `stringSchema` variable.


colinhacks:
Thanks!

PR #1700, opened 2022-12-15
fix installation links in table of contents

DetachHead:
i suggest removing the table of contents entirely, since it's so easy to accidentally break and github now has them built in anyway:

image


KotlinIsland:
Thanks, but we are migrating to `.rst` for the readme so this PR is kinda redundant...

PR #1711, opened 2022-12-16
add checks @ `ZodBigInt`.

### LINKED ISSUE #1592: BigInt refinements
My guess is to include most of what's available on `z.number()`

```ts
z.bigint().min(5);
z.bigint().max(5);
z.bigint().gt(5);
z.bigint().lt(5);
z.bigint().positive(); //     > 0
z.bigint().nonnegative(); //  >= 0
z.bigint().negative(); //     < 0
z.bigint().nonpositive(); //  <= 0
z.bigint().multipleOf(5);
```

igalklebanov:
Hey 👋 

This PR adds `.gt`, `.gte`, `.lt`, `.lte`, `.positive`, `.negative`, `.nonpositive`, `.nonnegative` & `.multipleOf` to `ZodBigInt`.

closes #1592.


colinhacks:
Thanks Igal! I like a lot of this but I made some changes. 

I consolidated `ZodBigIntTooSmallIssue` into `ZodTooSmallIssue` by converting `minimum: number` to `minimum: number | bigint`. I appreciate the effort to avoid breaking changes on the existing issue types, but I'd rather keep `too_big/too_small/etc` consolidated into a single issue type.

I got rid of all the `.safe()/.unsafe()` stuff, since it doesn't apply to bigints - the point of using a bigint is to not worry about being inside the safe integer range. I'd accept a PR to add `.safe()` to `ZodNumber` where it'd have more utility. But `.gigantic/.unsafe` seem unnecessary to me.

igalklebanov:
Thanks Colin! ⚡ 

> I got rid of all the .safe()/.unsafe() stuff, since it doesn't apply to bigints - the point of using a bigint is to not worry about being inside the safe integer range. I'd accept a PR to add .safe() to ZodNumber where it'd have more utility. But .gigantic/.unsafe seem unnecessary to me.

The use case I was thinking about was a 3rd party library, e.g. a database driver, returning `bigint` values, but zod consumer knows the underlying value should be in the safe integer range (`select count(*) as friends from people left join friendships group by people.id` - friends column might return as bigint, but although there's lots of people, most people don't have that many friends). So a helper such as `ZodBigInt.safe()` allows to describe this explictly and chain with `.transform(Number)` knowing the parsed underlying value is unchanged.

I'll add a separate PR for `ZodNumber.safe()` as js behavior with anything beyond `Number.MAX_SAFE_INTEGER` behaves weirdly.

colinhacks:
> The use case I was thinking about was a 3rd party library, e.g. a database driver, returning `bigint` values, but zod consumer knows the underlying value should be in the safe integer range (`select count(*) as friends from people left join friendships group by people.id` - friends column might return as bigint, but although there's lots of people, most people don't have that many friends). So a helper such as `ZodBigInt.safe()` allows to describe this explictly and chain with `.transform(Number)` knowing the parsed underlying value is unchanged.

This makes a lot of sense, sorry.

Issue #1693, opened 2022-12-14
Is it normal that a Tuple is valid if a value is empty ? 


nmeuni:
[Zod v.3.20.2]

Hello, I'm a in use case where I can have tuple like that: `[<1 empty>, "toto"]`
In this case the safeParse method return a success with `["toto"]` in `result.data`. Is this the expected behavior ?
I found no "simple" way to make the parser return an error. (Maybe by using .refine or by creating a custom rule)

Code to reproduce in RunKit
```ts
  var z = require("zod")
  
  const test = []
  
  test[1] = "toto"
  
  const ZodTupleTest = z.tuple([z.number(), z.string()]);
  
  ZodTupleTest.safeParse(test)
```

That said, it's a very usefull lib ;)

JacobWeisenburger:
It looks like there was a merged PR fixing this. Can I close this issue?

PR #1712, opened 2022-12-16
fix: [#1693] Tuple with empty items

metuan:
`map()` as referenced in [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) is is invoked only for array indexes which have assigned values. It is not invoked for empty slots in sparse arrays. 

In some cases, these slots behave as `undefined` (that would be caught by existing implementation) but in others (e.g. array iteration methods), empty slots are skipped. More info [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#array_methods_and_empty_slots).

Destructing will result in usage of property enumeration instead of iterator which should fix this issue.  

Screenshot for reference.
Screenshot 2022-12-17 at 00 19 40

Fixes both: `ZodTuple` and `ZodArray`. 
Reported in #1693


colinhacks:
Gotta love JavaScript. Thanks!

PR #1714, opened 2022-12-17
add `ZodNumber.isFinite`, make `ZodNumber.isInt` true if `.multipleOf(int)`.

igalklebanov:
Hey 👋 

- Forgot to add `.isFinite` @ #1546.

- A multiple of an integer is an integer, so let's make `.isInt` truthy in that case. This might allow people to ditch `.int()` if all they need is `.multipleOf(int)` but had something that uses `isInt`.


colinhacks:
I dig it. Thanks!

PR #1719, opened 2022-12-18
Improve accuracy of the `isAsync` type guard

aaronccasanova:
This PR updates the `isAsync` type guard to check the `typeof Promise` result against a string literal instead of the `undefined` data type. This change should improve the accuracy and reliability of the function when determining if a value is a Promise.


santosmarco-caribou:
LGTM. Thanks for fixing this

PR #1733, opened 2022-12-21
fix: Fix `ZodCatch`

santosmarco-caribou:
This PR fixes `ZodCatch`.

It looks like `ZodCatch` was a product of a quick copy-paste from `ZodDefault`, which introduced some bugs to it.

1. The `output` of a `ZodCatch` was previously defined as `util.noUndefined<T["_output"]>` (from `ZodDefault`) when it should actually be `T["_output"] | <<catch value>>`.
2. The `input` of a `ZodCatch` was previously defined as `T["_input"] | undefined` (also from `ZodDefault`) when it should actually be the `T["_input"]` unchanged.
3. Renamed props and methods from `default` to `catch`.


colinhacks:
I pushed a commit to fix the tests, since those weren't updated. I also got rid of the second generic on `ZodCatch` - the value passed into `.catch()` should match `T['output']` exactly and there's no need for additional inference there. Good PR!

popoleeMaster:
Which release is this planned to be put into ?

colinhacks:
Landed in `zod@3.20.3`

PR #1736, opened 2022-12-22
feat: Add extra coercions

santosmarco-caribou:
This PR adds the following extra coercions:

1. `Set` -> `Array` in `ZodArray`
2. `Set` -> `Array` in `ZodTuple`
3. `Array` -> `Set` in `ZodSet`
4. `Object` -> `Map` in `ZodMap`
5. `Map` -> `Object` in `ZodObject`


colinhacks:
This is interesting but not worth the squeeze, in my opinion. We need to be conservative about what sorts of features make it into core as Zod gets larger, and this use case feels very niche and edge casey. It opens up Zod to a bunch of feature requests of the form "Why isn't there a coercion mode from X to Y?". Keeping `z.coerce` as primitive-only is the best way to draw a line in the sand.

santosmarco-caribou:
> This is interesting but not worth the squeeze, in my opinion. We need to be conservative about what sorts of features make it into core as Zod gets larger, and this use case feels very niche and edge casey. It opens up Zod to a bunch of feature requests of the form "Why isn't there a coercion mode from X to Y?". Keeping `z.coerce` as primitive-only is the best way to draw a line in the sand.

Ok, I get what you are saying and agree with the decision.

Closing PR.

PR #1738, opened 2022-12-22
feat(#1690): Add `ZodRequired`

### LINKED ISSUE #1690: required should remove defaults
At the moment required only removes optionality rather than defaults.
This leads to somewhat nonintuitive behavior:

```ts
const t = z.object({ foo: z.string().default("") });  // foo is optional in input
const t2 = z.required({ foo: true }); // foo is still optional in input
t2.parse({}); // is ok but shouldn't
```

santosmarco-caribou:
## Closes #1690 

This PR adds the `ZodRequired` type, which removes any `undefined` from both the schema's output _and_ input.

> ~Note: I couldn't add `.required()` to the root `ZodType` class because it conflicts with the `.required()` method of `ZodObject`. Suggestions are welcome. I feel like it should live in the root `ZodType` class...~

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -361,6 +361,7 @@ There are a growing number of tools that are built atop or support Zod natively!
 - [`zodix`](https://github.com/rileytomasek/zodix): Zod utilities for FormData and URLSearchParams in Remix loaders and actions.
 - [`formik-validator-zod`](https://github.com/glazy/formik-validator-zod): Formik-compliant validator library that simplifies using Zod with Formik.
 - [`zod-i18n-map`](https://github.com/aiji42/zod-i18n): Useful for translating Zod error messages.
+- [`@modular-forms/solid`](https://github.com/fabian-hiller/modular-forms): Modular form library for SolidJS that supports Zod for validation.
 
 #### Zod to X
 
--- a/deno/lib/ZodError.ts
+++ b/deno/lib/ZodError.ts
@@ -16,6 +16,7 @@ export type typeToFlattenedError<T, U = string> = {
 };
 
 export const ZodIssueCode = util.arrayToEnum([
+  "required",
   "invalid_type",
   "invalid_literal",
   "custom",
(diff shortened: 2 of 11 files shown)


maxArturo:
@santosmarco-caribou looks good at a glance! 

I haven't taken an in depth look - but I wonder if, for the sake of keeping `.required()` on the root `ZodType`, it would be possible to bring in the specific `ZodObject` funtionality at the top `ZodType` so that there's no confilct in the namespace? or via a `_required()` on the ZodObject with a `instanceof` check at the `ZodType` level? Just some ideas.

santosmarco-caribou:
> @santosmarco-caribou looks good at a glance!
> 
> I haven't taken an in depth look - but I wonder if, for the sake of keeping `.required()` on the root `ZodType`, it would be possible to bring in the specific `ZodObject` funtionality at the top `ZodType` so that there's no confilct in the namespace? or via a `_required()` on the ZodObject with a `instanceof` check at the `ZodType` level? Just some ideas.

I'm afraid it isn't possible to bring the `ZodObject` logic up to the root `ZodType` class since the return type depends on the `schema`, `unknownKeys`, and `catchall`, which `ZodType` does not have access to.

santosmarco-caribou:
I came up with a solution using overloads. Don't know if that's the best way to handle this, but well, it works, passes all tests, and does not create any conflict.

Screenshot 2022-12-22 at 11 22 06 AM

maxArturo:
Very neat! Good job @santosmarco . To be sure, it would be a good idea to add tests that flex the interaction of `ZodObject` with `.require()` specifically, I think we're missing those.

colinhacks:
On further reflection, I think Zod should possibly consider generalizing this to `ZodExclude`. Though I'm still on the fence about whether this is worthwhile. 

Note that the types in this PR are technically wrong. For instance, if `.required()` behaved as implemented in this PR, the following schema would have `Output` type `string`, when it should be `string | undefined`.

```ts
z.string().transform(val => Math.random() > 0.5 ? val : undefined).required()
```

There's some inherent ambiguity here, which is why I've been hesistant to merge this. The open question is this: does the "required" check happen before or after the `innerType` parsing? I think different people will have different expectations about how this should work, unfortunately.

colinhacks:
I'm going to close this for now, but I'll keep this in mind over the development cycle for Zod 4. I need to do some restructuring and see if a more natural API emerges for making object schemas required (which is the fundamental issue solved by this PR). I also need to think more about how this will interact with Zod 4's new approach to `exactOptionalPropertyTypes`. I'll re-open if I decide to go ahead with `ZodRequired`. Thanks for the great work Marco 🙌

Issue #1508, opened 2022-10-20
what's the best way to deep merge objects ?


err0r500:
Hi,
Here's an example of the `.merge` behaviour : 

```
const childA = z.object({
  childA: z.string(),
});

const parentA = z.object({
  a: z.string(),
  child: childA,
});

const childB = z.object({
  childB: z.string(),
});

const parentB = z.object({
  b: z.string(),
  child: childB,
});

const notWorking = parentA.merge(parentB);
type NotWorking = z.infer<typeof notWorking>;
//type NotWorking = {
//  a: string;
//  child: {
//      childB: string;
//  };
//  b: string;
//}

const working = parentA
  .merge(parentB)
  .merge(z.object({ child: parentA.shape.child.merge(parentB.shape.child) }));
type Working = z.infer<typeof working>;
//type Working = {
//  a: string;
//  child: {
//      childA: string;
//      childB: string;
//  };
//  b: string;
//}
```

the working / not working is a poorly chosen way to express that I was expecting the `merge` function behaviour to propagate to children.

is there a recommended way to get the "working" one easily ? (maybe I didn't find the function in the doc)

temoncher:
I think I got a similar issue with array intersections
```ts
import { z } from 'zod';

const arrayWithIds = z
    .object({ id: z.number() })
    .array();

const arrayWithNames = z
    .object({ name: z.string() })
    .array();

arrayWithIds
    .and(arrayWithNames)
    .parse([])
    // @ts-expect-error Can only access `id` here, `name` is not present on `el`
    .forEach((el) => el.name);
z.intersection(arrayWithIds, arrayWithNames)
    .parse([])
    // @ts-expect-error The same issue with `z.intersection`
    .forEach((el) => el.name);
```
zod version: 3.19.1

santosmarco-caribou:
Hi @err0r500, thanks for your question. Here's a PR that implements `mergeDeep` in `ZodObject`

https://github.com/colinhacks/zod/pull/1739

lovetingyuan:
I need merge deeply function

ozyman42:
The PR was closed but I think it should be moved into it's own library if the maintainer doesn't want to deal with the method as I expect many others have this use-case. My use-case is implementing a utility whose main purpose is to let people deeply merge configurations and have them be typed

PR #1739, opened 2022-12-22
feat(#1508): Add mergeDeep to ZodObject

santosmarco-caribou:
This PR adds `mergeDeep` to `ZodObject`. It's now possible to merge two `ZodObjects` deeply.

The method follows the same rules as `merge`—keys of `B` have priority over those from `A`, and the `B`'s `unknownKeys` and `catchall` schema are preserved.

```ts
const A = z.object({
  a: z.string(),
  b: z.object({
    c: z.string(),
    d: z.string(),
    e: z.object({
      f: z.string(),
      g: z.string(),
      h: z.string(),
    }),
  }),
});

const B = z.object({
  b: z.object({
    i: z.string(), // should add this prop deeply inside the shape's `b` key
    j: z.string(), // same
    e: z.object({ // does not completely overwrite this key; instead add properties to it.
      k: z.string(), // will be added
      l: z.string(), // will be added
      h: z.bigint(),  // Overwrite from string to bigint
    }),
  }),
});

const mergedDeep = A.mergeDeep(B);
```

Screenshot 2022-12-22 at 5 35 10 AM

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -361,6 +361,7 @@ There are a growing number of tools that are built atop or support Zod natively!
 - [`zodix`](https://github.com/rileytomasek/zodix): Zod utilities for FormData and URLSearchParams in Remix loaders and actions.
 - [`formik-validator-zod`](https://github.com/glazy/formik-validator-zod): Formik-compliant validator library that simplifies using Zod with Formik.
 - [`zod-i18n-map`](https://github.com/aiji42/zod-i18n): Useful for translating Zod error messages.
+- [`@modular-forms/solid`](https://github.com/fabian-hiller/modular-forms): Modular form library for SolidJS that supports Zod for validation.
 
 #### Zod to X
 
--- a/deno/lib/__tests__/object.test.ts
+++ b/deno/lib/__tests__/object.test.ts
@@ -389,3 +389,73 @@ test("unknownkeys merging", () => {
   util.assertEqual<mergedSchema["_def"]["catchall"], z.ZodString>(true);
   expect(mergedSchema._def.catchall instanceof z.ZodString).toEqual(true);
 });
+
+const A = z.object({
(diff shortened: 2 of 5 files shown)


colinhacks:
Just because it can be done doesn't mean it should. The logic in `MergeZodObjectsDeep` is too complicated and limited. I regret adding `.deepPartial` because it remains incomplete and confusing to people who think it will "just work" when you have something like this:

```ts
z.object({
  outer: z.object({
    inner: z.string()
  }).optional()
})
```

The logic (both in `.deepPartial` and in your PR` isn't smart enough to properly handle the nested `ZodOptional`. In fact to be fully complete you'd need to account for every possible subclass of `ZodType` which isn't possible because Zod is designed to be extensible! People can subclass `ZodType` if they want and create new schema types. 

Anyway I'm very allergic to things like this since `deepPartial` and `discriminatedUnion` have been a huge burden and a point of confusion for a long time.

santosmarco-caribou:
> Just because it can be done doesn't mean it should. The logic in `MergeZodObjectsDeep` is too complicated and limited. I regret adding `.deepPartial` because it remains incomplete and confusing to people who think it will "just work" when you have something like this:
> 
> ```ts
> z.object({
>   outer: z.object({
>     inner: z.string()
>   }).optional()
> })
> ```
> 
> The logic (both in `.deepPartial` and in your PR`isn't smart enough to properly handle the nested`ZodOptional`. In fact to be fully complete you'd need to account for every possible subclass of `ZodType`which isn't possible because Zod is designed to be extensible! People can subclass`ZodType` if they want and create new schema types.
> 
> Anyway I'm very allergic to things like this since `deepPartial` and `discriminatedUnion` have been a huge burden and a point of confusion for a long time.

Closing PR and _no more deep things_ :( But I your concern is valid.

PR #1742, opened 2022-12-22
docs: Instruct users to return z.NEVER in .superRefine() when providing a type predicate

### LINKED ISSUE #1704: [Suggestion] Instruct users to return `z.NEVER` in `.superRefine()` when providing a type predicate
As the validation result **solely depends on whether `ctx.addIssue()` is called**,

https://github.com/colinhacks/zod/blob/ad4385439d7c0d7a584052dee5aa578fde5a6efd/README.md?plain=1#L1891

and the return value is only used to satisfy the type predicate and is **always thrown away**,

https://github.com/colinhacks/zod/blob/c9e4ed4095d77f5b43a5ba11d8f76b56ea48e5c6/src/types.ts#L3821-L3822

to prevent users from **accidentally forgetting to call `ctx.addIssue()` while returning a boolean value** as described in the example,

https://github.com/colinhacks/zod/blob/ad4385439d7c0d7a584052dee5aa578fde5a6efd/README.md?plain=1#L1931-L1945

instructing users to **return `z.NEVER` instead of a boolean** at the end (indicating the value is ***never*** used) just like the **early return** case when providing a type predicate may be a good idea.

```ts
  .superRefine((arg, ctx): arg is { first: string; second: number } => {
    if (!arg) {
      ctx.addIssue({
        code: z.ZodIssueCode.custom, // customize your issue
        message: "object should exist",
      });
    }

    return z.NEVER;  // this should satisfy the typing
  })
```

zetaraku:
This resolves #1704.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1919,7 +1919,7 @@ const schema = z.number().superRefine((val, ctx) => {
 
 #### Type refinements
 
-If you provide a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) to `.refine()` or `superRefine()`, the resulting type will be narrowed down to your predicate's type. This is useful if you are mixing multiple chained refinements and transformations:
+If you provide a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) to `.refine()` or `.superRefine()`, the resulting type will be narrowed down to your predicate's type. This is useful if you are mixing multiple chained refinements and transformations:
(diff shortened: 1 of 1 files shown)


maxArturo:
Thanks @zetaraku . Feel free to update the tests to demonstrate this best practice.

zetaraku:
I found that there is another `README.md` in `deno/lib`.
Why are there two `README.md`? I'm afraid it's out of sync.

PR #1749, opened 2022-12-23
Fix datetime offset without comma

ooga:
An ISO datetime with an offset without comma should be valid

https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators


colinhacks:
Looks good!

Issue #1747, opened 2022-12-23
String `.endsWith` & `.startsWith` could perhaps be typed more strongly


alii:
If the passed value was a literal, we could have a generic in ZodString that will include the prefix.

Use case:
At the moment, I am using the code below to validate that a value is a specific ID format that has a constant prefix

```ts
export function id<P extends IdPrefixes>(prefix: P) {
	return z
		.string()
		.refine((value): value is `${P}_${string}` => pika.validate(value, prefix));
}
```

Imagine if this could be done with just `.startsWith` and `.endsWith` — ZodString could then have a single generic of the output, which would default to just a string and can be made more specific with these two methods.

<img width="785" alt="CleanShot 2022-12-23 at 14 02 47@2x" src="https://user-images.githubusercontent.com/25351731/209348315-61d308e4-6581-4210-91ed-b02398adb85c.png">

santosmarco-caribou:
I've opened a PR for that. Let's see what people think.

https://github.com/colinhacks/zod/pull/1751

colinhacks:
I have some reservations discussed here: Thanks for the PR. I have some reservations discussed here: https://github.com/colinhacks/zod/pull/1730#issuecomment-1364490384

JacobWeisenburger:
I really like the idea of better static types for `.startsWith` and `.startsWith`. But I think @colinhacks is right, it would be quite hard to do this in a backwards compatible way.

@alii or @santosmarco-caribou, do you have some ideas on how to address backwards compatibility?

santosmarco-caribou:
One way would be to add an overload to these methods (almost like I did on my PR for `ZodRequired`).

We could leave the current implementation at the top of the overloads so the users will not feel any difference. This type of "exact" inference would be turned off by default.

Then, as a second overload, we add a definition that forces the user to pass `true` in one of the properties of the `options` object, turning the inference on.

Something like this:

```ts
startsWith(prefix: string, options?: { message?: string; exact?: never }): ZodString
startsWith<Start extends string>(prefix: Start, options: { message?: string; exact: true }): ZodString<Start, End>
startsWith(prefix: string, options?: { message?: string; exact?: true }): ZodString {
  // no change in implementation, the `exact` is solely for turning the inference on.
}
```

JacobWeisenburger:
I like it.

Any objections/reservations @colinhacks?

tachibanayui:
[...]

PR #1751, opened 2022-12-24
feat(#1747): Add `Start` and `End` to ZodString

santosmarco-caribou:
This PR adds two type parameters to `ZodString`: `Start` and `End`. They are populated by the `.startsWith()` and `.endsWith()` methods to improve inference.

Screenshot 2022-12-24 at 2 39 56 AM

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -557,8 +557,11 @@ const datetimeRegex = (args: { precision: number | null; offset: boolean }) => {
   }
 };
 
-export class ZodString extends ZodType<string, ZodStringDef> {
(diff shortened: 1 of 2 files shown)


colinhacks:
Thanks for the PR. I have some reservations discussed here: https://github.com/colinhacks/zod/pull/1730#issuecomment-1364490384

colinhacks:
Interesting idea, but I'd like to avoid adding type parameters onto the primitive schema types like `ZodString` to keep the types simpler in Intellisense. I generally think that's more important here, and this use case will be covered by a future template literal API anyway. But thanks!

spiftire:
I would love to have this feature. Any plans to add this?

Issue #1682, opened 2022-12-12
A preprocess within a schema breaks relationship between TypeOf<T> and ZodType<T>


ConnorSinnott:
Given a generic T, I'm trying to type the result of a function as the schema which can create T. This breaks however if the schema has a preprocessed field. I've broken the issue down to:

```Typescript
const noPreprocessObjSchema = z.object({
  date: z.date(),
});
type NoPreprocessObj = z.infer<typeof noPreprocessObjSchema>;

const preprocessObjSchema = z.object({
  date: z.preprocess(
    (arg) => (typeof arg === 'string' ? new Date(arg) : arg),
    z.date(),
  ),
});
type PreprocessObj = z.infer<typeof preprocessObjSchema>;

const noPreprocessParser: z.ZodType<NoPreprocessObj> = noPreprocessObjSchema;
const preprocessParser: z.ZodType<PreprocessObj> = preprocessObjSchema; // Error: The types of '_input.date' are incompatible between these types. Type 'unknown' is not assignable to type 'Date'
```

Is the recommend solution that I type the prepocessor as `ZodType<Date, ZodTypeDef, Date>`?

ConnorSinnott:
Wanted to bump this since I see its getting a little attention but no answers. [I see this change was implemented rather specifically on September 11th](https://github.com/colinhacks/zod/commit/2893e1659875944198aff28a3ba404858f834ed8). Normally for a primitive parser like `string()` the zod type would have both its input and output marked as `string` correct? 
```
ZodString extends ZodType<string, ZodStringDef>
...
ZodType<Output = any, Def extends ZodTypeDef = ZodTypeDef, Input = Output>
```
But even with Input typed as string, the `.parse` and `.safeParse` methods both accept `unknown` as the parse argument. So I'm not sure what value having `.preprocesses` input as `unknown` brings. Conceptually, yeah anything could go into a preprocessor, but the same could be said for the primitive parsers. I'd expect them all to fail if the value can't be parsed.

santosmarco-caribou:
@ConnorSinnott This is _partially_ correct. Read until the end because there's a PR to fix this.

When you create a schema with `preprocess`, there's no way of knowing which input you will receive on runtime. I mean,

```ts
  date: z.preprocess(
    (arg) => (typeof arg === 'string' ? new Date(arg) : arg),
    z.date(),
  ),
```

[...]

PR #1752, opened 2022-12-24
fix(#1682): Fix `preprocess`

santosmarco-caribou:
This PR fixes `preprocess`.

Previously, it was setting the schema's `Input` to `unknown`, even in cases where we were certain of the input type.

### Before the fix

Screenshot 2022-12-24 at 3 09 06 AM

Even though I'm coercing the input to a `Date` object, `preprocess` is unaware of that and is typing the `Input` of my schema as `unknown`.

## After the fix

Screenshot 2022-12-24 at 3 11 14 AM

Now, `preprocess` knows that the input will always be a `Date`, so it correctly sets the `Input` type of my schema to `Date`, _**and**_ also enforces that I pass a schema that is able to receive `Date` as input.

For example, the following throws a TS error:

Screenshot 2022-12-24 at 3 12 54 AM
Screenshot 2022-12-24 at 3 13 09 AM

Of course, because `z.string()` can't receive a `Date` object—it would certainly fail to parse.

Now, if I explicitly type the return of the preprocessor function, I can bypass this enforcement (this is not recommended, though):

Screenshot 2022-12-24 at 3 15 11 AM

### DIFF
--- a/deno/lib/__tests__/transformer.test.ts
+++ b/deno/lib/__tests__/transformer.test.ts
@@ -202,7 +202,7 @@ test("preprocess", () => {
 
   const value = schema.parse("asdf");
   expect(value).toEqual(["asdf"]);
-  util.assertEqual<typeof schema["_input"], unknown>(true);
+  util.assertEqual<typeof schema["_input"], unknown[]>(true);
 });
 
 test("async preprocess", async () => {
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -3889,7 +3889,7 @@ export class ZodEffects<
   };
 
   static createWithPreprocess = <I, S extends ZodType<any, ZodTypeDef, I>>(
-    preprocess: (arg: unknown) => I,
+    preprocess: (arg: unknown) => I | Promise<I>,
     schema: S,
     params?: RawCreateParams
   ): ZodEffects<S, S["_output"], I> => {
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -3888,11 +3888,11 @@ export class ZodEffects<
     });
   };
 
-  static createWithPreprocess = <I extends ZodTypeAny>(
-    preprocess: (arg: unknown) => unknown,
-    schema: I,
+  static createWithPreprocess = <I, S extends ZodType<any, ZodTypeDef, I>>(
+    preprocess: (arg: unknown) => I,
+    schema: S,
     params?: RawCreateParams
-  ): ZodEffects<I, I["_output"], unknown> => {
+  ): ZodEffects<S, S["_output"], I> => {
     return new ZodEffects({
       schema,
       effect: { type: "preprocess", transform: preprocess },
--- a/src/__tests__/transformer.test.ts
+++ b/src/__tests__/transformer.test.ts
@@ -201,7 +201,7 @@ test("preprocess", () => {
 
   const value = schema.parse("asdf");
   expect(value).toEqual(["asdf"]);
-  util.assertEqual<typeof schema["_input"], unknown>(true);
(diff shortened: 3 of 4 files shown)


colinhacks:
This is behaving as intended. The `Input` of a schema defines what data types can be passed into the schema and be parsed successfully. In other words, the `Input` should line up with the type of `arg` in your example. Then `arg` is coerced to a `Date` *during the parsing step of your `ZodPreprocess` instance* then passed into the inner schema (`z.date()`).

santosmarco-caribou:
> This is behaving as intended. The `Input` of a schema defines what data types can be passed into the schema and be parsed successfully. In other words, the `Input` should line up with the type of `arg` in your example. Then `arg` is coerced to a `Date` *during the parsing step of your `ZodPreprocess` instance* then passed into the inner schema (`z.date()`).
> 
> 

@colinhacks I understand but I don't know if I agree. I'm not too familiarized with ZodEffects but I feel like the `Input` of the preprocessed schema shouldn't line up with the type of `arg`. I mean, why?

In my opinion, it doesn't matter when or where `arg` gets coerced. It will always be coerced _**before**_ being passed to the schema, i.e., the input will always be coerced. And if the input is always coerced, why don't type the `Input` accordingly?

Can you provide an example of a preprocessed schema where `arg` gets coerced to `Date` but receives a different type of `Input`?

JacobWeisenburger:
@santosmarco-caribou, you make a compelling argument. I really like your ideas. You got my vote.

---

> Can you provide an example where `arg` gets coerced to `Date` but even after that, it still would make sense to pass any other kind of schema different than `ZodType<any, any, Date>` (e.g., `ZodDate`)?

I can't think of any situation where you would want to use something other than ZodDate for this.

ConnorSinnott:
> The Input of a schema defines what data types can be passed into the schema and be parsed successfully.

While I have [raised an issue related to this pr](https://github.com/colinhacks/zod/pull/1752), I do understand this viewpoint and would fully concede the solution if Zod typing did a little more lifting to determine the desired input type. However, if the goal is to ensure "The Input of a schema defines what data types can be passed into the schema and be parsed successfully," than I'd argue that typing the input as unknown moves us further away. Truly the only type we know for sure can be passed into the schema and parsed successfully, is Date. 

After all is said and done, I _expect_ a Date to be passed into this object. Sure I might do a little lifting to massage some bad data in, but the expected input is still Date. If I expected a collection of different inputs, a union might be a better fit... Perhaps on that note, preprocessor might benefit from having the potential inputs explicitly defined i.e.
`z.preprocess([z.string(), z.date()], (arg: string | Date) => ..., z.date())`. This differs from `z.string().or(z.date()).transform(arg => new Date(arg))` as we'd still be validating the result of `new Date(string)`

Also, I don't see mentioned here that the behavior in question is very new, and [was added only 4 months ago](https://github.com/colinhacks/zod/commit/2893e1659875944198aff28a3ba404858f834ed8).

colinhacks:
> It will always be coerced before being passed to the schema
The coercion happens during the parse step of ZodEffects. 

> I also expect a Date to be passed into this object
This just isn't how most people are using preprocess. The whole idea is to expand the set of allowable inputs by preprocessing them before relaying them through to the inner schema.

> a union might be a better fit
If you want the input type to be more specific than `unknown`, then a union schema + transform would definitely do the job.

This is still very cut and dry to me. 

```ts
const schema = z.preprocess(arg => arg instance of Date ? arg : new Date(), z.date())
schema.parse(/* data */)
```

The input represents the set of types that can passed into `.parse` and parsing will pass. This is what it always is across all schemas. In this scenario you can pass anything in and it will pass, so input should be `unknown`.

Issue #1743, opened 2022-12-22
`CreateParams` get lost when calling any of the root-level methods


santosmarco-caribou:
```ts
const MySchema = z.string({ description: 'this is a string schema' }).optional()
// => No longer has the description associated with it
```

This is because we are not passing the `CreateParams` to these methods on the root-level `ZodType` class.

Instead of:

```ts
nullable() {
  ZodNullable.create(this);
}
```

We should find a way to pass the `CreateParams` as well.

```ts
nullable() {
  ZodNullable.create(this, /* => */ params);
}
```

JacobWeisenburger:
Confirmed

```ts
const schema = z.string( { description: 'this is a string schema' } )
console.log( schema.description ) // this is a string schema
console.log( schema.optional().description ) // undefined
```

santosmarco-caribou:
Here's a PR to fix that:

https://github.com/colinhacks/zod/pull/1756

PR #1756, opened 2022-12-24
fix(#1743): Fix passing params in root class

santosmarco-caribou:
This PR fixes the cascading of the `createParams` to other schemas when using the helper methods provided by `ZodType`.

Previously,

```ts
const MySchema = z
  .string({ description: "this is a string schema" })
  .optional();
// => No longer has the description associated with it
```

The params would get lost if we called any of the helper methods (`.optional()`, `.nullable()` etc.).

Now:

Screenshot 2022-12-24 at 7 43 18 PM

Screenshot 2022-12-24 at 7 43 29 PM

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -47,8 +47,9 @@
   - [Sponsors](#sponsors)
   - [Ecosystem](#ecosystem)
 - [Installation](#installation)
-  - [Node/npm](#nodenpm)
-  - [Deno](#deno)
+  - [Requirements](#requirements)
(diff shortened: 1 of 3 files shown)


JacobWeisenburger:
this looks good, except can you tell me why there are changes to `deno/lib/README.md`?

JacobWeisenburger (changes requested):
I don't think this change should affect the readme, so remove those changes and then I think I can approve this.

JacobWeisenburger (approved):

PR #1766, opened 2022-12-27
add `ZodString.date()` & `ZodString.time()`.

### LINKED ISSUE #1676: [Date] Parsing of DateOnly strings
3.20 introduced `z.string().DateTime()` for cleaner string->DateTime parsing. However, it only supports strings containing both Date _and_ Time

An API I'm working with includes only the Date part (today would be `2022-12-12`). This API is written in C# using the [DateOnly](https://learn.microsoft.com/en-us/dotnet/api/system.dateonly?view=net-7.0) struct.

The built-in JavaScript `new Date(<string>)` does work with DateOnly strings, so I don't see a huge reason not to support it as well. I don't know what the API for it would be though, but the two most obvious ones that I could think of are the following:

1. An additional option in the `string().DateTime()` function: `dateOnly: boolean`
2. A separate `string().Date()` function (or `string().DateOnly()`)

(As an aside, C# also has a [TimeOnly](https://learn.microsoft.com/en-us/dotnet/api/system.timeonly?view=net-7.0) struct, but I don't really see how that would be parseable into a JavaScript primitive)

igalklebanov:
Hey 👋 

This PR resolves #1676, as suggested by @colinhacks @ [comment](https://github.com/colinhacks/zod/issues/1676#issuecomment-1347144589).

- [x] implement `ZodString.date()`.
- [x] implement `ZodString.time()`.
- [x] unit tests.
- [x] documentation.


src/types.ts
```diff
@@ -529,36 +539,43 @@ const emailRegex =
...
+
+  if (!args.withTime) {
+    return new RegExp(`${regex}$`);
+  }
+
+  regex = `${regex}\\d{2}:\\d{2}:\\d{2}`;
+
+  if (args.precision) {
+    regex = `${regex}\\.\\d{${args.precision}}`;
+  } else if (args.precision == null) {
```
maxArturo, line 566: is the `==` intended? if so, could this be just `} else {`? I think it depends if we want to implicitly allow arbitrary precision

maxArturo (approved):
hey @igalklebanov I promise I'm not following you 😆 I'm just going down the PRs to check them out! 

👍 LGTM, just a couple of comments you may want to consider!

deno/lib/README.md
```diff
@@ -541,6 +588,8 @@ z.string().startsWith(string);
 z.string().endsWith(string);
 z.string().trim(); // trim whitespace
 z.string().datetime(); // defaults to UTC, see below for options
+z.string().date(); // ISO short date format.
+z.string().time(); // time of day in 24-hour format. same options as datetime.
```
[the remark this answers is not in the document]
igalklebanov: Good idea! have added this to the existing section.

src/types.ts
igalklebanov: > is the == intended? if so, could this be just } else {?

This is intentional to support both `precision = 0` (no dot, no anything) and `precision = null` OR `precision = undefined` (the don't care arbitrary default mode). It can't be a simple `else` since it'll catch the 0 case, add a `.` and allow arbitrary precision.

I'll add a comment and a test case for the 0 case.

> I think it depends if we want to implicitly allow arbitrary precision

Currently `.datetime()` defaults to arbitrary precision so had to align.

igalklebanov:
@maxArturo Thanks again for the thorough review and great feedback! 💪

feychenie:
Hi @igalklebanov, there seems to be just a linting/prettier error preventing the merge :) Would be amazing if this could get in soon!

KennethHoff:
>as suggested by @colinhacks

😢 

(Although, I guess he did suggest it, I just created the issue about something entirely different 🤷 )

igalklebanov:
> > as suggested by @colinhacks
> 
> 😢
> 
> (Although, I guess he did suggest it, I just created the issue about the general idea )

Did I do anything wrong? 🤷

cyrusPIF:
Any updates on when this will be merged?

diego-escobedo:
Would be awesome to have this!!

softcam-claiton:
Hello, any updates on when this will be merged?

madi-tracksuit:
This would be a huge win, anything blocking or that needs assistance? Happy to dive in and help get it merged

kimsey0:
@colinhacks, can this merge? We'd also like to use it as soon as it's available. 😄

kimsey0:
@colinhacks, anything we can do to help this merge?

colinhacks:
Thanks!

deno/lib/__tests__/string.test.ts
```diff
@@ -470,6 +470,88 @@ test("datetime parsing", () => {
...
+  expect(() => date.parse("2000-01-011")).toThrow();
+  expect(() => date.parse("2000/01/01")).toThrow();
+  expect(() => date.parse("01-01-2022")).toThrow();
+  expect(() => date.parse("01/01/2022")).toThrow();
+  expect(() => date.parse("2000-01-01 00:00:00Z")).toThrow();
+  expect(() => date.parse("2020-10-14T17:42:29+00:00")).toThrow();
+  expect(() => date.parse("2020-10-14T17:42:29Z")).toThrow();
+  expect(() => date.parse("2020-10-14T17:42:29")).toThrow();
+  expect(() => date.parse("2020-10-14T17:42:29.123Z")).toThrow();
+});
```
szamanr, line 498: how about invalid dates?

```
  expect(() => date.parse("2000-01-32")).toThrow();
  expect(() => date.parse("2000-13-01")).toThrow();
  expect(() => date.parse("2001-02-29")).toThrow();
```

colinhacks:
This has landed in Zod 3.23.

PR #1781, opened 2022-12-30
Add unit test 'string coercion with min length validation'

joeydebreuk:
`undefined` parses to `undefined` for `z.coerce.string().min(1)`, I'd expect this to fail? 

So this unit test fails:

```ts
  const schema = z.coerce.string().min(1);
  expect(() => schema.parse(undefined)).toThrow();
```

### DIFF
--- a/deno/lib/__tests__/coerce.test.ts
+++ b/deno/lib/__tests__/coerce.test.ts
@@ -28,6 +28,12 @@ test("string coercion", () => {
   );
 });
 
+test("string coercion with min length validation", () => {
+  const schema = z.coerce.string().min(1);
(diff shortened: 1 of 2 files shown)


JacobWeisenburger:
Everything seems to be working as expected. When the value `undefined` is coerced to the string `'undefined'` then it has 1 or more characters and passes the `min(1)` check. Am I misunderstanding something?

```ts
console.log(
    z.coerce.string().min( 1 ).safeParse( undefined ).success
) // true
console.log(
    z.string().min( 1 ).safeParse( undefined ).success
) // false
```

joeydebreuk:
Understood. I didn't realize the result was `'undefined'` as type `string`.

PR #1786, opened 2022-12-31
Add `z.literal.template`

### LINKED ISSUE #419: [V3] Support string literals
Hello :wave: 

I recently encountered a case, where my schema returned a string, but a function accept only a TS template string

Here is an example showing the issue: https://codesandbox.io/s/snowy-river-n2ih2?file=/src/index.ts

image

Is there a way, currently or in the future, to be able to do this:

```ts
const str = z.template(`${z.number()}.${z.number()}.${z.number()}`)
type Str = z.infer<typeof str>
// => Str: `${number}.${number}.${number}`
```

igalklebanov:
Hey 👋 

Closes #419.

- [x] implement.
- [x] unit test.
- [x] document.

---

[EDIT by @colinhacks]

This PR has been updated to support a new API. The parsing remains early identical. The OP's original comment is below.

```ts
import * as z from "zod";

const url = z.literal.template([
  "https://",
  z.string(),
  ".",
  z.enum(["com", "net", "org"]),
]);

type url = z.infer<typeof url>;
// `https://${string}.com` | `https://${string}.net` | `https://${string}.org`
```
---

This PR adds `ZodTemplateLiteral` for building template literals. Should be used when a consumer wants to explictly get a template literal typescript type / union.

A [typescript template literal](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) consists of string literal types and types in interpolated positions (the `${}` slots).

Thus this new class has the following methods:

- `.interpolated(type)` accepts zod types with a string literal fitting output type (basically primitives).
- `.literal(value)` accepts primitives. A shorthand for `.interpolated(z.literal(value))`.

These methods append the argument to the accumulated template literal.
Validation is regular expression based.

API:

```ts
import { z } from 'zod'

const url = z.templateLiteral()
             .literal('https://')
             .interpolated(z.string().min(1))
             .literal('.')
             .interpolated(z.enum(['com', 'net']))
type URL = z.infer<typeof url> // `https://${string}.com` | `https://${string}.net`
```

(see unit tests for more examples).


igalklebanov:
Infer works thus far. Going to bed, will continue tomorrow. 💤

naorpeled (commented):
LGTM! 😎

naorpeled (approved):

src/__tests__/template-literal.test.ts
```diff
@@ -0,0 +1,595 @@
+// @ts-ignore TS6133
+import { expect, test } from "@jest/globals";
```
maxArturo, line 2: One other thing I ran into: number representations of exponents, e.g. `2e5` will evidently not work with `addInterpolatedPosition(z.number())`. I think this is probably okay, and its inherent with the limitation of treating everything as a string first and then applying the regex. However it _is_ a valid TS number, would it be worth it to include as a test to explicitly state that we're aware of it or TODO/FIXME?

maxArturo (changes requested):
@igalklebanov thanks! This is some awesome functionality 🎉 . I added a couple of things I found but overall this is great!

src/__tests__/template-literal.test.ts
```diff
@@ -0,0 +1,595 @@
...
+  expect(() => cuidZZZ.parse("cjld2cyuq0000t3rmniod1foy"));
+  expect(() => cuidZZZ.parse("cjld2cyuq0000t3rmniod1foyZZY"));
+  expect(() => cuidZZZ.parse("cjld2cyuq0000t3rmniod1foyZZZ1"));
+  expect(() => cuidZZZ.parse("1cjld2cyuq0000t3rmniod1foyZZZ"));
+  expect(() => datetime.parse("2022-01-01 00:00:00"));
+  expect(() => email.parse("info@example.com@"));
+  expect(() =>
+    z.templateLiteral().addInterpolatedPosition(z.string().url())
+  ).toThrow();
+  expect(() => uuid.parse("808989fd-3a6e-4af2-b607-737323a176f6Z"));
```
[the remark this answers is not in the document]
igalklebanov: Nice catch, this should throw, it has an extra character. I blame copilot. 😀

src/types.ts
```diff
@@ -4290,6 +4290,395 @@ export class ZodPipeline<
...
+    return regexString.replace(/(^\^)|(\$$)/g, "");
+  }
+
+  protected _getEscapedStringForRegex(str: unknown): string {
+    if (typeof str !== "string") {
+      str = `${str}`;
+    }
+
+    return (str as string).replace(
+      /(\(|\)|\[|\]|\{|\}|\.|\\|\/|\^|\-|\$|\?|\:|\=|\!|\*|\+|\,|\|)/g,
```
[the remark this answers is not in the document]
igalklebanov: Nice one, didn't know this is documented.

src/types.ts
```diff
@@ -4290,6 +4290,395 @@ export class ZodPipeline<
...
+  }
+
+  protected _transformZodStringPartToRegexString(part: ZodString): string {
+    let maxLength = Infinity,
+      minLength = 0,
+      endsWith = "",
+      startsWith = "";
+
+    for (const ch of part._def.checks) {
+      if (ch.kind === "cuid") {
```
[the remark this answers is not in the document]
igalklebanov: > I think here we may be not handling the trim() ZodStringCheck path.

Haven't thought about this use case. 🤔
The general rule is, any transforms are prohibited (see `ZodEffects`/`ZodPipeline` ban), let's go with that.. and see if consumers are interested in a different behavior here?

> I found it while checking if this could be rewritten as an exhaustive switch() check, which I suggest we do to find precisely these specific scenarios!

I tried to steer away from this, because it'll end up hurting every new feature. It'll make contributing and reviewing new features in other parts of the codebase harder, because now a lot of PRs will have to deal with template literal support - painful context switch.

I find that slowly adding support for new things is also safer for this feature. New things will get more focused PRs with attention to properly testing standalone cases, and mixed cases (different types/checks).

WDYT?

src/types.ts
```diff
@@ -4290,6 +4290,395 @@ export class ZodPipeline<
...
+        min = Math.max(min, ch.value);
+
+        if (ch.value >= 0) {
+          canBeNegative = false;
+
+          if (ch.value === 0 && !ch.inclusive) {
+            canBeZero = false;
+          }
+        }
+      } else if (ch.kind === "multipleOf") {
```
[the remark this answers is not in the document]
igalklebanov: I agree, I think we should ban this one.

src/__tests__/template-literal.test.ts
igalklebanov: Great catch!

By definition, everything that has typeof number at runtime, expect for NaN (since zod has special treatment for it), should be supported. Let's add support for this.

src/types.ts
```diff
@@ -4290,6 +4290,395 @@ export class ZodPipeline<
...
+    let canBeNegative = true,
+      canBePositive = true,
+      min = -Infinity,
+      max = Infinity,
+      canBeZero = true,
+      isFinite = false,
+      isInt = false,
+      acc = "";
+
+    for (const ch of part._def.checks) {
```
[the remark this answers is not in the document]
igalklebanov: added specific errors for unexpected checks.

deno/lib/__tests__/template-literal.test.ts
```diff
@@ -534,6 +572,52 @@ test("template literal parsing - success - complex cases", () => {
...
+  measurement.parse("1vmin");
+  measurement.parse("1.1vmin");
+  measurement.parse("0vmin");
+  measurement.parse("-1vmin");
+  measurement.parse("-1.1vmin");
+  measurement.parse("1vmax");
+  measurement.parse("1.1vmax");
+  measurement.parse("0vmax");
+  measurement.parse("-1vmax");
+  measurement.parse("-1.1vmax");
```
maxArturo, line 619: 💯 for thoroughness

src/types.ts
```diff
@@ -4525,6 +4549,11 @@ export class ZodTemplateLiteral<Template extends string = ""> extends ZodType<
         minLength = Math.max(minLength, ch.value);
       } else if (ch.kind === "startsWith") {
         startsWith = ch.value;
+      } else {
+        throw new ZodTemplateLiteralUnsupportedCheckError(
```
maxArturo, line 4553: 👍 this works!

src/types.ts
```diff
@@ -4329,6 +4329,38 @@ export interface ZodTemplateLiteralDef extends ZodTypeDef {
   typeName: ZodFirstPartyTypeKind.ZodTemplateLiteral;
 }
 
+export class ZodTemplateLiteralUnsupportedTypeError extends Error {
```
maxArturo, line 4332: Nice! I like having explicit errors for issues encountered on schema creation. With that said, I'm a little concerned if we're adding a lot of new functionality in a single PR. However, also we're having delays getting PRs reviewed so.. 🤷 

WDYT re: putting this in the `ZodError` file?

src/types.ts
```diff
@@ -4572,6 +4601,7 @@ export class ZodTemplateLiteral<Template extends string = ""> extends ZodType<
     return `{${minLength},}`;
   }
 
+  // FIXME: we do not support exponent notation (e.g. 2e5) since it conflicts with `.int()`.
```
maxArturo, line 4604: 👍

maxArturo (approved):
👍 ! Looks good to me. Just one thought on the new error classes but its not a blocker for this to be merged IMO. Thank you again for your timely fixes 🙏

src/types.ts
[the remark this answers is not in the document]
maxArturo: > The general rule is, any transforms are prohibited (see ZodEffects/ZodPipeline ban), let's go with that

sounds good to me!

> I tried to steer away from this, because it'll end up hurting every new feature.
> We could throw on unsupported checks.. without forcing to implement every check

Hmm. In one hand, _because_ of the exhaustive checks I found the `trim()` scenario. But on the other hand, I do see what you mean in terms of making each feature more complicated, especially with all the definitions being in one big file... :) 

The catch-all you added at the end is a good middle ground. Thank you!

src/types.ts
igalklebanov: Sounds good to me! 👍

igalklebanov:
@maxArturo Thank you for the thorough review and great feedback! 💪

spiftire:
Any plans to add this?

olehmisar:
Did you consider this API?
image

A quick dirty implementation:
```ts
import { z } from "zod";

// this is already present in zod codebase
type Writeable<T> = { -readonly [P in keyof T]: T[P] };

type _TemplateLiteralPartOutput =
  | string
  | number
  | boolean
  | bigint
  | null
  | undefined;
type _TemplateLiteralPartInput =
  | _TemplateLiteralPartOutput
  | z.ZodType<_TemplateLiteralPartOutput>;
type _MapPart<T extends _TemplateLiteralPartInput> = T extends z.ZodTypeAny
  ? T["_output"]
  : T;
type _ConcatParts<T extends _TemplateLiteralPartInput[]> = T extends [
  infer THead extends _TemplateLiteralPartInput,
  ...infer TTail extends _TemplateLiteralPartInput[]
]
  ? `${_MapPart<THead>}${_ConcatParts<TTail>}`
  : "";

function templateLiteral<T extends readonly _TemplateLiteralPartInput[]>(
  parts: T
): _ConcatParts<Writeable<T>> {
  throw "implementation here";
}

const a = templateLiteral([
  `hello `,
  z.string(),
  "! I am ",
  z.number(),
  " years old",
] as const);
```

igalklebanov:
@olehmisar Yeah. Read @colinhacks's comments [here](https://github.com/colinhacks/zod/issues/1563).

KholdStare:
Hey @igalklebanov, any updates on this? Great work btw. 😄 

Are you still considering @olehmisar 's API?

igalklebanov:
> Hey @igalklebanov, any updates on this? Great work btw.

None.

> Are you still considering @olehmisar 's API?

Never considered. It goes against the author's current design principles.

KholdStare:
Thanks! I think I misread the comment regarding the other API. Anything stopping this from being merged? Looks like everything is passing (except prettier check on README)

IanVS:
I was looking for exactly a feature like this.  Is there something blocking this PR, or just lack of bandwidth to review it?

MikeRippon:
As time goes on I feel like this would be more and more useful for us every day. I've now got tons of cases where we use templated strings to differentiate different types of ids such as built-in vs custom entites, or for type-safe hierarchies (or both):

```typescript
type ContentTypeId = "Link" | "Video" | "Custom:${string}"
type ContentId = "${ContentTypeId}:${string}"

type CategoryId = "Sponsor" | "Resource" | "Custom:${string}"
type SubCategoryId = "${CategoryId}:${string}"
```

Manually managing `regex().transform()` is just becoming increasingly painful and error prone, especially with the nested/hierarchical types. I think this is literally the only thing that we're really missing from Zod.

colinhacks:
This is a tough one. For starters, this isn't the API I would advocate for - we can use tagged template literals here to make the API more isomorphic with the represented type. I've done some experiments and I believe this is workable.

```ts
z.template`asdf${z.string()}${z.number()}`;
```
 
That said, this is a lot of code and complexity (2k LOC) for a pretty niche feature. I see all the upvotes/reactions, but the total numbers are still fairly small. This also falls into the set of features (like z.discriminatedUnion) that are inelegant to implement, because they require a bunch of switch-like logic over multiple Zod subclasses to implement correctly. 

All told I'm not super enthusiastic about this in the short term. Zod 4 will have better treeshakability characteristics, which changes the calculus on whether to merge obscure-ish features like this. I'll leave this open and try to keep people posted as Zod 4 comes along.

igalklebanov:
@colinhacks Your suggested API is cleaner for sure.

How do you infer the literal text outside the interpolated positions?

image
https://tsplay.dev/N5agMN

colinhacks:
Oops, yeah the tagged template literal API won't work. I got excited that the interpolations could be inferred but apparently the literal parts can't be as Igal said. Bummer.

I just rebased this onto v4 and made some changes:


```ts
z.literal.template([ "asdf", z.number() ]);
```

### `z.literal.template`

I like the dot-chaining thing. And since this is ultimately a special case of `z.literal`, I think it makes sense to use `z.literal.template`. 

### Array API

I see this was discussed above. In this case, I think an array-based API is fine here. My concerns about using arrays in `.pick/.omit` do not apply to the same extent. Users often have arrays of keys laying about from external tools or `Object.keys()`. Trying to pass these into `.pick()` would be problematic. I also like the isomorphism of using object masks to pick/omit an object schema. 

In this case though, the `Parts` argument is a specific mix of literals and schemas that's less likely to be declared externally. And an array is more isomorphic to the syntax of a template literal. 

### Looser typings 

The `z.literal.template` API no longer tries to statically prevent `ZodNan`, `ZodPipeline`, etc. I'm increasingly of the opinion it's better to provide an informative runtime error here, instead of an obscure assignment error. 

---

I tried using `const` parameters for the first time, since they were introduced in TS 5.0, and Zod 4 will require TS 5.0+. But for some reason the const inference didn't seem to work properly until TS 5.3 🤷‍♂️ It's a bit disappointing, and I couldn't see anything in the release notes for 5.3 to explain this. Strange.

colinhacks:
Gonna merge this into `v4` now. If there are any serious concerns with the new API, etc, there's plenty of time to address that before the v4 release.

Igal, amazing work on this! 🙌

PR #1793, opened 2023-01-01
Making `coerce` to be a type argument

### LINKED ISSUE #1760: 3.20: Input type of a schema with a coercion
I believe there is a confusion with the way how the new `coerce` feature works together with the inferring of the input type.

Here is the example with the `ZodString` having `coerce` enabled:

```ts
const schema = z.coerce.string(); // ZodString
console.log(schema.parse(123)); // "123", no errors
type SchemaInput = z.input<typeof schema>; // string
const input: SchemaInput = 123; // TS2322: Type 'number' is not assignable to type 'string'.
```

Under the hood, as per my understanding, the schema actually accepts `any` type of the input, making a `String` out of them.
However, inferring the input type of such schema returns `string`, so the `coerce` feature is not taken in account on the typescript level.

I believe, the `z.coerce.string()` should create `ZodType<string, ZodStringDef, any>` (input type — `any`, not equal to the output type — `string`).

what do you think, @colinhacks ?

RobinTail:
Fixes #1760 

This PR makes the schema types, that support `coerce` feature, to have the correct type of the input schema if this feature is enabled.

Thus, `z.coerce.number()` will have the type compatible to `ZodType<number, ZodNumberDef, any>`.
Third type argument stands for input type, which in case of `coerce` accepts anything.

Therefore, taking `z.input<>` on such schema will return `any`.

The behavior of schemas and methods without `coerce` feature, of having that feature disabled, remains the same.

Thanks to the default values of the new type argument, this change is backward compatible.

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -505,10 +505,11 @@ export type ZodStringCheck =
       message?: string;
     };
 
-export interface ZodStringDef extends ZodTypeDef {
+export interface ZodStringDef<Coerce extends boolean = false>
+  extends ZodTypeDef {
   checks: ZodStringCheck[];
   typeName: ZodFirstPartyTypeKind.ZodString;
-  coerce: boolean;
+  coerce: Coerce;
 }
 
 const cuidRegex = /^c[^\s-]{8,}$/i;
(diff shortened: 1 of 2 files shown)


RobinTail:
The `test with Deno` fails due to some issue within the `master` branch

RobinTail:
Should be fixed after merging #1794

RobinTail:
@JacobWeisenburger  || @colinhacks , please review

JacobWeisenburger:
Unfortunately, I don't think I have enough experience to approve this. Perhaps @colinhacks or @scotttrinh would be able to help.

RobinTail:
Ok. I updated the description for the clarity.

RobinTail:
@colinhacks || @scotttrinh  , please review

RobinTail:
image

aleclarson:
Why not create a new `ZodCoercedString` type (plus the others) instead of complicating the existing types?

colinhacks:
Discussed this under the original issue: https://github.com/colinhacks/zod/issues/1760#issuecomment-1407808567

PR #1797, opened 2023-01-02
Allow also "[+-]hh" as datetime offset

rafw87:
It adds support for `[+-]hh` datetime offset, according to https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators.

Example:
```
datetime.parse("2020-10-14T17:42:29+03");
```


colinhacks:
Looks good, thanks!

colinhacks:
So I've rolled this back in 35f0a38 given that (as @samchungy pointed out) this format can't be parsed by the `Date` constructor in any of the major JS engines.

Naapperas:
@colinhacks I was just now writing an issue regarding this issue because the documentation says that this format passes validation. While it is a valid ISO 8601 date string, we should perhaps update the documentation to reflect this problem, at least while the major JS engines still can't parse this format (that is if they ever will).

PR #1806, opened 2023-01-03
Update README.md: Tuple to Array

### LINKED ISSUE #1805: Fix zod enum `.option` property documentation in Readme
> You can also retrieve the list of options as a tuple with the `.options` property:

According to the doc in the Readme, the `.option` property is supposed to return a tuple, but in the actual implementation, it returns an array. This is also illustrated by the example juste bellow in the doc:

> FishEnum.options; // ["Salmon", "Tuna", "Trout"]);

The doc should be updated to state that the `.options` properties return an array instead of a tuple

chabroA:
fixes #1805 

Update zod enum `option` property documentation to better match reality. And fix typo in example code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -782,10 +782,10 @@ FishEnum.enum;
 */
 ```
 
-You can also retrieve the list of options as a tuple with the `.options` property:
+You can also retrieve the list of options as an array with the `.options` property:
 
 ```ts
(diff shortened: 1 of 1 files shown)


JacobWeisenburger:
I'm going to close this for now. You are more than welcome to see if someone else will merge it for you. But I just don't think this is needed.

PR #1813, opened 2023-01-04
Add string cuid2() validation

joulev:
[`cuid`](https://github.com/paralleldrive/cuid) has been deprecated in favour of [`cuid2`](https://github.com/paralleldrive/cuid2). Now it's not yet widely used but I believe it will catch up with `cuid` soon. This PR adds validation for `cuid2`.

It's my first time contributing here so please tell if I did something wrong. I did update the README with the additional method but feel free to revert the change there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -535,6 +535,7 @@ z.string().email();
 z.string().url();
 z.string().uuid();
 z.string().cuid();
+z.string().cuid2();
 z.string().regex(regex);
 z.string().startsWith(string);
 z.string().endsWith(string);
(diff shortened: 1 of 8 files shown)


colinhacks:
Looks great, thanks!

ciscoheat:
Great news, I've been waiting for this! Is a npm release far away?

joulev:
@ciscoheat Of course I don't know when a new version will be released but if you need it immediately you can use the regex directly

```ts
const cuid2Regex = /^[a-z][a-z0-9]*$/;
```

ciscoheat:
Yep, I'll do that in the meantime. It's kind of amusing that a one-char id is technically possible with this one...!

PR #1841, opened 2023-01-10
Readme: Reference `coerce` in "Supporting date strings" section

tordans:
Am I right in assuming that the `coerce` is the new way to solve the question answered by @JacobWeisenburger in https://github.com/colinhacks/zod/discussions/879#discussioncomment-2036276 – if so, this PR offers an update to the docs.

(Let me know if I should squash the typo commit or if you can squash merge via Github.)

---

Ping https://github.com/colinhacks/zod/pull/881

### DIFF
--- a/README.md
+++ b/README.md
@@ -731,18 +731,9 @@ z.date().max(new Date(), { message: "Too young!" });
 
 **Supporting date strings**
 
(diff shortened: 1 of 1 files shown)


JacobWeisenburger:
thanks for your suggestion. I used what you had and added an example. https://github.com/colinhacks/zod#dates

Issue #798, opened 2021-11-29
partial() doesn't work


nemanjam:
I use `userUpdateSchema` to make all fields optional and pass undefined `password` field but keep getting `Should be at least 3 characters` for `password` field. The schemas are bellow:

```typescript
export const userLoginSchema = z.object({
  email: z.string().email(),
  password: z.string().min(3).max(20), // I send undefined for this field
});

export const userRegisterSchema = userLoginSchema.extend({
  name: z.string().min(3).max(15),
  username: z.string().min(3).max(15),
});

export const userUpdateSchema = userRegisterSchema
  .extend({
    avatar: isBrowser() ? z.instanceof(FileList) : z.any(),
  })
  .omit({ email: true })
  .partial(); // make all fields optional

...

const Settings: React.FC<Props> = ({ user }) => {

  const { register, handleSubmit, formState, watch, getValues } = useForm({
    resolver: zodResolver(userUpdateSchema), // use this schema
    defaultValues: {
      username: user.username,
      name: user.name,
      avatar: user.image,
      password: undefined, // like this, just submit initial values
    },
  });
```

scotttrinh:
I'm not familiar with this hook, but are you certain it's not trying to pass `""` for `password` rather than `undefined` during the validation phase?

From our perspective, your code does work as intended:

[CodeSandbox](https://codesandbox.io/s/snowy-mountain-uzhtk?file=/src/index.ts)
```ts
import test from "tape";
import { z } from "zod";

export const userLoginSchema = z.object({
  email: z.string().email(),
  password: z.string().min(3).max(20)
});

export const userRegisterSchema = userLoginSchema.extend({
  name: z.string().min(3).max(15),
  username: z.string().min(3).max(15)
});

export const userUpdateSchema = userRegisterSchema
  .extend({
    avatar: z.any()
  })
  .omit({ email: true })
  .partial();

test("valid", (t) => {
  t.true(userUpdateSchema.safeParse({}).success); // this succeeds
  t.end();
});

test("invalid", (t) => {
  t.false(userUpdateSchema.safeParse({ password: "" }).success); // this fails
  t.end();
});
```

Closing as this appears to be more of an issue with integrating Zod into your form validation tool than any problem with Zod itself. Feel free to follow up here if you get something working or have any other questions that we might be able to help with from our side.

nemanjam:
[...]

PR #1849, opened 2023-01-10
Readme: Add example for string validation for an optional field to chapter "Unions"

tordans:
This is based on https://github.com/colinhacks/zod/issues/798#issuecomment-991003983.

I consider it a pretty useful addition which will like save some people some manual research. 

Especially since the solution was not what I was expecting and testing first…


JacobWeisenburger (approved):

JacobWeisenburger:
great work! keep it up.

PR #1875, opened 2023-01-13
fix: make sure only mask keys with truthy values are respected at runtime @ `.pick`, `.omit`, `.partial` & `.required`.

### LINKED ISSUE #1854: Method `.pick` / `.omit` might be misleading
Hi there,

I was using zod with some great methods and found out `.pick/ .omit` might be a bit misleading.

[According to the document](https://zod.dev/?id=pickomit), both of the methods accept a object that contains a key with a value which looks like this:
```ts
const Recipe = z.object({
  id: z.string(),
  name: z.string(),
  ingredients: z.array(z.string()),
});

const JustTheName = Recipe.pick({ name: true });
const NoIDRecipe = Recipe.omit({ id: true });
```

but the [actual behavior](https://github.com/colinhacks/zod/blob/master/src/types.ts#L2152) (not sure was intended or a logic error) is just looping through the given object's key. No matter the value is there or not (truthy or falsy), as long as the key exist in the object it will just effect the schema which seems a bit pointless to having the value with it.

I suggest the value should be also consider, for example:
```ts
const shouldOmit = true

Schema.omit({
   name: shouldOmit, //true
   age: !shouldOmit //false
})

Schema.parse({ 
  name: 'Charlie',
  age: 20
})  // { age: 20 }
```

Any thoughts?

igalklebanov:
resolves #1854.


wlcharlie:
@igalklebanov Thanks for helping out my issue 😄 

Just curious, should `Mask` also extends not just `true` but `boolean` ?

igalklebanov:
> Just curious, should `Mask` also extends not just `true` but `boolean` ?

I think this adds type complexity, and I can't see the benefits.. what are some benefits?

wlcharlie:
> I think this adds type complexity, and I can't see the benefits.. what are some benefits?

Well, since the PR fix the runtime issue, just trying to make code more easy to read.

If one use the method like this (the code under), typescript throw error since the argument of the type shouldn't have other type expect `true` or `undefined`.
```typescript
schema.pick({ name: true, age: false })
//  error on here         ^^^
``` 

Currently, it can fix by comment @ts-ignore, @ts-expect-error, or add more logic:
```typescript
schema.pick({ name: true, ...(shouldPick && { age: true } })
// this work
```
or this is acceptable?

igalklebanov:
> > I think this adds type complexity, and I can't see the benefits.. what are some benefits?
> 
> Well, since the PR fix the runtime issue, just trying to make code more easy to read.
> 
> If one use the method like this (the code under), typescript throw error since the argument of the type shouldn't have other type expect `true` or `undefined`.
> 
> ```ts
> schema.pick({ name: true, age: false })
> //  error on here         ^^^
> ```

This works as intended. You don't pass things you don't want to affect schema with. Same as how typescript `Pick<O, K>` or lodash `pick(obj, keys)` work.

Mask objects were chosen since the alternative, key array, is a footgun (breaks inference quite easily, and users were having trouble with array arguments in other zod APIs) according to @colinhacks.

> Currently, it can fix by comment @ts-ignore, @ts-expect-error, or add more logic:
> 
> ```ts
> schema.pick({ name: true, ...(shouldPick && { age: true } })
> // this work
> ```
> 
> or this is acceptable?

The code example would work.

You can also:

```ts
const subSchema = schema.pick({ name: true });

if (!shouldPick) {
  return subSchema;
}

return subSchema.addKey('age', schema.shape.age);
```

wlcharlie:
@igalklebanov Cool, then the current usage are good!

> This works as intended. You don't pass things you don't want to affect schema with. Same as how typescript Pick<O, K> or lodash pick(obj, keys) work.
> 
> Mask objects were chosen since the alternative, key array, is a footgun (breaks inference quite easily, and users were having trouble with array arguments in other zod APIs) according to @colinhacks.

This is really good point, bad array!
Thanks for having a discuss with me. Appreciated! 😄

src/types.ts
```diff
@@ -2198,24 +2203,16 @@ export class ZodObject<
...
-      });
-      return new ZodObject({
-        ...this._def,
-        shape: () => newShape,
-      }) as any;
-    } else {
-      for (const key in this.shape) {
-        const fieldSchema = this.shape[key];
+
+    util.objectKeys(this.shape).forEach((key) => {
```
maxArturo, line 2207: I think we could add a test to flex this new behavior here

src/types.ts
```diff
@@ -2239,30 +2236,22 @@ export class ZodObject<
...
-          let newField = fieldSchema;
-          while (newField instanceof ZodOptional) {
-            newField = (newField as ZodOptional<any>)._def.innerType;
-          }
-          newShape[key] = newField;
-        }
-      });
-    } else {
-      for (const key in this.shape) {
+
```
maxArturo, line 2239: likewise here

maxArturo (changes requested):
@igalklebanov We meet again! You've been adding a bunch of functionality lately! 🥇 
I suggest we add a couple of tests that I think are missing, but other than that LGTM. 👍

src/types.ts
igalklebanov: You're right!

Damn I remember adding these test cases.. 😞

igalklebanov:
@maxArturo Thank you once again! 💪

PR #1876, opened 2023-01-13
fix: refined nested object keys are not optional when using `deepPartial`.

### LINKED ISSUE #1835: `refine()` breaks `deepPartial()`
`deepPartial()` is not actually deep, if nested object has `refine()` on it.
Example 1, works:
```ts
const test1 = z
  .object({
    inner: z.object({
      data: z.string(),
    }),
  })
  .deepPartial();

type Test1 = z.infer<typeof test1>;

/*
type Test1 = {
    inner?: {
        data?: string | undefined;
    } | undefined;
}
*/
```
Example 2, doesn't work, `innerWithRefine` is not partial:
```ts
const test2 = z
  .object({
    innerWithRefine: z
      .object({
        data: z.string(),
      })
      .refine((value) => true),
  })
  .deepPartial();

type Test2 = z.infer<typeof test2>;

/*
type Test2 = {
    innerWithRefine?: {
        data: string; // is not optional, but it is supposed to be
    } | undefined;
}
*/
```

igalklebanov:
resolves #1835.

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -52,6 +52,7 @@
   - [Deno](#from-denolandx-deno)
 - [Basic usage](#basic-usage)
 - [Primitives](#primitives)
+- [Coercion for primitives](#coercion-for-primitives)
 - [Literals](#literals)
 - [Strings](#strings)
 - [Numbers](#numbers)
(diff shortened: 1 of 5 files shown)


1valdis:
Does this fix the implementation too or just typing? Afaik it currently validates in non-optional way too, not just has the wrong type.

igalklebanov:
> Does this fix the implementation too or just typing? Afaik it currently validates in non-optional way too, not just has the wrong type.

The issue did not discuss runtime. I can expand this PR to deal with the runtime side too I guess.

colinhacks:
As mentioned any PR here would need to handle both static and runtime - they must agree.

More broadly though, this fixes one edge case of dozens. A "full solution" would need to account for every possible subtype of ZodType in `deepPartialify` and even then you aren't accounting for user-defined subtypes so things could still break.  Doing this would also add a lot of code to implement a feature that should probably never have been added.

igalklebanov:
> As mentioned any PR here would need to handle both static and runtime - they must agree.

My bad. 😢 

> More broadly though, this fixes one edge case of dozens. A "full solution" would need to account for every possible subtype of ZodType in deepPartialify and even then you aren't accounting for user-defined subtypes so things could still break. Doing this would also add a lot of code to implement a feature (deepPartial) that should probably never have been added. I'll likely deprecate it because it only throws people off.

Sounds like its totally not worth addressing. I'm closing this PR then. Thank you! 💪

PR #1877, opened 2023-01-13
Readme: Add intro to chapter Literals

tordans:
I was trying to understand what zod literals are. The docs could benefit from an intro sentence. This is a suggestion for such an intro. However, this can probably be more specific. For example, which differences are there to TS literal types? Eg. with Symbol and BigInt (which I both know nothing about…).


JacobWeisenburger:
Great work

PR #1926, opened 2023-01-23
fix: [#1839] remove caught errors from issues

### LINKED ISSUE #1839: Caught errors are not distinguishable from uncaught errors
In v3.20.x errors handled via the `catch` operator are logged along with errors that actually cause the parsing to fail. This makes it difficult to differentiate "fatal", uncaught errors from non-fatal, caught/handled errors. This isn't a huge deal in smaller schemas, but becomes very tedious with larger ones.

Is there a way to suppress caught errors from `parse` and `safeParse`, or perhaps mark them as caught?

### Example
See the [interactive example on Stackblitz](https://stackblitz.com/edit/typescript-oxmrfa?file=index.ts), but here's the code for a minimal reproduction:
```typescript
const schema = z.object({
  stringProp: z.string(),
  numberPropWithCatch: z.number().catch(0),
});

schema.parse({
  // This value is of incorrect type and causes the parsing to fail
  stringProp: {},
  // This value is of incorrect type as well, but DOESN'T
  // cause parsing to fail as it's caught in the schema.
  // However, the error is reported in the overall list
  // of parsing errors.
  numberPropWithCatch: '',
});
```

Running this code results in an error **as expected**, but notice how both the fatal and caught errors are displayed:
image

maxArturo:
Fixes #1839. Includes tests.

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -52,6 +52,7 @@
   - [Deno](#from-denolandx-deno)
 - [Basic usage](#basic-usage)
 - [Primitives](#primitives)
+- [Coercion for primitives](#coercion-for-primitives)
 - [Literals](#literals)
 - [Strings](#strings)
 - [Numbers](#numbers)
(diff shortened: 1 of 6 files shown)


colinhacks:
Your original implementation didn't account for the possibility that multiple issues might be added by the ZodCatch instance's inner schema. Fixed. Other than that, looks good, thanks!

PR #1930, opened 2023-01-25
Updates documentation for `z.coerce.date`

### LINKED ISSUE #1929: `z.coerce.date().parse(null)` validates successfully
This is a bit of a gotcha:
```typescript
z.coerce.date().safeParse(null).success // true
```

It's not technically unexpected, though:
```typescript
new Date(null) // Thu Jan 01 1970 10:00:00 GMT+1000 (Australian Eastern Standard Time)
```

I'm happy to PR a docs update briefly describing this gotcha, unless the project wants to address this by changing how `z.coerce.date()` works internally.

(IMO it should _not_ change `z.coerce.date()` as just calling `new Date()` without any other logic is very tidy.)

psychedelicious:
Updates documentation for `z.coerce.date()`.

Resolves #1929

### DIFF
--- a/README.md
+++ b/README.md
@@ -755,6 +755,9 @@ console.log( dateSchema.safeParse( new Date( '1/10/23' ) ).success ) // true
 /* invalid dates */
 console.log( dateSchema.safeParse( '2023-13-10' ).success ) // false
 console.log( dateSchema.safeParse( '0000-00-00' ).success ) // false
+
+/* one gotcha - null is a valid Date */
+console.log( dateSchema.safeParse( null ).success ) // true
 ```
 
 For older zod versions, use [`z.preprocess`](#preprocess) like [described in this thread](https://github.com/colinhacks/zod/discussions/879#discussioncomment-2036276).
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -52,6 +52,7 @@
   - [Deno](#from-denolandx-deno)
 - [Basic usage](#basic-usage)
 - [Primitives](#primitives)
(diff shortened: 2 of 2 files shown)


JacobWeisenburger:
considering this is a gotcha with javascript and not necessarily a problem with zod. I'm not sure this needs to be fixed. But I'm open to hearing a rebuttal or other people's opinions.

```ts
new Date(null) // Thu Jan 01 1970 10:00:00 GMT+1000 (Australian Eastern Standard Time)
```

JacobWeisenburger:
To me it seems out of zod's scope to include notes in the docs about something that is a quirk of javascript and not zod. Because if we include this, then why don't we include other quirks of javascript?

Perhaps we should include these things too:
```ts
console.log([] == true) // false
console.log([] == false) // true
console.log(Boolean([])) // true

console.log({} == true) // false
console.log(Boolean({})) // false
```

psychedelicious:
I agree it's far outside scope to mention every quirk, but the docs *do* provide a number of examples of valid and invalid dates. So why not add the one really unexpectedly-valid date in there? That was my thinking.

colinhacks:
I'm also opposed to this - the docs mention that `z.coerce.date` passes the value into `new Date()`. This is JS, so all sorts of crazy nonsense will happen. We don't want to give the impression that this is a quirk in Zod.

sschneider-ihre-pvs:
that would render z.coerce.date be always not nullable? So there is a possibility to get `Invalid Date` for some string input values and for undefined, and null is not one of them.
Is it possible to coerce null into undefined or filter out null before coerce.date ?

PR #1945, opened 2023-01-27
Minor typo fix

iamchandru6470:
Minor typo fix under Basic usage "Literals" section

### DIFF
--- a/README.md
+++ b/README.md
@@ -558,7 +558,7 @@ z.coerce.boolean().parse(null); // => false
 
 ## Literals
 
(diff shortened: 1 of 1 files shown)


JacobWeisenburger (changes requested):
Actually I think it should be 'allow' because it's being used with a plural 'literals'.

iamchandru6470:
> Actually I think it should be 'allow' because it's being used with a plural 'literals'.

I had confusion between 'allow' or 'allows', as you suggested 'allow' makes sense.
Updating PR with the same.

JacobWeisenburger:
Thanks

Issue #1979, opened 2023-02-03
Able to validate invalid email


imran-vz:
Currently, I'm able to pass this as email and get no issues.
```
imran@345934898929^&^&*&**#$%*&%#imran.com
```

```
pawan.anand@%9y8325&%#%$&#$R%&#$%R&#$%R&%#$R%%^^%5rw3ewe.r.ef.d.d.d.d.d.aaaa.wef.co
```

JacobWeisenburger:
Are you able to give a code example?

fvckDesa:
```js
const scheme = z.string().email();
const email = "pawan.anand@%9y8325&%#%$&#$R%&#$%R&#$%R&%#$R%%^^%5rw3ewe.r.ef.d.d.d.d.d.aaaa.wef.co";
scheme.parse(email); //don't throw an error
```

According to this site ([link](https://snov.io/knowledgebase/what-is-a-valid-email-address-format/#:~:text=Every%20email%20address%20consists%20of,.com%E2%80%9D%20is%20the%20domain.)) the domain can contains:

- max 63 characters
- uppercase and lowercase Latin letters
- numbers 0-9 (not only numbers)
- one "-" not in start or end

JacobWeisenburger:
Behind the scenes zod uses this regex to validate email addresses:
https://github.com/colinhacks/zod/blob/a57fc2da3c48a5a2270e9684017d8bc056af8a40/src/types.ts#L523-L524

If you would like a different regex to do this, you can use something like these:

```ts
// completely custom regex
const schema = z.string().regex( /some regex/ )

// custom regex that must match the zod email regex
const schema = z.string().email().regex( /some regex/ )
```

fvckDesa:
I created a pull request and for me this is a valid change because can prevent some issues

florent-martineau:
Another example of email that works but, I believe, is invalid:
``invalid@foo.....com``

M0hammedImran:
> Another example of email that works but, I believe, is invalid: `invalid@foo.....com`

This doesn't pass the test for the default regex.

JacobWeisenburger:
> This doesn't pass the test for the default regex.

I disagree.

```ts
const schema = z.string().email()
console.log( schema.safeParse( 'invalid@foo.....com' ).success ) // true
```

Issue #1515, opened 2022-10-25
Update email regex to use W3C's one?


timojuonoja:
There is no 100% bullet proof email regex, but should you update it to use W3C's one? Currently I'm facing a bug that `a\b@c.com` is passed, when W3C's regex wouldn't pass it.

Current zod regex for email is:
```
(/^(([^<>()[\]\.,;:\s@\"]+(\.[^<>()[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$/i)
```

When [W3C suggests](https://html.spec.whatwg.org/multipage/input.html#email-state-(type=email)):
```
/^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/
```

Tyler-Murphy:
See also: https://github.com/colinhacks/zod/pull/1435

timojuonoja:
@Tyler-Murphy PR #1435 does not fix this issue.

Tyler-Murphy:
Sorry, I should have been clearer. I'm just pointing out its existence, since the W3C regex appears not to work with some of the test cases in #1435.

mpartel:
This would also allow `+` in addresses like name+test123@example.com, which would be good.

fvckDesa:
@timojuonoja "\" non è un carattere email valido, l'unico modo per usarlo è con le virgolette doppie intorno alla parte locale dell'email

timojuonoja:
> @timojuonoja "\" is not valid email character, the only way to use it is with the double quotes around the local-part of email

Yes, you are absolutely right by the way.
We seem to have an error in our test cases after all... I'm closing this due to incorrect assumptions.

PR #1982, opened 2023-02-03
fix[#1979]: Increment Email validation

fvckDesa:
solves issues #1979, #1515 and #1435


colinhacks:
Looks great, thanks!

Tyler-Murphy:
This only gets part of #1435 I think, since it still doesn't allow IP addresses without square brackets around the IP address, e.g. `someone@123.123.123.123`. Only `someone@[123.123.123.123]` is allowed by the new regex here.

ggmartins091:
@colinhacks @fvckDesa fyi

fvckDesa:
@ggmartins091 Yes, the problem is here:
``` ts
([A-Za-z0-9]([A-Za-z0-9-]*[A-Za-z0-9])*(\.[A-Za-z]{2,})+)
```
As soon as I can I try to fix it

PR #2013, opened 2023-02-08
Changed order of createZodEnum overloads to make ZodEnum.extract match the first declaration

### LINKED ISSUE #2011: type error
in type.d.ts file, Some type errors occur here： Writeable<FilterEnum<T, ToExclude[number]>>
version: '3.20.3'
image

SimplyLinn:
So, in the new extract feature for ZodEnums, the order of the createZodEnum overloads meant that the output of FilterEnum matched the first overload `readonly [U, ...U[]]`, since writeable arrays extend readable arrays but not vice versa.

Overloads get read top to bottom, and the first match that works is selected, as per my understanding.

Changing the order means that it will not try to make already writeable arrays writeable a second time, which in the case of the FilterEnums output, seems to matter.

Resolves #2011

### DIFF
--- a/deno/lib/__tests__/catch.test.ts
+++ b/deno/lib/__tests__/catch.test.ts
@@ -48,7 +48,7 @@ test("catch with transform", () => {
   );
 
   type inp = z.input<typeof stringWithDefault>;
-  util.assertEqual<inp, string>(true);
+  util.assertEqual<inp, unknown>(true);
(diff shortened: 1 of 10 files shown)


SimplyLinn:
Oh, my editor removed an eslint-disable-next-line comment, since eslint reported it as extraneous. If that is undesired, let me know and I'll revert that change.

fatihaziz:
great, can't wait it got merged

Issue #1981, opened 2023-02-03
Add warning for long parses


rubber-duck-software:
To ensure users are aware of when zod is causing performance issue, add `console.warn` when a parse takes longer than 500ms (or pick a number). Sometimes this can be difficult to discover since javascript is IO heavy, cpu is rarely the performance bottleneck, but when it is, it would be good to know.

mohitmohlia:
Hey @rubber-duck-software @JacobWeisenburger I am picking this issue, it could be my first open source contribution.
Can you please guide me how to stress test the parse() method . 
Basically i am trying this 

<img width="663" alt="Screenshot 2023-02-12 at 3 29 38 PM" src="https://user-images.githubusercontent.com/46368810/218304469-2ae78b62-85fa-4e85-8987-c941a5b2d88a.png">

JacobWeisenburger:
@mohitmohlia looks good to me.

mohitmohlia:
@JacobWeisenburger  Please assign this to me I am raising the PR. Thanks

mohitmohlia:
@JacobWeisenburger Hey can you review the Pull request

JacobWeisenburger:
I don't have the authority to approve this type of thing. sorry

mohitmohlia:
@rubber-duck-software @colinhacks Hey guys, can you please review this PR

colinhacks:
- This would be more okay if you could opt into it with a `debug: true` flag passed into `.parse()`. But I generally don't love per-parse config options.
- Choosing an arbitrary cutoff would please no one, so this would also need to be configurable with another flag
- I'm not sure this should be in Zod core. Applications with requirements like this can create their own timeouts and have total freedom to implement their own logging/fallback logic. A Zod-native API doesn't improve on this materially in my opinion.

rubber-duck-software:
You're right - we can't just turn this on and expect everyone to be happy. It would have to be configurable. 

Perhaps configuring this globally would be good enough. A `z.configure()` which set some module scope variables in zod. Something like 
```js
z.configure({
    longParseWarnTiming: 500
})
```

[...]

PR #2034, opened 2023-02-12
console warning added for parse() method

mohitmohlia:
Fixes : #1981


colinhacks:
Sorry, I don't think this feature should be in Zod core. If we do add it, the implementation will need to be quite a bit more customizable than this and probably use timeouts.

PR #2038, opened 2023-02-15
lowercase method for ZodString

Levin779:
This adds `lowerCase()` to `ZodString`.

```ts
z.string().lowerCase().parse('LOWERCASE') = 'lowercase'

```


JacobWeisenburger:
This already works:
```ts
console.log( z.string().parse( 'LOWERCASE' ).toLowerCase() ) // 'lowercase'
```

So why do we need another way of doing this?

Levin779:
Than why is there this:

```ts
console.log(z.string().trim().parse(' trim ')) // 'trim'
```
If you could just do this:

```ts
console.log(z.string().parse(' trim ').trim()) // 'trim'
```

JacobWeisenburger:
Personally, I don't know why we need the trim method, because like you said, we can just do this:
```ts
console.log(z.string().parse(' trim ').trim()) // 'trim'
```

But ultimately, it's not my call weather or not things like this make it in. I'm just trying to ask for your reasoning about why we need `lowerCase`. To me, adding `lowerCase` simply because we have `trim` is not a good enough reason. Because then where do we stop adding things like this? Seems like a lot of [scope creep](https://en.wikipedia.org/wiki/Scope_creep).

Levin779:
Obviously you are right but `trim()` and `lowerCase()` could be used in scenarios where you have to parse the same zod schema several times.
So you dont have to call `toLowerCase()` everytime.

I know it's not ideal but imagine something like this:

```ts
const UserSchema = z.object({
	name: z.string().trim().lowerCase(),
	email: z.string().trim().lowerCase().email(),
)}

const User = UserSchema.parse(userData)

// or

const User = {
	name: userData.name.trim().toLowerCase(),
	email: userData.email.trim().toLowerCase(),
}
```

With zod you could reuse the Schema and don't have to pay attention to things like `trim()` or `toLowerCase()`.

colinhacks:
> So why do we need another way of doing this?

There's value in only needing to declare this transform once, instead of per-parse.

I reluctantly added `.trim()` because I was resistant to modifying the input inside the parsing logic for `ZodString`. But ultimately it's a really clean API that's useful, so I decided to be less dogmatic about siloing all transformation logic in `ZodEffect`. We now have `.endsWith` and `.startsWith` too so it makes sense to add `.toLowerCase()` and `.toUpperCase()` too imo.

PR #2042, opened 2023-02-15
feat: Add support for arbitrary schema annotations.

nicknotfun:
The 'description' property is useful in that it permits a non-functional tag in the schema that could be used by external tools; there are other arbitrary use-cases where a similar pattern could be useful, for example:

- Titles or names for an editor context
- Marking fields as sensitive [e.g. PIII]
- Tooltips or longer-form descriptions

To this end, add a new method annotate() that supports arbitrary key-value data in a schema.

Note: It would be possible (but potentially an interface change) to re-implement description ontop of this - I have not done that in this PR but can if desired.


nicknotfun:
@colinhacks wondering your thoughts?

juanrgm:
Why not change the name of `annotate` to `meta`?

nicknotfun:
> Why not change the name of `annotate` to `meta`?

Sure, i'm not too fussed on the name- just felt the convention leant towards full words (and 'metadata' wasn't a verb)

nicknotfun:
@colinhacks just checking if this makes sense? It's proven very useful to us [and I imagine could re-implement definition on top of it too]

extradosages:
Seems a like overkill that we're requiring the key to be a symbol, here, instead of just a string.

IvanovES:
> Seems a like overkill that we're requiring the key to be a symbol, here, instead of just a string.

Here's an example that supports this statement. There are multiple OpenAPI generating libraries. With the support for metadata OpenAPI implementors can agree on the key names and then users should be able to switch between implementations. With just symbols this will not be possible. Keep symbols though, people will still probably need them.

What also seems a bit off is that you need to call `annotate` multiple times if you have more than one annotation. Let's just accept the annotations object. This should reduce the schema copying a bit, but the copying is probably not necessary to begin with. Metadata doesn't change the actual schema, so it's not like you can break someone by mutating the annotations object.

nicknotfun:
@extradosages @IvanovES slow turn-around on my part, but agree completely - updated to support bulk assignment of objects (so permitting any object-key as a key)

Wondering if anyone else has comments? We're using this happily

lvisb:
I need this so badly!

nicknotfun:
@colinhacks any input?

gabrielbauman:
I see this feature isn't master, and there are a number of people with different takes on it. FOr now I'm just monkey-patching along the lines of:

`import { z, ZodType } from 'zod';
import { ZodTypeDef } from 'zod/lib/types';

declare module 'zod' {
  interface ZodType<
    Output = any,
    Input = Output
  > {
    annotate(symbol: string, value: any): this;
    meta: { [key: string]: any };
  }
}

// Extend ZodType's prototype
Object.defineProperty(ZodType.prototype, 'annotate', {
  value: function (key: string, value: any) {
    if (!this.meta) {
      this.meta = {};
    }
    this.meta[key] = value;
    return this;
  },
  writable: true
});

// Use
const annotatedString = z.string().annotate('someSpecialThing', true);

// Access meta
console.log(annotatedString.meta['someSpecialThing']); // true
`

Any chance of an official solution being rolled into zod?

nelisbijl:
> I see this feature isn't master, and there are a number of people with different takes on it. For now I'm just monkey-patching along the lines of:

I agree that having some form of meta data could be helpful. For instance whether a number represents a currency or a percentage. That could be used in a custom ZodErrorMap

Unfortunately the suggested solution fails when you add a validation like min or max after the annotation. The ZodType is recreated with an additional check. The meta prop is not transferred to the new object

PR #2049, opened 2023-02-17
feat: z.string().ulid() - add support for ulids

chaunceyau:
# Overview

Adds support for ulids https://github.com/ulid/spec. Shopify has really popularized these and have a lot of benefits https://news.ycombinator.com/item?id=34294693


chaunceyau:
cc @colinhacks @hugo-vrijswijk - not sure how to assign reviewers or if it's something only your team does

deno/lib/types.ts
```diff
@@ -903,6 +918,9 @@ export class ZodString extends ZodType<string, ZodStringDef> {
   get isCUID2() {
     return !!this._def.checks.find((ch) => ch.kind === "cuid2");
   }
+  get isULID() {
+    return !!this._def.checks.find((ch) => ch.kind === "cuid2");
```
Frantss, line 922: probably a copy-paste mistake
```suggestion
    return !!this._def.checks.find((ch) => ch.kind === "ulid");
```

src/types.ts
```diff
@@ -903,6 +918,9 @@ export class ZodString extends ZodType<string, ZodStringDef> {
   get isCUID2() {
     return !!this._def.checks.find((ch) => ch.kind === "cuid2");
   }
+  get isULID() {
+    return !!this._def.checks.find((ch) => ch.kind === "cuid2");
```
Frantss, line 922: Same here
```suggestion
    return !!this._def.checks.find((ch) => ch.kind === "ulid");
```

deno/lib/types.ts
chief-austinc: good catch, addressed!

src/types.ts
chief-austinc: good catch, addressed!

PR #2058, opened 2023-02-18
Fix email issues in pull request #1982

fvckDesa:
- [x] Invalidate emails with dot next other
- [x] Invalidate IPv4 address whit numbers bigger than 255
- [x] Add IPv6 domain


colinhacks:
Looks good!

PR #2063, opened 2023-02-19
Implemented `ZodNumber#multipleOfValue` (closes #1889)

### LINKED ISSUE #1889: Add a getter for `multipleOf` on `ZodNumber`
While creating [`uniforms-bridge-zod`](https://www.npmjs.com/package/uniforms-bridge-zod), I had a need to get the resulting `multipleOf` to use it in a `<input type"number">`'s `step` attribute. Right now, I have to work around it a little:
```ts
type ZodNumberCheck = ZodNumberDef['checks'][number];
const step = field._def.checks.find(
  (check): check is Extract<ZodNumberCheck, { kind: 'multipleOf' }> =>
    check.kind === 'multipleOf',
);
```

_([source](https://github.com/vazco/uniforms/blob/6353c831673adc9a786932c4ab60483c59de3d9c/packages/uniforms-bridge-zod/src/ZodBridge.ts#L173-L181))_, but I thought that exposing it similarly to `maxValue` for `max` and `minValue` for `min` would be a nice feature.

I have to problems with that, though:
1. What if there are multiple `multipleOf`s? Ideally we'd need to calculate least common multiple of them, but maybe using the first one is enough?
2. What should be the name? `multipleOfValue` or `stepValue` stay in-align with `max`/`maxValue`, but aren't ideal.

I can file a PR for that.

radekmie:
In this pull request, I added a getter for `multipleOf` value on `ZodNumber`, as proposed in #1889. I decided to go with the simple solution first, i.e., take only the first value - if needed, it can be changed in the future to calculate the Least Common Multiple (it's mentioned in the tests).

I wasn't sure about two things, though:
1. It looks like other getters (e.g., `maxValue`) are not documented. If there's a place where I could add documentation for `multipleOfValue`, I'll gladly add it as well.
2. There's only a few usages of `Array#find` in the project, and all of them should actually be replaced with `some` (as all of them do `!!this._def.checks.find(...)`). That's why I decided to go with `for..of` and return `null` instead of `undefined` when nothing is found. Let me know whether a different implementation is preferred.

### DIFF
--- a/deno/lib/__tests__/number.test.ts
+++ b/deno/lib/__tests__/number.test.ts
@@ -166,3 +166,16 @@ test("finite getter", () => {
   expect(multipleOfFive.isFinite).toEqual(true);
   expect(z.number().min(5).max(10).isFinite).toEqual(true);
 });
+
+test("multipleOf getter", () => {
+  expect(z.number().multipleOfValue).toEqual(null);
+  expect(z.number().multipleOf(2).multipleOfValue).toEqual(2);
+  expect(z.number().multipleOf(-2).multipleOfValue).toEqual(-2);
+
+  // Only the first value is used. In the future it can be extended to calculate
+  // Least Common Multiple of them.
+  expect(z.number().multipleOf(2).multipleOf(3).multipleOfValue).toEqual(2);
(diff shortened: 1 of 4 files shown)


radekmie:
It's still valid.

radekmie:
It's still valid.

Issue #2059, opened 2023-02-18
Support IP address validation


fvckDesa:
I have been adding email validation with ip address, I can make IP validation a build-in feature.
The result is like:
```js
z.string().ip({ version: "v4" });
```
or
```js
z.string().ipv4();
z.string().ipv6();
```

PR #2066, opened 2023-02-19
feat(#2059): z.string.ip() - add support for IP address

fvckDesa:
This pull request solves #2059 adding [IP Address](https://en.wikipedia.org/wiki/IP_address) validation in zod string

- [x] Basic implementation
- [x] Choose what type of ip 
- [x] Add IPv6 address
- [x] Add ip to docs


colinhacks:
Immaculate.

PR #2087, opened 2023-02-22
feat: add `.catch` error

0xWryth:
# Introduction

Hey 👋🏻

This PR introduces a way to collect a `ZodError` when using `.catch` method, which can be useful in many case when wanting to collect logs whenever default value is used.

# Example
```ts
const numberWithErrorCatch = z.number().catch((e) => {
  console.error(e);

  return 42;
});

// No error shown in console.
numberWithErrorCatch.parse(5); // => 5

// Default value is returned, an error will be shown in console.
numberWithErrorCatch.parse("tuna"); // => 42
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -2124,29 +2124,17 @@ numberWithCatch.parse(5); // => 5
 numberWithCatch.parse("tuna"); // => 42
 ```
 
-Optionally, you can pass a function into `.catch` that will be re-executed whenever a default value needs to be generated:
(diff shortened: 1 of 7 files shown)


colinhacks:
Great stuff! Made a small tweak - passing a `ctx` object into the catch handler to leave room for future APIs and so we can avoid instantiating a ZodError instance if the validation passes. Creating Errors in JavaScript is slow/expensive.

PR #2090, opened 2023-02-23
FIX: emoji regex and tests

joseph-lozano:
so it was really late when Theo put the bounty up on the stupid emoji validation. It had some errors but got merged in while I was sleeping. This fixes some of the issues that were still on the PR. See #2045 

It also turns out the regex can be made much nicer using unicode codepoints in the Regex, see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes and https://unicode.org/reports/tr51/#Emoji_Properties


src/__tests__/string.test.ts
```diff
@@ -101,13 +101,11 @@ test("more email validations", () => {
 
 test("url validations", () => {
   const url = z.string().url();
-  try {
-    url.parse("http://google.com");
-    url.parse("https://google.com/asdf?asdf=ljk3lk4&asdf=234#asdf");
-    expect(() => url.parse("asdf")).toThrow();
-    expect(() => url.parse("https:/")).toThrow();
-    expect(() => url.parse("asdfj@lkjsdf.com")).toThrow();
-  } catch (err) {}
```
joseph-lozano, line 110: My original PR #2045 copied from this test, which had this empty `catch`, so I though it would also fix this in the follow up.

colinhacks:
I'd fixed this with the somewhat hacky per-character validation, but this is definitely better. I was briefly worried that unicode property escapes might not be supported across all engines/runtimes (we've had issues with lookbehind assertions) but it seems to be. 👍 Thanks!

rrebase:
Seems that the following line now makes `zod` incompatible with React Native as it uses the `Unicode property escape` syntax, which is currently not supported by Hermes (the default React Native JS engine)
```ts
const emojiRegex = /^(\p{Extended_Pictographic}|\p{Emoji_Component})+$/u;
```

PR #2094, opened 2023-02-23
docs: Fix docs for `.default()`

tansonlee:


JacobWeisenburger:
Thanks for your help with this PR. `default` only does it's thing if the value passed to parse is undefined. So it doesn't work with empty objects. Please update your PR to reflect this.

```ts
const objectWithDefault = z.object( { fish: z.string() } ).default( { fish: 'tuna' } )
console.log( objectWithDefault.parse( undefined ) ) // { fish: 'tuna' }

const result = objectWithDefault.safeParse( {} )
!result.success && console.log( result.error.issues )
// [
//     {
//         code: 'invalid_type',
//         expected: 'string',
//         received: 'undefined',
//         path: [ 'fish' ],
//         message: 'Required'
//     }
// ]
```

JacobWeisenburger (changes requested):
Thanks for your help with this PR. `default` only does it's thing if the value passed to parse is undefined. So it doesn't work with empty objects. Please update your PR to reflect this.

```ts
const objectWithDefault = z.object( { fish: z.string() } ).default( { fish: 'tuna' } )
console.log( objectWithDefault.parse( undefined ) ) // { fish: 'tuna' }

const result = objectWithDefault.safeParse( {} )
!result.success && console.log( result.error.issues )
// [
//     {
//         code: 'invalid_type',
//         expected: 'string',
//         received: 'undefined',
//         path: [ 'fish' ],
//         message: 'Required'
//     }
// ]
```

colinhacks:
The documentation is correct as-is

PR #2097, opened 2023-02-23
BRAND Record to Non Partial

Hiroshiba:
This PR is an artifact of the discussion.
https://github.com/colinhacks/zod/discussions/2069#discussioncomment-5091931

A `Record` keyed by `string`, `number` or `symbol` is `Non Partial`, but `BRAND` (which should behave similarly) is `Partial`.
This PR ensures that a key with `BRAND` will also be `Non Partial`.

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -1409,7 +1409,7 @@ type NumberSet = z.infer<typeof numberSet>;
 // type NumberSet = Set<number>
 ```
 
-Set schemas can be further contrainted with the following utility methods.
+Set schemas can be further constrained with the following utility methods.
 
 ```ts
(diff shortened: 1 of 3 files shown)


colinhacks:
Looks good!

wdanilo:
It doesn't work:

```ts
const zodKey = Z.string().brand('zodKey')
const zodRecord = Z.record(zodKey, Z.number())
type ZodRecord = Z.infer<typeof zodRecord>
```

still infers as possibly undefined values.

fwoelffel:
I am also observing this unexpected behaviour where using a branded type as the record key would result in a partial record 🤔

Issue #1741, opened 2022-12-22
Zod v3.20.2 causing stalled build in CRA project


tornvallalexander:
Started using Zod in a create-react-app project in combination with `react-hook-form`. Everything works fine except building the project.

No error is thrown, the build simply freezes and is stalled. When running `npm run build --verbose` it just pretty much stops completely after the first initial logs that happen within approx. 20ms. 

It seems to be a problem with the, at this time, latest version v3.20.2. After downgrading Zod to v3.19.1, it builds normally.

xkcm:
I've been experiencing a similar performance issue.

TypeScript autocompletion is extremely slow in VS Code when `zod@3.20.2` package is installed, the CPU usage spikes to 100% and it usually takes 3-4 seconds to load the suggestions. After removing `zod` or downgrading it to `v3.19.1` the issue is gone. You can see more details in the related stackoverflow post I created: https://stackoverflow.com/questions/74881472/slow-typescript-autocompletion-in-vs-code-for-zod.

It's worth mentioning that this issue also occurs in Webstorm, so I'm convinced it's not VSCode-related.

bioxeed:
I'm experiencing similar and if I leave it long it enough it will error with
`error TS2589: Type instantiation is excessively deep and possibly infinite.`

kibertoad:
@colinhacks You are still unconvinced it should have been a semver major :)?

tornvallalexander:
It seems like I'm unable to reproduce the behaviour within a new basic implementation with Zod and React Hook Form. The project builds normally and runs as it should. The error is still persistent in the original project where I found the issue. Then again, project where I found it is more complex in terms of general structure for forms and input fields.

Matan18:
I had the same problem when I added de 3.20.0+ version (using nextjs + next-auth), I was using `coerce` to validate query params in a api request handler, I apparently solved the problem by upgrading my typescript version to 4.9.4.

stephank:
We have the same issue in a project. The 'possibly infinite' notice takes roughly 1 hour to appear in CI.

Our project was pinned to TypeScript 4.8.2. This seems to be fixed in TypeScript 4.8.3 and 4.9.4, so perhaps try upgrading that if you're still seeing this issue.

emilbon99:
[...]

PR #2107, opened 2023-02-27
Refactoring of ZodFormattedError type to improve tsc check time

gydroperit:
This small refactoring of ZodFormattedError fix several  performance issues in some edge cases.

I did some benchmarks with simple case with typescript compiler option **strict: false**   https://github.com/gydroperit/zod-test-regression

1. Now amount of generated types three times less than before, while  building typescript project, thus almost halves **Check time**

Without fix

```Files:              72
Lines:           30507
Identifiers:     49830
Symbols:         47942
Types:           17794
Instantiations:  29270
Memory used:    77234K
I/O read:        0.00s
I/O write:       0.00s
Parse time:      0.31s
Bind time:       0.13s
Check time:      0.31s
Emit time:       0.01s
Total time:      0.75s
```
with this fix
```
Files:              72
Lines:           30508
Identifiers:     49824
Symbols:         44537
Types:            5659
Instantiations:  17489
Memory used:    70989K
I/O read:        0.00s
I/O write:       0.00s
Parse time:      0.31s
Bind time:       0.12s
Check time:      0.20s
Emit time:       0.01s
Total time:      0.63s
```

2. Second issue is maybe more  important than 15 ms of build time. On some reason VS Code on latest  version of zod has very laggy intellisense autocompletion(about 1 to 3 seconds to show autocomplete list) and slow ESLint check time in source files with zod used(About 8 seconds in my case). Even this strict option set to true.
After fix - it will run smooth like in previous version,

This problem also described in https://stackoverflow.com/questions/74881472/slow-typescript-autocompletion-in-vs-code-for-zod
https://github.com/colinhacks/zod/issues/1741
https://github.com/microsoft/TypeScript/issues/45824

### DIFF
--- a/deno/lib/ZodError.ts
+++ b/deno/lib/ZodError.ts
@@ -170,15 +170,17 @@ export const quotelessJson = (obj: any) => {
   return json.replace(/"([^"]+)":/g, "$1:");
 };
 
+type recursiveZodFormattedError<T> = T extends [any, ...any[]]
+  ? { [K in keyof T]?: ZodFormattedError<T[K]> }
+  : T extends any[]
+  ? { [k: number]: ZodFormattedError<T[number]> }
+  : T extends object
+  ? { [K in keyof T]?: ZodFormattedError<T[K]> }
+  : unknown;
+
 export type ZodFormattedError<T, U = string> = {
   _errors: U[];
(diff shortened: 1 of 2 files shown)


colinhacks:
Makes sense, looks good. Thanks!

PR #2126, opened 2023-03-02
Revert to original input data if coerced date results in invalid date.

Pingviinituutti:
Hi!

So this started with the idea that it would be nice to now what the value of an invalid date was, e.g.,

``` typescript
const d = z.coerce.date();

const customErrorMap: z.ZodErrorMap = (issue, ctx) => {
    if (issue.code === z.ZodIssueCode.invalid_date) {
        return { message: `Could not parse '${(ctx.data)}' to a valid date!` };
    }
    return { message: ctx.defaultError };
};

z.setErrorMap(customErrorMap);
```
When we then parse an invalid date, the expected outcome (personally) is
``` ts
d.parse('very invalid date') // -> 'Could not parse 'very invalid date' to a valid date!'
```
however, we actually get the response `Could not parse 'Invalid Date' to a valid date!`, because `coerce` modifies the value before the type check. 

This change would greatly help if, let's say, I have hundreds of dates with only some being invalid. In this case I could find much faster which entries were the ones that didn't validate successfully.

Last, I didn't manage yet to build and test this change. But if someone already sees some benefit or alternative suggestions, please comment.


JacobWeisenburger:
Could you do something like this instead of needing this PR?

```ts
const datelikeToDate = z.date().or( z.string().datetime() ).pipe( z.coerce.date() )

console.log( datelikeToDate.safeParse( new Date ).success ) // true
console.log( datelikeToDate.safeParse( '2023-03-03T19:21:34.169Z' ).success ) // true
console.log( datelikeToDate.safeParse( 'invalid date' ).success ) // false
```

Pingviinituutti:
Hmmm the problem isn't actually really in `coerce`, but more in parsing any invalid `date`. Maybe the test cases below demonstrate better what I'm seeking for. 

The issue is that the initial invalid date value passed to `parse` (or `safeParse`) isn't propagating to `ctx.data` when a custom `ZodErrorMap` is used. My proposed fix only makes this work when `coerce` is used, but it might be worth to take a look at how this should behave when invalid dates are parsed without `coerce` too?

``` ts
const customErrorMap: z.ZodErrorMap = (issue, ctx) => {
  if (issue.code === z.ZodIssueCode.invalid_date) {
    return { message: ctx.data };
  }
  return { message: ctx.defaultError };
};

test("invalid date in union string and date", () => {
  const schema = z.date().or( z.string().datetime());

  expect(typeof schema.parse('2023-03-03T19:21:34.169Z')).toEqual('string');

  try {
    schema.parse("very invalid date", { errorMap: customErrorMap });
  } catch (error) {
    expect(error).toBeInstanceOf(z.ZodError);
    if (error instanceof z.ZodError) {
      expect(error.flatten()).toEqual({
        formErrors: ["very invalid date"], // fails because received "Invalid datetime" instead of the value I tried to parse
        fieldErrors: {},
      });
    }
  }
  // a date in the format dd.mm.yyyy hh:mm is not parsable by new Date()
  try {
    schema.parse("31.12.2023 23:59", { errorMap: customErrorMap });
  } catch (error) {
    expect(error).toBeInstanceOf(z.ZodError);
    if (error instanceof z.ZodError) {
      expect(error.flatten()).toEqual({
        formErrors: ["31.12.2023 23:59"], // fails because got "Invalid datetime" instead of the value I tried to parse
        fieldErrors: {},
      });
    }
  }
})

test("invalid date in union string and date piped into date coercion", () => {
  const schema = z.date().or( z.string().datetime() ).pipe( z.coerce.date() );

  expect(schema.parse('2023-03-03T19:21:34.169Z')).toBeInstanceOf(Date);

  try {
    schema.parse("very invalid date", { errorMap: customErrorMap });
  } catch (error) {
    expect(error).toBeInstanceOf(z.ZodError);
    if (error instanceof z.ZodError) {
      expect(error.flatten()).toEqual({
        formErrors: ["very invalid date"], // fails because got "Invalid datetime", instead of the value I tried to parse
        fieldErrors: {},
      });
    }
  }
  // a date in the format dd.mm.yyyy hh:mm is not parsable by new Date()
  try {
    schema.parse("31.12.2023 23:59", { errorMap: customErrorMap });
  } catch (error) {
    expect(error).toBeInstanceOf(z.ZodError);
    if (error instanceof z.ZodError) {
      expect(error.flatten()).toEqual({
        formErrors: ["31.12.2023 23:59"], // fails because got "Invalid datetime" instead of the value I tried to parse
        fieldErrors: {},
      });
    }
  }
})
```

JacobWeisenburger:
The problem that I can see with your `errorMap` is that it assumes it will receive an error code of `invalid_date`. However, from my testing it gets `invalid_string` instead.
```ts
const datelikeToDate = z.date().or( z.string().datetime() )

const result = datelikeToDate.safeParse( 'invalid date' )
!result.success && console.log( result.error.issues.map( x => x.code ) )
// [ 'invalid_string' ]
```

So you need to change your `errorMap` to handle both cases.
```ts
const errorMap: z.ZodErrorMap = ( issue, ctx ) => {
    if (
        issue.code === z.ZodIssueCode.invalid_date
        || issue.code === z.ZodIssueCode.invalid_string
    ) {
        return { message: ctx.data }
    }
    return { message: ctx.defaultError }
}

const datelikeToDate = z.date().or( z.string().datetime() )

const result = datelikeToDate.safeParse( 'very invalid date', { errorMap } )
!result.success && console.log( result.error.format() )
// { _errors: [ 'very invalid date' ] }
```

If you found my answer satisfactory, please consider supporting me. Even a small amount is greatly appreciated. Thanks friend! 🙏
https://github.com/sponsors/JacobWeisenburger

JacobWeisenburger:
I think I see what you are saying now as shown in the example below.

```ts
const errorMap: z.ZodErrorMap = ( issue, ctx ) => {
    if ( issue.code === z.ZodIssueCode.invalid_date ) {
        return { message: ctx.data }
    }
    return { message: ctx.defaultError }
}

const toDate = z.coerce.date()

const result = toDate.safeParse( 'very invalid date', { errorMap } )
!result.success && console.log( result.error.issues[ 0 ].message )
// Invalid Date
// expected 'very invalid date'
```

Pingviinituutti:
> I think I see what you are saying now as shown in the example below.
> 
> ```ts
> const errorMap: z.ZodErrorMap = ( issue, ctx ) => {
>     if ( issue.code === z.ZodIssueCode.invalid_date ) {
>         return { message: ctx.data }
>     }
>     return { message: ctx.defaultError }
> }
> 
> const toDate = z.coerce.date()
> 
> const result = toDate.safeParse( 'very invalid date', { errorMap } )
> !result.success && console.log( result.error.issues[ 0 ].message )
> // Invalid Date
> // expected 'very invalid date'
> ```

Yeah the original value is not in the context even with a simple custom `ZodErrorMap`, e.g.

``` ts
const customErrorMap: z.ZodErrorMap = (_issue, ctx) => {
  return { message: ctx.data };
};
```

When I dig more into what's happening in the background, I can still find the original value pretty far away in the process. It's just not included when in the `issue` when it's pushed into `ctx.issues` at https://github.com/colinhacks/zod/blob/master/src/helpers/parseUtil.ts#L12. Maybe the `issue` instance could have one more property called `data`?

JacobWeisenburger:
Thanks so much for the sponsorship! 🙏

colinhacks:
The core thing is that Zod intentionally doesn't expose the data in the ZodError objects, because it's a security vulnerability. Many people requested this because they need the ability to print their logs (which may include ZodErrors) and don't want to leak user information.

There may be a way to add `data` as a non-enumerable property of ZodIssue though, which would make it available in error maps but wouldn't expose it in logs. I'm looking into it.

Pingviinituutti:
Okay, I did more testing and noticed I was wrong in saying that there is an issue when not using `coerce`. My custom error map was bad. Only the first test fails in the code below.

``` ts
const customErrorMap: z.ZodErrorMap = (_issue, ctx) => {
  return { message: ctx.data };
};

test("original value is available in custom ZodErrorMap for coerced date", () => {
  const stringDate = z.coerce.date();
  try {
    stringDate.parse('very invalid date', 
      { errorMap: customErrorMap }
    );
  } catch (err) {
    const zerr: z.ZodError = err as any;
    expect(zerr.issues.length).toEqual(1);
    expect(zerr.issues[0].message).toEqual(
      "very invalid date" // fails
    );
  }
});

test("parsed value is available after pipe with customZodErrorMap", () => {
  const stringDate = z.date().or( z.string().datetime() ).pipe( z.coerce.date() );
  try {
    stringDate.parse("very invalid date", 
      { errorMap: customErrorMap }
    );
  } catch (err) {
    const zerr: z.ZodError = err as any;
    expect(zerr.issues.length).toEqual(1);
    expect(zerr.issues[0].message).toEqual(
      "very invalid date" // correct
    );
  }
});

test("invalid date in union string and date", () => {
  const schema = z.date().or( z.string().datetime());
  try {
    schema.parse("very invalid date", { errorMap: customErrorMap });
  } catch (err) {
    const zerr: z.ZodError = err as any;
    expect(zerr.issues[0].message).toEqual(
      'very invalid date' // correct
    );
  }
});
```

Maybe these comments serve as documentation for whomever tries to have custom `ZodErrorMap`s with coerced dates. 

Thanks a bunch for helping out here! I'll modify this pull request to only contain tests and add a note into the documentation.

PR #2132, opened 2023-03-02
re-introduce `ZodBigInt.(un)safe`.

igalklebanov:
Re-introducing (un)safe checks @ `ZodBigInt` following [comment](https://github.com/colinhacks/zod/pull/1711#issuecomment-1445514486).

### DIFF
--- a/README.md
+++ b/README.md
@@ -745,6 +745,11 @@ z.bigint().negative(); // < 0n
 z.bigint().nonpositive(); // <= 0n
 
 z.bigint().multipleOf(5n); // Evenly divisible by 5n.
+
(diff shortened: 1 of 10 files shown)


colinhacks:
Sorry, I'm backtracking again. I think the right way to address the use case described in #1711 is:

```ts
z.bigint().transform(Number).pipe(z.number().safe())
```

Ultimately almost no one will use `z.bigint()` and I'm increasingly conservative when it comes to bundle size, so I'm hesitant to add relatively obscure features like this.

igalklebanov:
I totally understand. I'll close this one.

PR #2137, opened 2023-03-03
Cache the evaluation of ParseInputLazyPath.path() for a moderate perf improvement

jussisaurio:
Hi, thanks for the great library, I adore the API :)

I was looking for ways to make some unintrusive improvements to the performance, and noticed since a ParseInput's path may be evaluated multiple times, caching the evaluation result avoids some extra array allocations.

Around a 10% speedup on 'realworld' on my computer.

Dunno what happened with the rest of the diff, I don't think I touched anything or had "format on save" on...

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -599,11 +599,11 @@ z.string().regex(regex);
 z.string().startsWith(string);
 z.string().endsWith(string);
 z.string().trim(); // trim whitespace
-z.string().toLowerCase(); // toLowerCase
-z.string().toUpperCase(); // toLowerCase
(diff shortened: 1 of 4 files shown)


colinhacks:
Thanks! Simplified the implementation a bit.

jussisaurio:
Hi! The changed code is definitely simpler, but at least on my machine it wipes out the perf improvement almost completely. I am guessing that Array.prototype.concat is just fundamentally a lot slower than allocating an empty array and then pushing into it.

colinhacks:
Ah interesting. Guess getting rid of `.concat` matters more than the caching bit. Sorry, I should have re-run the benchmarks.  I'll revert.

PR #2144, opened 2023-03-05
Proposal: ZodObject.omitIfExists()

jussisaurio:
I found myself wanting to predefine a set of keys that would always be omitted from certain schemas if present. Example scenario:

```
const OMIT_AUTOGENERATED = {
  id: true,
  uuid: true,
  createdAt: true,
  updatedAt: true
}

const someSchema = someSchemaRaw.omit(OMIT_AUTOGENERATED);
const someSchema2 = someSchema2Raw.omit(OMIT_AUTOGENERATED);
```

If there are certain "Raw schemas" (using the above nomenclature) that don't have all the properties defined in OMIT_AUTOGENERATED, then I can't use it on them.

With something like ZodType.omitIfExists() one could blanket-exclude a set of keys from schemas without necessarily requiring them to be present in all of them.

### DIFF
--- a/deno/lib/__tests__/pickomit.test.ts
+++ b/deno/lib/__tests__/pickomit.test.ts
@@ -77,6 +77,28 @@ test("omit parse - fail", () => {
   expect(bad4).toThrow();
 });
 
+test("omitIfExists type inference", () => {
+  const nonameFish = fish.omitIfExists({ name: true, nonExistent: true });
+  type nonameFish = z.infer<typeof nonameFish>;
+  util.assertEqual<nonameFish, { age: number; nested: {} }>(true);
+});
+
+test("omitIfExists parse - success", () => {
+  const nonameFish = fish.omitIfExists({ name: true, nonExistent: true });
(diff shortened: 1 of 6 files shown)


colinhacks:
Hm this is a decent argument for reverting the behavior introduced in https://github.com/colinhacks/zod/pull/1564. I think I'm in favor of reverting, esp. since TypeScript itself doesn't care:

```ts
type asdf = Omit<{ a: string }, "b">;
// no error
```

Thoughts @igalklebanov?

igalklebanov:
Let's revert it on all mask methods. It hurts compiler performance.

jussisaurio:
I let this stew overnight in my head and I also think it's not great to have two subtly different method names. This is a small convenience thing for me so I don't feel strongly about reverting to `.omit()` original behavior or leaving it as is based on that alone, but if the current implementation indeed carries a compile-time perf penalty, then I'd also lean towards reverting tbh.

colinhacks:
I've reverted these changes in the latest patch release, so this should now work @jussisaurio

PR #2173, opened 2023-03-09
Add support coercion parse for Map and Set

itok01:
This change will add support coercion parse for Map and Set.

When use `JSON.parse(data)`, Map in data is parsed as Object.
If we parse with a schema containing `ZodMap`, it will fail.
The problem is solved by adding a function to parse Object as Map.

```ts
z.coerce.map(z.string(), z.string()).parse({ first: "foo", second: "bar" });
```

Also, we can't use a set in JSON.
That's why in JSON, it's common practice to treat Set as Array.
ZodSet doesn't support to parse Array.
The problem is solved by adding a function to parse Array as Set.

```ts
z.coerce.set(z.string()).parse(["first", "second"]);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -598,6 +598,8 @@ z.coerce.number(); // Number(input)
 z.coerce.boolean(); // Boolean(input)
 z.coerce.bigint(); // BigInt(input)
 z.coerce.date(); // new Date(input)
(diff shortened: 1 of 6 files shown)


itok01:
I'll close this PR because a similar PR was already closed.
https://github.com/colinhacks/zod/pull/1736

Issue #2168, opened 2023-03-08
`z.enum().extract()`, and `z.enum().exclude()` should take variadic args


JacobWeisenburger:
Before:
```ts
z.enum( [ 'foo', 'bar', 'baz' ] ).extract( [ 'foo', 'bar' ] )
```

After:
```ts
z.enum( [ 'foo', 'bar', 'baz' ] ).extract( 'foo', 'bar' )
```

I think this looks a lot cleaner, but I'm not sure of the implications for typescript. Any thoughts?

galmarz:
@JacobWeisenburger z.enum accepts second a optional param, which I think makes rest params problematic here.

JacobWeisenburger:
fair enough, perhaps just `extract` and `exclude` should take variadic args

galmarz:
Small change so I created the PR to show the code. Personally I don't just mind passing array, but its not a huge change to support variadic args aswell.

colinhacks:
I'd rather leave this as an array so we can support a parameters object down the road if necessary. Best to leave our options open for future API expansion.

PR #2174, opened 2023-03-09
Add variadic args overload to enum().extract(), and enum().exclude()

galmarz:
Created this PR to show how changes proposed in #2168 might look like in the code in my understanding. 
Added overload for `z.enum().extract` and `z.enum().exclude`, to also accept variadic args.
@JacobWeisenburger thoughts?


JacobWeisenburger:
looks good. But it will need to be approved by someone with the proper authority.

Issue #2192, opened 2023-03-13
refine() function gets called even when parser already failed within translate()/refine()/regex()


ctollerud:
This is very similar to this one:
https://github.com/colinhacks/zod/issues/2113

Happening in version 3.21.4

refine() function still runs even though a previous translate has already invalidated it.  It's getting passed some aribtrary object
instead of the string that it expects.
```typescript
const parseResult = z
	.unknown()
	.transform((x, ctx) => {
		if (typeof x === 'string') return x;

		console.log('I fail because input is a number (correct behavior)');
		ctx.addIssue({
			code: 'custom',
		});
		return z.NEVER;
	})
	.transform((x: string) => {
		console.log("I don't get executed (correct behavior since transform above invalidated the input)");
		return x;
	})
	.refine((x: string) => {
		//BUG
		console.log("I shouldn't get called but I do!!!!");
		console.log(`I should be a string but I am a ${typeof x}!!!`); //'object'
		console.log(x); // some sort of '{status:'aborted'} object (the value underlying z.NEVER?)
	})
	.safeParse(42);

console.log(`succeeded:  ${parseResult.success}`); // false (correct behavior)
```
EDIT:  looks like there are more scenarios that are causing this 'zombie refine' scenario.  Here's two more:
- A failure within a `refine` causing it:
```typescript
import * as z from 'zod';

const parseResult = z
	.string()
	.refine(x => false) // force anything/everything to fail
	.transform(x => {
		console.log("I don't get called"); // correct
		return x;
	})
	.refine((x: string) => {
		console.log("I shouldn't get called!!!!!!");
		console.log(typeof x); // string
		console.log(x); // '123'
	})
	.safeParse('123');

console.log(`succeeded:  ${parseResult.success}`); // false (correct behavior)
```
- a failure within a `regex` causing it:
```typescript
import * as z from 'zod';

const parseResult = z
	.string()
	.regex(/abc/) // fail because input doesn't match regex
	.transform(x => {
		console.log("I don't get called"); // correct
		return x;
	})
	.refine((x: string) => {
		console.log("I shouldn't get called!!!!!!");
		console.log(typeof x); // string
		console.log(x); // '123'
	})
	.safeParse('123');

console.log(`succeeded:  ${parseResult.success}`); // false (correct behavior)
```

These were pretty easy to stumble upon, so there are likely more out there as well.

Darkhogg:
@ctollerud That is actually very useful, thank you.

[...]

PR #2217, opened 2023-03-19
Fix refinement abort early behavior and update tests

poteboy:
## Overview

This PR fixes the refinement abort early behavior and updates the corresponding tests (Issue #2192). The previous behavior caused unexpected continuation of the refinement process in some cases even if a previous refinement failed.

## Changes

- Updated the refinement handling code to allow early termination when a refinement fails.
- Updated test cases to reflect the correct behavior and expectations.
- Renamed the test "no abort early on refinements" to "unreachable second refinement if the first one fails" for better clarity.

## Impact

With these changes, the library should now handle refinements more consistently and provide accurate validation results as expected. This also resolves Issue #2192.

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -4277,7 +4277,6 @@ export class ZodEffects<
           parent: ctx,
         });
         if (inner.status === "aborted") return INVALID;
-        console.log(ctx.common.issues.length);
         if (inner.status === "dirty" && ctx.common.issues.length > 0) {
           status.dirty();
           return inner;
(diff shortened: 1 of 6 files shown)


src/types.ts
```diff
@@ -4277,7 +4277,11 @@ export class ZodEffects<
           parent: ctx,
         });
         if (inner.status === "aborted") return INVALID;
-        if (inner.status === "dirty") status.dirty();
+        console.log(ctx.common.issues.length);
```
ctollerud: console.log ;)

deno/lib/types.ts
```diff
@@ -4277,7 +4277,11 @@ export class ZodEffects<
           parent: ctx,
         });
         if (inner.status === "aborted") return INVALID;
-        if (inner.status === "dirty") status.dirty();
+        console.log(ctx.common.issues.length);
```
ctollerud: console.log

PR #2221, opened 2023-03-20
Testing on Typescript 5.0

RobinTail:
Since TS 5.0 was recently released, I think it's worth it to confirm the general operation of `zod` running this version.

https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/


colinhacks:
Thank you1

PR #2224, opened 2023-03-21
Update the email regex to allow dashes in domain names when subdomains are present

TomasHubelbauer:
Hi, as of current, tom@test.test.com passes the email validation regex but tom@test.te-st.com doesn't.

This PR fixes that.

I see you already have #2157 going so maybe this PR is not all that needed. I will comment on that PR, too. But I had to figure this regex out to make it work for my usecase (where I will switch to using `.regex`) so I figured I'd share my finding anyway and if there's no interest in changing the regex at this point, no hard feelings.


shroudedcode:
Hey @colinhacks, would really appreciate it if this could be merged soon! 🙏

The current regex is failing for a lot of valid email addresses and this PR seems like an uncontroversial quick fix before any larger changes discussed in #2157.

PR #2225, opened 2023-03-21
fix ulid regex

r0ots:
Ulid regex is not restrictive enough and can capture longer strings than expected.

`[0-9A-HJKMNP-TV-Z]{26}`

With this regex, an ulid like`01ARZ3NDEKTSV4RRFFQ69G5FAVA` is considered valid.
This contains 27 chars, but at least 26 are validating the regex, so this returns true.

Adding `^` at the start and `$` at the end ensures that the regex only validates exact matching strings.


nathanchapman:
@colinhacks able to merge this soon?

nathanchapman:
https://github.com/colinhacks/zod/issues/2820 no joke 🥴

Issue #2191, opened 2023-03-13
New signature for Object.partial to exclude certain keys (opposite mask)


darkbasic:
```
partial<Mask extends {
        [k in keyof T]?: true;
    }>(mask: Mask): ZodObject<objectUtil.noNever<{
        [k in keyof T]: k extends keyof Mask ? ZodOptional<T[k]> : T[k];
    }>, UnknownKeys, Catchall>;
```

Right now it's possible to provide a mask to `Object.partial()` to apply it only to certain keys.
I would like to be able to also do the opposite: make all properties optional except the provided ones.
Basically the additional signature would have `Mask extends { [k in keyof T]?: false }` instead of `true`.
Obviously you can't mix true and false because it would make no sense.

darkbasic:
See https://github.com/colinhacks/zod/pull/2234

darkbasic:
Not sure, has this been implemented in Zod 4? @colinhacks

colinhacks:
All else equal, I prefer there to be one correct way to do things in Zod's API surface. Not likely to happen, I'm afraid.

Issue #2190, opened 2023-03-13
Object.nullable and Object.nonNullable


darkbasic:
`Object.partial()` is very handy to make properties optional, but sometimes you want to make properties nullable instead (for example when you make SQL unions) and there is no method to do so.

I'm achieving the same result with custom functions but it's very inconvenient:

```
export function toNullableOrOptionalShape(
  shape: ZodRawShape,
  type: "nullable" | "optional"
): ZodRawShape {
  return Object.entries(shape).reduce<ZodRawShape>((acc, [key, value]) => {
    acc[key] = value[type]();
    return acc;
  }, {});
}

export function toNullableObject<
  T extends ZodRawShape,
  UnknownKeys extends UnknownKeysParam = UnknownKeysParam,
  Catchall extends ZodTypeAny = ZodTypeAny
>(
  object: ZodObject<T>
): ZodObject<
  {
    [k in keyof T]: ZodNullable<T[k]>;
  },
  UnknownKeys,
  Catchall
> {
  return new ZodObject({
    ...object._def,
    shape: () => toNullableOrOptionalShape(object.shape, "nullable"),
  }) as any;
}
```

I would love to have an `Object.nullable()` method and I could PR it if you are interested in having such feature.

We also need the equivalent of `Object.required` (`Object.nonNullable`?) to mark fileds as non-nullable.

pleunv:
Also interested in this, have a similar use-case where the input schema is more permissive than the output schema (e.g. `initialValues` is inferred from an input schema and can have nullable values while submission data is using an output schema and _needs_ to have its values set). Would be handy if there was a way to base these schemas on one another.

darkbasic:
@pleunv you're in luck then because I've just implement this and much more: https://github.com/colinhacks/zod/pull/2234

ShravanSunder:
Would be great to have #2234

darkbasic:
> Would be great to have https://github.com/colinhacks/zod/pull/2234

Would be great to at least know @colinhacks opinion about that PR.
I'm currently already using that PR via yarn patches but if he doesn't envision ever merging it I might start looking for alternatives.
If it's just a matter of him still not having enough time to review it I might be fine with the patches though.

sezanzeb:
nonNullable should be equivalent to typescripts `NonNullable<>` type in my opinion. I have created a pull request for this: https://github.com/colinhacks/zod/pull/3141

colinhacks:
[...]

PR #2234, opened 2023-03-23
[partial, required]: implement falsy masks and optional/nullable/nullish flavors

darkbasic:
Closes https://github.com/colinhacks/zod/issues/2191 and https://github.com/colinhacks/zod/issues/2190

Previously a mask allowed you to to require/partialize only certain properties while leaving the others intact. I called such old behaviour _thruthy masks_ because you selected the properties assigning a `true` value to them.
I implemented the opposite, meaning that you can require/partialize all properties **except** certain ones. I called this new behaviour _falsy masks_ because you select the properties assigning a `false` value to them.
I've made care of making everything backwards compatible, with the small exception that you can't mix `true` and `false` values in the same mask anymore (previously `false` values were simply ignored) because that would be ambiguous: how should it behave, like a falsy or thruthy mask? I could theoretically treat it as a truthy mask and ignore falsy values but that would be a bad behaviour in my opinion because it wouldn't be intuitive. Also the breaking change is small enough because I see no valid reason why anybody would be mixing true and false values anyway.
Since we're at it I'm also forbidding empty masks (previously `{}` was a valid mask) because again it's not intuitive whether you should treat it as a truthy or falsy mask (I think that particular fix ended up in the second commit, but since I'm upstreaming both of them it shouldn't matter. I can rebase it into the first one if you prefer).

The second feature I'm upstreaming are optional/nullable/nullish flavors for partial and required. That depends on the previous commit because it implements such behaviour for both truthy and falsy masks. Previously partial transformed a field of type `T` into `T | undefined` while required transformed a field of type `T | undefined` into `T`. Since along with the optional method we also have nullable and nullish methods it makes sense to have nullable and nullish flavors of partial and required as well. Naming starts to get messy though, so I've decided to not add any more methods but instead select the "flavor" via parameters. I've made sure to keep everything 100% backwards compatible so the default is still the optional flavor and the number and the position of the parameters changes depending on the overloaded function: parameters get processed and mask and type get reassembled in the function implementation so this is completely transparent to the end user.
A nullable flavor of required will strip out ZodNullable instead of ZodOptional (`T | null` becomes `T`) while the nullish flavor will 
strip out both ZodNullable and ZodOptional (`T | undefined | null` becomes `T`).
Similarly a nullable flavor of partial will wrap with ZodNullable instead of ZodOptional (`T` becomes `T | null`) while the nullish flavor will wrap with both ZodNullable and ZodOptional (`T` becomes `T | undefined | null`).

While bad behaviours are already enforced via typings (not mixing falsy and truthy masks, not passing empty masks which are ambiguous and cannot be discerned to a specific mask type) I also enforce these checks via type assertions (`assertMaskIsNotMixed`). While special care has been taken to optimize such function, a small (IMO negligible) performance impact is inevitable. If you prefer we can get rid of it and assume all our users will use TypeScript, but in my opinion it's better to keep it. I've also been careful to compute conditions outside of for loops to achieve some small performance improvements whenever possible, so I wouldn't worry about that.

I might have been over zealous  with the tests and they are probably too much and too many, but I didn't want you to get mad about testing every single possible corner case so the more the better. Feel free to cut them down as you see fit.

I still didn't update the documentation because I was looking for some feedback first. You didn't answer to my feature requests yet so I hope this will be enough to get your attention because I'd really love to have these features merged ASAP.
I'm currently working on a zod types generator so I will hopefully be able to test this on a larger project of my own soon. Tests are probably enough to cover every possible corner case, but the more testing the better.

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -623,7 +623,7 @@ z.coerce.boolean().parse(null); // => false
 
 ## Literals
 
-Literal schemas represent a [literal type](https://www.typescriptlang.org/docs/handbook/literal-types.html), like `"hello world"` or `5`.
+Literal schemas represent a [literal type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types), like `"hello world"` or `5`.
 
 ```ts
 const tuna = z.literal("tuna");
--- a/deno/lib/__tests__/partials.test.ts
+++ b/deno/lib/__tests__/partials.test.ts
@@ -5,7 +5,7 @@ const test = Deno.test;
 import { util } from "../helpers/util.ts";
 import * as z from "../index.ts";
 import { ZodNullable, ZodOptional } from "../index.ts";
-import { MaskErrorType } from "../types.ts";
+import { MaskErrorType, PartialType } from "../types.ts";
 
 const nested = z.object({
   name: z.string(),
@@ -171,6 +171,62 @@ test("required inference", () => {
   util.assertEqual<expected, required>(true);
 });
 
+test("required without mask and with nullable type", () => {
+  const object = z.object({
+    name: z.string(),
+    age: z.number().optional(),
+    field: z.string().optional().default("asdf"),
+    nullableField: z.number().nullable(),
+    nullishField: z.string().nullish(),
+  });
+
+  const requiredObject = object.required(PartialType.NULLABLE);
+
+  type required = z.infer<typeof requiredObject>;
+  type expected = {
+    name: string;
+    age?: number;
+    field: string;
+    nullableField: number;
+    nullishField?: string;
+  };
+  util.assertEqual<expected, required>(true);
+
+  expect(requiredObject.shape.name).toBeInstanceOf(z.ZodString);
+  expect(requiredObject.shape.age).toBeInstanceOf(z.ZodOptional);
+  expect(requiredObject.shape.field).toBeInstanceOf(z.ZodDefault);
+  expect(requiredObject.shape.nullableField).toBeInstanceOf(z.ZodNumber);
+  expect(requiredObject.shape.nullishField).toBeInstanceOf(z.ZodOptional);
+});
+
+test("required without mask and with nullish type", () => {
+  const object = z.object({
+    name: z.string(),
+    age: z.number().optional(),
+    field: z.string().optional().default("asdf"),
+    nullableField: z.number().nullable(),
+    nullishField: z.string().nullish(),
+  });
+
+  const requiredObject = object.required(PartialType.NULLISH);
+
+  type required = z.infer<typeof requiredObject>;
+  type expected = {
+    name: string;
+    age: number;
+    field: string;
+    nullableField: number;
+    nullishField: string;
+  };
+  util.assertEqual<expected, required>(true);
+
+  expect(requiredObject.shape.name).toBeInstanceOf(z.ZodString);
+  expect(requiredObject.shape.age).toBeInstanceOf(z.ZodNumber);
+  expect(requiredObject.shape.field).toBeInstanceOf(z.ZodDefault);
+  expect(requiredObject.shape.nullableField).toBeInstanceOf(z.ZodNumber);
+  expect(requiredObject.shape.nullishField).toBeInstanceOf(z.ZodString);
+});
+
 test("required with emtpy mask -- throws", () => {
   const object = z.object({
     name: z.string(),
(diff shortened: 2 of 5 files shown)


darkbasic:
I've manually updated my projects' zod types to make use of the methods introduced by this PR (instead of my custom helpers) and the resulting types still match with the database schema.
By the way we will probably want to rename `assertMaskIsNotMixed` to `assertMaskIsValid` and `PartialType` to `FilterType` or something else (because it's being used by both partial and required).

t-animal:
This is really extremely useful, especially in cases where a backend doesn't support (or doesn't want to) serializing non-existing (null) values to undefined (missing in json). I'd love to see this merged soon, because it would allow us to get rid of some ugly code.

darkbasic:
> because it would allow us to get rid of some ugly code

I know the feeling, in fact I have more things I would love to get merged but I'm waiting for this before pouring more time into tests/etc. With this PR, the automatic zod types generator that I've just finished and the remaining features which I'm currently shipping in my own zod fork I've basically cut down several hundreds lines of code into just a few dozens, plus I've uncovered several bugs. One of these additional features is built-in support for SQL UNIONs: whenever you need to UNION two different tables everything but the intersection must be nullable and it is extremely annoying and error prone having to do so manually.

darkbasic:
Not stale, I'm still waiting for any kind of feedback from the maintainer.

osdiab:
ping @colinhacks

nicktrn:
This is great stuff! Hope this gets reviewed soon.

If it isn't, it could be solely down to that wall of text in the initial post. Headings, bullet points, and code blocks go a long way - particularly for complex issues. Would be a shame to see this go to waste.

colinhacks:
I've had a tremendous amount of trouble deciding what to do with this. For starters, it's a really well-made PR and it implements features that I think are useful. But there are some things that are giving me pause which I'll outline below. All told though, I'm going to close this, but I *really* appreciate the great work here @darkbasic, and the patience from everyone on this thread.

### Modes

I don't like the idea of "modes" like `PartialType` generally, and usually I can think of an API that avoids them. (More on that later.)

### Bundle size

I was hesitant to add this in v3 because it's a lot of code for a feature that isn't widely requested. Zod 4 will be redesigned to be more tree-shakable so this becomes a little more okay, but since most users will use `ZodObject`, these methods will end up in their bundle, even if unused.

### What is "optional" anyway?

Zod 4 will make a distinction between different types of "optional". There's the "question mark" flavor (`{name?: string}`) and there's the "or-undefined" flavor (`{name: string | undefined }`). In Zod 4 you'll be able to specify one or the other  flavor (or both) when defining object schemas. Any API that deals with partial/required needs to be able to handle that distinction. It's tricky.

Here's a rough initial proposal for the set of "partial" APIs. All of these optionally accept (truthy) masks.

```ts
const schema = z.object({ name: z.string() });

schema.partial();  // { name?: string | undefined } 
schema.partial.except({ ... }) // equivalent of falsy masks

schema.allow.empty(); // { name?: string }
schema.allow.undefined(); // { name: string | undefined }
schema.allow.null(); // { name: string | null }
schema.allow.nullish(); // { name: string | null | undefined }
```

The proposed set of "required" APIs. All of these optionally accept (truthy) masks.

```ts
const schema = z.object({ name: z.string().optional().nullable().or(z.undefined()) });
// { name?: string | undefined | null }

schema.required();  // { name: string } 
schema.required.except({ ... }) // equivalent of falsy mask

// mask supported but optional
schema.not.empty(); // { name: string | undefined }
schema.not.undefined(); // { name?: string | null }
schema.not.null(); // { name: string | undefined }
schema.not.nullish(); // { name?: string }
```

Optionality is complicated. Whatever ends up landing in Zod 4 will be quite a bit different from this PR. I don't feel that good about my proposal either. I'll keep this all this in mind as I rethink things for Zod 4, maybe something more elegant will materialize.

Thanks again @darkbasic! And feel free to chime in with thoughts.

darkbasic:
Thanks for reviewing my PR, even if you didn't end up accepting my changes in the end.

Unfortunately I don't have the time right now to discuss eventual API changes, but I'd like to add my two cents about its utility.

> I was hesitant to add this in v3 because it's a lot of code for a feature that isn't widely requested

I suspect that this feature is of particularly high value to library/tools developers rather than end users.
In particular as a mikro-orm collaborator I've written a typings generator that creates a zod schema from the ORM entity definitions, with the goal of generating slonik types via the ORM metadata to manually write composable SQL queries that are type safe at runtime and build time:
https://github.com/mikro-orm/mikro-orm/discussions/5102#discussioncomment-8076746
https://the-guild.dev/blog/accounts.js-1.0-rc#about-me

It relies on this feature for which I'm currently maintaining my own zod fork, which is the main reason why I haven't released it yet.

That being said, even if its main utility would probably be limited to libraries/tooling developers, I can still see 11 hearts just in this PR, which is massive wall of text that would scare 90% of the users just by the sheer size of it.

Issue #2227, opened 2023-03-21
Proto pollution vulnerability when using `z.record(z.object())` 


marcus13371337:
Hello,

I've found a minor security vulnerability when using `z.record(z.object())` with proto pollution. Consider the following example:
```tsx
  const schema = z.record(
    z.object({
      b: z.string(),
    })
  );

  const data = JSON.parse(`{
    "c": {
      "b": "world"
    },
  "__proto__": {
    "b": "world2"
  }
}`);

  const result = schema.parse(data);

  console.log(result.b); // world2
```

I would assume `result.b` should be `undefined` after the parsing, but instead the data in `__proto__` is polluted to the object, i.e accessing `result.b` gives me `world2`

marcus13371337:
Fixed in https://github.com/colinhacks/zod/pull/2239

PR #2239, opened 2023-03-24
Fix proto pollution vulnerability

marcus13371337:
Fixes the minor vulnerability described [here](https://github.com/colinhacks/zod/issues/2227)


colinhacks:
Thank you! Very interesting bug.

Issue #2106, opened 2023-02-27
Deprecating `z.discriminatedUnion`?


colinhacks:
Superceded by https://github.com/colinhacks/zod/issues/3407

---

I'm planning to deprecate `z.discriminatedUnion` in favor of a "switch" API that's cleaner and more generalizable. You can dynamically "switch" between multiple schemas at parse-time based on the input.

```ts
const schema = z.switch(()=>{
  return Math.random() ? z.string() : z.number()
});
```

I expand more on the `z.switch` API later. Let's talk about `z.discriminatedUnion`.

## Why

 `z.union` naively tries each union element until parsing succeeds. That's slow and bad. Zod needed some solution.

`z.discriminatedUnion` was a mistake. The API was good but I had reservations about the implementation. It required fiddly recursive logic to exrtract a literal discriminator key from each union element.

<img width="584" alt="Screenshot 2023-02-27 at 2 07 02 AM" src="https://user-images.githubusercontent.com/3084745/221534750-4652a10c-0c6d-4c62-a06b-918ca112c250.png">

It's a bad sign when a method or class requires weird recursive traversal of other schemas. For starters, Zod is designed to be *subclassable*. Users can theoretically subclass `ZodType` to implement custom schema types. But logic like this `instanceof` switch statement don't and can't account for any user-land schema types. 

But the main problem is just that this kind of pattern is bad and introduces a lot of edge cases. It means that only certain kinds of schemas are allowed as discriminators, and others will fail in unexpected ways. There are now dozens of issues that have been opened regarding these various edge cases. The PRs attempting to solve this problem are irredeemably complex and introduce even more edge cases.

- https://github.com/colinhacks/zod/issues/1773
- https://github.com/colinhacks/zod/issues/1667
- https://github.com/colinhacks/zod/issues/1384
- https://github.com/colinhacks/zod/issues/1757
- https://github.com/colinhacks/zod/issues/1504
- https://github.com/colinhacks/zod/issues/1136
- https://github.com/colinhacks/zod/issues/1477
- https://github.com/colinhacks/zod/issues/1424
- https://github.com/colinhacks/zod/pull/1213
- https://github.com/colinhacks/zod/issues/1490
- https://github.com/colinhacks/zod/issues/1181
- https://github.com/colinhacks/zod/issues/1075
- https://github.com/colinhacks/zod/issues/1444
- https://github.com/colinhacks/zod/issues/1868
- https://github.com/colinhacks/zod/issues/1878

[...]

PR #2287, opened 2023-04-03
Fix branded Record keys in ZodRecord

googol:
Because of a quirk of the extends syntax of typescript conditional
types, the previous version of this check for a branded key didn't work
as expected.

Thank you @akomm for the explanation of why the extends sides need to be
swapped at https://github.com/colinhacks/zod/pull/2287#issuecomment-1689922672

Originally I intended to remove the RecordType type entirely, since it
would seem initially that it is only there to work around earlier
versions of typescript not having the `noUncheckedIndexedAccess` option.
However, it is really meant to work around a problem with using ZodEnum
as the key, see this comment: https://github.com/colinhacks/zod/pull/2287#issuecomment-1676624932

~The RecordType conditional type seems to exist for handling the partialness of indexed objects. However that is handled by the tsconfig option noUncheckedIndexedAccess in TS@4.1 and later.~

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -721,7 +721,7 @@ z.string().regex(regex);
 z.string().includes(string);
 z.string().startsWith(string);
 z.string().endsWith(string);
-z.string().datetime(); // defaults to UTC, see below for options
+z.string().datetime(); // ISO 8601; default is without UTC offset, see below for options
 z.string().ip(); // defaults to IPv4 and IPv6, see below for options
 
 // transformations
@@ -760,7 +760,7 @@ z.string().ip({ message: "Invalid IP address" });
 
 ### ISO datetimes
 
-The `z.string().datetime()` method defaults to UTC validation: no timezone offsets with arbitrary sub-second decimal precision.
+The `z.string().datetime()` method enforces ISO 8601; default is no timezone offsets and arbitrary sub-second decimal precision.
 
 ```ts
 const datetime = z.string().datetime();
@@ -2825,10 +2825,9 @@ This more declarative API makes schema definitions vastly more concise.
 
 [https://github.com/pelotom/runtypes](https://github.com/pelotom/runtypes)
 
-Good type inference support. They DO support readonly types, which Zod does not.
+Good type inference support.
 
 - Supports "pattern matching": computed properties that distribute over unions
-- Supports readonly types
 - Missing object methods: (deepPartial, merge)
 - Missing nonempty arrays with proper typing (`[T, ...T[]]`)
 - Missing promise schemas
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -3340,7 +3340,7 @@ export type RecordType<K extends string | number | symbol, V> = [
   ? Record<K, V>
   : [symbol] extends [K]
   ? Record<K, V>
-  : [BRAND<string | number | symbol>] extends [K]
+  : [K] extends [BRAND<string | number | symbol>]
   ? Record<K, V>
   : Partial<Record<K, V>>;
 export class ZodRecord<
--- a/src/types.ts
+++ b/src/types.ts
@@ -3340,7 +3340,7 @@ export type RecordType<K extends string | number | symbol, V> = [
   ? Record<K, V>
   : [symbol] extends [K]
   ? Record<K, V>
-  : [BRAND<string | number | symbol>] extends [K]
+  : [K] extends [BRAND<string | number | symbol>]
   ? Record<K, V>
   : Partial<Record<K, V>>;
 export class ZodRecord<


sangxxh (approved):
Awesome thank you! This will bring back the TypeScript compile time safety that we're desperately missing 🙂

googol:
Many thanks for approving this! Super happy to get this annoyance fixed :)

googol:
Just to clarify for myself, when can I expect this to get merged and published?

googol:
@sangxxh Hey! Could we get this merged soon?

colinhacks:
Unfortunately this is currently intentional. We need the `Partial` here because Zod doesn't do exhaustiveness checking across the values when you pass a `ZodEnum` as the key schema.  in as the `keySchema` for a ZodRecord, Zod  

```ts
const schema = z.record(z.enum(['a', 'b']), z.number());
type schema = z.infer<typeof schema>;

schema.parse({ a: 5 }); // passes
var x: schema = { a: 5 } // missing key 'b'
```

I agree this isn't ideal, and this will likely change in Zod v4.

googol:
Ahh I see, I guess I never figured out what the original reason for this weird type is. The problem I'm having with it is that it's messing up records that are keyed by a branded string. These cases are not meant to have partial applied to them, right?

[This line](https://github.com/colinhacks/zod/pull/2287/files#diff-52632a4861fc9d7dc2dacef13cd91d60286dd706c1bb57438b8ee6a579a8796aL3326) probably was an earlier attempt at fixing this, however it doesn't work for me. Do you have an idea on how to fix this?

```ts
const schema = z.record(z.string().brand('some-branded-string'), z.number())
type schema = z.infer<typeof schema> // expected Record<string & BRAND<'some-branded-string'>, number>, got Partial<Record<string & BRAND<'some-branded-string'>, number>>
```

googol:
Thanks for your comment @akomm!  That makes a lot of sense as to why the earlier merged fix didn't really fix the problem.

I've included your suggested solution in a commit on this branch now, to see how it passes the test suite.

I'll do testing on my codebase tomorrow (UTC+3) with this idea, and let you all know

googol:
I've verified that the change suggested by @akomm does fix the problem we're having in our codebase, so I've amended the branch to fix that one check only, so that this won't affect the case with ZodEnum

@colinhacks would this be ok to merge now

Stefandasbach:
What's the update here, @colinhacks?

akomm:
Is there something besides "no time" blocking this? Some doubts or issues? Maybe we can help.

googol:
ping @colinhacks. is there anything I could do to help this get merged?

dkrieger:
@googol do you have a sense of how a `Record<someStringLiteralUnion, boolean>` could be enforced via the current state of zod? ran into this issue just now and a bit discouraged that this has been stale for so long. asking both to anticipate potential objections from @colinhacks and to see if you've found a better short-term alternative to manually defining this schema as an object

googol:
Unfortunately I have nothing :/

fwoelffel:
Having partial records when the keys are a literal union is by design. It reflects the way zod parses the records : it does not make sure that every value of the union is represented in the record's keys.

colinhacks:
There's clearly a lot of confusion about why ZodRecord is the way it is. The comment by @fwoelffel is exactly right:

> Having partial records when the keys are a literal union is by design. It reflects the way zod parses the records : it does not make sure that every value of the union is represented in the record's keys.

It's impossible (in the general case) for Zod to take an arbitrary schema and make a list of literals that will pass validation. As such, it isn't possible for Zod to do exhaustiveness checking on the keys of a given input object, to make sure, say, all the elements of a ZodUnion appear in the object. That's why Zod makes all fields optional for literal keys.

> My argument for this approach is that it would mimic 1:1 TS type safety level to runtime safety level in zod

Exactly, this is the goal. For maximum type safety, Zod should add `| undefined` when there are non-finite keys in `z.record()`. In practice most people don't currently have `noUncheckedIndexedAccess` defined and found this behavior surprising before the `RecordType` change. 

It's important to note that Zod's isn't able to alter its runtime behavior based on your `tsconfig.json`. Your tsconfig doesn't exist once your code is bundled. So Zod needs to make a decision about what the inferred type is, and its runtime behavior should agree with the inferred type. This is also why it's not easy for Zod and `z.object()` to magically support `exactOptionalPropertyTypes`.

Let's talk about improvements I'd like to make to `ZodRecord`.

### `noUncheckedIndexedAccess` 

As @akomm has suggested, Zod can check when `noUncheckedIndexedAccess` is enabled (in the static domain, not runtime) and add `| undefined` to *all* inferred ZodRecord types *including* non-finite keys. I'm in favor of that.

Practically speaking most people don't have `noUncheckedIndexedAccess` enabled currently, so I still stand behind the decision to introduce `RecordType` and remove optionality from non-finite keys. But per @akomm's suggestion the optionality should still remain for users with `noUncheckedIndexedAccess: true`.


After some tinkering I found this can be done statically with the type alias below. If someone finds a less hacky way let me know.

```ts
declare const arg: {[k:string]: string};
type noUncheckedIndexedAccess = (typeof arg.qwerty) extends string ? false : true;
```

This is a breaking change I'd land it in Zod 4 because it's potentially breaking.

### Improved exhaustiveness checking 

I'm not very enthusiastic about this, and it's very nontrivial. It's the same difficulty that exists with ZodDiscriminatedUnion, which I described here: https://github.com/colinhacks/zod/issues/2106 

The problem boils down to this: given an arbitrary Zod schema, determine the set of literal values that exist that will pass validation. You need to do this in both the runtime and static domains. 

In the static domain, this requires a recursive type alias. See `partialUtil.DeepPartial` for an example of what it looks like. It's *very* annoying and error-prone to unwrap all the `ZodOptional`, `ZodNullable`, `ZodEffects`, `ZodReadonly`, `ZodUnion`, `ZodDiscriminatedUnion`, instances that may be wrapping your `ZodLiteral`, `ZodEnum`, and `ZodNativeEnum` instances. 

Same problem at runtime, this requires a recursive visitor pattern. Out of the gate, there's no way to make this compatible with any user-defined `ZodType` subclasses, since the logic would necessarily be strongly-coupled to the set of first-party Zod-defined subclasses. 

Plus there are schemas for which this problem is literally unsolvable like `z.literal("asdf").transform(val => val.toUpperCase() as "ASDF")`. There's no way for Zod to determine at runtime that `"ASDF"` is a potentially valid input, so Zod can't correctly due exhaustiveness checking. 

I could add in some custom logic to special-case the major use cases (passing a ZodEnum in as the key to z.record) but a half-baked solution like that is just a recipe for maintenance headaches and user frustration. Look at all the issues surrounding ZodDiscriminatedUnion to see that up close. 

I will experiment with this regardless and see if I can land on something that addresses major use cases without undue complexity.

### Broken overload

Turns out there's a bug that allows you to pass in an arbitrary schema as the key schema into `z.record()`...whoops. 

```ts
const schema = z.record(z.object({ name: z.number() }), z.number());
```

This should enforcing that the first argument conforms to `type KeySchema = ZodType<string | number | symbol, any, any>;` 

In order to fix it I need to remove the "single argument" shorthand: `z.record(z.number)`. I'll do this in Zod 4 as well.

Issue #2285, opened 2023-04-01
New feature: z.date().utc()


thelinuxlich:
Because of the differences when the date field goes from backend(probably UTC) to client(local timezone) it's easy to get a date(YYYY-mm-dd) 1 day behind/forward and have to deal with it manually, specially when using form inputs with type=date.

Therefore I propose that this transform:

```ts
z.date().transform((val) => convertDateToUTC(val));

function convertDateToUTC(date: Date) {
	return new Date(
		date.getUTCFullYear(),
		date.getUTCMonth(),
		date.getUTCDate(),
		date.getUTCHours(),
		date.getUTCMinutes(),
		date.getUTCSeconds()
	);
}
```

Becomes this core function:
```ts
z.date().utc()
```

JacobWeisenburger:
Good idea.

If this gets added to zod core, I'm pretty sure it should also include the milliseconds.

```ts
const convertDateToUTC = ( date: Date ) => new Date(
    date.getUTCFullYear(),
    date.getUTCMonth(),
    date.getUTCDate(),
    date.getUTCHours(),
    date.getUTCMinutes(),
    date.getUTCSeconds(),
    date.getUTCMilliseconds(), // should also include
)

z.date().transform( convertDateToUTC )
```

thelinuxlich:
Agree!

joaoGabriel55:
I can work on that @JacobWeisenburger and @thelinuxlich

thelinuxlich:
Nice catch, to mantain the date, convertDateToUTC should be changed to:

```ts
const convertDateToUTC = ( date: Date ) => new Date(Date.UTC(
    date.getUTCFullYear(),
    date.getUTCMonth(),
    date.getUTCDate(),
    date.getUTCHours(),
    date.getUTCMinutes(),
    date.getUTCSeconds(),
    date.getUTCMilliseconds(), // should also include
))
```

joaoGabriel55:
Cool guys! Thanks for the tips. I think that today I will start to work on that!

JacobWeisenburger:
Now this seems to NOT solve the original problem.

```ts
const convertDateToUTC = ( date: Date ) => new Date(
    Date.UTC(
        date.getUTCFullYear(),
        date.getUTCMonth(),
        date.getUTCDate(),
        date.getUTCHours(),
        date.getUTCMinutes(),
        date.getUTCSeconds(),
        date.getUTCMilliseconds(),
    )
)

[...]

PR #2294, opened 2023-04-04
[Issue #2285] New feature: z.date().utc()

joaoGabriel55:
Closes: #2285

**Overview**

Because of the differences when the date field goes from backend(probably UTC) to client(local timezone) it's easy to get a date(YYYY-mm-dd) 1 day behind/forward and have to deal with it manually, specially when using form inputs with type=date.

**What was made:**
- New Code implemented
- Added new test case on `src/__tests__/date.test.ts`

Thanks for helped me @ramos-ph

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -1693,6 +1693,38 @@ export class ZodDate extends ZodType<Date, ZodDateDef> {
     });
   }
 
+  utc() {
+    const convertDateToUTC = (date: Date) => {
+      const utc = new Date(
+        Date.UTC(
+          date.getUTCFullYear(),
+          date.getUTCMonth(),
+          date.getUTCDate(),
+          date.getUTCHours(),
+          date.getUTCMinutes(),
+          date.getUTCSeconds(),
+          date.getUTCMilliseconds()
+        )
+      );
+
+      utc.toString = function () {
+        const weekDay = utc.toLocaleString("default", { weekday: "short" });
+        const monthName = utc.toLocaleString("default", { month: "short" });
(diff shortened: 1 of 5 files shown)


src/__tests__/date.test.ts
```diff
@@ -32,3 +32,11 @@ test("min max getters", () => {
     beforeBenchmarkDate
   );
 });
+
+test("UTC parser", () => {
+  const schemaUTCDate = z.date().utc();
+
+  const date = new Date("2023-04-04");
+
+  expect(schemaUTCDate.parse(date).toString()).toEqual('Mon Apr 04 2023 00:00:00 GMT+0000 (Coordinated Universal Time)');
```
ramos-ph, line 41: I think this spec is decent for `toString`, but perhaps we can go a little further by testing the resulting `Date` object itself

joaoGabriel55: Yeah, I think that we can try another `Date` object scenarios. Like:
`new Date(2023, 4, 3)`
`new Date('2023/04/04')`
`new Date('04/04/2023')`

deno/lib/types.ts
```diff
@@ -1693,6 +1693,38 @@ export class ZodDate extends ZodType<Date, ZodDateDef> {
...
+          date.getUTCMonth(),
+          date.getUTCDate(),
+          date.getUTCHours(),
+          date.getUTCMinutes(),
+          date.getUTCSeconds(),
+          date.getUTCMilliseconds()
+        )
+      );
+
+      utc.toString = function () {
```
igalklebanov, line 1710: Oh boy, this one's a stinker.

PR #2309, opened 2023-04-07
Lowercase/Uppercase Validators

### LINKED ISSUE #2300: `.toLowerCase/UpperCase()` should infer to `Lowercase/Uppercase<string>`
The [Lowercase](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html#lowercasestringtype) and [Uppercase](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html#uppercasestringtype) types are relatively new to TS and already have [analogs in Zod](https://github.com/colinhacks/zod#strings).

When I do this:
```ts
const MySchema = z.string().toLowerCase();
type MyType = z.infer<typeof MySchema>;
```
I should get this:
```ts
type MyType = Lowercase<string>
```
In zod v3.21.4, I get this:
```ts
type MyType = string
```

brianblakely:
This change achieves two objectives:

1. It validates strings for being lowercase or uppercase by adding the `lowercase()` and `uppercase()` validators to `ZodString`.
2. If these validators are used, the resulting inference will use the appropriate utility type, `Lowercase<string>` or `Uppercase<string>` (fixes #2300 but not in the way that issue suggests). It also enables utility types to be used for other string validators, should that ever be desired.


brianblakely:
This PR was created naively of #2307, which also implements these validators. I am leaving this open because that PR does not close issue #2300 by also inferring the utility types.

igalklebanov:
Closed my PR. 

I don't feel strongly about the validation aspect of this anymore.
Types didn't really work, kept returning `string` after `.parse` - you should add type assertion, we did the same thing and I suspect it doesn't work in this PR either atm.

brianblakely:
> Closed my PR.
> 
> I don't feel strongly about the validation aspect of this anymore. Types didn't really work, kept returning `string` (instead of `Lowercase<string>`) after `.parse` - you should add type assertions in your tests, we implemented pretty much the same and I suspect it doesn't work in this PR either atm.

Thank you for your input, Igal. I used your method of validation instead of a regex, because it's faster (tested in V8). I can confirm that `.parse` resolves to a `Lowercase/Uppercase<string>`.

brianblakely:
Resolved a minor conflict with master.

brianblakely:
Is there interest in merging this feature @colinhacks

colinhacks:
There are a few things giving me pause here.

One, this could be initially confusing with `.toLowerCase()` and `.toUpperCase()`. This isn't a huge deal. But perhaps now that `z.coerce` has landed, maybe this should be `z.uppercase()` (to validate) and `z.coerce.uppercase()` (to transform). Hm. This loses the parallelism with the built-in string functions `toLowerCase`/`toUpperCase` though, so probably not.

Two, I don't like the idea of having a `lowerCase` and a `toLowerCase` method side-by-side. The mismatch in capitalization triggers my OCD and feels inelegant. Why can't JavaScript use snake case :(

Three (this is the main one) I don't love the idea of muddying `ZodString` with a generic. It would be a breaking change, for starters. Also, to be truly correct, you'd need two generics, one for the output type and one for the input type. Once you do this, the result of `z.string()`  changes from `ZodString` to `ZodString<string, string>` in everyone's Intellisense, forever. That's redundant, and looks bad, and it's a bad tradeoff (imo) for a minor usability improvement relating to a niche feature. 

So at the moment, I'm out on this. It's possible Zod 4 will add these generics anyway for a range of other reasons; in that case I'll revisit this, but at that point the implementation will be quite different. So closing for now. Thanks for the great work here @brianblakely 👍👍

brianblakely:
I am open to pursuing this again after Zod 4 lands, in a way that considers your comments. Thanks.

Issue #2183, opened 2023-03-10
`z.inferFlattenedErrors` does not account for transforms/effects


qw-in:
Originally reported in `zod-form-data`: https://github.com/airjp73/remix-validated-form/issues/250#issuecomment-1454827269

[Playground reproduction](https://www.typescriptlang.org/play?ssl=5&ssc=47&pln=5&pc=25#code/JYWwDg9gTgLgBAbzgLzgXzgMyhEcBEyEAJvgNwBQFAxhAHYDO8EARgFZwC8KAdK2wFNqMABRJMECAC5eTKMDoBzEQEp0KnjCgBDRhKggRELgD44YuC21QZEHhONoVKqjACeYAXABq2gDbAxNowwPQAolA4UAxcvAqYAlAAYn7BMAJ0AsQRUQwAPO6eEJhw-CaUNPRMcFACMdz8PAzaCQAK1gwCYk5UwCUiAIS1DE0ArtTUdQxqCBRwcLSM8Jip8NzDPIlR9qvpdKqU8wD0R-PzAHoA-FTzi9Vb0AwyvgFBIeGRj7ErwZRoQA)

I'm not entirely clear on the details, but it appears that when using a `transform` (that seems to use `z.Effect` under the hood) `z.inferFlattenedErrors` does not match the type of `error.flatten()`

```ts
import { z } from "zod";

const obj = z.object({ foo: z.string() }).transform(o => ({ bar: o.foo }))

type ValidationErrors = z.inferFlattenedErrors<typeof obj>;

const res = obj.safeParse({})

if (!res.success) {
  const flat = res.error.flatten();
  const errors: ValidationErrors = flat;
  // ^ type error here
}
```

JacobWeisenburger:
I have confirmed this

ytsunekawa:
I made a PR so that inferFlattenedErrors will be of the type regarding input, even for schemas where transform is defined.
I'm a beginner so the result may not be what you expect.

https://github.com/colinhacks/zod/pull/2321

aaronadamsCA:
Well actually, this is probably a problem:

[...]

PR #2321, opened 2023-04-13
Fix[#2183] inferFlattenedErrors should be a type about input type

ytsunekawa:
This PR is to resolve issue #2183, which is to make inferFlattenedErrors be of type input, even for schemas with transforms defined.
I'm new to writing PRs, so I wasn't sure of the etiquette, and I'm not sure of the differences in the README, so I'd be glad to hear your comments.

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -71,16 +71,11 @@
 - [Coercion for primitives](#coercion-for-primitives)
 - [Literals](#literals)
 - [Strings](#strings)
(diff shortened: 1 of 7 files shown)


aaronadamsCA:
Just adding a comment to note this isn't stale.

deno/lib/README.md
suhaibmujahid, line 1: The changes in the `README.md` file seem unrelated to what this PR is fixing.

Issue #2284, opened 2023-04-01
Equality between properties of the schema


aminnairi:
In the case of comparing two passwords for instance (`password` & `passwordConfirmation`), it would be great to have a way of comparing the two inside the schema.

```javascript
const registrationSchema = z.object({
  password: z.string(),
  passwordConfirmation: z.equals(schema => schema.password.value)
});
```

The schema is deliberately simplified for demonstration purposes.

Here above, the `equals` could be an operator that helps check if the value is equal (`===`) to the wanted property of the schema.

This could also be great to enable this operator with a `refine` when needed.

```javascript
z.string().refine((value, schema) => {
  return schema.password.value === value;
})
```

This would not mean breaking the API if this is another parameter that the `refine` method accepts.

Or, it could be a completely new method.

```javascript
z.string().compare((value, schema) => {
  return schema.password.value === value
})
```

> This would also be useful for unlocking cool things for dates for instance `z.date().after(schema => schema.beginDate.value)`

Either way, the point is that it would be awesome to be able to reflectively inspect values from our own schema for that particular case and others to come.

*Note: the code excerpts above are only pseudo-code blocks, the API might be changed completely, I didn't read the full source-code to adapt these examples.*

joaoGabriel55:
Hello @aminnairi I sent a PR for that, but my solution is not type safe, cause is impossible get the current schema type in a `z.string()` method, something like:
```ts
equalsTo(field: CurrentSchema, message?: errorUtil.ErrMessage);

// usage:
const registrationSchema = z.object({
    password: z.string(),
    passwordConfirmation: z.string().equalsTo("password"), // impossible get the current schema type ({ password: string; passwordConfirmation: string; })
});
```

Demivan:
Downside of the `refine` method is that it waits for the entire schema to be valid before it runs.
How about having `partialRefine` that will execute when specified fields are valid?
It can be made typesafe by using `keyof` because it is specified on the object schema.

```
const registrationSchema = z.object({
  name: z.string(),
  email: z.string().email(),
  password: z.string(),
  passwordConfirmation: z.string()
})
.partialRefine(
  ['password', 'passwordConfirmation'], // Does no require name and email to be valid to run
  (values) => { // `values` can be typesafe too by extracting only specified keys
    return values.password === values.passwordConfirmation;
  })
```

aminnairi:
[...]

PR #2323, opened 2023-04-13
feat: implement equality between properties from schema

joaoGabriel55:
From issue: #2284

**Overview**
In the case of comparing two passwords for instance (password & passwordConfirmation), it would be great to have a way of comparing the two inside the schema.

**Solution**
Was created a new `ZodString` method: `equalsTo(field: string, message?: errorUtil.ErrMessage)`

Test file to look at:
`src/__tests__/string.test.ts`

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -57,8 +57,8 @@
 - [Coercion for primitives](#coercion-for-primitives)
 - [Literals](#literals)
 - [Strings](#strings)
-  - [Datetime](#datetime-validation)
-  - [IP](#ip-address-validation)
+  - [Datetime](#iso-datetimes)
+  - [IP](#ip-addresses)
 - [Numbers](#numbers)
 - [BigInts](#bigints)
 - [NaNs](#nans)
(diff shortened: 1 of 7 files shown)


igalklebanov:
Hey 👋 

I don't think this fits. It's not type-safe. 😞 
I could sneeze and accidentally add an emoji, and it'll still pass all CICD checks, get to production, and I'll lose customers.

joaoGabriel55:
> Hey wave
> 
> I don't think this fits. It's not type-safe. disappointed I could sneeze and accidentally add an emoji, and it'll still pass all CICD checks, get to production, and I'll lose customers.

Yeah, that is true. But it is possible get the current schema type and change the type of `field` param? e.g.:
```ts
equalsTo(field: CurrentSchema, message?: errorUtil.ErrMessage)
```
However, if this was not possible I think just using `refine` is enough.

What do you think @igalklebanov ?

igalklebanov:
> > Hey wave
> > 
> > I don't think this fits. It's not type-safe. disappointed I could sneeze and accidentally add an emoji, and it'll still pass all CICD checks, get to production, and I'll lose customers.
> 
> Yeah, that is true. But would be possible get the current schema type and change the type of `field` param? e.g.:
> ```ts
> equalsTo(field: CurrentSchema, message?: errorUtil.ErrMessage)
> ```
> However, if this was not possible I think just using `refine` is enough.
> 
> What do you think @igalklebanov and @aminnairi?

That seems impossible.

PR #2331, opened 2023-04-17
fix ZodInvalidEnumValueIssue.options typings

IlyaSemenov:
ZodInvalidEnumValueIssue.options is incorrectly typed as if it could have numbers:
https://github.com/colinhacks/zod/blob/502384e56fe2b1f8173735df6c3b0d41bce04edc/src/ZodError.ts#L74

this contradicts `zod.enum` contract (which only accepts strings for options) and even contradicts the error handling documentation:

https://github.com/colinhacks/zod/blob/502384e56fe2b1f8173735df6c3b0d41bce04edc/ERROR_HANDLING.md?plain=1#L48

The PR updates the typings to only allow `string[]`.

### DIFF
--- a/deno/lib/ZodError.ts
+++ b/deno/lib/ZodError.ts
@@ -69,9 +69,9 @@ export interface ZodInvalidUnionDiscriminatorIssue extends ZodIssueBase {
 }
 
 export interface ZodInvalidEnumValueIssue extends ZodIssueBase {
-  received: string | number;
+  received: string;
   code: typeof ZodIssueCode.invalid_enum_value;
-  options: (string | number)[];
+  options: string[];
 }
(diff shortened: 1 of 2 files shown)


IlyaSemenov:
This is a very simple PR that makes perfect sense in fixing wrongly advertised types.

colinhacks:
The `ZodInvalidEnumValueIssue` is also used for ZodNativeEnum, where number values are possible. This is the kind of edge case where I decided it's worthwhile to consolidate these issue types into one for convenience, despite the slight loosening of the type signature. Messing with the various Issue types tends to break a lot of user code so this is unlikely to change. Thanks for your work on this though!

Issue #2318, opened 2023-04-11
Add z.select (similar to z.enum but accepting anything as options)


IlyaSemenov:
### Problem

Imagine I have a schema that validates a value to be one of the predefined integers. Like when a user chooses the length of the secret key to be either 512, 1024 or 2048 bits (just random example). Naturally, I would imagine there was a Zod schema for that... like `z.enum([512, 1024, 2048])`? Unfortunately, that doesn't work, as `z.enum` only accepts string values.

Another scenario is where the list of options is coming from the variable and is only known in runtime:

```ts
const options = await getOptions(); // string[] or number[] or even any[]
const schema = z.enum(options); // doesn't work
```

### Current workaround

One can use `z.union([z.literal(512), z.literal(1024), z.literal(2048)]` but it's cumbersome, it's arguably slow (as it validates each sub-schema independently) and it doesn't work with option values other than strings and numbers.

### Considered solutions

At first, I was thinking that `z.enum` could be extended to allow any list of anything as valid options (#2317). That is not a viable approach as `z.enum(...).enum` will break then (thanks @igalklebanov for pointing to that).

Another suggested approach was to extend `z.union` to accept literals. I don't like that for two reasons:
* `z.union` is advertised as an aggregating schema builder. Mixing unrelated concepts (nested schemas vs. literals) should be avoided.
* If `z.union` were to accept arbitrary options but treat zod schemas separately, then zod schemas themselves would not be valid options to choose from. (Not that it's a huge tragedy but that is a limitation that could be avoided).

### Suggested approach

I propose to add `z.select` which will be basically `z.enum` without `schema.enum` but accepting anything as the list of options (not even necessarily primitives):

```ts
const Select = z.select([1, 5, "test", true]) // or: z.select(options) where options extends any[]
type Select = z.infer<typeof Select> // 1 | 5 | "test" | true
Select.options // [1, 5, "test", true]
const option = Select.parse(...) // one of options
```

In fact, `z.enum` will probably be a subclass of `z.select` in that case (they will share `schema.options` and probably the actual validation implementation).

The resulting type could be derived from the options with [`ElementOf<...>`](https://github.com/ts-essentials/ts-essentials/tree/afccd3550cc1c03b87d1df4bd590b1e1caa0c930/lib/element-of) recipe.

IlyaSemenov:
This is a valid issue with a reasoning and having a PR.

IlyaSemenov:
[...]

PR #2333, opened 2023-04-17
Add zod.select

IlyaSemenov:
This adds `zod.select` as suggested in #2318.


IlyaSemenov:
😔

colinhacks:
My current plan is to overload `z.literal()` with this functionality. It's a breaking change so it will land in Zod 4. But thanks for the PR!

PR #2338, opened 2023-04-18
Make EnumValues generic

### LINKED ISSUE #2280: Restrict z.enum options to existing type
### Problem

I want to restrict options passed to `z.enum` to the existing type.

```ts
type Status = 'new' | 'accepted' | 'closed'

z.enum(['new', 'accepted']) // OK
z.enum(['new', 'approved']) // Invalid choice - should be somehow disallowed
```

Is there an elegant way to statically validate the typings of enum options?

### Solutions tried

```ts
z.enum(['new', 'accepted'] as Status[]) // Argument of type 'Status[]' is not assignable to parameter of type '[string, ...string[]]'.
z.enum(['new', 'accepted'] satisfies Status[]) // Argument of type '("new" | "acepted")[]' is not assignable to parameter of type '[string, ...string[]]'.
z.enum<Status>(['new', 'accepted']) // Expected 2 type arguments, but got 1
z.enum(['new', 'accepted'] as [Status, ...Status[]]) // works, but ugly
z.enum(['new', 'accepted'] satisfies [Status, ...Status[]]) // works, but ugly
z.enum(['new', 'accepted'] as const satisfies readonly Status[]) // works, but could it be simpler?
```

at least a built-in helper would be useful:

```ts
z.enum(['new', 'accepted'] as z.EnumOptions<Status>) // self explanatory
```

IlyaSemenov:
Allow statically validate enum values against external type. Fix #2280.


IlyaSemenov:
🙄

lnfel:
Please merge this, `readonly [string, ...string[]]` is an absurd way of typing

jakubwolny:
can we move forward with this? are there any objections?

PR #2341, opened 2023-04-19
update uuid regex to handle v6, v7, v8 uuids and max UUID

gnz00:
Just need to include the new versions, defined here:
https://www.ietf.org/archive/id/draft-peabody-dispatch-new-uuid-format-04.html#name-new-formats


johanneslumpe:
@colinhacks Sorry for the ping. Is it possible to merge this PR? We're currently running into the issue of having to validate v7 uuids and would like to prevent having to use a custom type for them.

colinhacks:
I generalized the UUID regex here in response to some other feedback https://github.com/colinhacks/zod/commit/37e9c550460e4edd144da90d903e878c119c5cc1

Should be available in `zod@canary` until the next minor lands.

kibertoad:
@colinhacks Any timeline on when new version will be released?

PR #2356, opened 2023-04-20
Add getters to Map for key + value?

bbar:
For Records we have access to `keySchema` and `valueSchema`, like this:

```
const schema = z.record(z.string(), z.number());
console.log({
  key: schema.keySchema,
  value: schema.valueSchema,
});
```

I've added the same methods to Map so users don't have to use `_def.keyType` and `_def.valueType`.


colinhacks:
Beautiful, thanks!

PR #2359, opened 2023-04-21
docs: adapt demonstrative example

TkDodo:
To the best of my knowledge, it is not true that zod will throw an error for extra fields per default. According to the docs, to achieve this, we'd need to use `.strict()` our object. See: https://zod.dev/?id=strict

I've opted for just removing this one validation from the example, but we can replace it with a different one if you prefer to have 3 errors in the result set.

### DIFF
--- a/ERROR_HANDLING.md
+++ b/ERROR_HANDLING.md
@@ -130,12 +130,6 @@ Here are the errors that will be printed:
     message: "Invalid input: expected string, received number",
   },
   {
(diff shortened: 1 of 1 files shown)


colinhacks:
Good catch. Ended up taking a different approach: https://github.com/colinhacks/zod/commit/a3c525658bc43edf40747a99b8f882d8d3d1e0c7

PR #2360, opened 2023-04-21
feat: add `partialRefine` method

Demivan:
`partialRefine` is equivalent to `refine` but does not require entire object schema to be valid. It will run when all specified fields are valid instead. This allows for validation of dependent fields.

```ts
const validationSchema = z
  .object({
    email: z.string().email(),
    password: z.string(),
    confirmPassword: z.string(),
  })
  .partialRefine(
    ["password", "confirmPassword"],
    data => data.password === data.confirmPassword,
    {
      path: ["confirmPassword"],
      message: "Both password and confirmation must match",
    }
  );
```

Closes: #2284

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -57,8 +57,8 @@
 - [Coercion for primitives](#coercion-for-primitives)
 - [Literals](#literals)
 - [Strings](#strings)
-  - [Datetime](#datetime-validation)
-  - [IP](#ip-address-validation)
+  - [Datetime](#iso-datetimes)
+  - [IP](#ip-addresses)
 - [Numbers](#numbers)
 - [BigInts](#bigints)
 - [NaNs](#nans)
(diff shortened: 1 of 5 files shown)


Demivan:
@colinhacks Sorry for ping. Just want to check if this is something you are willing to merge.

If it is, I can work on adding documentation for this feature.

colinhacks:
Apologies for the lateness @Demivan. This is an interesting idea, and definitely important for this use case. I don't immediately love the idea of a special `refine` method just to accommodate this use case. Zod 4 will introduce a concept of a "check" that generalizes the current "refinement" concept. I think this would fit into that nicely.

```ts
z.object({ ... }).check({
  fields: ["password", "confirmPassword"],
  run(ctx){
    ctx.input; // Pick<T, "password" | "confirmPassword">
    // run validation
    
  }
})
```

PR #2363, opened 2023-04-23
feat: Improve async tooling with new options `cache`, `debounce`, and `freshest`

roguealexander:
Hello, first time contributing to zod so let me know if I'm stepping on any toes.

_Forward: This work in this PR has been extremely helpful in validating forms in my own project, and figured it would be a great addition to zod at large, and hopefully help some other people. This is somewhat of a departure from the current pure zod, but with things like `zodValidator` for `react-hook-forms` (and others) it can help make a lot of devs lives easier. I would understand if you choose that you don't want this PR to be added, no hard feeling either way._

-------

This PR is intended to improve the tooling of async `superRefine` and `transform` operations, particularly with rapidly changing inputs (think forms). 

I've added a second option to `superRefine` and `transform`, called `AsyncRefinementOptions`:
```ts
export type AsyncRefinementOptions = {
  cache?: boolean;
  debounce?: { ms: number };
  freshest?: boolean;
};
```

As part of this work, I've added introspection of `issues` in `RefinementCtx`, and a `fatal` flag to the options of all base validations. Here's an example of both:

I've written both tests and documentation for both of these feature sets, and I think the easiest and best way to explain them is to copy paste the docs I've added:

-------

**Within `.superRefine` docs section:**

### Async Refinement Options

`superRefine` takes an optional second argument of type `AsyncRefinementOptions`. These options are `cache`, `debounce`, and `freshest`. AsyncRefinementOptions are most useful in rapidly changing data where only the most recent result is needed (think forms).

**`cache`** will store the most recent result of the `superRefine`, and any subsequent parses will return that result if the input data has not changed. The cache check and return happens before the async operations within the `superRefine` are executed, and before any async operations up the validation chain are executed:

```ts
const cacheSchema = z
  .string()
  .email()
  .superRefine(async (val) => {
    await serverTrip1();
  })
  .superRefine(async (val) => {
    await serverTrip2();
  }, { cache: true });

// Executes both serverTrip1 and serverTrip2
await cacheSchema.parseAsync('test@gmail.com');

// Cache hit, neither serverTrip executes
await cacheSchema.parseAsync('test@gmail.com');
```

> ⚠️ `cache` is a naive caching implementation, only storing a single cached result.

**`freshest`** ensures that only the most recent result is returned from rapidly changing inputs. A validation pass is marked as stale by a fresher pass resolving, this ensures that the result of a slower pass won't ever replace the fresher one when it finally resolves:

```ts
const freshestSchema = z
  .string()
  .email({ fatal: true })
  .superRefine(async (val) => {
    // Only called if email is valid (see fatal flag)
    await serverEmailValidation(val)
  }, { freshest: true })

// Chronological events:

  // User form entry 1
  // Passes email validation, starts server validation
  freshestSchema.parseAsync('test@gmail.com')

  // User form entry 2 (while server validation still running)
  // Fails email validation immediately, exits
  freshestSchema.parseAsync('test@gmail')

  // Entry 2 resolves first and exits with error: "Invalid email"
  // Entry 2 resolving marks Entry 1 as stale

  // Entry 1 resolves with valid, but is superseded by the fresher error "Invalid email"
```

**`debounce`** extends the functionality of `freshest`, and takes a duration param: `ms`. This option allows superRefinements to be debounced by that duration. Validation passes will become stale if they are replaced by a fresher pass within the debounce duration, and will return the most recent resolved result instead:

```ts
const debounceSchema = z
  .string()
  .email({ fatal: true })
  .superRefine(async (val) => {
    await serverEmailValidation(val)
  }, { debounce: { ms: 250 } })

// Chronological events:
  // User form entry 1 (has resolved as valid)
  debounceSchema.parseAsync("test@gmail.org")

  // Changing entries 2-4 in quick succession
  debounceSchema.parseAsync("test@gmail.c") // Does not call `serverEmailValidation`
  debounceSchema.parseAsync("test@gmail.co") // Does not call `serverEmailValidation`
  debounceSchema.parseAsync("test@gmail.com") // Passes debounce, calls `serverEmailValidation`
  // Result of "test@gmail.com" async validation is returned
```

AsyncRefinementOptions can be combined:

`{ debounce: { ms: 250 }, cache: true }`

`{ freshest: true, cache: true }`

> ⚠️ Technically `debounce` and `freshest` can be combined, but because debounce is a superset of freshest, it is the same as just `debounce`.

**Within `.transform` docs section**

You may also want to abort early before the transform fires:

```ts
const IdToUser = z
  .string()
  .uuid({ fatal: true })
  .transform(async (id) => {
    // Will not be called if string is not a valid uuid
    return await getUserById(id);
  });
```

> ⚠️ If your schema contains asynchronous transforms, you must use .parseAsync() or .safeParseAsync() to parse data. Otherwise Zod will throw an error.

#### Async Transform Options

`transform` takes an optional second argument of type `AsyncRefinementOptions`. These options are `cache`, `debounce`, and `freshest`. AsyncRefinementOptions are most useful in rapidly changing data where only the most recent result is needed (think forms).

`AsyncRefinementOptions` work the same way on `transform`s as they do on `superRefine`ments, and can be read about more in the `superRefine` section above. As an overview:

- **`cache`** stores the result of the freshest `transform` call, and will return that value if the input hasn't changed, without calling any async operations before or within the `transform`.

- **`freshest`** will always return the freshest resolved result. A slow validation pass will be marked stale if a newer pass resolves first, either through a faster async operation, or through a fatal issue that resolves the pass instantly. It prevents a slow validation pass from overwriting the result of a fresher pass when it finally resolves.

- **`debounce`** is a superset of `freshest`, but takes a duration param `ms`. It will wait for that duration before calling the async `transform` function. If a new validation pass is started before the debounce duration is finished, the old pass is marked as stale, and will return the most recent resolved result.

- `transform`s with these options set can still be layered and interleaved with other `transform`s, `superRefine`ments, or basic validations.

- `freshest` or `debounce` can be combined with `cache`. (debounce is a superset of freshest, so the freshest flag does nothing when combined with debounce)

------

Thanks for taking the time to read through this, let me know what you think.


colinhacks:
I've been sitting on this PR for a long time because it's very well-implemented and I think the core proposal is really interesting. Ultimately I think it's a bit too sophisticated for Zod core, and the relatively small demand for this feature can't justify the increase in bundle size. 

Many things from this will land in Zod 4, like `fatal`. I also like the changes you made like `errorUtil.normalize`, etc. Refinements/transforms will be getting a big refactor in Zod 4, so I'll keep these ideas in mind. 

Thanks for the great work on this, and apologies for the lateness in responding. It's a great PR.

Issue #2267, opened 2023-03-28
Error in regex for email validation in PR #1982 and #2058


fvckDesa:
It was pointed out to me (https://github.com/colinhacks/zod/pull/1982#issuecomment-1487157041, #2154) that the regex I wrote has problems with some valid emails, I looked it again and I found the problem in the final part:
``` ts
([A-Za-z0-9]([A-Za-z0-9-]*[A-Za-z0-9])*(\.[A-Za-z]{2,})+))
```
I got confused between domain name, subdomains and top level domain.
I also saw PR #2157 from @colinhacks who wants to make the email regex simpler so if he decides he doesn't want to change it I could fix it

Shreyas-SAS:
So are you willing to change it or should I work towards fixing the regex issue?

fvckDesa:
Yes, I fix it

Shreyas-SAS:
> Yes, I fix it

oh ok. If you are confused at any step please ping me, maybe I can help with something.

Issue #2205, opened 2023-03-17
Email validation rejects email with special characters


timheerwagen:
For example info@räucher.de gets rejected because of the "ä".

In Chrome, special characters are automatically converted to ASCII on the client side and back. However, this is not the case with Firefox and Safari. So I have no problems in Chrome browsers, but in the others.

Im using simple z.string().email(), dont want to transform each to ASCII by myself and use my own validator.

kasbah:
Emails with special characters still get rejected.

Soviut:
I have a similar issue with emails that contain multiple subdomains with the first level subdomain (closest to the domain) containing numbers.

```
PASS test@example.com
PASS test@test.example.com
PASS test@test0.example.com
PASS test@test.test.example.com
PASS test@test0.test.example.com

FAIL test@test.test0.example.com
FAIL test@test.test.test0.example.com
FAIL test@test.test0.test.example.com
```

PR #2380, opened 2023-04-28
emails support subdomains

trevoro:
This addresses some issues raised by #2267 and #2205 with email addresses not supporting subdomains.
While I'm not completely certain the regex needs to retain this level of complexity (I think it supports IPv6 which is impressive yet... wild) it now supports the potentially slightly more likely situation of an email address for a subdomain which is entirely valid. 

The changes in the deno README seem to be attributed to the commit hooks; they were likely forgotten on a previous merge.

I'm aware that this might overlap with the work on simplifying the email regex in which case, feel free to reject this PR.


colinhacks:
Fixed in #2157, but thanks~

PR #2394, opened 2023-05-02
docs: add `zodock` to mocking ecosystem

ItMaga:
First of all, I want to thank you for this great library! 🚀 

I have implemented the following utility to generate dummy data based on Zod schemas. Unlike similar utilities already presented in the ecosystem, this utility covers most of the functionality of Zod schemes, for example, `transform`, `refine`, `preprocess`, recursive types and all complex combinations of these methods. 

I will be very happy if this library is useful to the community, and I will try to actively support it.

---
In addition, I wanted to ask for help with the `z.never()` type. What data needs to be passed in order to successfully pass the check when parsing a schema with `z.never()`?


colinhacks:
Love it, thanks!

PR #2410, opened 2023-05-06
docs: add tapiduck to API libraries

sumukhbarve:
Zod is brilliant. Thank you for creating and maintaining it.

tRPC is amazing too. It gives you fullstack intellisense without the need for GraphQL (or codegen.) But it still uses GraphQL-like "query" and "mutation" terminology. And [it isn't natively compatible](https://github.com/trpc/trpc/issues/755) with OpenAPI.

Tapiduck is a simpler and RESTful alternative to tRPC. It uses Zod for inferring shape definitions, and for enforcing runtime type validations. It is OpenAPI compatible, built on Express, and should feel familiar to most Node devs.


colinhacks:
Love it, thanks!

PR #2420, opened 2023-05-11
fix: super refinement function types

carl0s-sa:
Typescript was complaining about the `.superRefine` argument returning a `Promise` when it was expecting a `void` return.

This fixes it so it accepts `Promise<void>` as a valid return value

Attachment of error:
Screenshot 2023-05-11 at 11 36 09

### DIFF
--- a/deno/lib/__tests__/refine.test.ts
+++ b/deno/lib/__tests__/refine.test.ts
@@ -156,6 +156,35 @@ test("superRefine", () => {
   Strings.parse(["asfd", "qwer"]);
 });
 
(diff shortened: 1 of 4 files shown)


carl0s-sa:
ccing @colinhacks for review 🙏

colinhacks:
Awesome thanks! Guess there's not a lot of async super-refiners out there.

rjmackay:
Any chance this will be included in a release soon?

colinhacks:
Landed in 3.22

PR #2422, opened 2023-05-11
Adds `.assert` schema method

StefanTerdell:
### Reasoning
I find myself doing these types of "quick and dirty" checks a lot, especially when first transitioning old code from the dreaded `"x" in y and typeof y.x === "object" && y.x !== null && z in y.x && [etc, etc]`, especially when there's already error handling present. This would speed up the process a bit and leave the flow of the old code a bit more intact.

#### From the readme:

> ### `.assert`
> 
> Uses safeParse under the hood to assert the input type of the schema. Useful for inlining quick checks on unknown data that does not need to be transformed.
> 
> ```ts
> const data = JSON.parse('"billie"') as unknown
> 
> if (z.string().assert(data)) {
>   // data is now recognized as a string by TS
>   data.toUpperCase().startsWith("B")
> }
> ```

#### Comparison to safeParse

Example legacy code:
```ts
if (typeof data === "object" && data !== null && "prop" in data && typeof data.prop === "string") {
  handleData(data)
} else {
  handleError()
}
```

Example using assert:
```ts
const schema = z.object({ prop: z.string() })

if (schema.assert(data)) {
  handleData(data)
} else {
  handleError()
}
```

Example using safeParse:
```ts
const schema = z.object({ prop: z.string() })

const safeParseResult = schema.safeParse(data)

if (safeParseResult.success) {
  const data = safeParseResult.data;

  handleData(data)
} else {
  handleError()
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1922,6 +1922,18 @@ For convenience, this has been aliased to `.spa`:
 await stringSchema.spa("billie");
 ```
 
+### `.assert`
+
+Uses safeParse under the hood to assert the input type of the schema. Useful for inlining quick checks on unknown data that does not need to be transformed.
+
+```ts
+const data = JSON.parse('"billie"') as unknown
+
+if (z.string().assert(data)) {
+  // data is now recognized as a string by TS
+  data.toUpperCase().startsWith("B")
+}
+```
 ### `.refine`
 
 `.refine(validator: (data:T)=>any, params?: RefineParams)`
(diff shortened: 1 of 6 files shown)


StefanTerdell:
Just now saw the discussion in #1113. However, the implementation in this PR targets the `Input` type, so any type transformations should be properly ignored.

PR #2426, opened 2023-05-12
adds ctx to preprocess

Pingviinituutti:
This PR would allow `context`/`ctx` to be used in `preprocess` in the same way it is currently usable in `transform`. 
This let's the user, for example, add issues/errors already in the preprocessing state.

Related to https://github.com/colinhacks/zod/issues/696


colinhacks:
Made a couple tweaks but looks good.

Issue #2122, opened 2023-03-01
UUID validation fails


cplummer-linq:
Using Zod 3.20.6.

I have a guid that doesn't pass validation. Unit tests:

```
import { z } from 'zod';

describe('Validation Tests', () => {
  // f4405c78-1a39-4d59-9d8a-7e56d5e13294
  // 852e395d-693e-ed11-809a-c971856d9437

  it('should succeed, correctly', async () => {
    const schema = z.object({
      value: z.string().trim().uuid(),
    });
    const parsed = schema.safeParse({ value: 'f4405c78-1a39-4d59-9d8a-7e56d5e13294' });
    expect(parsed.success).toStrictEqual(true);
  });

  it('should succeed, but does not', async () => {
    const schema = z.object({
      value: z.string().trim().uuid(),
    });
    const parsed = schema.safeParse({ value: '852e395d-693e-ed11-809a-c971856d9437' });
    expect(parsed.success).toStrictEqual(true);
  });
});
```

WORMSS:
```diff
- expect(parsed.success).toStrictEqual(true);

+ expect(parsed.success).toStrictEqual(true);
```

cplummer-linq:
Hmmm... it's a generated guid from a SQL database. Can uuid() not be used on those types of values?

WORMSS:
GUID or UUID? which is it making?

cplummer-linq:
I'd like validate a GUID which can come from the NEWID() or NEWSEQUENTIALID() functions in SQL.
https://learn.microsoft.com/en-us/sql/t-sql/functions/newsequentialid-transact-sql?view=sql-server-ver16

They even call it UUID in their docs:
https://learn.microsoft.com/en-us/windows/win32/api/rpcdce/nf-rpcdce-uuidcreatesequential

cplummer-linq:
I just retested, and using all upper case for that value didn't matter.

cplummer-linq:
https://learn.microsoft.com/en-us/sql/t-sql/functions/newid-transact-sql?view=sql-server-ver16

![image](https://user-images.githubusercontent.com/103345240/222234896-62d76d5f-dfd8-4298-aced-b20ecc35c3e6.png)

https://www.rfc-editor.org/rfc/rfc4122

No, I haven't read this RFC, just providing links for reference.

cplummer-linq:
Clarification. NEWSEQUENTIALID() is not RFC4122 compliant.

[...]

PR #2430, opened 2023-05-15
Add guid to string validation

dca123:
I was running into the same issue discussed here #2122 for my GUID 'd89e7b01-7598-ed11-9d7a-0022489382fd'. 

Im using a MSSQL database with a newsequentialid column to generate my guids. For some of the generated id's as the example above the UUID error was being flagged. As discussed in the issue, this is because while NEWID is compliant with 4122, NEWSEQUENTIALID is not ! 

I lifted the regex straight from here - https://ihateregex.io/expr/uuid/ 🙏🏼
It would be great to have this general-esque guid parser, even though it ain't 4122 compliant. 

Lemme know how I could improve this PR
Cheers,
Dev

### DIFF
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -57,8 +57,8 @@
 - [Coercion for primitives](#coercion-for-primitives)
 - [Literals](#literals)
 - [Strings](#strings)
-  - [Datetime](#datetime-validation)
-  - [IP](#ip-address-validation)
+  - [Datetime](#iso-datetimes)
+  - [IP](#ip-addresses)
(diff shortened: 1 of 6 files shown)


colinhacks:
I think I'll just generalize the `uuidRegex` to use this. Thanks for alerting me to this!

dca123:
ty ty <3

25khattab:
Hello Just wanted to say that the GUID doesn't pass I don't know why 

but the regex works fine when using it with `z.string().regex()`

but doesn't work using `z.string().uuid()`

PR #2432, opened 2023-05-15
docs: allow Zod to be used in dev tools at site

jrel:
This change allows us to test Zod on the documentation site for a simple playground test!

### DIFF
--- a/index.html
+++ b/index.html
@@ -256,5 +256,6 @@
           .prepend(document.querySelector(".theme-btn"));
       });
     </script>
+    <script src="https://www.unpkg.com/zod/lib/index.umd.js"></script>
   </body>
 </html>


index.html
```diff
@@ -256,5 +256,6 @@
           .prepend(document.querySelector(".theme-btn"));
       });
     </script>
+    <script src="https://www.unpkg.com/zod@3.21.4/lib/index.umd.js"></script>
```
samchungy, line 259: issue is you'd probably want that version number to be dynamic

jrel: You are sure. I will change!

jrel: After reading the post about how the docs site works [here](https://colinhacks.com/essays/docs-the-smart-way), I think we can use the built lib instead of the published lib. This allows us to test the PR changes on the Deploy Preview site.

The current problem with doing this is that the build stage runs on the GitHub action, and Netlify loads the resources directly from GitHub.
To fix this, we need to configure the build command in Netlify and use `/index.umd.js` as src!

Does it make sense to make this kind of change?

colinhacks:
Great idea

Issue #2440, opened 2023-05-17
superRefine breaks compatibility with discriminatedUnion


cdaringe:
# Problem

Here's an easy demo, using zod@3.21.4

```ts
import * as z from "zod";

const schemaA = z.object({
  version: z.literal("a"),
}); // observation, no use of superRefine

z.discriminatedUnion("version", [schemaA]); // works

const schemaB = z
  .object({
    version: z.literal("b"),
  })
  .superRefine(() => {}); // observation, use of superRefine

z.discriminatedUnion("version", [schemaB]); // doesn't work

/*
Type 'ZodEffects<ZodObject<{ version: ZodLiteral<"b">; }, "strip", ZodTypeAny, { version: "b"; }, { version: "b"; }>, { version: "b"; }, { version: "b"; }>' is missing the following properties from type 'ZodObject<{ version: ZodTypeAny; } & ZodRawShape, UnknownKeysParam, ZodTypeAny, { [x: string]: any; version?: any; }, { [x: string]: any; version?: any; }>': _cached, _getCached, shape, strict, and 14 more.ts(2740)
*/
```

The type-error is correct, and it breaks at runtime as well.

In otherwords, ZodEffect<ZodObject> isn't supported... but it should be! A schema that has assertions on top of it is still a compatible schema.

# Expectation

`z.discriminatedUnion` to still work on `.superRefine`d schemas.

cdaringe:
ah jamon. have had a patch open for this issue for months

jonavila:
Just hit this bug. Any workaround?

cdaringe:
None. I submitted a patch, but it has idled. I ended up just using .union

johnslemmer:
I have ran in to this as well. I was about to submit my own issue with this example: https://codesandbox.io/p/sandbox/suspicious-bash-4z8ptf but found this issue instead.

colinhacks:
Closing per https://github.com/colinhacks/zod/pull/2441#issuecomment-2044094604

premdasvm:
I've ran into the same issue, is there a work around?

```ts
const SubPartQuestionZodSchema = QuestionSchema.extend({
	Type: z.literal(QuestionTypes.SUB_PART),
	Questions: z.array(z.lazy(() => QuestionZodSchema)),
});

export const QuestionZodSchema = z.discriminatedUnion("Type", [
	FillInTheBlankQuestionZodSchema,
	EssayQuestionZodSchema,
[...]

PR #2441, opened 2023-05-17
fix: ZodEffect in discriminatedUnion

cdaringe:
# Problem

`.refine` and `.superRefine`d schemas cannot be used in `discriminatedUnion`, but should be able to be.

See https://github.com/colinhacks/zod/issues/2440

# Discussion

The fix is very simple.

1. Support `ZodEffect`s where in the input type is also a `ZodObject`. Add a union member that says "not only accept ZodObject, but ALSO accept ZodEffect\<ZodObject\>"
2. When looking up the schema shape, do a very efficient check to see what type we are, and get the schema shape based on whether we are a ZodEffect or ZodObject

Add a test. Blamo. Easy!

There's a lot of discussion of "problems with z.discriminatedUnion". I see this as _independent_ from those critiques. This, I perceive, was just a basic oversight in the typing of the original implementation. The fix is simple and carries minimal overhead.


deno/lib/README.md
```diff
@@ -57,8 +57,8 @@
 - [Coercion for primitives](#coercion-for-primitives)
 - [Literals](#literals)
 - [Strings](#strings)
-  - [Datetime](#datetime-validation)
-  - [IP](#ip-address-validation)
+  - [Datetime](#iso-datetimes)
+  - [IP](#ip-addresses)
```
cdaringe, line 61: uh errr uhhh i didn't do this. this was the git hooks? should i purge this?

deno/lib/types.ts
```diff
@@ -2882,13 +2882,20 @@ const getDiscriminator = <T extends ZodTypeAny>(
   }
 };
 
-export type ZodDiscriminatedUnionOption<Discriminator extends string> =
```
cdaringe, line 2885: this is a bad diff 😆 . this type is still present on L2895

deno/lib/types.ts
```diff
@@ -2882,13 +2882,20 @@ const getDiscriminator = <T extends ZodTypeAny>(
...
     UnknownKeysParam,
     ZodTypeAny
   >;
 
+type ZodDiscriminatedUnionOptionEffectsObject<Discriminator extends string> =
+  ZodEffects<ZodDiscriminatedUnionOptionObject<Discriminator>>;
+
+export type ZodDiscriminatedUnionOption<Discriminator extends string> =
+  | ZodDiscriminatedUnionOptionObject<Discriminator>
+  | ZodDiscriminatedUnionOptionEffectsObject<Discriminator>;
```
cdaringe, line 2897: the creation of this union is the **core change** in this PR!

cdaringe:
hey friends, this is a pretty simple PR to review. let me know if there's anything i can do to reduce your review burden. i annotated the changes, for example, as a mechanism to make your review more straight forward. i know maintenance is a huge effort, so I want to be courteous of your precious time!

cdaringe:
Hey all, friendly bump. Thanks for your time and support

cdaringe:
Hey all, friendly bump. Thanks for your time and support

ebramanti:
I have run into this issue too, would be really helpful if this change could land soon!

k0rnl:
I have run into this issue too, would be really helpful if this change could land soon ! 🙏

colinhacks:
This is a well-made PR, but the root problem here is that refinements should be stored *inside* the schema instead of wrapping the schema with ZodEffects. I'm hesitant to merge because this adds more complexity to a feature (discriminated union) that likely won't exist in Zod 4. The Zod 4 implementation will avoid the fundamental messiness of discriminatedUnion and it's right around the corner, so I'm gonna close. Thanks @cdaringe, this was a great PR.

ms-ati:
Almost a year later, checking in?

PR #2459, opened 2023-05-26
Add typeName

samchungy:


scammi:
Hey, yes, we've encounter this, would be a good addition on our opinion.

dooleyb1:
Any chance of getting this pull request released?

colinhacks:
This isn't in the type definition for the same reason the enum is called `ZodFirstPartyTypeKind` — Zod is designed to be subclassable. In theory anyone can subclass `ZodType` for custom parsing behavior. In practice, not many people do it, but it's still part of the core design. This may change in the future but until then `ZodTypeDef` is intended to be a generalized interface that applies beyond Zod's first-party subclasses.

PR #2467, opened 2023-05-30
Add forms context section

### LINKED ISSUE #2466: z.string() validates empty strings
This is created after viewing the conversation on this issue: #63

Currently, the result of `z.string()` validation of an empty string `""` leads to a pass instead of a fail, even though the field itself is required. This is not documented anywhere except github issues and apparently an old changelog, but it is not very intuitive in some cases and leads to potential issues especially when trying to validate forms.

Potential solutions:
1. Update `z.string()` to not pass validation for `""` unless `.optional()` is provided (breaking change)
2. Update the zod strings documentation section with a note specifically about this functionality, providing the alternative proposed solution. specifically:
```
z.string().trim().min(1, { message: "Required" })
// or
const notEmpty = z.string().trim().min(1, { message: "Required" });
z.string().pipe(notEmpty);
```
Credit for the above solutions go to the folks in issue #63

I expect to get resistance to option 1 so hopefully option 2 is a reasonable middle ground 🙂

mikeybinns:
Closes #2466

This PR adds a short explainer section to ensure people who are using Zod with forms have the correct context when working with forms and don't expect a "required string" Zod schema to prevent empty strings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2644,7 +2644,7 @@ if (!result.success) {
 
 TypeScript and HTML forms share some common terms, but they do not mean the same thing, which can lead to some confusion when using Zod which uses terms consistent with TypeScript's usage.
 One common misstep is that the schema `z.string()` will allow empty strings for a required HTML form field, which matches the TypeScript type system, but might not be what you expect when you consider the `z.string()` is "required" be default. However, "required" in this context means "a value of this type is present" and the empty string is a value of type `string`.
-```
+```ts
 const str:string = "" // This is okay
 z.string().parse("") // Which means this shouldn't throw an error
 ```
@@ -2652,11 +2652,15 @@ This means if you intend to use Zod with forms, you need to prepare your form da
 
 Option 1, Make empty required fields null when collecting the data
 
-`const requiredFieldValue = isFieldRequired && fieldValue === "" ? null : fieldValue;`
+```ts
+const requiredFieldValue = isFieldRequired && fieldValue === "" ? null : fieldValue;
+```
 
 Option 2, Update the zod schema to fail if field is empty
 
-`z.string().trim().min(1, { message: "Required" }).parse("") // Throws "Required" error.`
+```ts
+z.string().trim().min(1, { message: "Required" }).parse("") // Throws "Required" error.
+```
 
 If you plan to use Zod with forms, you may find the [form integrations](#form-integrations) list helpful.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2638,6 +2639,25 @@ if (!result.success) {
   // => ["Expected string, received number"]
 }
 ```
+### Using Zod with forms
+
+Zod and HTML forms share some common terms, but they do not mean the same thing, which can lead to some confusion.
```
scotttrinh, line 2644:
```suggestion
TypeScript and HTML forms share some common terms, but they do not mean the same thing, which can lead to some confusion when using Zod which uses terms consistent with TypeScript's usage.
```

README.md
```diff
@@ -2638,6 +2639,25 @@ if (!result.success) {
...
+One common misstep is that the schema `z.string()` will allow empty strings for a required HTML form field, which matches the typescript type system, but might not be what you expect.
```
scotttrinh, line 2645:
```suggestion
One common misstep is that the schema `z.string()` will allow empty strings for a required HTML form field, which matches the TypeScript type system, but might not be what you expect when you consider the `z.string()` is "required" be default. However, "required" in this context means a value of this type is present and the empty string is a value of type `string`.
```

README.md
```diff
@@ -2638,6 +2639,25 @@ if (!result.success) {
   // => ["Expected string, received number"]
 }
 ```
+### Using Zod with forms
```
scotttrinh, line 2642:
```suggestion

### Using Zod with forms
```

README.md
```diff
@@ -2638,6 +2639,25 @@ if (!result.success) {
...
+
+Zod and HTML forms share some common terms, but they do not mean the same thing, which can lead to some confusion.
+One common misstep is that the schema `z.string()` will allow empty strings for a required HTML form field, which matches the typescript type system, but might not be what you expect.
+```
```
scotttrinh, line 2646: It's hard to use GitHub's suggestion feature on code blocks, but can you make the following changes:

1. Use ` ```ts`
2. The options you list below should also use ` ```ts` code fences in the examples.
3. Can you format the code examples like prettier with no configuration would? Our formatting is pretty inconsistent, but I'd like to improve that in general.

scotttrinh (changes requested):
This is really great, thanks so much for doing this. Some minor edits requested.

mikeybinns:
I have added all the suggested changed, just waiting for you to approve and merge :)

mikeybinns:
@scotttrinh Hi this is all ready and updated, I'm not sure why it says "Changes requested" still, but I've done all the requested changes and marked the conversations as resolved. Can we get this merged in?

colinhacks:
I appreciate the effort here, but I don't think something like this worth extending the already-very-long README.

Issue #2470, opened 2023-05-31
There's no ergonomic way to rename object properties or do `.partial()` etc on object transformation pipelines


jedwards1211:
Consider the case where you want to build an API response with camelCase properties in JS, and then send it over the wire with snake_case properties: 

```ts
const Post =
  z.object({
    id: z.string(),
    body: z.string(),
    userId: z.string(),
  })
  .transform(
    ({ userId, ...rest }) => ({ ...rest, user_id: userId })
  )
  .pipe(z.object({
    id: z.string(),
    body: z.string(),
    user_id: z.string(),
  }))
```

You might want to do `.partial()` if you have another API that allows the user to pick a selection of fields from a `Post`:

```ts
const PostSelection = Post.partial() // plz?
```

One could create the following helper function for this.  But this all feels very cumbersome.  Maybe in an alternate universe, Zod has a more ergonomic way to do all of this?

```ts
function pipelinePartial(schema) {
  if (schema instanceof z.ZodObject) {
    return schema.partial()
  }
  if (schema instanceof z.ZodPipeline) {
    return z.pipeline(pipelinePartial(schema._def.in), pipelinePartial(schema._def.out))
  }
  if (schema instanceof z.ZodEffects) {
    return z.effects(pipelinePartial(schema._def.schema), schema._def.effect)
  }
  throw new Error(`unsupported schema type: ${schema.constructor.name}`)
}
```

JacobWeisenburger:
Is this what you are looking for?

```ts
import { z } from 'zod'
import _ from 'lodash'

const snakeCaseKeys = ( data: object ) => _.mapKeys( data, ( val, key ) => _.snakeCase( key ) )

const Post = z.object( {
    id: z.string(),
    body: z.string(),
    userId: z.string(),
} )

console.log(
    Post.parse( {
        id: '1',
        body: 'body',
        userId: '1',
    } )
)
// {
//     id: "1",
//     body: "body",
//     userId: "1"
// }

const snakeCasedPost = Post.transform( snakeCaseKeys )
console.log(
    snakeCasedPost.parse( {
        id: '1',
        body: 'body',
        userId: '1',
    } )
)
// {
//     id: "1",
//     body: "body",
//     user_id: "1"
// }

const snakeCasedPartialPost = Post.partial().transform( snakeCaseKeys )
console.log(
    snakeCasedPartialPost.parse( {
        userId: '1',
    } )
)
// {
//     user_id: "1"
// }
```

[...]

PR #2471, opened 2023-05-31
feat: support renaming object properties with .from() and add metadata

jedwards1211:
This enables the following:

```ts
const Response = z.object({
  userId: z.string().from("user_id"),
});
const response = Response.parse({
  user_id: "foo",
});
// => { userId: 'foo' }
type ResponseOutput = z.output<typeof Response>;
// equivalent to:
type ResponseOutput = {
  userId: string;
};
type ResponseInput = z.input<typeof Response>;
// equivalent to:
type ResponseInput = {
  user_id: string;
};
```

### Details

This PR also adds a `metadata` property to all `ZodTypeDef`s; `.from('foo')` clones the schema with `metadata: { from: 'foo' }` on the def.  There have been multiple issues requesting support for attaching metadata to nodes, and I know you've encouraged users to wrap Zod schemas so that you can keep Zod simple, but we haven't thought of any satisfactory ways to accomplish a lot of our goals without metadata:

- Associating OpenAPI metadata like examples to a schema used for several responses' properties, and being able to generate an OpenAPI spec from the combined response schemas (a number of libs out there monkeypatch `.openapi()` onto `ZodType`).
- Marking certain response properties as includable/selectable
- Renaming properties like this

The `ZodObject` types and parsing code are able to unwrap `ZodOptional/ZodNullable/ZodDefault` so that `.from('foo').optional()` works as well as `.optional().from('foo')`.

I added type parameters everywhere for metadata, which is kind of awkward.  It's not strictly necessary, even without all those type parameters,`ExtractFrom<ZodOptional<WithMetadata<ZodString, {from: 'foo'}>>>` should work.  But maybe having the metadata type parameters makes sense in an imagined ecosystem of third-party libs that operate on attached metadata? 

### Notes

I acknowledge this is a big change, and I know you've been wary about attaching metadata to zod schemas in the past, but I wanted to see what you think about this approach.  This is the solution to #2470 that we've come up with to use in our own fork, if nothing else.  For the API framework we're building we want to provide much cleaner ways of declaring input/output property names

I haven't updated the Deno code yet... waiting to get some feedback on this approach before I proceed

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "ts-node": "^10.9.1",
     "tslib": "^2.3.1",
     "tsx": "^3.8.0",
-    "typescript": "~4.5.0"
+    "typescript": "^5.0.0"
   },
   "exports": {
     ".": {
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@
 - [Nullables](#nullables)
 - [Objects](#objects)
   - [`.shape`](#shape)
+  - [`.from`](#from)
   - [`.keyof`](#keyof)
   - [`.extend`](#extend)
   - [`.merge`](#merge)
@@ -1129,6 +1130,32 @@ Use `.shape` to access the schemas for a particular key.
 ```ts
 Dog.shape.name; // => string schema
 Dog.shape.age; // => number schema
+```
+
+### `.from`
+
+Use `.from` to rename an input property to a different name in the output.
+
+```ts
+const Response = z.object({
+  userId: z.string().from("user_id"),
+});
+const response = Response.parse({
+  user_id: "foo",
+});
+// => { userId: 'foo' }
+
+type ResponseOutput = z.output<typeof Response>;
+// equivalent to:
+type ResponseOutput = {
+  userId: string;
+};
+
+type ResponseInput = z.input<typeof Response>;
+// equivalent to:
+type ResponseInput = {
+  user_id: string;
+};
 ```
 
 ### `.keyof`
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -82,6 +82,7 @@
 - [Nullables](#nullables)
 - [Objects](#objects)
   - [`.shape`](#shape)
+  - [`.from`](#from)
   - [`.keyof`](#keyof)
   - [`.extend`](#extend)
   - [`.merge`](#merge)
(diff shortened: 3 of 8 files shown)


package.json
```diff
@@ -35,7 +35,7 @@
     "ts-node": "^10.9.1",
     "tslib": "^2.3.1",
     "tsx": "^3.8.0",
-    "typescript": "~4.5.0"
+    "typescript": "^5.0.0"
```
jedwards1211, line 38: I was getting a possibly infinite recursion error on ZodDiscriminatedUnion until I upgraded...

scotttrinh:
I would imagine that the mapping is metadata on the ZodObject, not on the individual properties. I'm picturing an API more like:

```ts
const Response = z.object({
  userId: z.string(),
}).mapKeys({
  userId: "user_id",
});
```

Mostly because each individual property is itself a schema and for composability's sake you might be taking a schema from some object's `shape` and putting into another schema which doesn't have the expected key, or maybe it ends up as an array element, etc.

jedwards1211:
@scotttrinh yes that's probably better in a lot of ways...it feels dissatisfying for `userId` to be duplicated, but I guess there's just no perfect solution.

jedwards1211:
@scotttrinh would the `.mapKeys` information be preserved by subsequent `.partial`, `.extend`, etc?  In my use case at least, I would want it to

scotttrinh:
@jedwards1211 

> would the .mapKeys information be preserved by subsequent .partial, .extend, etc? In my use case at least, I would want it to

Yeah, although `pick`/`omit` would need to remove it in some cases so you don't have surprises if you later merge an omitted object schema with a new one that you didn't define a map for and suddenly your keys is getting mapped because of some deep ancestor 😂

jedwards1211:
@scotttrinh well I just mean if you pick the `userId` property it would keep the `userId: user_id` mapping, to me that would be the expected behavior.  Whereas if you omit `userId` it would also discard the `userId: user_id` mapping.

jedwards1211:
The bigger problem with metadata typing I'm running into though is that it increases the size of the graph TS has to traverse to measure the variance of types that contain ZodObject schemas etc. and sometimes hits the limit where it bails and says the type instantiation is possibly infinite.  I haven't gotten to the bottom of whether there's a mistake in my approach that causes true infinite recursion in some cases or if it just pushes the type graph size a bit over the threshold.

rattrayalex:
@colinhacks @scotttrinh setting aside the `.mapKeys()` API question, I'm curious for your thoughts on the `metadata` approach of this PR, and whether it's something you might be open to?

In the past, Colin has said he's not interested in allowing third-party libraries which build on zod to embed their own custom metadata in zod schemas – since then, in practice, many libraries which build on zod have had a hard time providing great APIs to users whilst building on top of zod. For ourselves, I feel fairly confident we'll need to fork zod to provide a good user experience without a clean way to attach custom metadata to zod schemas in a typesafe way.

Zod can serve as the basis for various other DSLs (ORMs, API DSLs like ours, etc) and I think it'd really benefit the community to have a single base for this rather than requiring a bunch of forks to keep builders nice and fluent.

Of course, it's your project, so if this isn't a direction you're interested in, we'll figure something else out!

rattrayalex:
Not stale

colinhacks:
Closing in favor of #2496

Issue #1630, opened 2022-12-04
[Suggestion] Stricter boolean coerce option


pheuter:
Since we're currently using Svelte Kit with form actions that make heavy use of the [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) browser api, the new Zod 3.20 release is a very much welcome improvement, specifically the new coercion functionality (since all form post values are serialized as strings over the wire).

One suggestion/request I have is to allow boolean coerce to take a `strict` option that would check for explicit boolean strings:

_Current:_

```js
const schema = z.coerce.boolean();
schema.parse("true"); // => true
schema.parse("false"); // => true
```

_Suggestion:_

```js
const schema = z.coerce.boolean({ strict: true });
schema.parse("true"); // => true
schema.parse("false"); // => false
schema.parse("123"); // => ZodError
```

colinhacks:
Hm :(

This is definitely an important use case but I think the proposed behavior is pretty weird. It's inconsistent with how coercion is handled with other data types. I don't think `z.coerce` is good for this particular use case.

Seems like a better approach for the example you mentioned is something like this:

```ts
z.string().transform(JSON.stringify).pipe(z.boolean());
```

I've been thinking about ways to make it easier to parse `FormData` with Zod for a while but struggling. `FormData` is so astonishingly bad as to be nearly unsalvagable and I hate it 🤷‍♂️

colinhacks:
Perhaps a `zod-formdata` lib is called for

```ts
import * as zfd from "zod-formdata";

zfd.formdata({
  checkbox: zfd.boolean(),
  field: zfd.string(),
  array: zfd.all(z.string()) // uses `FormData.getAll()`,
  jsonified: zfd.json() // input must be string, Zod calls `JSON.parse` automatically
})
```

IDK this is a very early

pheuter:
> I think the proposed behavior is pretty weird.

I haven't used Zod nearly enough to have a good sense of what's idiomatic, so completely defer to your intuitions here, was mainly trying to convey intent and not implementation.

> `FormData` is so astonishingly bad as to be nearly unsalvagable and I hate it 🤷‍♂️

Couldn't agree more :)

> Perhaps a `zod-formdata` lib is called for

[...]

PR #2488, opened 2023-06-06
fix: Stricter boolean coerce option (#1630)

f4lcrum:
I am addressing this error that I encountered during the development of a school project. 
More details here: https://github.com/colinhacks/zod/issues/1630


kerolloz (approved):

colinhacks:
Thanks for the PR, but I think the current behavior is cleaner, even if it is quirky. IMO any truthy value should coerce to `true`. There's now a warning of this potential pitfall in the README, which should help others avoid this issue.

Issue #507, opened 2021-06-30
Data representation of schema?


jasonkuhrt:
Hey,

Given a Zod schema e.g.:

```ts
const a = z.object({
  foo: z.string(),
})
```

Is there a way to get a plain data representation of this?

I am trying to map a Zod schema to a GraphQL schema (specifically an input object type).

scotttrinh:
@jasonkuhrt 

Perhaps you're looking for `ZodObject.shape`?
[TypeScript Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAQwM5IKazgMyhEcBEyasA9EjFMAMYz4DcAUKJBgN5wBecAvljnvg4QAJvQYMqEAHblEcALycAdBABGAK1Q0AFKwZwsECAC5l5SlIDm2gJQAaBtxvii6GCoDW2hEqQALBDBUJUwjOGAZGAQpKlQITGUALREAZQoIy2cgA)

```ts
import assert from "assert/strict";
import { z } from "zod";

const a = z.object({
  foo: z.string(),
})

assert.ok(a.shape.foo instanceof z.ZodString)
```

jasonkuhrt:
@scotttrinh perhaps, I'll check that out thanks!

jasonkuhrt:
@scotttrinh there is a use-case I'm missing.

I need a way to attach metadata to zod schema fields. I would like to attach something like this:

```ts
.metadata({
  graphqlMapping: 'Foo'
})
```

Where `Foo` is a name of a type in the GraphQL API.

This is because I have some fields that I want to map to/re-use existing GraphQL types in the schema.

scotttrinh:
@jasonkuhrt I don't think you're going to find a way to "attach metadata" to a `ZodObject` since that is something that Zod itself defines and doesn't know anything about carrying around additional user-defined metadata.

[...]

PR #2496, opened 2023-06-08
Metadata take 2

jedwards1211:
My approach to metadata in #2471 has major TS performance problems because of the way it uses intersection types.

This is an alternative approach where `z.string().withMetadata({ openapi: { refId: 'user' } })` creates a `ZodMetadata<ZodString, { openapi: { refId: 'user' } }>` type.

Metadata nested within `ZodOptional`, `ZodNullable`, `ZodPromise` etc can be extracted with `z.extractMetadata`; `z.extractDeepMetadata` would also extract out of `ZodArray` element and `ZodSet` value schemas.

This would provide an alternative to monkeypatching for libraries like [`zod-to-openapi`](https://github.com/asteasolutions/zod-to-openapi).  Wrapping schemas wouldn't work for their use case of making sure `$ref`s get generated in the OpenAPI schema for references to a given Zod schema.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@
 - [Native enums](#native-enums)
 - [Optionals](#optionals)
 - [Nullables](#nullables)
+- [Metadata](#metadata)
 - [Objects](#objects)
   - [`.shape`](#shape)
   - [`.keyof`](#keyof)
@@ -1101,6 +1102,88 @@ Extract the inner schema with `.unwrap()`.
 const stringSchema = z.string();
 const nullableString = stringSchema.nullable();
 nullableString.unwrap() === stringSchema; // true
+```
+
+## Metadata
+
+You can associate metadata with a schema with `z.withMetadata()`.
+
+```ts
+const user = z.withMetadata(
+  z.object({
+    id: z.string(),
+    name: z.string().optional(),
+  }),
+  {
+    openapi: {
+      refId: "user",
+    },
+  }
+);
+```
+
+Or use the `.withMetadata()` method.
+
+```ts
+const user = z
+  .object({
+    id: z.string(),
+    name: z.string().optional(),
+  })
+  .withMetadata({
+    openapi: {
+      refId: "user",
+    },
+  });
+```
+
+Extract the inner schema with `.unwrap()`.
+
+```ts
+const user = z.object({
+  id: z.string(),
+  name: z.string().optional(),
+});
+const userWithMetadata = user.withMetadata({
+  openapi: { refId: "user" },
+});
+userWithMetadata.unwrap() === user; // true
+```
+
+You can extract metadata nested inside `.optional()`,
+`.nullable()`, `.default()`, `z.lazy()`, `.transform()`,
+`.catch()`, `.brand()`, `.pipeline()` output, or `.promise()`
+with `z.extractMetadata`:
+
+```ts
+const user = z.object({
+  id: z.string(),
+  name: z.string().optional(),
+});
+const metadata = {
+  openapi: { refId: "user" },
+};
+const userWithMetadata = user.withMetadata(metadata);
+const maybeUserPromise = userWithMetadata.nullable().promise();
+z.extractMetadata(maybeUserPromise) === metadata; // true
+type M = z.extractMetadata<typeof maybeUserPromise>; // typeof metadata
+```
+
+You can extract metadata nested inside the above schemas or
+also `.array()` or `z.set()` with `z.extractDeepMetadata`:
+
+```ts
+const user = z.object({
+  id: z.string(),
+  name: z.string().optional(),
+});
+const metadata = {
+  openapi: { refId: "user" },
+};
+const userWithMetadata = user.withMetadata(metadata);
+const maybeUserArray = userWithMetadata.array().nullable();
+z.extractDeepMetadata(maybeUserArray) === metadata; // true
+type M = z.extractDeepMetadata<typeof maybeUserArray>; // typeof metadata
 ```
 
 ## Objects
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -80,6 +80,7 @@
 - [Native enums](#native-enums)
 - [Optionals](#optionals)
 - [Nullables](#nullables)
+- [Metadata](#metadata)
 - [Objects](#objects)
   - [`.shape`](#shape)
   - [`.keyof`](#keyof)
@@ -1101,6 +1102,88 @@ Extract the inner schema with `.unwrap()`.
 const stringSchema = z.string();
 const nullableString = stringSchema.nullable();
 nullableString.unwrap() === stringSchema; // true
+```
+
+## Metadata
+
+You can associate metadata with a schema with `z.withMetadata()`.
+
+```ts
+const user = z.withMetadata(
+  z.object({
+    id: z.string(),
+    name: z.string().optional(),
+  }),
+  {
+    openapi: {
+      refId: "user",
+    },
+  }
+);
+```
+
+Or use the `.withMetadata()` method.
+
+```ts
+const user = z
+  .object({
+    id: z.string(),
+    name: z.string().optional(),
+  })
+  .withMetadata({
+    openapi: {
+      refId: "user",
+    },
+  });
+```
+
+Extract the inner schema with `.unwrap()`.
+
+```ts
+const user = z.object({
+  id: z.string(),
+  name: z.string().optional(),
+});
+const userWithMetadata = user.withMetadata({
+  openapi: { refId: "user" },
+});
+userWithMetadata.unwrap() === user; // true
+```
+
+You can extract metadata nested inside `.optional()`,
+`.nullable()`, `.default()`, `z.lazy()`, `.transform()`,
+`.catch()`, `.brand()`, `.pipeline()` output, or `.promise()`
+with `z.extractMetadata`:
+
+```ts
+const user = z.object({
+  id: z.string(),
+  name: z.string().optional(),
+});
+const metadata = {
+  openapi: { refId: "user" },
+};
+const userWithMetadata = user.withMetadata(metadata);
+const maybeUserPromise = userWithMetadata.nullable().promise();
+z.extractMetadata(maybeUserPromise) === metadata; // true
+type M = z.extractMetadata<typeof maybeUserPromise>; // typeof metadata
+```
+
+You can extract metadata nested inside the above schemas or
+also `.array()` or `z.set()` with `z.extractDeepMetadata`:
+
+```ts
+const user = z.object({
+  id: z.string(),
+  name: z.string().optional(),
+});
+const metadata = {
+  openapi: { refId: "user" },
+};
+const userWithMetadata = user.withMetadata(metadata);
+const maybeUserArray = userWithMetadata.array().nullable();
+z.extractDeepMetadata(maybeUserArray) === metadata; // true
+type M = z.extractDeepMetadata<typeof maybeUserArray>; // typeof metadata
 ```
 
 ## Objects
--- a/deno/lib/__tests__/metadata.test.ts
+++ b/deno/lib/__tests__/metadata.test.ts
@@ -1,0 +1,213 @@
+import { expect } from "https://deno.land/x/expect@v0.2.6/mod.ts";
+const test = Deno.test;
+
+import { util } from "../helpers/util.ts";
+import * as z from "../index.ts";
+
+const metadata = { foo: "bar" as const };
+
+test("extractMetadata from ZodTypeAny", () => {
+  const schema = z.any();
+  util.assertEqual<z.extractMetadata<typeof schema>, never>(true);
+  expect(z.extractMetadata(schema)).toEqual(undefined);
+});
+test("extractMetadata from ZodObject", () => {
+  const schema = z.object({ a: z.string().withMetadata(metadata) });
+  util.assertEqual<z.extractMetadata<typeof schema>, never>(true);
+  expect(z.extractMetadata(schema)).toEqual(undefined);
+});
+test("extractMetadata from ZodUnion", () => {
+  const schema = z.union([
+    z.string().withMetadata(metadata),
+    z.number().withMetadata(metadata),
+  ]);
+  util.assertEqual<z.extractMetadata<typeof schema>, never>(true);
+  expect(z.extractMetadata(schema)).toEqual(undefined);
+});
+test("extractMetadata from ZodLazy of ZodAny", () => {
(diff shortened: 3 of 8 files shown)


bb010g:
In #2471, you provide a sample usage of metadata with `.from()`. What would that look like on this implementation?

colinhacks:
Here's some things I don't like:

- `ZodMetadata` existing as a no-op class just to hold data
- People getting confused and mad because they can't do stuff like`z.number().metadata({}).min(1)` 
- Code that looks like `extractMetadata`. This is the same concern I've expressed with `deepPartial` and `discriminatedUnion` - if you need to do a huge switch statement over every Zod subclass, you should probably try to think of a better API

I haven't done a good job of expressing all of this. I've also been horribly busy but now I have a window to work on Zod 4 full time! So I'm going to do my darnedest to explain why I've been so against metadata, and hopefully propose a solution that will make people happy,

I have some ideas around strongly-typed metadata in Zod 4 that I'll write up as an RFC soon. I agree this is becoming more and more important as @rattrayalex and others have mentioned. People clearly want & need to attach metadata to their schemas. But letting them attach any untyped object with a `.metadata()` property doesn't help anyone and leads to bad/messy code.

98% of people want metadata so they can use their Zod schemas as a source of truth, and generate other artifacts (JSON Schemas, etc) from them. I may be over-indexing on JSON schema, but generally there are only two places you want to attach metadata: on *objects* and on *properties*. (Right? RIGHT!?)

It would be easy to let users specify metadata on *objects* in a typesafe way. You use [*declaration merging*](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) to let users specify an interface, then `z.object()` can detect that and accept metadata in a typesafe way. This is the same principle `zod-to-openapi` uses to add methods to `ZodObject`: https://github.com/asteasolutions/zod-to-openapi/commit/dc01991ec8e8e8fbe1e0707ab9607d4fead29ec9#diff-10348b603df97619f3ed4b79057f539ca90ef8ea17f228a11212c9150605b4b1L37-L58

Here's how it would look to an end user:
```ts
declare module "zod" {
  interface ZodObjectMetadata {
    $id: string
  }
}

z.object({ name: z.string() }, {
  metadata: { $id: "User" } // fully typed
})
```

Keep in mind that this would apply to *all invokations* of `z.object()`. There's no way to specify different metadata on a per-invokation or per-module basis. 

I'm also not sure if I could prevent a user from doing `z.object({ name: z.string() })` — i.e. not specifying a params object at all if `ZodObjectMetadata` is defined — or if that would even be desirable.

---

There's already a lot of issues there, and we haven't even discussed *property metadata* :(  Here's a naive solution:

```ts
declare module "zod" {
  interface ZodObjectMetadata {
    description: string
  }
}

z.object({
  name: z.string().metadata({ 
    description: "First name" // typesafe!
  })
})
```

We can provide some typesafety in the `.metadata()` call, but we can't prevent users from simply *not calling* `.metadata()`.

```ts
z.object({
  name: z.string()
})
```

Or rather, we probably could but it would require one of two things: adding a `ZodMetadata` class (no) or adding a statically tracked generic to every `ZodType` subclass that would track the inferred metadata type of each schema (also no). 

To explain that second point a little deeper. Right now when you call `z.string()`, you get out a `ZodString`. If you want the ability to infer metadata and have it be strongly typed, you would need to add a generic onto that class:

```ts
class ZodString<Meta = null> {
    _metadata: Meta
  metadata<T>(data: T): ZodString<T>
}
```

Now when you call `z.string()`, you see this in Intellisense:

```ts
z.string(); // ZodString<null>
```

It seems like a little thing but this would have to be added to *every subclass*, muddy up every Intellisense autocompletion, and make life a little bit worse for all users of Zod.

---

To add complexity to this, Zod is *compositional*. That means when you call `z.string().optional()`, you get back a `ZodOptional` instance wrapping a `ZodString` instance. So what happens if you have this?

```ts
z.string().metadata({ description: "Sup" }).optional()
```

That's right, you have a `ZodOptional` instance with *no metadata*, because the metadata was attached to the inner `ZodString`. That means Zod would have to *bubble up* the metadata through all these method calls, copying it each time to the new outer-layer. 

But then there are also edge cases. Should the metadata survive across calls to `.transform()`? The data structure *could* have been dramatically changed inside the transform to something totally unrecognizable, making the metadata inapplicable. How about `.and()`? Should we merge the metadata of the two intersected types? Who wins in the case of a conflict? 

To top it all off, it would still annoying for a third-party library to write a function can *statically enforce the metadata type*. The only way this works is if places constraints 


I suppose this is the kind of stuff that only is apparent when you start thinking through the internals of the implementation. It's complicated to explain, but people don't tend to be satisfied when I say "trust me, it's complicated and probably a really bad idea".

---

In case it isn't obvious, every idea so far is really really bad. The only solution, as far as I can see, is to introduce a generalized `ZodObject`-like "schema collection" type with first-class support for configurable metadata. It would look something like this:

```ts
// "User" is a collection that contains properties
// each property is associated with a FieldMetadata object
// and a schema that conforms to z.ZodTypeAny (any Zod schema)
type FieldMetadata = { description: string; };
const User = z.collection<FieldMetadata, z.ZodTypeAny>()
  .add("firstName", z.string(), { description: "First name" })
  .add("lastName", z.string(), { description: "Last name" });
```

In JSON Schema, you often want *collections of schemas*. Each schema should have some known, well-typed metadata, as well as each field of each schema.

The `z.collection` API can compose to accommodate this.

```ts
// "User" is a collection that contains properties
// each property is associated with a FieldMetadata object
// and a schema that conforms to z.ZodTypeAny (any Zod schema)
type FieldMetadata = { description: string; };
const User = z.collection<FieldMetadata, z.ZodTypeAny>(); 
const BlogPost = z.collection<FieldMetadata, z.ZodTypeAny>(); 

// "SchemaCollection" is a collection that contains collections
// Each  property is associated with a FieldMetadata object
// and a schema that conforms to z.ZodTypeAny (any Zod schema)
type ObjectMetadata = { $id: string };
const SchemaCollection = z
  .collection<ObjectMetadata, z.ZodCollection<FieldMetadata, z.ZodTypeAny>>()
  .add("User", User, { $id: "User" })
  .add("BlogPost", BlogPost, { $id: "BlogPost" });
```

Collections require each element to have a label. You can access this with `.items`.

```ts
User.items.firstName; // { schema: ZodString, metadata: FieldMetadata }
SchemaCollection.items.User; // { schema: User; metadata: ObjectMetadata }
``` 

When you call `.parse()`, it acts like a `ZodObject`, using the labels as keys. (Admittedly it's a little weird that `SchemaCollection` above has a `.parse()` method but whatever.)

```ts
User.parse({
  firstName: "Colin", 
  lastName: "Sleepy"
})
```

---

I've been saying since [at least 2021](https://github.com/colinhacks/zod/issues/507#issuecomment-873473128) that the right way to attach metadata to Zod schemas is not to do it at all. Instead, you should be using *composition*. Let your users pass in a Zod schema to specify *types*. If you need any additional metadata, it should be specified *alongside* the Zod schema. 

If you are building a library on top of Zod that, say, generates JSON Schema specs, just let people pass in an object that looks like this:

```ts
interface {
  schema: z.ZodType,
  metadata: WhateverYouWant
} 
```

This `ZodCollection` concept is just a general-purpose (and I suppose "first party") way to do the compositional thing I've been recommending for years.

---

There's also this approach for specifying metadata to fields using something that's closer to the existing `z.object()` API, but it's a bit verbose/obvious. I'm not a fan, but perhaps this will spark ideas in others.

```ts
z.meta<{label: string}>().object({
  firstName: { 
    $type: z.string(),
    label: "First name"
  },
  age: {
    $type: z.number(),
    label: "Age"
  }
})
```

lo1tuma:
I’m using zod already since a few years and few days ago it was the first time I had the need for adding metadata to schemas for a [graphql query builder](https://github.com/enormora/schema-hub/tree/906279a8275f8b6b3b6a6cc3aaaea1a356a1a6c3/source/zod-graphql-query-builder). The approach I ended up with, is almost completely independent from `zod` (inspired by this [comment](https://github.com/colinhacks/zod/issues/2917#issuecomment-1797153924)) utilizing `WeakMap`:

```typescript
type MyMetadata = { foo: string; };
const myMetdataRegistry = new WeakMap<z.ZodTypeAny, MyMetdata>;

function addMetadata<Schema extends z.ZodTypeAny>(schema: Schema, metadata: MyMetadata): ZodLazy<Schema> {
    const wrappedSchema = z.lazy(() => {
         return schema;
    });
     myMetdataRegistry.set(wrappedSchema, metadata)
     return wrappedSchema;
}

const mySchema = z.object({
    someProperty: addMetadata(z.string(), { foo: 'bar' })
})
```

So far it works quite well for my use-case. However I would also be happy about an easier way to handle metadata. Relying on declaration merging etc doesn’t sound easier though.
Regarding `ZodCollection` so far I like the idea. I’m not quite sure we the chaining API is needed and why it can’t look like similar to `ZodObject` where we just define the shape as an object literal. I’m also curious how inference would work via `z.infer`, do we need some kind of `build()` function when using the chaining API?

> I may be over-indexing on JSON schema, but generally there are only two places you want to attach metadata: on objects and on properties. (Right? RIGHT!?)

So far I also only had the use-case of object properties. A while ago I had the need to add a custom property to a union type so I can create better error messages, but I guess this is another topic because we would need to pass through those custom properties to `ZodIssue`.


--------------

I think it is also worth to have look at how `@effect/schema` handles [annotations](https://github.com/Effect-TS/effect/tree/9881f26ee6c44491e600c9b5206a03a3598a0e04/packages/schema#annotations).

colinhacks:
Thanks for the thoughtful response. 

> I think it is also worth to have look at how @effect/schema handles [annotations](https://github.com/Effect-TS/effect/tree/9881f26ee6c44491e600c9b5206a03a3598a0e04/packages/schema#annotations).

This seems to be totally untyped and not reflected in the type signature of the resulting schema. So there would be no way for tooling authors to statically enforce something like "give me a schema with metadata that matches SomeInterface".

This is true for your solution as well (though it does get the job done).

rattrayalex:
Really appreciate all the thought put into this @colinhacks ! I'm sadly at the end of a busy day and not able to do a thorough writeup, but in case it helps, a few gut reactions:

1. It's a hard requirement to expose a fluent API to my library's users, so `mylib.addMetadata(z.string(), {…})` is a no-go for me. I want it to to look and feel like zod, because that's what people are familiar with. If you make me do that, I'll just extend/monkeypatch/fork zod (as we've done so far, but hope not to do).
2. The `interface ZodObjectMetadata {` proposal you have is something I wouldn't love, but could probably live with. 
  a. In practice, we'd again probably extend/monkeypatch/fork zod to add things like `z.object({}).myMetaField(123)` as sugar for `z.object({}, { metadata: { $myLib: { myMetaField: 123 } } })` but at least the internals would be much cleaner / more peaceful.
3. just objects & properties sounds like an OK limitation to me at first blush
4. "we can't prevent users from simply not calling .metadata()" - this sounds fine to me fwiw. In general I think it's reasonable to not expect tremendous typesafety of metadata.
5. "adding a ZodMetadata class (no)" – FWIW it wasn't clear to me on a first skim of this document why this is a hard no (in a sea of bad options). I'm not saying I think it's _good_, nor that you failed to justify it (maybe I missed it), just that I think a strong justification is needed.

I'm quite curious what @jedwards1211 thinks, if he's around to take a look

jedwards1211:
@colinhacks 

> I've been saying since https://github.com/colinhacks/zod/issues/507#issuecomment-873473128 that the right way to attach metadata to Zod schemas is not to do it at all. Instead, you should be using composition.

I think this keeps coming up despite what you've said because composition doesn't work for what some people are trying to accomplish.

> I may be over-indexing on JSON schema, but generally there are only two places you want to attach metadata: on objects and on properties. (Right? RIGHT!?)

I'm afraid not, at least when it comes to generating OpenAPI specs from zod schemas.  One simple real-world example is a shared enum type:

```ts
const PetType = z.enum(['dog', 'cat']).metadata({ openapi: { ref: 'PetType' } })

const AddUserRequest = z.object({
  name: z.string(),
  pets: z.array(z.object({
    name: z.string(),
    type: PetType,
  })).optional(),
})

const GetUserPetsRequest = z.object({
  userId: z.string(),
  petType: PetType.optional(),
})

const GetUserPetsResponse = z.array(z.object({
  name: z.string(),
  type: PetType,
}))

...
```

This way of doing things allows `PetType` to be used in many different places, while still being able to output `{ "ref": "#/components/schemas/PetType" }` in all the corresponding places in the generated OpenAPI spec, instead of inlining a copy of the whole type in each place.

Without being able to attach the metadata to `PetType` itself, you'd have to attach it to every property that uses the enum type, which would be a lot of repeating yourself.

Can you illustrate how you would solve a problem like this with composition?  I can't imagine a useful way to do it.

jedwards1211:
> That's right, you have a ZodOptional instance with no metadata, because the metadata was attached to the inner ZodString. That means Zod would have to bubble up the metadata through all these method calls, copying it each time to the new outer-layer.

That's okay and often makes the most sense.  To take my last example, `z.enum(['dog', 'cat']).metadata({ openapi: { ref: 'PetType' } }).optional()`, you don't want the `PetType` in `#/components/schemas` to be optional.  Instead you want to wrap `$ref`s to it in optional schemas.

colinhacks:
On further reflection, I think there should probably just be new subclasses for `ZodJsonSchema` and `ZodJsonSchemaProperty`. These can be designed in Zod core, or a new package (`zod-json-schema`), or some kind of plugin. This would make it possible to design the API for exactly this use case. They'll extend `ZodType` so they'll be compatible with all existing Zod schemas. Maybe something like:

```ts
import * as z from "zod";
import "zod/plugins/json-schema";

const User = z.jsonschema.schema({ 
  $id: "User",
  properties: {
    firstName: z.jsonschema.property({
      description: "First name",
      type: z.string()
    })
  }
})
```

rattrayalex:
@colinhacks Yeah, I think something that could work well! (Assuming that third parties can add other things like `z.mylib.property({})` and those can bear typesafe metadata).

rattrayalex:
In case it's useful, I just came across a similar project to ours, which wraps zod with a `.openapi()` helper: https://hono.dev/snippets/zod-openapi 

I haven't yet dug into the source code to see how it's implemented, but @colinhacks it may be interesting to you.

colinhacks:
I'm going to close this. Some combination of a new `ZodJsonSchema` class and [plugins](https://github.com/colinhacks/zod/pull/3445) will be the recommended way to handle metadata in Zod 4.

dlvhdr:
@colinhacks Any news on the RFC? something we can follow

Issue #2062, opened 2023-02-19
Strict by default or strict deep


hornta:
Hi, I would like to either have strict on all of my types or something like `.strict()` but that propagates down on all sub objects in my schema. Now it's too easy to miss out on defining a schema as strict I think.

gawi:
I also think it would be worthwhile to have this in the API (and possibly `deepPassthrough()` and `deepStrip()`). 

If you're in a hurry, I think you can implement it yourself. Take a look at `deepPartial`+`deepPartialify` for inspiration.

```js
  deepPartial(): partialUtil.DeepPartial<this> {
    return deepPartialify(this) as any;
  }
```

```js
function deepPartialify(schema: ZodTypeAny): any {
  if (schema instanceof ZodObject) {
    const newShape: any = {};

    for (const key in schema.shape) {
      const fieldSchema = schema.shape[key];
      newShape[key] = ZodOptional.create(deepPartialify(fieldSchema));
    }
    return new ZodObject({
      ...schema._def,
      shape: () => newShape,
    }) as any;
  } else if (schema instanceof ZodArray) {
    return ZodArray.create(deepPartialify(schema.element));
  } else if (schema instanceof ZodOptional) {
    return ZodOptional.create(deepPartialify(schema.unwrap()));
  } else if (schema instanceof ZodNullable) {
    return ZodNullable.create(deepPartialify(schema.unwrap()));
  } else if (schema instanceof ZodTuple) {
    return ZodTuple.create(
      schema.items.map((item: any) => deepPartialify(item))
    );
  } else {
    return schema;
  }
}

```

gawi:
I've tried to generalize to all kind of deep unknown policies. Here's what I got:

First, you need 2 utilities: a function and a type.

```ts
type ZodObjectMapper<T extends ZodRawShape, U extends UnknownKeysParam> = (
  o: ZodObject<T>
) => ZodObject<T, U>;

function deepApplyObject(
  schema: ZodTypeAny,
  map: ZodObjectMapper<any, any>
): any {
  if (schema instanceof ZodObject) {
    const newShape: Record<string, ZodTypeAny> = {};
    for (const key in schema.shape) {
      const fieldSchema = schema.shape[key];
      newShape[key] = deepApplyObject(fieldSchema, map);
    }
    const newObject = new ZodObject({
      ...schema._def,
      shape: () => newShape,
    });
    return map(newObject);
  } else if (schema instanceof ZodArray) {
[...]

PR #2512, opened 2023-06-19
[docs] add notice about strictObject

cupcakearmy:
Found `strictObject` in [this ticket](https://github.com/colinhacks/zod/issues/2062) and i wondered if i missed it in the docs, but it cannot be found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1340,6 +1340,8 @@ person.parse({
 // => throws ZodError
 ```
 
+For convenience, you can use `z.strictObject({})`, which is the equivalent of `z.object({}).strict()`.
+
 ### `.strip`
 
 You can use the `.strip` method to reset an object schema to the default behavior (stripping unrecognized keys).


cupcakearmy:
Def not stale my bot

vixalien:
In the issue, I also saw that it is "the recommended approach". I just find it weird that it's not recommended anywhere 🤔

cupcakearmy:
@colinhacks ping ❤️

colinhacks:
I'd prefer people not use this, tbh. It opens up the door to a bunch of top-level APIs that are better kept as methods (e.g. `z.optionalString()`). I generally don't like having multiple ways of achieving the same thing.

Also, there will likely be a way to configure this setting globally in Zod 4, which solves the core problem better than `z.strictObject()` anyway.

---

> In the issue, I also saw that it is "the recommended approach". I just find it weird that it's not recommended anywhere 🤔

cc @vixalien  It is the recommended approach for achieving the OP's suggestion of a `deepStrict()`, not in the general case.

cupcakearmy:
Sure :) 
Global option would be reaaallyy nice!
Thanks for the time, so you on v4 🤘

paradoxloop:
@colinhacks 

Did there end up being a way to configure this globally in Zod 4 or a way to set a schema to deep strict?

> I appreciate the PR, though tbh I'd prefer people not use this. It opens up the door to a bunch of top-level APIs that are better kept as methods (e.g. `z.optionalString()`). I generally don't like having multiple ways of achieving the same thing.
> 
> Also, there will likely be a way to configure this setting globally in Zod 4, which solves the core problem better than `z.strictObject()` anyway.

PR #2517, opened 2023-06-20
feat: add unique modifier to array

jrel:
Resolves  #2316


colinhacks:
This is a good PR, but I don't think this feature necessarily merits getting added to Zod core. Zod needs to be careful about bundle size, and even small features like this add up. 

It also introduces some potential for confusion from people who may expect Zod to perform deep equality on objects, instead of simple referential equality (as implemented in your PR). 

I appreciate the PR though, it was excellent.

PR #2522, opened 2023-06-20
fix: support local time if no offset

### LINKED ISSUE #2385: Datetime validation without offset should allow omission of Z zone designator
ISO-8601 allows datetime strings to not have any timezone offset or _Z_ zone designator. In these instances, it is assumed to be in local time. (Reference: https://en.wikipedia.org/wiki/ISO_8601#Local_time_(unqualified))

However, the datetime validator currently marks these as invalid.

```ts
const schema = z.string().datetime() // offset: false is default behavior
console.log( schema.safeParse( '2022-01-01T00:00:00' ).success ) // false
```

bchrobot:
See https://en.wikipedia.org/wiki/ISO_8601#Local_time_(unqualified)

Closes #2385

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,9 @@
 - [Coercion for primitives](#coercion-for-primitives)
 - [Literals](#literals)
 - [Strings](#strings)
-  - [ISO datetimes](#iso-datetimes)
+  - [Datetimes](#datetimes)
+  - [Dates](#dates)
+  - [Times](#times)
   - [IP addresses](#ip-addresses)
 - [Numbers](#numbers)
 - [BigInts](#bigints)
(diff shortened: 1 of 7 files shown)


jonathanj:
@bchrobot Out of interest, why does the case where precision is 0 and offset is not required not make `Z` optional too?

What is still outstanding to get this change reviewed?

jonathanj:
> > @bchrobot Out of interest, why does the case where precision is 0 and offset is not required not make `Z` optional too?
> 
> Is your question about `zod`'s current implementation? Or the ISO standard?

Sorry, I wanted to comment on the exact line but because it falls into a "not changed" area, I couldn't do that.

I'm referring to [this code](https://github.com/colinhacks/zod/blob/5078c8ff72512788f44866c11a5862cb46f7fc37/src/types.ts#L586-L594). Which looks like it's part of the same case as `if (args.precision)` but just caters for a difference in the regex. I would have expected `Z` to be optional in this case too (at least when `args.offset` is false).

I'm just curious if there is a reason it's not.

> My guess is difficulty getting tests to run/pass locally but it's been a little bit since I last worked on this. I can look later this week.

That would be very much appreciated, thank you!

colinhacks:
I'd say this should be behind a `unqualified: true` flag, in the same way that `offset: true` means "allow offset". The default behavior should allow `Z` only.

colinhacks:
This has landed in Zod 3.23.

buti1021:
Thank you for zod and this PR!

This should be documented, had to look into the source code to find this option. Short on time right now, but might find the opportunity soon™

PR #2620, opened 2023-07-31
perf: speed up `addQuestionMarks`

Andarist:
Just an experiment to speed up `addQuestionMarks`. I wonder if the change is 100% semantically the same and what your CI might tell us about it.

### DIFF
--- a/src/helpers/util.ts
+++ b/src/helpers/util.ts
@@ -101,20 +101,21 @@ export namespace objectUtil {
     [k in Exclude<keyof U, keyof V>]: U[k];
   } & V;
 
-  type optionalKeys<T extends object> = {
-    [k in keyof T]: undefined extends T[k] ? k : never;
-  }[keyof T];
-  type requiredKeys<T extends object> = {
-    [k in keyof T]: undefined extends T[k] ? never : k;
-  }[keyof T];
-  type pickRequired<T extends object, R extends keyof T = requiredKeys<T>> = {
-    [k in R]: T[k];
+  // type optionalKeys<T extends object> = {
+  //   [k in keyof T]: undefined extends T[k] ? k : never;
+  // }[keyof T];
+
+  type keepRequiredKeys<T extends object> = {
+    [k in keyof T as undefined extends T[k] ? never : k]-?: T[k];
   };
-  type pickOptional<T extends object, O extends keyof T = optionalKeys<T>> = {
-    [k in O]?: T[k];
-  };
-  export type addQuestionMarks<T extends object> = pickRequired<T> &
-    pickOptional<T> & { [k in keyof T]?: unknown };
+
+  // type alkjsdf = addQuestionMarks<{ a: any }>;
+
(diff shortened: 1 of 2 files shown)


src/helpers/util.ts
```diff
@@ -104,17 +104,16 @@ export namespace objectUtil {
   //   [k in keyof T]: undefined extends T[k] ? k : never;
   // }[keyof T];
 
-  type requiredKeys<T extends object> = {
-    [k in keyof T]: undefined extends T[k] ? never : k;
-  }[keyof T];
+  type keepRequiredKeys<T extends object> = {
+    [k in keyof T as undefined extends T[k] ? never : k]-?: T[k];
```
Andarist, line 108: if some semantic difference is found using this approach then we could also try this
```suggestion
    [k in keyof T as undefined extends T[k] ? never : k]-?: unknown;
```

colinhacks:
@Andarist Thanks for this! The `addQuestionMarks` (and associated test suite) have evolved underneath this PR. tldr: I couldn't get this implementation to work with the current set of tests (mostly added a few weeks ago) in `generics.test.ts` and `object.test.ts`. 

You might get some mild interest out of the failing test cases. Here's one that's surprising tricky to pass (from generics.test.ts):

```ts
test("assignability", () => {
  const createSchemaAndParse = <K extends string, VS extends z.ZodString>(
    key: K,
    valueSchema: VS,
    data: unknown
  ) => {
    const schema = z.object({
      [key]: valueSchema,
    });
    const parsed = schema.parse(data);
    const inferred: z.infer<z.ZodObject<{ [k in K]: VS }>> = parsed;
    return inferred;
  };
  createSchemaAndParse("foo", z.string(), { foo: "" });
});
```

It seems the main issue with your `keepRequiredKeys` approach is that there's no longer an assertion that the keys of the resulting object extend `keyof T` due to the `as` clause. When writing to do inference or type manipulation inside generic functions, that seems to be important 🤷‍♂️

For reference here's the current implementation:

```ts
type optionalKeys<T extends object> = {
  [k in keyof T]: undefined extends T[k] ? k : never;
}[keyof T];
type requiredKeys<T extends object> = {
  [k in keyof T]: undefined extends T[k] ? never : k;
}[keyof T];
type pickRequired<T extends object, R extends keyof T = requiredKeys<T>> = {
  [k in R]: T[k];
};
type pickOptional<T extends object, O extends keyof T = optionalKeys<T>> = {
  [k in O]?: T[k];
};
export type addQuestionMarks<T extends object> = pickRequired<T> &
  pickOptional<T> & { [k in keyof T]?: unknown };
```

Andarist:
> It seems the main issue with your keepRequiredKeys approach is that there's no longer an assertion that the keys of the resulting object extend keyof T due to the as clause.

It should understand this but:
- it could depend on TS version
- I wouldn't be surprised that something might stay unhandled on TS side of things here

I'm interested in learning more about this but I also don't have time to spend on this... what to do, what to do 😭

PR #2636, opened 2023-08-06
combine shared fields into one object

oussamaBGZ:
I combined multiple union objects into one and making all the values optional for more readability also created a separated discriminated union for "kind" property.


colinhacks:
Thanks, but the current version is more correct and typesafe, as evidenced by the number of type casts required in your PR. It should really be a top-level discriminated union.

Issue #2639, opened 2023-08-09
String format request: xid


redxtech:
In line with `UUID`, `ULID`, `CUID`, and `CUID2`, I think it would be great to add an additional format as well: `xid`.

It seems like it's fairly simple and well supported, and I'm looking to use it in a project that I use `zod` for, so support would be great.

Here's the source repo: https://github.com/rs/xid
Here's a typescript implementation: https://github.com/yiwen-ai/xid-ts

I can start on a PR if that would be appreciated!

PR #2640, opened 2023-08-09
add xid string validator

redxtech:
I've added an additional string validator to validate `xid` formatted strings.

Implements the feature requested by #2639 

I've pretty much just done a global search on the codebase for `cuid2`/`ulid`, and added a matching entry for `xid` at each match, making slight changes where necessary.

Let me know if I'm missing anything!

### DIFF
--- a/README.md
+++ b/README.md
@@ -865,6 +865,7 @@ z.string().nanoid();
 z.string().cuid();
 z.string().cuid2();
 z.string().ulid();
+z.string().xid();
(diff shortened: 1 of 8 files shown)


colinhacks:
Thanks this looks great! I'm merging this into the v4 branch. Some version of this will land in Zod 4, but (as a heads up) it may be a while, and this feature/API will likely undergo some changes in the interim.

PR #2648, opened 2023-08-11
Update README for z.enum dynamic values 

moseslucas:
# What this PR do

* Updates the `README.md` and add details on how to let `z.enum()` accepts dynamic values

## Current Doc
It says about `zod.enum` not able to infer 
Screenshot 2023-08-11 at 9 06 10 PM

## Updated Doc
Suggests type assertion that will allow dynamic values to `z.enum()`
Screenshot 2023-08-11 at 9 06 05 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -962,6 +962,13 @@ const fish = ["Salmon", "Tuna", "Trout"];
 const FishEnum = z.enum(fish);
 ```
 
+But you may assert with `[string]` to let `z.enum` accept dynamic values.
+
+```ts
+const fish = ["Salmon", "Tuna", "Trout"];
+const FishEnum = z.enum(fish as [string]);
+```
+
 **Autocompletion**
 
 To get autocompletion with a Zod enum, use the `.enum` property of your schema:
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -962,6 +962,13 @@ const fish = ["Salmon", "Tuna", "Trout"];
 const FishEnum = z.enum(fish);
 ```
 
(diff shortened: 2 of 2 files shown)


colinhacks:
This pattern is still not recommended so I'd rather not document it. I can't see any reason not to use `as const` here, especially since you're already doing a cast. Is there are reason you're trying to do this?

moseslucas:
> This pattern is still not recommended so I'd rather not document it. I can't see any reason not to use `as const` here, especially since you're already doing a cast. Is there are reason you're trying to do this?

Hi @colinhacks , I think my use case is rare but here's how I end up doing it that way.

I have a json file that stores a list of slugs
```
/* dataSlugs.json */
["one", "two", "three"]
```

Then I import this json and pass it as parameter to z.enum
```
import slugs from './dataSlugs.json'

export const mySchema = z.enum(slugs) /* this throws a type error */
export const mySchema = z.enum(slugs as const) /* this throws a type error */
```

This is the error.
Screenshot 2023-08-22 at 10 48 57 AM


You can't parse `as const` when the value came from a json file. But parsing it with `as [string]` works for me.
Additional context why I'm importing the json file is that I'm working on a monorepo. So I'm sharing a list of slugs between the backend api and the frontend UI app.

colinhacks:
I'm not a fan of this because `as [string]` (a 1-element tuple) isn't a correct type signature for `["Salmon", "Tuna", "Trout"]`.

If the elements of the tuple can't be statically known, you shouldn't be using `z.enum()`. The implement of ZodEnum expects to statically know all the possible values. 

Perhaps Zod should implement `z.string().oneOf()` to support this use case?

PR #2652, opened 2023-08-14
Initial prototype fix for issue #2651

### LINKED ISSUE #2651: zod.function as property of zod.object loses `this`
Quoting myself from discord:

```
const MyObj = z.object({ foo: z.function() });
class MyObjImpl implements z.infer<typeof MyObj> { private bar: string; constructor(){this.bar = 'baz';} foo() { console.log(this.bar); } };
const inst = MyObj.parse(new MyObjImpl()); //ok
inst.foo(); // kablooey :( "this is undefined"
```

The problem is that the parsed object's parsed method is an arrow function, and thus has a "this" of the ZodFunction (see: https://github.com/colinhacks/zod/blob/1ecd6241ef97b33ce229b49f1346ffeee5d0ba74/src/types.ts#L3755). A similar problem applies to async functions.

I can think of two possible solutions here. First is to make the parsed result a normal function instead of an arrow function, then apply its `this` to the underlying `fn`. The second is to use proxies on either the parsed function, the parsed object, or both to get a similar behavior.

I'll have a not-yet-mergeable PR for the former strategy up for this shortly. Would love help getting it over the line.

ddurschlag:
Uses a non-arrow function to maintain `this`.

Closes #2651

### DIFF
--- a/deno/lib/__tests__/function.test.ts
+++ b/deno/lib/__tests__/function.test.ts
@@ -29,6 +29,19 @@ test("function inference 1", () => {
   util.assertEqual<func1, (k: string) => number>(true);
 });
 
+test("method parsing", () => {
+	const methodObject = z.object({
+		property: z.number(),
+		method: z.function().args(z.string()).returns(z.number())
+	});
+	const methodInstance = {
+		property: 3,
+		method: function(s: string) { return s.length + this.property; }
+	};
+	const parsed = methodObject.parse(methodInstance);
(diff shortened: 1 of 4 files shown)


ddurschlag:
Known problems:
1) Only applies to sync functions (easy fix, just haven't done it)
2) Not clear if there's a performance impact
3) Not clear if this is better than a proxy-based approach
4) Involves `this`-aliasing that eslint does not approve of
5) Test might be considered anemic? Not sure if I should be adding other tests.

Would love maintainer opinions or code suggestions on any of the above.

PR #2656, opened 2023-08-15
Enhance ZodArray `.min()` and `.length()` Types

### LINKED ISSUE #2658: Stricter Array Types
Hey crew!

We rely on [noUncheckedIndexedAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) on our repos so we rely on nonempty array types.

image

This makes enforcing array type lengths very important.

This proposes enhancements to `min` and `length` checks to create even more accurate types which can work well with `noUncheckedIndexedAccess`.

```ts
z.string().min(3);
// [string, string, string, ...string[]];

z.string().length(3);
// [string, string, string];
```

We could type the following
```ts

type MaxArray<
  T,
  Max extends number,
  Min extends number = 0,
  R extends T[] = Tuple<T, Max>
> = R["length"] extends Min
  ? Min extends 0
    ? []
    : never
  : R extends [infer U, ...infer V]
  ? U extends undefined
    ? never
    : V extends T[]
    ? R | MaxArray<T, Max, Min, V>
    : never
  : never;

z.string().max(3);
// [] | [string] | [string, string] | [string, string, string];
```

But that would require a little bit of rework to pipe min and max cardinality around and probably isn't all that helpful in general.

samchungy:
Resolves #2658

### DIFF
--- a/README.md
+++ b/README.md
@@ -1425,11 +1425,15 @@ const nonEmptyStrings = z.string().array().nonempty({
 
 ```ts
 z.string().array().min(5); // must contain 5 or more items
-z.string().array().max(5); // must contain 5 or fewer items
-z.string().array().length(5); // must contain 5 items exactly
-```
+// the inferred type is now
+// [string, string, string, string, string, ...string[]]
 
-Unlike `.nonempty()` these methods do not change the inferred type.
+z.string().array().max(5); // must contain 5 or fewer items
+
+z.string().array().length(5); // must contain 5 items exactly
+// the inferred type is now
+// [string, string, string, string, string]
+```
 
 ## Tuples
 
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -1425,11 +1425,15 @@ const nonEmptyStrings = z.string().array().nonempty({
 
 ```ts
 z.string().array().min(5); // must contain 5 or more items
-z.string().array().max(5); // must contain 5 or fewer items
-z.string().array().length(5); // must contain 5 items exactly
-```
+// the inferred type is now
+// [string, string, string, string, string, ...string[]]
 
-Unlike `.nonempty()` these methods do not change the inferred type.
+z.string().array().max(5); // must contain 5 or fewer items
+
+z.string().array().length(5); // must contain 5 items exactly
+// the inferred type is now
+// [string, string, string, string, string]
+```
 
 ## Tuples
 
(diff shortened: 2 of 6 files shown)


colinhacks:
I dig the concept but this particular incarnation won't fly. That recursive alias can really hurt performance under the best of circumstances, and I'm sure there are many usages of `z.number().length(10000)` in the wild.

Since this ultimately resolves to a tuple type I'd be more amenable to keeping separate from ZodArray. (In retrospect even `.nonempty()` should probably have returned a `ZodTuple`). 

Maybe a `.tuple()` method on ZodType?

```ts
z.string().tuple(3);
// ZodTuple<[ZodString, ZodString, ZodString]>
```

samchungy:
Closing to rethink

damian-balas:
@samchungy @colinhacks Any progress on this?

I need this type 
```ts
type TypeINeed = [number, number];
type TypeItProduces = [number, ...number[]];
```

I've used this methonds, but none returned infered type as: `[number,number]`
```ts
{
  rooms: z.array(z.number()).nonempty().min(2).max(2).length(2),
  rooms: z.number().array().min(2).nonempty(),
  rooms: z.number().array().min(2).max(2).nonempty(),
}
```

Is there a way in the current version?

samchungy:
> @samchungy @colinhacks Any progress on this?
> 
> I need this type
> 
> ```ts
> type TypeINeed = [number, number];
> type TypeItProduces = [number, ...number[]];
> ```
> 
> I've used this methonds, but none returned infered type as: `[number,number]`
> 
> ```ts
> {
>   rooms: z.array(z.number()).nonempty().min(2).max(2).length(2),
>   rooms: z.number().array().min(2).nonempty(),
>   rooms: z.number().array().min(2).max(2).nonempty(),
> }
> ```
> 
> Is there a way in the current version?

Sounds like you're after a tuple rather than an array

dca123:
I've a usecase where I'm using length to ensure that I only have 8 elements in the array. Figured i should use tupples instead. However, I don't wanna wanna write those elements, 8 times in the array in my code (would be quite verbose) so i built this lil helper function to do that for me.

```
type TimesTuple<
  N extends number,
  T,
  R extends Array<T> = [],
> = R['length'] extends N ? R : TimesTuple<N, T, [...R, T]>;

function times<N extends number, T>(n: N, obj: T): TimesTuple<N, T> {
  const returnArr = [];
  for (let i = 0; i < n; i++) {
    returnArr.push(obj);
  }
  return returnArr as TimesTuple<N, T>;
}

```

would it make sense to have an optional param as the second parameter to have this repeating feature ? wonder how people commonly use the tuples feature 🤔

PR #2659, opened 2023-08-15
Use Set.has instead of Array.indexOf for enum comparison (perf improvement)

jmike:
Improve zod's performance by using `Set.has` instead of `Array.indexOf` when checking for the presence of a value inside `ZodEnum` or `ZodNativeEnum`.

`Array.indexOf` is especially problematic when passing multiple (e.g. 30+) values to an enum. This PR introduces a very simple change that turns an `O(n)` operation to `O(1)`.

#### Before this PR

benchmark-results-latest-master

#### After this PR

benchmark-results-using-set

### Conclusions

- Using `Set.has` we were able to increase the performance of the `parse` method in `ZodEnum` and `ZodNativeEnum` by 20%.
- The more the enum values the greater the benefit by using `Set` instead of `Array`.

### DIFF
--- a/deno/lib/benchmarks/index.ts
+++ b/deno/lib/benchmarks/index.ts
@@ -43,3 +43,9 @@ if (!argv.length) {
 for (const suite of suites) {
   suite.run();
 }
+
+// exit on Ctrl-C
+process.on("SIGINT", function () {
+  console.log("Exiting...");
+  process.exit();
+});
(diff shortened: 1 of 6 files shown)


colinhacks:
Thanks this is excellent! I switched to the `#cache` approach primarily to avoid including `valueSet` in the def. The def should be the minimum set of runtime information required to instantiate the schema, ideally with no redundancy. Thanks again!

colinhacks:
This has landed in Zod 3.23.

PR #2666, opened 2023-08-18
Add string jwt

### LINKED ISSUE #2946: Add support for jwt
Add support to verify jwt strings. For example
```ts
const jwtSchema = zod.string().jwt()
jwtSchema.parse()
```

danilomourelle:
Add a `.jwt()` validator in string type.

It expects for a string with 3 parts splited by `.` with the first part being a base64 encoded string that represents a objetc wit a property `type` with value `JWT`.

**It validates just format NOT signature.**

Resolves #2946

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -680,7 +680,7 @@ function isValidJwt(token: string) {
     }
 
     const [header] = tokensParts;
-    const parsedHeader = JSON.parse(Buffer.from(header, "base64").toString());
+    const parsedHeader = JSON.parse(atob(header));
 
     if (!("type" in parsedHeader) || parsedHeader.type !== "JWT") {
       return false;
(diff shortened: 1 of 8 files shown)


hudson-bruno:
Just a reminder that you left some `console.logs` in the code

danilomourelle:
Ooopss... thanks man, already removed.

deno/lib/types.ts
```diff
@@ -619,6 +620,26 @@ function isValidIP(ip: string, version?: IpVersion) {
...
 
+function isValidJwt(token: string) {
+  try {
+    const tokensParts = token.split(".");
+    if (tokensParts.length !== 3) {
+      return false;
+    }
+
+    const [header] = tokensParts;
+    const parsedHeader = JSON.parse(Buffer.from(header, "base64").toString());
```
kanweiwei, line 631: It should works both in node.js and modern browsers, so you can't use `Buffer`. You can take a look at the code. https://github.com/validatorjs/validator.js/blob/master/src/lib/isJWT.js

danilomourelle: Sorry, not a front guy 🙈. Replace `Buffer` for `atob`. Also add option to pass algorithm used in JWT

nikelborm:
Why wasn't this PR merged yet? Does it need some improvements?

danilomourelle:
Hey guys, anything else I should do to it be accept?

danilomourelle:
Hello @colinhacks 

I believe this PR are good to go. Let me know if need anything else.

colinhacks:
Thanks, this is excellent stuff! Apologies for the long turnaround time. I'm hesitant to merge features like this into Zod 3 for bundle size reasons. Zod 4 will have better tree shaking characteristics so I'm merging this into the v4 branch. Some version of this will land in Zod 4, but (as a heads up) it may be a while, and this feature/API will likely undergo some changes in the interim.

PR #2675, opened 2023-08-22
fix: .partial() produced "undefined" in props #635

### LINKED ISSUE #635: .optional() and --exactOptionalPropertyTypes
Currently, .optional() behaves like:

```typescript
const user = z.object({
  username: z.string().optional(),
});
type C = z.infer<typeof C>; // { username?: string | undefined };
```

This results in a type mismatch when you're using --exactOptionalPropertyTypes in TypeScript 4.4 and expecting type ```C```  to match an interface definition like:

```typescript
interface User {
  username?: string;
}
```

.partial(), .partialBy(), .deepPartial() all have the same issue.

It would be nice to unbundle the optionality of the key from the union type with undefined for the value.

I suggest that these methods be changed to specify the optional absence of the key by default, and perhaps accept an option to restore the old behavior of adding ```.or(z.undefined())``` to the value schema(s). This would unfortunately be a breaking change, but it makes more sense than the current behavior, especially as more projects adopt --exactOptionalPropertyTypes.

bazuka5801:
⚠️  There may have been a  breaking change in PR, I don't know.
Fixes #635
All tests passed except `all-errors`, now a message can have undefined type and it should be checked by ? (question mark, before referring to it)

This patch can be installed for testing by adding the following lines to `package.json`.
```json
"zod": "npm:@bazuka5801/zod@3.22.2"
```

### DIFF
--- a/configs/tsconfig.base.json
+++ b/configs/tsconfig.base.json
@@ -13,6 +13,7 @@
     "noUnusedParameters": true,
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
+    "exactOptionalPropertyTypes": true,
     "downlevelIteration": true,
     "isolatedModules": true
   },
--- a/deno/lib/ZodError.ts
+++ b/deno/lib/ZodError.ts
@@ -38,7 +38,7 @@ export type ZodIssueCode = keyof typeof ZodIssueCode;
 
 export type ZodIssueBase = {
   path: (string | number)[];
-  message?: string;
+  message?: string | undefined;
 };
 
 export interface ZodInvalidTypeIssue extends ZodIssueBase {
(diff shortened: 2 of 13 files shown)


colinhacks:
I appreciate the work on this, but this is not a complete solution to #635.

Zod needs to provide a way to opt into `exactOptionalPropertyTypes` at runtime, so `ZodObject` will properly parse optional values. That is, differentiate between a literal `undefined` and a key missing entirely from the input. There also needs to be a way to differentiate between these two types (as defined w/ Zod schemas):

```ts
{ name?: string }
{ name?: string | undefined }
```

Solving this is one of the main goals of Zod 4 and I'll be publishing an RFC with my proposal soon.

ScreamZ:
@colinhacks now that zod4 is released i still can't see `exactOptionalPropertyTypes` any update ? :)

PR #2708, opened 2023-09-05
feat: supports optional tuple elements

morgs32:
I see that the last test on tuples *for this feature* was commented out... hopefully just for want of a little more work?
Let me know?! Forgive me if there are any instructions I should have followed in creating this PR.

Also see the type assertions. I'm trying to use tuples as a schema for the params of a function with optional parameters. So `[number | undefined]` won't match, it has to be something like `[number?]`.


morgs32:
@colinhacks I think this one is a small footprint, no new features, just a fix. It adds functionality that many might presume is true of the z.tuple() method.
I'm sure you're busy - this'll be the only time I ping you!

colinhacks:
Thanks! I'm merging this as a starting point, but it'll undergo a lot of change before then. Some variant of this will land in Zod 4 but it'll be a while.

Issue #1983, opened 2023-02-03
.deepPartial with union support


aphex:
Given the following example comparing `.deepPartial` to `PartialDeep` from type-fest.

```ts
import { z } from 'zod'
import type { PartialDeep } from 'type-fest'

const ZodFoo = z.object({
  bar: z.object({
    zed: z.union([
      z.object({
        a: z.string(),
      }),
      z.object({
        b: z.string(),
      }),
    ]),
  }),
})

type Foo = z.infer<typeof ZodFoo>

const ZodPartialFoo = ZodFoo.deepPartial()
type PartialFoo = z.infer<typeof ZodPartialFoo>

let f: PartialFoo
// f.bar?.zed // a and b are not optional

let g: PartialDeep<Foo>
// g.bar?.zed // a and b are optional
```

Is it possible zod `deepPartial` can handle unions making all properties of each type in the union optional as well?

galmarz:
@colinhacks should this be closed as `deepPartial` is now deprecated?

PR #2714, opened 2023-09-06
Add support for unions in `deepPartial`

kaciakmaciak:
Resolves https://github.com/colinhacks/zod/issues/1983

I could do a similar PR for the discriminated union.


colinhacks:
`.deepPartial` has been deprecated for reasons described elsewhere. tldr: it's too complex to implement properly and there are too many edge cases (including unions).

PR #2717, opened 2023-09-07
Export jsdoc with `@deprecated` when building

amotarao:
I didn't know for a long time that `z.string().nonempty()` was deprecated.

- JSdoc was deleted in `build:cjs`, so prevented it.
- JSdoc was output in the constructor when arrow functions were used in the class, so prevented it.
- Fixed comment not in JSdoc format.

## before
スクリーンショット 2023-09-07 11 44 55

## after
with strike
スクリーンショット 2023-09-07 11 46 12


colinhacks:
Nice, thanks!

Issue #2718, opened 2023-09-07
Update to Custom Schemas example in Documentation


mitchgollub:
Hi, 

I'm interested in updating the Custom Schemas example in the documentation.  The function type takes in an unknown, so the example casts the value to a string in order to operate on it with a RegExp

```typescript
const px = z.custom<`${number}px`>((val) => {
  return /^\d+px$/.test(val as string);
});
```

I want to propose using Type Narrowing. It's helped me catch errors from validating a non-expected type, and I believe it would help others as well as a good practice with interacting with the unknown type.

```typescript
const px = z.custom<`${number}px`>((val) => {
  return typeof val === "string" ? /^\d+px$/.test(val) : false;
});
```

Please let me know if this is a welcome change.  I can make a fork of the repo and create a PR!

JacobWeisenburger:
Sounds good to me. Please make the PR and I'll take a look.

mitchgollub:
Thank you for the opportunity to contribute @JacobWeisenburger!  I've made my PR here: https://github.com/colinhacks/zod/pull/2778

mitchgollub:
PR has been merged - closing issue

PR #2778, opened 2023-09-23
#2718 Updated Custom Schemas documentation example to use type narrowing

mitchgollub:
Hi,

I'm interested in updating the Custom Schemas example in the documentation. The function type takes in an unknown, so the example casts the value to a string in order to operate on it with a RegExp

```typescript
const px = z.custom<`${number}px`>((val) => {
  return /^\d+px$/.test(val as string);
});
```

I want to propose using Type Narrowing. It's helped me catch errors from validating a non-expected type, and I believe it would help others as well as a good practice with interacting with the unknown type.

```typescript
const px = z.custom<`${number}px`>((val) => {
  return typeof val === "string" ? /^\d+px$/.test(val) : false;
});
```

This is a PR for issue #2718

### DIFF
--- a/README.md
+++ b/README.md
@@ -1879,7 +1879,7 @@ You can create a Zod schema for any TypeScript type by using `z.custom()`. This 
 
 ```ts
 const px = z.custom<`${number}px`>((val) => {
-  return /^\d+px$/.test(val as string);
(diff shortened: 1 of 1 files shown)


JacobWeisenburger (approved):

JacobWeisenburger:
Thanks @mitchgollub

PR #2827, opened 2023-10-01
Allow undefined values in record, similar to object.

### LINKED ISSUE #2762: keep undefined values in .record()
### Discussed in https://github.com/colinhacks/zod/discussions/2760

<sup>Originally posted by **DanielHoffmann** September 22, 2023</sup>
is there a way where I can type a record while keeping the keys that have `undefined` as a value?

For example, how could I set up this schema so this test passes?

```typescript
  import { z } from 'zod'
  test('allow undefined values', () => {
    const keySchema = z.string().startsWith('_')
    const valueSchema = z.union([z.number(), z.undefined()])
    const schema = z.record(keySchema, valueSchema)
    expect(
      Object.keys(
        schema.parse({
          _test: undefined,
        }),
      ),
    ).toEqual(['_test'])
  })
```

in this case the `Object.keys(schema.parse(...))` is returning empty array

Could this be a bug? For example, this tests passes:

```typescript
import { z } from 'zod'
test('allow undefined values', () => {
    const schema = z.object({
      a: z.union([z.number(), z.undefined()]),
    })
    expect(
      Object.keys(
        schema.parse({
          a: undefined,
        }),
      ),
    ).toEqual(['a'])
})
```

I would kind of expect `.record()` and `.object()` to behave the same in regards to the values</div>

galmrz:
`ZodRecord` and `ZodObject` `_parse` methods are dealing differently with undefined values which seems inconsistent.
Problem described further in this issue: #2762 .
I added the same logic to the `_parse` method in `ZodRecord` as implemented  in `ZodObject`.

closes #2762

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -3753,7 +3753,7 @@ export class ZodRecord<
     const pairs: {
       key: ParseReturnType<any>;
       value: ParseReturnType<any>;
-      alwaysSet: boolean;
+      alwaysSet?: boolean;
     }[] = [];
 
     const keyType = this._def.keyType;
--- a/src/types.ts
+++ b/src/types.ts
@@ -3765,7 +3765,7 @@ export class ZodRecord<
         value: valueType._parse(
           new ParseInputLazyPath(ctx, ctx.data[key], ctx.path, key)
         ),
-        alwaysSet: key in ctx.data,
+        alwaysSet: true,
       });
     }
 
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -868,11 +868,8 @@ z.string().nanoid();
 z.string().cuid();
 z.string().cuid2();
 z.string().ulid();
-<<<<<<< HEAD
 z.string().xid();
-=======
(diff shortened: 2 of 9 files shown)


src/types.ts
```diff
@@ -3382,6 +3383,7 @@ export class ZodRecord<
         value: valueType._parse(
           new ParseInputLazyPath(ctx, ctx.data[key], ctx.path, key)
         ),
+        alwaysSet: key in ctx.data,
```
andersk, line 3386: We just got `key` from `for (const key in ctx.data)`, so this will always be `true`, right?

galmrz: Yes, should be `alwaysSet: true` here.
Also currently only `ParseStatus.mergeObjectSync` supports `alwaysSet`, so I updated the async one as well `ParseStatus.mergeObjectAsync`(and added a test for it).

colinhacks:
Thanks! This is really great. Big oversight on `mergeObjectSync`...thanks for catching.

PR #2839, opened 2023-10-04
TS compilation perf: steal faster extendShape type from tRPC

jussisaurio:
I have some empirical evidence in a decently sized closed-source project that liberal use of `ZodObject.extend()` seems to degrade TS compilation and intellisense performance quite rapidly. I remembered seeing an equivalent implementation of `extendShape` in tRPC (in that project, called `Overwrite`) that looked simpler. I did some light profiling on it using said closed-source project and here are the results:

```
// Zod@jussisaurio

% npx tsc --noEmit --extendedDiagnostics --incremental false

Files:                        1248
Lines of Library:            39145
Lines of Definitions:       126023
Lines of TypeScript:         40594
Lines of JavaScript:             0
Lines of JSON:                   0
Lines of Other:                  0
Identifiers:                183660
Symbols:                    290202
Types:                      147133
Instantiations:            2361427
Memory used:               345043K
Assignability cache size:    34109
Identity cache size:          1470
Subtype cache size:            708
Strict subtype cache size:   22394
I/O Read time:               0.04s
Parse time:                  0.38s
ResolveModule time:          0.09s
ResolveTypeReference time:   0.01s
ResolveLibrary time:         0.01s
Program time:                0.60s
Bind time:                   0.17s
Check time:                  1.73s
printTime time:              0.00s
Emit time:                   0.00s
Total time:                  2.49s

// Zod@3.21.1

% npx tsc --noEmit --extendedDiagnostics --incremental false

Files:                        1248
Lines of Library:            39145
Lines of Definitions:       126004
Lines of TypeScript:         40594
Lines of JavaScript:             0
Lines of JSON:                   0
Lines of Other:                  0
Identifiers:                183556
Symbols:                    286851
Types:                      152232
Instantiations:            7692583
Memory used:               356978K
Assignability cache size:    33682
Identity cache size:          1464
Subtype cache size:            708
Strict subtype cache size:   22394
I/O Read time:               0.04s
Parse time:                  0.39s
ResolveModule time:          0.10s
ResolveTypeReference time:   0.01s
ResolveLibrary time:         0.01s
Program time:                0.60s
Bind time:                   0.17s
Check time:                  2.80s
printTime time:              0.00s
Emit time:                   0.00s
Total time:                  3.57s

// Zod@jussisaurio, run #2

% npx tsc --noEmit --extendedDiagnostics --incremental false

Files:                        1248
Lines of Library:            39145
Lines of Definitions:       126023
Lines of TypeScript:         40594
[...]

PR #2845, opened 2023-10-06
TS compilation perf: faster objectUtil.addQuestionMarks

jussisaurio:
I'm not sure what the precise reasons for this being faster are, but consistently benchmarking in my project about 50% more type instantiations with the original version vs. the one proposed in this commit; plus the compilation time is 20% longer in the original.

EDIT: check out [this issue comment](https://github.com/microsoft/TypeScript/issues/56017#issuecomment-1755995893) from a `tsc` maintainer

Before:
```
Files:                        1251
Lines of Library:            39145
Lines of Definitions:       126004
Lines of TypeScript:         11878
Lines of JavaScript:             0
Lines of JSON:                   0
Lines of Other:                  0
Identifiers:                183217
Symbols:                    285778
Types:                       99275
Instantiations:            7734511
Memory used:               340757K
Assignability cache size:    34242
Identity cache size:          1483
Subtype cache size:            724
Strict subtype cache size:      99
I/O Read time:               0.03s
Parse time:                  0.38s
ResolveModule time:          0.10s
ResolveTypeReference time:   0.01s
ResolveLibrary time:         0.01s
Program time:                0.61s
Bind time:                   0.16s
Check time:                  2.53s
I/O Write time:              0.00s
printTime time:              0.02s
Emit time:                   0.02s
Total time:                  3.32s
```

After:
```
Files:                        1251
Lines of Library:            39145
Lines of Definitions:       126008
Lines of TypeScript:         11878
Lines of JavaScript:             0
Lines of JSON:                   0
Lines of Other:                  0
Identifiers:                183224
Symbols:                    286106
Types:                       86715
Instantiations:            5187916
Memory used:               334963K
Assignability cache size:    34946
Identity cache size:          1479
Subtype cache size:            724
Strict subtype cache size:      99
I/O Read time:               0.03s
Parse time:                  0.38s
ResolveModule time:          0.11s
ResolveTypeReference time:   0.01s
ResolveLibrary time:         0.01s
Program time:                0.62s
Bind time:                   0.18s
Check time:                  2.04s
I/O Write time:              0.00s
printTime time:              0.02s
Emit time:                   0.02s
Total time:                  2.86s
```

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,11 @@
+# Zod Pull Request
+
+**IMPORTANT:** Development of the next major version of Zod (`v4`) is currently ongoing. If your PR implements new functionality, it should target the `v4` branch, NOT the `master` branch. (If it's a bugfix, the `master` branch is fine.)
+
+## Overview
+
+Thank you for your contribution to our project! Before submitting your pull request, please ensure the following:
+
+- [ ] Your code changes are well-documented.
+- [ ] You have tested your changes.
+- [ ] You have updated any relevant documentation.
--- a/README.md
+++ b/README.md
@@ -867,16 +867,18 @@ z.string().includes(string);
 z.string().startsWith(string);
 z.string().endsWith(string);
 z.string().datetime(); // ISO 8601; by default only `Z` timezone allowed
-z.string().date(); // ISO date format (YYYY-MM-DD)
-z.string().time(); // ISO time format (HH:mm:ss[.SSSSSS])
-z.string().duration(); // ISO 8601 duration
 z.string().ip(); // defaults to allow both IPv4 and IPv6
-z.string().base64();
 
 // transforms
 z.string().trim(); // trim whitespace
 z.string().toLowerCase(); // toLowerCase
 z.string().toUpperCase(); // toUpperCase
+
+// added in Zod 3.23
+z.string().date(); // ISO date format (YYYY-MM-DD)
+z.string().time(); // ISO time format (HH:mm:ss[.SSSSSS])
+z.string().duration(); // ISO 8601 duration
+z.string().base64();
 ```
 
 > Check out [validator.js](https://github.com/validatorjs/validator.js) for a bunch of other useful string validation functions that can be used in conjunction with [Refinements](#refine).
@@ -913,10 +915,6 @@ z.string().ip({ message: "Invalid IP address" });
 
 As you may have noticed, Zod string includes a few date/time related validations. These validations are regular expression based, so they are not as strict as a full date/time library. However, they are very convenient for validating user input.
 
-The `z.string().date()` method validates strings in the format `YYYY-MM-DD`.
-
-The `z.string().time()` method validates strings in the format `HH:mm:ss[.SSSSSS][Z|(+|-)hh[:]mm]` (the time portion of [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)). It defaults to `HH:mm:ss[.SSSSSS]` validation: no timezone offsets or `Z`, with arbitrary sub-second decimal.
-
 The `z.string().datetime()` method enforces ISO 8601; default is no timezone offsets and arbitrary sub-second decimal precision.
 
 ```ts
@@ -948,15 +946,11 @@ const datetime = z.string().datetime({ precision: 3 });
 datetime.parse("2020-01-01T00:00:00.123Z"); // pass
 datetime.parse("2020-01-01T00:00:00Z"); // fail
 datetime.parse("2020-01-01T00:00:00.123456Z"); // fail
-
-const time = z.string().time({ precision: 3 });
-
-time.parse("00:00:00.123"); // pass
-time.parse("00:00:00"); // fail
-time.parse("00:00:00.123456"); // fail
 ```
 
 ### Dates
+
+> Added in Zod 3.23
 
 The `z.string().date()` method validates strings in the format `YYYY-MM-DD`.
 
@@ -969,6 +963,8 @@ date.parse("2020-01-32"); // fail
 ```
 
 ### Times
+
+> Added in Zod 3.23
 
 The `z.string().time()` method validates strings in the format `HH:MM:SS[.s+]`. The second can include arbitrary decimal precision. It does not allow timezone offsets of any kind.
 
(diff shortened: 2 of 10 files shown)


victorhnogueira:
interesting, how did you do this benchmark? how can I reproduce?

jussisaurio:
> interesting, how did you do this benchmark? how can I reproduce?

I'm basically just benchmarking it against one of the TS APIs at work that uses Zod quite heavily. Would probably be good to have some public `zod-kitchen-sink` type performance benchmark project though...

I'm having trouble getting any perf difference between these two implementations on just a few ZodObjects, but in my real world project it's quite clear as you can tell from the numbers cited in the OP

jussisaurio:
> interesting, how did you do this benchmark? how can I reproduce?

Script to generate a bunch of zod object schemas and .omit() & .extend() modifications for each:

```
import fs from "fs";

// Step 1: Possible Zod types
const possibleZodTypes = ["z.string()", "z.number()", "z.boolean()"];
const possibleChainMethods = ["", ".optional()", ".nullable()", ".nullish()"];

// Step 2: Generate a random string for keys and variable names
function generateRandomString(length) {
  const charset = "abcdefghijklmnopqrstuvwxyz";
  let result = "";
  for (let i = 0; i < length; i++) {
    const randomIndex = Math.floor(Math.random() * charset.length);
    result += charset[randomIndex];
  }
  return result;
}

// Step 3: Generate a random Zod schema
function generateRandomZodSchema() {
  const numberOfKeys = Math.floor(Math.random() * 30) + 1; // 1-30 keys

  let schema = `z.object({`;
  const keys = [];
  for (let i = 0; i < numberOfKeys; i++) {
    const key = generateRandomString(8); // Key name of 8 chars
    keys.push(key);
    const randomTypeIndex = Math.floor(Math.random() * possibleZodTypes.length);
    const randomChainMethodIndex = Math.floor(
      Math.random() * possibleChainMethods.length
    );
    const randomType = possibleZodTypes[randomTypeIndex];
    const randomChance = Math.random();
    if (randomChance > 0.98) {
      const { schema: nestedSchema } = generateRandomZodSchema();
      schema += ` ${key}: ${nestedSchema},`;
      continue;
    }
    const type = randomType + possibleChainMethods[randomChainMethodIndex];

    schema += ` ${key}: ${type},`;
  }

  schema += " })";
  return {
    schema,
    keys,
  };
}

// Step 4: Write the generated schemas to a file
function writeToFile(
  filepath = "randomZodSchemas.ts",
  numberOfSchemas = 100,
  numberOfOmits = 10,
  numberOfExtends = 10
) {
  let allSchemas = 'import * as z from "zod";\n\n';
  for (let i = 0; i < numberOfSchemas; i++) {
    const variableName = generateRandomString(7);
    const { schema, keys } = generateRandomZodSchema();
    allSchemas += `export const ${variableName} = ` + schema + ";\n";

    for (let i = 0; i < numberOfOmits; i++) {
      const omitSchemaVariableName = generateRandomString(7);
      const omitKeys = keys
        .slice(0, keys.length - 1)
        .filter(() => Math.random() > 0.5);
      allSchemas += `export const ${omitSchemaVariableName} = ${variableName}.omit({
      ${omitKeys.map((key) => `"${key}": true`).join(",\n")}
    });\n\n`;
    }
    for (let i = 0; i < numberOfExtends; i++) {
      const extendSchemaVariableName = generateRandomString(7);
      const extendKeys = Array(3)
        .fill(0)
        .map(() => generateRandomString(7));
      allSchemas += `export const ${extendSchemaVariableName} = ${variableName}.extend({
      ${extendKeys.map((key) => `"${key}": z.string()`).join(",\n")}
    });\n\n`;
    }
  }

  fs.writeFile(filepath, allSchemas, (err) => {
    if (err) {
      console.error("Error writing file:", err);
    }
  });
}

writeToFile("src/randomZodSchemas.ts");
```

Generate a file, then `import * from './thatFile.ts'` and run `npx tsc --noEmit --extendedDiagnostics` on it. This reproduces the perf difference quite clearly. 

Importantly: a meaningful amount of perf degradation only occurs when using .omit() and .extend(), but when you do, the difference is stark.

jussisaurio:
Turns out it was even possible to remove the `requiredKeys` helper with had yet a another small positive impact on performance; github user Andarist landed on this kind of solution earlier this year in #2620 !

Beyond this PR, it might be a good idea to build a separate compilation performance regression test suite for Zod. Something like the type generation script above (albeit cleaned up) might serve as a starting point.

jussisaurio:
I made a really braindead-simple benchmarking repo, some test runs here:

https://github.com/colinhacks/zod/pull/2839#issuecomment-1756110256

psychedelicious:
On zod 3.22.4, this patch breaks a ton of types.

jussisaurio:
> On zod 3.22.4, this patch breaks a ton of types.

Could you provide some example types so I can modify the PR accordingly?

psychedelicious:
I apologize, your PR works fine. I cherry-picked your commits onto upstream master and used that build successfully.

My problem was that I patched the build artefact directly with `pnpm patch`. That broke not only `zod` types, but also a ton of non-`zod` types with nested objects/records. I suppose there is a weird interaction with `pnpm patch` and TS. Spooky!

jussisaurio:
Was able to reduce the # of type instantiations some more by baking in flattening to `addQuestionMarks`; that type was always wrapped with `flatten` so inlining it was possible

colinhacks:
Thanks! I'd added some additional tests in `generics.test.ts` and it took some fiddling to get this to pass. The `zod-ts-perftest` repo is *so useful*! Thanks for putting that together. Got it down to ~3m instantiations from ~6m, sub 3s build times on my machine. 🚀

colinhacks:
This has landed in Zod 3.23.

colinhacks:
And...it broke JSDoc #3437 

Here's where I'm at in my investigations:

```ts
// slow (60% more instantiations), but preserves JSDoc
type extendShape2<A extends object, B extends object> = Pick<
  A,
  Exclude<keyof A, keyof B>
> & B;

// fast, but JSDoc is lost
type extendShape1<A extends object, B extends object> = {
  [K in keyof A | keyof B]: K extends keyof B
    ? B[K]
    : K extends keyof A
    ? A[K]
    : never;
};

// fast & preserves JSDoc! doesn't reduce object to simplest form
type extendShape3<A extends object, B extends object> = {
  [K in Exclude<keyof A, keyof B>]: A[K];
} & B;
```

jussisaurio:
```
  export type extendShape<A extends object, B extends object> = {
    [K in keyof A]: K extends keyof B ? never : A[K];
  } & {
    [K in keyof B]: B[K];
  };
```

This seems to preserve the JSDocs at least with the example given in #3437

PR #2860, opened 2023-10-11
Add brand to ZodBranded def at runtime

nahvego:
Make brand available at runtime when used as a parameter for .brand()

With this brand() can be used to enable external code to operate on types based on definition.

Typing is left as-is to be backwards-compatible so instead of overriding the BRAND symbol this just puts the brand parameter of the constructor in _def.

Let me know what you think! 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -2442,6 +2442,17 @@ type Cat = z.infer<typeof Cat>;
 
 Note that branded types do not affect the runtime result of `.parse`. It is a static-only construct.
 
+However when using the brand as a parameter the type will include it in the typing at runtime (but not the parse!)
+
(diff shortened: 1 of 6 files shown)


RobinTail:
@colinhacks , please make branded types programmatically distinguishable.
That would be a huge help!

colinhacks:
Branding is inherently a runtime construct, so I'm a little hesistant to do this. The fact that the type-only `.brand<T>()` API exists mean that there's no guarantee that and given `ZodBrand` instance

artalar:
@colinhacks What do you think about allowing to put "unique symbol" from a user space code?

PR #2872, opened 2023-10-17
Strict RFC4122 UUID validation and separate guid validation

jeansibelius:
I wanted to use Zod to validate UUIDs, but noticed that the current implementation passes also UUIDs that are actually not valid according to the spec [as defined in RFC4122](https://datatracker.ietf.org/doc/html/rfc4122#section-4).

Namely, the existing regex allows for version and variant values that should not be possible for a UUID.

I created this PR as a proposal in order to have UUID validation follow RFC4122 strictly and add a more generic `guid` string validator that allows for non-conforming UUID-like values (utilising the pre-existing regex making migration as simple as updating `string().uuid`s to `string().guid` where necessary). 

In regards to UUID v6+ I would suggest adding a `uuidv6` (or similar explicitly named) validator until the [RFC draft](https://www.ietf.org/archive/id/draft-peabody-dispatch-new-uuid-format-04.html#name-new-formats) gets approved. This is not implemented by this PR.

I updated the tests accordingly and made sure they pass.

I hope this is helpful! (At the very least, I hope to spark some discussion.)

If there is something that I overlooked, please let me know.

Thanks a lot for creating an amazing library!

Related issues I was able to find:
https://github.com/colinhacks/zod/issues/2122
https://github.com/colinhacks/zod/pull/2341
https://github.com/colinhacks/zod/pull/2430

### DIFF
--- a/deno/lib/ZodError.ts
+++ b/deno/lib/ZodError.ts
@@ -96,6 +96,7 @@ export type StringValidation =
   | "emoji"
   | "uuid"
   | "nanoid"
+  | "guid"
   | "regex"
   | "cuid"
   | "cuid2"
--- a/deno/lib/__tests__/string.test.ts
+++ b/deno/lib/__tests__/string.test.ts
@@ -307,14 +307,12 @@ test("bad nanoid", () => {
 });
 
 [
-  "9491d710-3185-0e06-bea0-6a2f275345e0",
   "9491d710-3185-1e06-bea0-6a2f275345e0",
   "9491d710-3185-2e06-bea0-6a2f275345e0",
   "9491d710-3185-3e06-bea0-6a2f275345e0",
   "9491d710-3185-4e06-bea0-6a2f275345e0",
   "9491d710-3185-5e06-bea0-6a2f275345e0",
   "9491d710-3185-5e06-aea0-6a2f275345e0",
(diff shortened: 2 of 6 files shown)


colinhacks:
I like this, but since it's a breaking change it'll have to wait until Zod 4 (soon).

jeansibelius:
Thanks a lot! Very happy to see that this was useful.

> I'm merging this into the v4 branch. Some version of this will land in Zod 4, but (as a heads up) it may be a while, and this feature/API will likely undergo some changes in the interim.

Nice! Looking forward to it! I'll try to keep an eye out, if I could be of further help. 

> Zod 4 will likely include z.uuidv1, z.uuidv2, etc to account for people who want to validate a specific version.

Do you think it would make sense for there to also be exist a catch-all z.uuid for any valid uuid?

> I added back a stricter check for the nil regex and adjusted tests according.

Thanks a lot! Sorry for the trouble.

> There were also some \b in the guidRegex that seemed out of place, feel free to chime in if there was a good reason for those I'm missing.

Hmm... Can't remember now, if there was some reason to check for word-boundaries in those spots. Looking at it now, they seem superfluous, indeed. (My best guess now is that it's just some cruft from when I made the regexes for uuid and guid as distinct as possible.)

Thanks a lot again for this great library!

PR #2896, opened 2023-10-23
fix: wrong typing

Fgc17:
The type of the check argument function must be a generic, otherwise the type you use for your own custom function would need to be unknown, and that's what is currently happening:

```ts
function customValidator(value: string) {
 ...
}

z.custom(customValidator) 
// this throws an error saying that the customValidator function `value` argument must be of type unknown.
```

After inserting the generic, everything works fine!


colinhacks:
Fixed in b2ed20cd703321466fbf499986245c42eae02900 but thanks!

PR #2897, opened 2023-10-24
docs: introduce ".nonstrict" method

exoego:
`z.object(...).nonstrict(...)` is available since [v1.0.11](https://github.com/colinhacks/zod/blob/master/CHANGELOG.md#zod1011) but not documented.

I found `.nonstrict` is helpful when I define type loosely and improve in incremental steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1336,6 +1336,10 @@ person.parse({
 // => throws ZodError
 ```
 
+### `.nonstrict`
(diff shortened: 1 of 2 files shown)


exoego:
Ah, I found `.passthrough` and `.nonstrict` is just an alias of it.

Issue #2904, opened 2023-10-26
Zod preprocess does not generate error


stackline:
I have been using zod as a validation library with react-hook-form. When I set up validation using preprocess in zod, I noticed that errors were not being detected as expected.

#### Similar issue

I have confirmed that this issue has been reported on the react-hook-form/resolver repository.

* https://github.com/react-hook-form/resolvers/issues/635

#### What I confirmed about this issue

Regarding this issue, I downgraded the zod version from 3.22.4 to 3.21.4 and confirmed that errors were detected as expected. I also tried several versions and found that the issue occurs starting from 3.22.0.

#### How to reproduce

Although this is a quote from the issue on the react-hook-form/resolver repository, here is a codesandbox that reproduces the phenomenon of errors not being detected (with zod version 3.22.4):

* https://codesandbox.io/s/react-hook-form-zod-preprocess-mn8sy4?file=/src/index.js

Additionally, here is a codesandbox with zod version 3.21.4, where errors are detected:

* https://codesandbox.io/s/react-hook-form-zod-preprocess-forked-cqm66q

Austin-Love:
I have noticed the same behavior, but it's actually a little more nuanced than that. I opened the same issue yesterday (https://github.com/colinhacks/zod/issues/2900)

Turns out preprocessers are canceled if a field declared before them in the schema without a preprocessor fails validation. 

My team found a temporary fix by replacing nested preprocessors with transforms and then piping them into a validator. We've only made this change to form resolvers because we are concerned about the side effects.

kevbook:
Yep, v3.22.0 introduced this bug!

aaronadamsCA:
This issue appears to be a duplicate of #2687.

Here is a one-line fix you can apply as a local patch to the Zod package:

- [For Deno users](https://github.com/colinhacks/zod/pull/2912/files#diff-52632a4861fc9d7dc2dacef13cd91d60286dd706c1bb57438b8ee6a579a8796a)
- [For everyone else](https://github.com/colinhacks/zod/pull/2912/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410a)

I would suggest lending your 👍 to #2912 to encourage a fix to be merged.

PR #2912, opened 2023-10-27
Reporting errors of the preprocess that is the second property of object

yukukotani:
Fix https://github.com/colinhacks/zod/issues/2904

This behavior was introduced in #2426. `ctx.common.issues` includes the errors of previous properties, so that they affects the status of preprocess.


yukukotani:
I'm not confident in my understanding of the whole codebase and the change in #2426.

@Pingviinituutti Could you take a look at this PR?

deno/lib/__tests__/transformer.test.ts
```diff
@@ -297,6 +297,23 @@ test("short circuit on dirty", () => {
...
+    nonEmptyStr: z.string().min(1),
+    positiveNum: z.preprocess((v) => Number(v), z.number().positive()),
+  });
+  const result = schema.safeParse({
+    nonEmptyStr: "",
+    positiveNum: "",
+  });
+  expect(result.success).toEqual(false);
+  if (!result.success) {
+    expect(result.error.issues.length).toEqual(2);
```
yukukotani, line 311: This fails in current master branch since `result.error.issues.length === 1`

Pingviinituutti:
Looks good to me, seems the simple fix is enough to :)

Are the current test cases enough to ensure it works as intended?

yukukotani:
@Pingviinituutti Thanks!

> Are the current test cases enough to ensure it works as intended?

The current test cases are fail on master branch, so I think it's minimum but enough.

yukukotani:
@Austin-Love This is ready and waiting for getting review

colinhacks:
Made some fairly major changes to `ctx.addIssue` in `preprocess` actually agrees with the runtime behavior in `.superRefine` and `.transform`. Namely, all issues are considered *non-fatal* — validation will continue even if `.addIssue` was called during the `preprocess`. This behavior can be modified by indicating `fatal: true` in `ctx.addIssue`:

```ts
const schema = z.preprocess((data, ctx) => {
      ctx.addIssue({
        code: "custom",
        message: `bad thing happened`,
        fatal: true
      });
      return data;
    }, z.string())
```

PR #2913, opened 2023-10-27
fix: support unqualified (local) datetime strings

### LINKED ISSUE #2385: Datetime validation without offset should allow omission of Z zone designator

0xturner:
FIxes #2385

Adds support for local unqualified date time strings. (the ones that don't end in "Z")

ISO-8601 allows unqualified date time strings, in which case the time is assumed to be in local time rather than UTC. [(See)](https://en.wikipedia.org/wiki/ISO_8601#Local_time_(unqualified)) 

To prevent breaking changes, this is an an additional opt-in option to the existing `string().dateTime()` method.

```
const datetime = z.string().datetime();
datetime.parse("2020-01-01T00:00:00"); // fail, same as before, requires trailing "z"

const datetimeUnqualified = z.string().datetime({ unqualified: true });
datetime.parse("2020-01-01T00:00:00"); // pass
```

I realized shortly after creating this PR that there is a similar effort at #2522. I've taken a slightly different approach in this PR. I am not sure what the etiquette is in terms of duplicate PRs, happy to collaborate and merge efforts with #2522 if that is possible/preferred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -742,7 +742,7 @@ z.string().uuid({ message: "Invalid UUID" });
 z.string().includes("tuna", { message: "Must include tuna" });
 z.string().startsWith("https://", { message: "Must provide secure URL" });
 z.string().endsWith(".com", { message: "Only .com domains allowed" });
-z.string().datetime({ message: "Invalid datetime string! Must be UTC." });
+z.string().datetime({ message: "Invalid datetime string!" });
 z.string().ip({ message: "Invalid IP address" });
 ```
 
@@ -754,6 +754,7 @@ The `z.string().datetime()` method enforces ISO 8601; default is no timezone off
 const datetime = z.string().datetime();
 
 datetime.parse("2020-01-01T00:00:00Z"); // pass
+datetime.parse("2020-01-01T00:00:00"); // pass (unqualified)
 datetime.parse("2020-01-01T00:00:00.123Z"); // pass
 datetime.parse("2020-01-01T00:00:00.123456Z"); // pass (arbitrary precision)
 datetime.parse("2020-01-01T00:00:00+02:00"); // fail (no offsets allowed)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -709,7 +709,7 @@ z.string().regex(regex);
 z.string().includes(string);
 z.string().startsWith(string);
 z.string().endsWith(string);
-z.string().datetime(); // ISO 8601; default is without UTC offset, see below for options
+z.string().datetime(); // ISO 8601; default is UTC with zero offset, see below for options
```
0xturner, line 712: Choosing more explicit wording here.

"Without UTC offset" is potentially misleading, as there is some nuance in the fact that "Z" is a zone indicator for the zero UTC offset as opposed to the absence of an offset.

src/types.ts
```diff
@@ -535,6 +535,7 @@ export type ZodStringCheck =
       kind: "datetime";
       offset: boolean;
       precision: number | null;
+      unqualified: boolean;
```
0xturner, line 538: Some other names I considered for this option, open to changing:
- `local`
- `utc` (defaults to true, set to false to allow unqualified dates)

MattSidor (approved):

Lephilosophe237:
Hi I got the issue with the offset , do you know when this will be release please

ahmed-abdelmotey:
any updates on when this is getting merged ?

Irvenae:
can we get this merged?

ezze:
Please, any feedback at least, guys, why it's not merged yet?

ericdew:
Also interested in this change - would save a lot of headaches when working with forms with datetime inputs.

colinhacks:
Thanks @0xturner — I ended up merging the other PR after modifying it to use an approach/API almost identical to yours 😅 Whoops. Great stuff here!

PR #2928, opened 2023-11-03
Add mandatory code: ZodeIssueCode to ZodIssueBase type.

sgpinkus:
Minor change. Just makes it easier to tell that all ZodIssue instances are actually going to have a code value.


colinhacks:
Thanks but this is currently intentional. I omitted this to allow for the possibility of user-defined issue types, but never really got around to defining an API for that. This is being entirely rewritten for Zod 4, so I'm afraid there's not much use in merging at this point, but I'll keep this PR in mind during the refactor 👍

PR #2950, opened 2023-11-09
update makeIssue to support the force parameter returned from setErrorMap and allow Custom Message Translations

dev2xl:
Now you can define error messages at the schema level (like a translation key), and use setErrorMap like this to lookup the translation.

Returning force: true will prevent this message of being overwritten by the schema message. This allow to create and return more custom translations for messages.

// Translation logic ex.
// error.message is the one defined at schema level; that now you can use it as the translation key

Usage:
```
z.setErrorMap((error, ctx) => {
  const message = getTranslatedMessageForKey(error.message);
  return { message, force: true };
});
```

### DIFF
--- a/src/helpers/parseUtil.ts
+++ b/src/helpers/parseUtil.ts
@@ -17,18 +17,24 @@ export const makeIssue = (params: {
   };
 
   let errorMessage = "";
+  // track intention of using error map result even when there is an error message defined at schema level
+  let force = false; 
+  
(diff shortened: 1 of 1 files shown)


colinhacks:
Thanks for this PR. The error map system is being totally rewritten for Zod 4 so I'm afraid there's little point in modifying this API now. I'll keep this use case under consideration.

PR #2961, opened 2023-11-13
Fix regex in cuid2 validation to be what cuid2 library expects

### LINKED ISSUE #2960: z.string().cuid2() validator doesn't conform to the upstream format
Regex which is defined for cuid2 format validation doesn't conform to the one defined in the cuid2 library itself. 
It bothers me as i can't use this validator with ids generated by this elixir implementation of cuid2: https://github.com/joseph-lozano/cuid2

Said regex in the zod codebase:
https://github.com/colinhacks/zod/blob/master/src/types.ts#L549

Regex in the upstream cuid2 codebase:
https://github.com/paralleldrive/cuid2/blob/main/src/index.js#L105

etareduction:
Closes #2960

### DIFF
--- a/deno/lib/__tests__/string.test.ts
+++ b/deno/lib/__tests__/string.test.ts
@@ -250,7 +250,6 @@ test("cuid2", () => {
   validStrings.forEach((s) => cuid2.parse(s));
   const invalidStrings = [
     "", // empty string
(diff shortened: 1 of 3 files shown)


colinhacks:
This has landed in Zod 3.23.

PR #2978, opened 2023-11-18
feat: Allow passing custom message to literal

### LINKED ISSUE #2974: Custom error messages don't work with z.literal
## Code
```ts
const schema = z.literal(true, {
  invalid_type_error: "Custom Error 1",
  required_error: "Custom Error 2",
})
// ...
schema.safeParse(false)
```

## Expection Result
```
ZodError: [
  {
    "received": false,
    "code": "invalid_literal",
    "expected": true,
    "path": [],
    "message": "Custom Error 1"
  }
]
```

## Actual Result
```
ZodError: [
  {
    "received": false,
    "code": "invalid_literal",
    "expected": true,
    "path": [],
    "message": "Invalid literal value, expected true"
  }
]
```

LiamMartens:
Fixes #2974

Uses a message key instead of "invalid_xx" key - because those are only used for required and invalid type using a message key is more in line with other checks


colinhacks:
Thanks for this! Error customization is going to be a lot better in Zod 4. I'll merge this but it'll likely undergo some changes before being released.

Issue #2686, opened 2023-08-27
Allow number in enums


m-dressler:
It would be helpful to allow numbered enums such as in this example

```typescript
const userMap = {
  1: { name: "John" },
  2: { name: "Bob" },
};

const ids = Object.keys(prices).map(Number) as [keyof typeof prices];
const idSchema = z.enum(ids);
```

While `"${number}"` could be used to index the userMap, if this were a network request, it would annoyingly also cause the request to format the numbers into strings. Respectively, I think there are valid use-cases where a numbered enum can be helpful.

joshuat:
I added a workaround for this issue in #1118. I would love to see this functionality supported by zod.

nghhoese:
Need this as well or a .exclude/.omit on the native enum. Right now im facing the problem that for Some methods i want to allow certain values of a enum. Right now im working around it using union and literals but that gives an ugly validation error

mnahkies:
I've just bumped up against this implementing zod schema generation from openapi definitions.

Consider the openapi schema:
```
redirectStatusCode:
  type:
    - integer
    - "null"
  enum:
    - 301
    - 302
    - 307
    - 308
```

I'd ideally like to be able to output:
```
redirectStatusCode: z.enum([301, 302, 307, 308]).nullable()
```
With the resulting typescript type `301 | 302 | 307 | 308 | null`

For now I've used the `z.literal` workaround mentioned in #1118
```
redirectStatusCode: z.union([z.literal(307), z.literal(301), z.literal(302), z.literal(308)]).nullable()
```

YaroslavFrolov:
`
export const STATUSES = {
  disabled: 0,
  active: 1,
  expired: 2,
} as const;

const status = zod.nativeEnum(STATUSES); // 0 | 1 | 2
`

cdtut:
@colinhacks can PR be merged?

Better if it can support not only number but all primitive type like true, false, null, and undefined so can check different values in array.

colinhacks:
Closing this per https://github.com/colinhacks/zod/pull/2980#issuecomment-2073499456

My plan for addressing this use case is to support an array values in ZodLiteral, which can then be expanded to support a variety of primitive types:

```ts
z.literal(["a", 2, 3n, false])
```

bernaferrari:
would `z.literal(["a", "b"] == z.enum(["a", "b"]) == z.union(["a", "b"])`?

DibyodyutiMondal:
[...]

PR #2980, opened 2023-11-18
feat: Support numeric enum values in addition to string values

LiamMartens:
Implements: #2686


colinhacks:
My current plan to support this behavior is to support an array values in `ZodLiteral`. ZodEnum has special properties like `.enum` that work best with string values. I'm aware numeric values also work with `.enum`, but I still don't think allowing numbers here is aligned with the goals of ZodEnum.

DibyodyutiMondal:
@colinhacks Is there an issue that tracks the implementation of `z.literal(array)`?

PR #2995, opened 2023-11-27
Add documentation note about eslint no-unsafe

ari-becker:
Documenting this because it took me way too long to understand how `any` was entering my codebase and triggering `eslint`.


colinhacks:
I'll fix the underlying issue in the next version of Zod. I'd rather not add sections/asides to the docs that aren't broadly applicable.

PR #3012, opened 2023-12-01
Add optional generic Data type to parse and safeParse

jadamduff:
Problem:
Currently, parse and safeParse allow any `unknown` type to be passed as input. This makes it extremely difficult to ensure type safety when using zod for type transformations. Without being able to restrict the type being passed to parse, misalignments in type definitions will only be caught at runtime via zod errors.

For example, let's say we have an address object that is of type Address:
```
type Address = {
  street: string,
  city: string,
  state: string,
  zip: string
};

const address: Address = {
  street: "1234 Example Rd",
  city: "New York",
  state: "NY",
  zip: "12345"
};
```

And we want to convert it to type Location:
```
const locationSchema = z.object({
  city: z.string(),
  state: z.string(),
  lat: z.number(),
  long: z.number()
});

type Location = z.infer<typeof locationSchema>;

const location: Location = locationSchema.parse(address); // no type error
```

We won't realize we're passing invalid data to parse until runtime.

Solution:
If you know ahead of time what type should be passed to parse, providing an optional generic Data type to parse allows this misalignment to be caught by the type checker.

Example:
```
const location: Location = locationSchema.parse<Location>(address); // type error
```

I've included implementations and tests for parse, safeParse, parseAsync, and safeParseAsync.

This would be extremely helpful.

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,24 +185,24 @@ Sponsorship at any level is appreciated and encouraged. For individual developer
 <table>
   <tr>
     <td align="center">
-      <a href="https://speakeasyapi.dev/">
+      <a href="https://speakeasyapi.dev/" target="_blank">
         <img src="https://avatars.githubusercontent.com/u/91446104?s=200&v=4" width="200px;" alt="Speakeasy API" />
       </a>
       <br />
       <b>Speakeasy</b>
       <br />
-      <a href="https://speakeasyapi.dev/">speakeasyapi.dev</a>
+      <a href="https://speakeasyapi.dev/" target="_blank">speakeasyapi.dev</a>
       <br />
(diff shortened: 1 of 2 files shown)


lo1tuma:
IMHO if runtime validation is not wanted I think `parse` or `safeParse` shouldn’t be used. Maybe a new function like `transform` could be introduced which automatically infers the required input type and output type from the schema, so a generic type parameter is not needed.

jadamduff:
@lo1tuma that makes sense. Are you imagining that the transform function would also validate? We would be looking to validate and transform at the same time.

lo1tuma:
Why do you want to validate something where you know it already has the correct type? 🤔 
Is it about non-schema validations such as `email()`?  Not sure if it is possible to provide an API that only executes validations and transformations but skips schema parsing 🤔.

Issue #2484, opened 2023-06-04
Common context for schema transformations


OiYouYeahYou:
## TL;DR

I would like to have a locals object for transformers so it's easier to deal with deeply nested objects. For example:

```ts
const mediaSchema = z
	.object({
		path: z.string(),
	})
	.transform((media, locals) => ({ ...media, path: locals.url + media.path }))

const locals = {
	url: 'https://example.com/',
}
mainSchema.parse(obj, locals)
```

## Problem

I have a media schema which comes with just the path, as such:

```ts
const mediaSchema = z.object({
	path: z.string(),
})
```

```json
{
	"path": "/media/some-image.jpg"
}
```

I'd like to dynamically transform the path into a full URL, so each parse could have a different domain, eg: `https://example.com/media/some-image.jpg` and `https://different-example.com/media/some-image.jpg`,

But, currently it is served in multiple places in a larger schema. For example:

```ts
const mainSchema = z.object({
	deeply: z.object({
		nested: z.object({
			media: mediaSchema,
		}),
	}),
	more: z.object({
		nested: z
			.object({
				media: mediaSchema.array(),
			})
			.array(),
	}),
})
```

There are some cumbersome ways to solve this, which all step around using Zod, instead of using Zod out the box. But the best way I think this could be done is to use some sort of common context passed down to transformation functions.

indianakernick:
It's a little bit hacky, but you can use the path to pass data from the parse function down the transformations (and refinements).

```ts
const mediaSchema = z
  .object({
    path: z.string(),
  })
  .transform((media, ctx) => ({
    ...media,
    path: (ctx.path[0] as any).url + media.path,
  }));

const locals = {
  url: 'https://example.com/',
};
schema.parse(obj, { path: [locals as any] });
```

You'll need to remove that extra path component from errors though. It would be nice if there was a more direct way of doing this.

indianakernick:
I'm using the technique I described in my previous comment. It's a bit gross. I don't like the `as any` (although a built-in way is probably going to use `any` since otherwise you'd need to add generic parameter to `parse` and propagate that all the way through). `any`way... I'd thought I talk about my use case and why I think this feature would be useful.

[...]

PR #3023, opened 2023-12-07
feat: meta in context for transform & refine

zce:
for #2484 

```ts
declare module 'zod' {
  interface ZodMeta {
    foo: string
  }
}

const basicSchema = z.string().transform((val, ctx) => {
  console.log(ctx.meta.foo)
  return val.toUpperCase()
})

const res1 = basicSchema.parse('zce', { meta: { foo: 'bar' } })
console.log(res1)

const objSchema = z
  .object({
    name: z.string().transform((val, ctx) => {
      console.log(ctx.meta.foo)
      console.log(ctx.path)
      return val.toUpperCase()
    }),
    age: z.number().superRefine((val, ctx) => {
      console.log(ctx.meta.foo)
      console.log(ctx.path)
      return val > 0
    })
  })

const res2 = objSchema.parse({ name: 'John', age: 30 }, { meta: { foo: 'bar' } })
console.log(res2)
```


ferm10n:
Nice work, you beat me to the punch! [I had a fork](https://github.com/ferm10n/zod/commit/eaa771cba72c931717144ca595c1698e7ae2facc) that has pretty much the same changes, except I added to the `RefinementCtx` type and called it `refinementCtxExtra` instead of `meta`. I think I like `meta` more though.

The only real difference was I added a unit test, the same one in `deno/lib/__tests__/transformer.test.ts` and `src/__tests__transformer.test.ts`. Using your `meta`, the test would look like this:
```ts
  test("meta", async () => {
    const expectedMeta = { abc: 123 };
    let actualMeta;
    z.string().transform((val, ctx) => {
      actualMeta = ctx.meta;
      return val;
    }).parse("asdf", { meta: expectedMeta });
    expect(actualMeta).toEqual(expectedMeta);
  });
```
Just a suggestion if you want to add it to your PR

pzuraq:
Would like to add a +1 for this, would be nice for defining schemas with encrypted values, so you can pass the encryption key in and have it parse out the value

john-twigg-ck:
Please Merge...

colinhacks:
This would be better characterized as "context". I suppose `ctx.context` is a little weird but the term "metadata" will mean something different in Zod 4. I also don't like that the types indicate that `ctx.meta` is always available in refinements/transforms, but in reality it may not be defined at all. It should be 

```ts
export type RefinementCtx = {
  addIssue: (arg: IssueData) => void;
  path: (string | number)[];
  meta?: ZodMeta; // this should be optional
};
```

Alternatively, if `ZodMeta` is defined in user-space, all calls to `.parse*()` could require `meta` to be specified.

zce:
> This would be better characterized as "context". I suppose `ctx.context` is a little weird but the term "metadata" will mean something different in Zod 4. 

I also think it is more appropriate to call it 'context'. 

In fact, users often just want to mount some additional relevant data in context.

Maybe change the 'RefinementCtx' type to interface and allow extension is a good choice.

colinhacks:
Yeah that's a good approach for letting people hack something.

If Zod tried to provide a first-party API for context, it wouldn't be a global interface like `ZodMeta`. It would be something schema-scoped that "bubbles up" to wherever you call `.parse`.

```ts
const schema = z.string({
  context: z.object({ stuff: z.string() })
}).refine((val, ctx)=>{
  ctx.context.stuff; // string
});

schema.parse("hello", { context: { stuff: "stuff" })
```

miserylee:
this feature is really what I want, pity of closed

Issue #3008, opened 2023-11-30
Support single value in enum's `.exclude` & `.extract`


MichaelDeBoey:
```ts
const granularitySchema = z.enum([
  'day',
  'week',
  'fortnight',
  'month',
  'quarter',
  'semester',
  'year',
]);
```

Right now I need to always pass an array to these methods even if I just want to exclude/extract 1 value

```ts
const notDaySchema = granularitySchema.exclude(['day']);
const daySchema = granularitySchema.extract(['day']);
```

It would be nice if I could just do

```ts
const notDaySchema = granularitySchema.exclude('day');
const daySchema = granularitySchema.extract('day');
```

rakheesingh:
@JacobWeisenburger can anyone help me, getting the error while running `yarn build`

<img width="1237" alt="Screenshot 2023-12-09 at 8 03 03 AM" src="https://github.com/colinhacks/zod/assets/22937223/1739dd3c-18ec-409f-802d-5a9adabdcfa8">

Panth977:
@MichaelDeBoey 
we can add this on package side, but you can also customize it on your project side,

make sure to run this before any usage of following methods
```ts
// have this code before or after, dosn't matter
const granularitySchema = z.enum(['day', 'week', 'fortnight', 'month', 'quarter', 'semester', 'year']); 
// add new methods as your need in developer experience 
z.ZodEnum.prototype.excludeSome = function (this, ...values) {
    return this.exclude(values);
};
z.ZodEnum.prototype.extractSome = function (this, ...values) {
    return this.extract(values);
};
// after adding prototype you can use those methods.
const notDaySchema = granularitySchema.excludeSome('day');
const daySchema = granularitySchema.extractSome('day');
```
```ts
// in your .d.ts, add following to get type def everywere
import { FilterEnum, Writeable, ZodEnumDef, ZodType, typecast } from "zod";

declare module 'zod' {
    interface ZodEnum<T extends [string, ...string[]]> extends ZodType<T[number], ZodEnumDef<T>> {
        excludeSome<ToExclude extends readonly [T[number], ...T[number][]]>(...values: ToExclude): ZodEnum<typecast<Writeable<FilterEnum<T, ToExclude[number]>>, [string, ...string[]]>>;
        extractSome<ToExtract extends readonly [T[number], ...T[number][]]>(...values: ToExtract): ZodEnum<Writeable<ToExtract>>;
    }
}
```

IN case you are using js,

```js
[...]

PR #3031, opened 2023-12-09
Support single value in enum's .exclude & .extract

rakheesingh:
#3008 
Added changes for exclude and extract filter, Once you receive the input checking wheater it is array or a single string and converting into array if single string.


colinhacks:
I'd rather keep this API consistent as an array. It also gives us more options for future API expansion, including the changes introduced here: https://github.com/colinhacks/zod/pull/3240

PR #3036, opened 2023-12-10
Remove eslint no-this-alias hackaround.

rotu:
This is a slight readability and style improvement. Instead of saving `this` to a temporary variable, save the `ZodFunctionDef` instance.

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -3746,15 +3746,14 @@ export class ZodFunction<
 
     const params = { errorMap: ctx.common.contextualErrorMap };
     const fn = ctx.data;
+    const fnDef = this._def;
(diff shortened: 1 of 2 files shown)


rotu:
I'm not quite sure I understand this code correctly and I think it may be broken in design. For instance if the return type is a `ZodUnion<[ZodNumber, ZodPromise<ZodNumber>]>` I don't know what it's expected to do.

colinhacks:
> I'm not quite sure I understand this code correctly and I think it may be broken in design

Yeah I believe you're right. I appreciate the PR but as you've observed `ZodFunction` needs to be totally re-thought, so I'm going to close. Thanks though 👍

Issue #2236, opened 2023-03-23
z.string().url() seems to accept any string


obrassard:
Hello, I think I may have found a bug with URL validation.

Indeed, I am trying to validate an object which has a property “ticketLink" that should be an URL (z.string().url())

This is my schema : 
```ts

const schema = z.object({
    projectId: z.string().uuid(),
    title: z.string().min(5).max(100),
    description: z.string().max(255).optional(),
    punchTypeId: z.string().uuid(),
    hours: z.number().min(0.25).max(24),
    ticketLink: z.string().url(),
    billable: z.boolean().optional(),
    date: z.date(),
});
```

However, no matter what I am setting as a value for the `ticketLink`property, the schema does not return an error for this property, even if it's not a valid URL.

For instance, validating this object :
```ts
const formState = {"date": 2023-03-23T20:00:05.466Z, "projectId": "28dfcf92-5b36-4976-1d48-08db1f3ca426", "ticketLink": "abc"}
schema.safeParse(formState)
```

doesn't raise an error for the `ticketLink` field, but "abc" should not be accpeted as an url.

We're using **"zod": "3.21.4"**

ahmafi:
This is working fine for me, I'm getting the `Invalid url` message in the errors list.

```ts
import { z } from 'zod';

const schema = z.object({
  projectId: z.string().uuid(),
  title: z.string().min(5).max(100),
  description: z.string().max(255).optional(),
  punchTypeId: z.string().uuid(),
  hours: z.number().min(0.25).max(24),
  ticketLink: z.string().url(),
  billable: z.boolean().optional(),
  date: z.date(),
});

const formState = {
  date: '2023-03-23T20:00:05.466Z',
  projectId: '28dfcf92-5b36-4976-1d48-08db1f3ca426',
  ticketLink: 'abc',
};

const result = schema.safeParse(formState);
if (!result.success) console.log(result.error.issues);
```

JacobWeisenburger:
Everything is working as expected for me. Please send a reproducible example if you are still having issues.

```ts
const schema = z.object( {
    ticketLink: z.string().url(),
} )

const formState = {
    ticketLink: 'abc',
}
const result = schema.safeParse( formState )
!result.success && console.log( result.error.issues )
// [
//     {
//         validation: 'url',
//         code: 'invalid_string',
//         message: 'Invalid url',
//         path: [ 'ticketLink' ]
//     }
// ]
```

obrassard:
[...]

PR #3049, opened 2023-12-13
fix: make URL validation more strict

RobertPechaCZ:
- allow only URLs with valid hostname and forbid TLD without subdomains

Fixes https://github.com/colinhacks/zod/issues/2236

### DIFF
--- a/deno/lib/__tests__/string.test.ts
+++ b/deno/lib/__tests__/string.test.ts
@@ -243,9 +243,17 @@ test("url validations", () => {
   const url = z.string().url();
   url.parse("http://google.com");
   url.parse("https://google.com/asdf?asdf=ljk3lk4&asdf=234#asdf");
+  url.parse(
+    "https://anonymous:flabada@developer.mozilla.org/en-US/docs/Web/API/URL/password"
+  );
   expect(() => url.parse("asdf")).toThrow();
+  expect(() => url.parse("http:.......///broken.com")).toThrow();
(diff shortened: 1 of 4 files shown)


Bartek532 (approved):

deno/lib/__tests__/string.test.ts
```diff
@@ -166,9 +166,17 @@ test("url validations", () => {
...
+    "https://anonymous:flabada@developer.mozilla.org/en-US/docs/Web/API/URL/password"
+  );
   expect(() => url.parse("asdf")).toThrow();
+  expect(() => url.parse("http:.......///broken.com")).toThrow();
+  expect(() => url.parse("c:")).toThrow();
+  expect(() => url.parse("WWW:WWW.COM")).toThrow();
   expect(() => url.parse("https:/")).toThrow();
+  expect(() => url.parse("https://asdf")).toThrow();
   expect(() => url.parse("asdfj@lkjsdf.com")).toThrow();
+  expect(() => url.parse("http://localhost")).toThrow();
```
m10rten, line 179: `http://localhost` is a valid URL.
It translates to `http://localhost:80/`
Same goes for `https://localhost`, that goes to `https://localhost:443`

colinhacks:
Thanks, I think this is a good idea, and perhaps a good middleground between the current `new URL` approach and the old URL regex (which was ultimately untenable imo).

PR #3051, opened 2023-12-13
feat: add support for KSUIDs

### LINKED ISSUE #3050: Support for KSUIDS
We've already got validators for other types of UID (UUID, ULID, CUID etc.), so rather than having to use a regex when working with KSUIDs, it would be nice if we had the same experience:

```ts
const ksuid = z.string().ksuid()
```

stefanlogue:
Resolves #3050

### DIFF
--- a/README.md
+++ b/README.md
@@ -869,6 +869,7 @@ z.string().cuid();
 z.string().cuid2();
 z.string().ulid();
 z.string().xid();
+z.string().ksuid();
 z.string().regex(regex);
 z.string().includes(string);
 z.string().startsWith(string);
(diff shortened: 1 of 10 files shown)


rixrix:
hey team

I hope you're all doing well. I wanted to check in regarding the status of this PR. I've been working on a project where I'm using KSUID, and I believe it's important to have validation in place for it.

colinhacks:
Thanks and apologies for the long delay in reviewing. This wasn't a feature I felt comfortable merging until Zod had a better tree-shaking setup. I've merged this into the `v4` branch where I'll be working on a big rewrite, including some work on treeshaking. Note that it may undergo some API changes before release, and it may take some time before Zod 4 lands.

PR #3067, opened 2023-12-18
Add docs for coercing nullish values

rbuetzer:
While the `pipe()` documentation contains a section on how to fix issues with nullish values in `coerce()`. The `coerce()` documentation does not mention that there are issues. This was a pitfall for us. Thus, I'd like to have the nullish value edge cases covered in the `coerce()` section.

Unfortunately, I do not have the language skills to update the Chinese documentation accordingly.


colinhacks:
Thanks, this is worth clarifying.

PR #3070, opened 2023-12-19
feat: add support for jwt

### LINKED ISSUE #2946: Add support for jwt

mirceaalexandru:
Resolves #2946

### DIFF
--- a/README.md
+++ b/README.md
@@ -741,6 +741,7 @@ z.string().uuid();
 z.string().cuid();
 z.string().cuid2();
 z.string().ulid();
+z.string().jwt();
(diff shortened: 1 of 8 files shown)


colinhacks:
Thanks! I'm merging this into the `v4` branch as a starting point, but be aware this feature may undergo some changes before shipping. It also may be some time before it lands in a stable release. This is a feature I didn't want to include until Zod has better tree-shaking, which will happen in v4.

PR #3078, opened 2023-12-21
fix wrong type in custom validator function data argument

Fgc17:
The data argument of the function passed to custom validator was causing the following issue:

```
Argument of type '(data: ...) => boolean' is not assignable to parameter of type '(data: unknown) => any'.
```

I solved it by using a ``D`` generic:
```ts
custom: <T>(check?: ((data: unknown) => any) ... -> custom: <T, D>(check?: ((data: D) => any) ...
```


colinhacks:
This has been fixed elsewhere. Thanks for the PR though!

PR #3081, opened 2023-12-22
refactor: refactor getparsedType

mohamadrezamomeni:
Hello
I recently came across Zod and used it in a project with great results and started reading Zod code.


mohamadrezamomeni:
@colinhacks can you give me your opinion?

colinhacks:
I appreciate this PR, but the parsing logic is going to be extensively rewritten in Zod 4 and that's my priority right now. Unless there's evidence that this is dramatically faster than the current implementation, I'd rather not mess with this too much.

Issue #3077, opened 2023-12-21
Exception handling in transform or safeParse


zaripych:
```
import { z } from "zod";

const invalidJson = '{"ax: 1}'

console.log(
  z
    .string()
    .transform((val) => JSON.parse(val))
    .safeParse(invalidJson)
);
```

--

Expected:
```json
{ "success": false, "error": ... }
```

--

Result:
It throws the exception. "Unexpected end of JSON input".

--

Basically exceptions are not handled in `transform` or `safeParse` as you would expect them to. 

I've checked the documentation and it doesn't mention anything about handling exceptions in `transform`.

There seem to be similar behaviour for `refine` as well.

zod package version - 3.22.4

JacobWeisenburger:
Perhaps this will help:
https://github.com/JacobWeisenburger/zod_utilz#json

zaripych:
Yeah, I ended up handling exception on my own. But it felt surprising that `safeParse` actually doesn't handle exceptions. 👍

JacobWeisenburger:
are you wanting this to stay open as an enhancement? or are you satisfied with your solution?

zaripych:
Thank you. Yes. It seems reasonable to me to handle the exception in `safeParse` at the very least or improve the `.transform` to handle the exceptions and add the error to the context automatically, but if that complicates the use of the library or makes breaking changes that people don't want - I'm fine with closing it.

I think the main benefit of handling the exception inside the `transform` by users is more fine grained control over the error message. But if that is important to users they can always add it later.

The benefit of handling the exception in Zod itself is brewety and simplicity. Ie:

```
.tranform((value) => JSON.parse(value))
```

is shorter than

```
.tranform((value, ctx) => {
  try {
    return JSON.parse(value)
  } catch (err) {
    ctx.addIssue('custom', ...);
    return z.NEVER;
  }
})
```

There is just so much going on in `.transform` now ... `ctx`, `addIssue`, `z.NEVER` ... many things to be done "correctly".

Does this seem like a good direction to you?

mmkal:
I agree, this is a common enough use-case that I think it makes sense for zod to handle correctly. Otherwise, the path of least resistance is to do `z.object({ fooConfig: z.string().transform(s => JSON.parse(s)).pipe(z.object({ abc: z.number() }))`

[...]

PR #3109, opened 2024-01-05
Add z.string().json(...) helper

mmkal:
Based on https://github.com/colinhacks/zod/issues/3077#issuecomment-1878470503. I started playing around to see how quick it would be to implement, and it ended up being very quick - so opening a PR in hopes the idea is accepted.

Copying from the readme for the `z.string().json(...)` method - the linked issue has more details on the motivation:

### JSON

The `z.string().json(...)` method parses strings as JSON, then pipes the result to another specified schema.

```ts
const Env = z.object({
  API_CONFIG: z.string().json(
    z.object({
      host: z.string(),
      port: z.number().min(1000).max(2000),
    })
  ),
  SOME_OTHER_VALUE: z.string(),
});

const env = Env.parse({
  API_CONFIG: '{ "host": "example.com", "port": 1234 }',
  SOME_OTHER_VALUE: "abc123",
});

env.API_CONFIG.host; // returns parsed value
```

If invalid JSON is encountered, the syntax error will be wrapped and put into a parse error:

```ts
const env = Env.safeParse({
  API_CONFIG: "not valid json!",
  SOME_OTHER_VALUE: "abc123",
});

if (!env.success) {
  console.log(env.error); // ... Unexpected token n in JSON at position 0 ...
}
```

This is recommended over using `z.string().transform(s => JSON.parse(s))`, since that will not catch parse errors, even when using `.safeParse`.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -581,7 +581,8 @@ export type ZodStringCheck =
     }
   | { kind: "duration"; message?: string }
   | { kind: "ip"; version?: IpVersion; message?: string }
-  | { kind: "base64"; message?: string };
+  | { kind: "base64"; message?: string }
+  | { kind: "json"; message?: string };
 
 export interface ZodStringDef extends ZodTypeDef {
   checks: ZodStringCheck[];
@@ -1029,6 +1030,18 @@ export class ZodString extends ZodType<string, ZodStringDef, string> {
           });
           status.dirty();
         }
+      } else if (check.kind === "json") {
+        try {
+          JSON.parse(input.data);
+        } catch (err) {
+          ctx = this._getOrReturnCtx(input, ctx);
+          addIssueToContext(ctx, {
+            code: ZodIssueCode.invalid_string,
+            validation: "json",
+            message: check.message,
+          });
+          status.dirty();
+        }
       } else {
         util.assertNever(check);
       }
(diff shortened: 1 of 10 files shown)


src/types.ts
```diff
@@ -957,6 +957,21 @@ export class ZodString extends ZodType<string, ZodStringDef> {
...
 
+  json<T extends ZodTypeAny>(shape: T) {
+    return this.transform((val, ctx) => {
+      try {
+        return JSON.parse(val);
+      } catch (error: unknown) {
+        ctx.addIssue({
+          code: ZodIssueCode.invalid_string,
+          validation: "json",
+          message: (error as Error).message,
```
m10rten, line 968: With a quick scan, can't find anywhere else where `error as Error` is used.
So if you need it to be a `message: string`, check for `instanceof` first, or use the `...errorUtil.errToObj`

mmkal: IMO `as` is better here than doing something at runtime. `JSON.parse(...)` only throws `Error`s (specifically `SyntaxError`s) as [documented by MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#exceptions). I also don't feel particularly strongly though.

colinhacks:
Thanks, great stuff. I'm merging this into the `v4` branch as a starting point, but the API is likely to change before this lands. I actually think this would be better as a top-level `z.json()`, and `z.string().json()` should be reserved for verifying that the string is in fact a valid JSON string.

m10rten:
@colinhacks what would be the feature of a `z.json`?
Parsing and stringify?

colinhacks:
Yes, a schema that encapsulates the `JSON.parse` step. `Input` is `string`, `Output` is inferred from the arguments:

```ts
const schema = z.json(z.object({ name: z.string() }))
schema.parse(`{ "name": "Maarten" }`)
```

m10rten:
That looks clean, however would that not be this: (?)

```ts
const jsonStr = `{ "name": "Jhon Doe" }`;
const schema= z.object({
  name: z.string(),
});
const anyJsonObj: unknown = JSON.parse(jsonStr);
const parsed = schema.parse(anyJsonObj);
// ^? type: z.infer<typeof schema>
```

I could be wrong, but the `.json` would be a replacement for `JSON.parse`, where this function: [MDN Docs: JSON.parse](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) also takes in a `reviver`, how are you planning to resolve that since any object can be different, users can make their own revivers and set custom properties to, for example, make a map or set in JSON.

So by making this 'replacement' as to call it, would you not need way more steps to have the same, generic, result?

mmkal:
Was just going to comment re: this, it would be good to have a method like

```ts
const Config = z.json(
  z.object({
    apiKey: z.string(),
    templateId: z.string(),
  })
).reviver((k, v) => typeof v === 'number' ? v.toString() : v)

const result = Config.parse(process.env.CONFIG) // '{"apiKey":"x", "templateId":123}' => {apiKey: 'x', templateId: '123'}
```

You could often achieve a similar result with `.transform` or `.preprocess` but there are still plenty of valid use cases for `reviver`.

But @m10rten it's not really the same as calling `JSON.parse(jsonStr)` - that will throw an error directly. It won't be encapsulated by `safeParse`, won't be wrapped as a zod error. And it becomes much more inconvenient to compose, e.g. if it's in a nested property like:

```ts
const BigSchema = z.object({
  type: z.string(),
  foo: z.number(),
  env: z.object({
    NODE_ENV: z.string(),
    CONFIG: z.json(z.object({apiKey: z.string()})).reviver(...)
  }),
})
```

Issue #2846, opened 2023-10-07
Build in File and FileList Support


iamalipe:
# Add a support for File and FileList.

example

```ts
const size = 10000 // byte
z.flle().size(10000).allowType(["image/jpeg","image/png"])

z.fileList().min(1).max(10).size(10000).allowType(["image/jpeg","image/png"])
```

kaciakmaciak:
Hey, I've created a PR for this and happy to hear from you. https://github.com/colinhacks/zod/pull/3118

colinhacks:
This PR has landed in the `v4` branch but it may take some time before it lands. In the meantime, I recommend `z.instanceof`:

```ts
z.instanceof(File)
```

colinhacks:
`z.file()` has landed in Zod 4: https://zod.dev/api#files

I recommend `z.instanceof(FileList)` if you need that functionality. Zod only added support for `File` now that it's implemented in virtually every major server- and browser-side runtime. The same can't be said for `FileList` so it won't be getting added to core.

PR #3118, opened 2024-01-07
Add support for `file` validation

kaciakmaciak:
https://github.com/colinhacks/zod/issues/2846

Keen on the feedback. Cheers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1184,6 +1184,42 @@ console.log(dateSchema.safeParse("0000-00-00").success); // false
 
 For older zod versions, use [`z.preprocess`](#preprocess) like [described in this thread](https://github.com/colinhacks/zod/discussions/879#discussioncomment-2036276).
 
+## Files (Browser only)
+
+Use z.file() to validate `File` instances.
+
+```ts
+z.file().safeParse(new File(["foobar"], "foobar.txt", { type: "text/plain" })); // success: true
+z.file().safeParse("foobar"); // success: false
+```
+
+You can customize certain error messages when creating a file schema.
+
+```ts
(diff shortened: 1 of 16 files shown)


colinhacks:
This is great! I have some minor API nitpicks, but more relevantly, I'm hesitant to merge this into Zod 3 due to bundle size constraints. I plan to ship this in Zod 4, which will have better tree-shaking.

colinhacks:
I'm pretty sure the `@web-std` types would have to be a `dependency` or `peerDependency`. Also `File` is supported in Node.js 20 and exists as a global. I have to do some experiments but I think a better solution may be to declare an empty `interface File` that extends any externally-defined global `File` interface, should one exist. I'll do some experiments.

colinhacks:
I made a few small changes to this (excellent) PR.

I've renamed the methods to match the properies of `File`.

```
.accept() -> .type()
.filename() -> .name()
```

I like the idea of `accept` and trying to agree with `input[type=file]`. but it's a little fraught. The behavior of the `accept` attribute in the browser is complicated. For now I'd rather provide a clean, well-defined API for validating MIME types, and we'll see if users request support for file extensions. The old API also muddies the water, in that it mixes together the validation logic for `.type` and `.filename`. 

For similar reasons, I changed `.filename()` to just `.name()` to agree with that property.

I added a new ZodIssue subtype for `invalid_file_name` to avoid that `custom` error.

---

I'm merging this now into the `v4` branch, but a) this API is still likely to undergo some modification and b) it may be a while before landing in a stable release. Thanks again for the excellent PR! 👍

PR #3124, opened 2024-01-10
Short circuit preprocess only on own errors

### LINKED ISSUE #3123: Order of z.object props with preprocess gives wrong parse results
Version: **3.22.4**

It seems that Zod is unable of properly validating props with `preprocess` if they are not placed at the beginning of the Zod Object.

The following schema should always return an error for `second`.
```ts
const schemaWrong = z.object({
  first: z.string(),
  second: z.preprocess(() => undefined, z.string()),
});

schemaWrong.safeParse({}) // Shows only 1 error in `first`
```

If we switch the order of the props, then it works: 
```ts
const schemaFixed = z.object({
  second: z.preprocess(() => undefined, z.string()),
  first: z.string(),
});

schemaWrong.safeParse({}) // Correctly shows 2 errors.
```

Playground: https://stackblitz.com/edit/typescript-wqj5ep?file=schemas%2FschemaWithDefaultType.ts

Note: Version **3.21** works correctly

selimb:
Fixes #3123

### DIFF
--- a/deno/lib/__tests__/transformer.test.ts
+++ b/deno/lib/__tests__/transformer.test.ts
@@ -279,6 +279,27 @@ test("z.NEVER in preprocess", async () => {
   }
 });
 
+// Fix for #3123
+test("preprocess only short circuit on its own issues", () => {
+  const schema1 = z.object({
+    first: z.string(),
+    second: z.preprocess(() => undefined, z.string()),
+  });
+  const schema2 = z.object({
+    second: z.preprocess(() => undefined, z.string()),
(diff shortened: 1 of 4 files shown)


haines:
We ran into #3123 when using the `zfd.text` helper from `zod-form-data`: https://github.com/airjp73/remix-validated-form/issues/349

I've applied this PR using `patch-package` and can confirm it resolves the issue for us.

selimb:
Duplicate of #2912. Looks like there were multiple issues/PRs for the same thing.

Issue #3103, opened 2024-01-03
add method to find if the given field is required


sajjathossain:
I recently ran into a scenario where I need to know if the given field is required or not. but I was unable to find any solution for it. something like this would be nice. e.g.:
```ts
ZodSchema.isRequired("fieldName")
```

JacobWeisenburger:
Is this what you are looking for?

```ts
const isFooRequired = !z.object( { foo: z.string() } ).shape.foo.isOptional()
```

If you found my answer satisfactory, please consider supporting me. Even a small amount is greatly appreciated. Thanks friend! 🙏
https://github.com/sponsors/JacobWeisenburger

JacobWeisenburger:
Are you looking for a workaround or a new feature?

sajjathossain:
> Are you looking for a workaround or a new feature?

a feature

anugoyal998:
Can I start working on it? Please assign it to me

anugoyal998:
Can I work on this feature?

fernandollisboa:
I can work on that!

colinhacks:
Sorry, this feature is too niche to merit inclusion in core, especially with `isOptional()` already existing. Closing 🤙

PR #3126, opened 2024-01-10
feat: add method to find if the given field is required #3103

anugoyal998:
Added the method to check if a given key is required or not.

```ts
const schema = z.object({
  name: z.string()
})

console.log(schema.isKeyRequired("name")) // true
console.log(schema.isKeyRequired("foo")) // false
```

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -2692,6 +2692,12 @@ export class ZodObject<
     ) as any;
   }
 
+  isKeyRequired(key: string) {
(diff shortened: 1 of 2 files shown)


colinhacks:
There doesn't seem to be a bunch of demand for this feature, so I'm going to close this. Thanks for the contribution though! 

PS You can already do this, which is pretty concise:

```ts
const schema = z.object({
  name: z.string()
})

!schema.shape.name.isOptional()
```

Issue #2654, opened 2023-08-14
Schema in object being inferred differently (and weirdly)


ericallam:
See the following TS snippet:

```ts
import { z } from "zod";

const EventNameSchema = z.string().or(z.array(z.string()));

type EventName = z.infer<typeof EventNameSchema>;
// EventName is string | string[]

const EventSchema = z.object({
  name: z.string().or(z.array(z.string())) // this is the same as the EventNameSchema
});

type EventWithName = z.infer<typeof EventSchema>;
type EventName2 = EventWithName["name"];
// EventName2 is (string | string[]) & (string | string[] | undefined)
```

And the TS playground: [link](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgLzgXzgMyhEcBEyEAJvgNwBQFAxhAHYDO8AogG4CmdMAcgIYjsAytQAW7ELzgBeFADomUYHQDmACgCUs6KuSzeUKLwCeO+TEUqN69ZQowjYdnDace-JzN1LM7KAB57RwhMZw4uPgFhMQkAPkoAenjQ1winYAY4BSVlOAAfTPNsgG0AXSpaRhYwmCjxSU8tACMAK3ZqGFUECjg4OncALjksy01tXX1DE11htWt1CjQbKkCnFy4AdWAYEVTpOW9fAId2YOSuWtjKFbO3AQAmPbWYTe3Uovw+gXwShKSn1Ie6TgqhmeQKFmUpXUcAAZMDQfkZqUwQBXOjEdiYJTsYjqIA)

I'm not sure if this is intended or a bug or maybe just a user error. Using zod `3.21.4` and TS `4.8.4`

algora-pbc:
💎 **$25** bounty created by @ericallam
👉 To claim this bounty, [submit your pull request on Algora](https://console.algora.io/bounties/cllo7okue0002mr0f69v7n51u)
📝 Before proceeding, please make sure you can [**receive payouts**](https://docs.algora.io/bounties/payments#country-support) in your country
💵 Payment arrives in your account 2-5 days after the bounty is rewarded
💯 You keep 100% of the bounty award
🙏 Thank you for contributing to colinhacks/zod!

kremedev:
Add `"strictNullChecks": false` to tsconfig.json

zcesur:
had a similar issue recently, seems to be due to the split & intersect approach for inferring object types

typing `addQuestionMarks` more precisely does resolve this issue:

```diff
[...]

PR #3138, opened 2024-01-14
Schema in object being inferred differently

abhi12299:
fixes: https://github.com/colinhacks/zod/issues/2654

Done by explicitly checking if there are any optional keys in the object and then adding to `addQuestionMarks` type in `utils`

### DIFF
--- a/deno/lib/__tests__/unions.test.ts
+++ b/deno/lib/__tests__/unions.test.ts
@@ -63,3 +63,11 @@ test("readonly union", async () => {
   union.parse("asdf");
   union.parse(12);
 });
+
+test("nested no undefined", () => {
+  const inner = z.string().or(z.array(z.string()));
+  const outer = z.object({ inner });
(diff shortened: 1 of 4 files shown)


abhi12299:
Unable to add/commit test case to the PR. Ran tests on local and it works fine. All tests pass!

abhi12299:
> Unable to add/commit test case to the PR. Ran tests on local and it works fine. All tests pass!

Was adding test in `deno/lib/__tests__`. Now, added the test in `src/__tests__/unions.test.ts`.

abhi12299:
@colinhacks your thoughts on this?

colinhacks:
Thanks, this is a great distillation of the problem and a smart fix. The `addQuestionMarks` alias has undergone quite a bit of iteration recently and the root problem has been solved. I added your test case in #0a80b66. Thanks again!

PR #3141, opened 2024-01-15
nonNullable to disallow null

sezanzeb:
```ts
z.any().nonNullable()
z.nonNullable(z.any())
```

can both be now used to make this throw:

```ts
z.any().nonNullable().parse(null)
```

This is not equivalent to unwrap (as proposed in https://github.com/colinhacks/zod/issues/2190). Instead, this is equivalent to typescripts NonNullable type https://www.typescriptlang.org/docs/handbook/utility-types.html#nonnullabletype

### DIFF
--- a/README.md
+++ b/README.md
@@ -1179,6 +1179,31 @@ const nullableString = stringSchema.nullable();
 nullableString.unwrap() === stringSchema; // true
 ```
 
+## NonNullables
+
+You can also make something NonNullable, using `z.nonNullable()`.
+
+```ts
+const nullableString = z.nonNullable(z.any());
+nullableString.parse("asdf"); // => "asdf"
+nullableString.parse(null); // fails
+```
+
+Or use the `.nonNullable()` method.
+
+```ts
+const E = z.any().nonNullable();
+type E = z.infer<typeof E>; // NonNullable<any>
+```
+
+Extract the inner schema with `.unwrap()`.
+
+```ts
+const stringSchema = z.any();
+const nullableString = stringSchema.nonNullable();
+nullableString.unwrap() === stringSchema; // true
+```
+
 ## Objects
 
 ```ts
--- a/deno/lib/README.md
+++ b/deno/lib/README.md
@@ -1179,6 +1179,31 @@ const nullableString = stringSchema.nullable();
 nullableString.unwrap() === stringSchema; // true
 ```
 
+## NonNullables
+
+You can also make something NonNullable, using `z.nonNullable()`.
+
+```ts
+const nullableString = z.nonNullable(z.any());
+nullableString.parse("asdf"); // => "asdf"
+nullableString.parse(null); // fails
+```
+
+Or use the `.nonNullable()` method.
+
+```ts
+const E = z.any().nonNullable();
+type E = z.infer<typeof E>; // NonNullable<any>
+```
+
+Extract the inner schema with `.unwrap()`.
+
+```ts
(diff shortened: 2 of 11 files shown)


colinhacks:
I appreciate the PR, but this is a little to niche to go into Zod core. This is better implemented with a transform. 

```ts
const A = z.string().nullable().transform((val, ctx) => {
  if(val === null) {
    ctx.addIssue({
      code: z.ZodIssueCode.custom,
      message: "Cannot be null",
      path: [],
    })
    return z.NEVER;
  }
  return val;
})

type A = z.infer<typeof A>;
// string
```

sezanzeb:
I obviuosly disagree, otherwise I wouldn't have made the PR. It would have come handy at work when I decided to work on this.

Furthermore it is compliant with typescripts NonNullable, and not a funky Idea that I just made up. So if NonNullable would ever be a thing, it would be in the way this PR proposes.

The typescript developers didn't think a NonNullable type is too niche to add it to typescript.

Transformers are flexible, but they are not nearly as readable and beautiful like the zods other features.

And also, this addition already has automated tests and is fully documented. This is a finished feature that doesn't add immediate development overhead for you. It's as good as it gets.

sezanzeb:
I think zod should be made as similar to typescripts types as possible. I want to be able to replace all interfaces and such with zod types with beautiful zod structures.

Well, my 2 cents.

sezanzeb:
I guess I should also add that I didn't encounter a situation in which I wanted to use this since making the PR, maybe someone else should comment on this instead. I can't give you a realistic estimate on how commonly this feature would be used. On the other hand, there are tons of other existing zod features that I didn't consider using since then as well, so...

michaelhays:
What's more is the transform workaround [here](https://github.com/colinhacks/zod/pull/3141#issuecomment-2014120888) doesn't seem to work for me with `z.any()`, as `nestedValue` is still optional:

```ts
const schema = z.object({
  nestedValue: z
    .any()
    .nullable()
    .transform((val) => (val == null ? z.NEVER : val)),
})

type SchemaType = z.infer<typeof schema>
```

```ts
type SchemaType = {
  nestedValue?: any;
}
```

sezanzeb:
nonNullable also restricts `undefined` (https://www.typescriptlang.org/docs/handbook/utility-types.html#nonnullabletype)

> It's as good as it gets.

That aged well

committed the change.

---

use case: A gateway passes along all payloads it gets, and it doesn't care for their content, but it does check if the payloads are set by using `z.unknown().nonNullable()`

colinhacks:
> nonNullable also restricts undefined

The fact that `.nonNullable()` would not be the inverse of `.nullable()` is quite annoying to me. But adding a `.notNullable()` method that only disallows `null` would be a footgun for those who expect it to behave like `NonNullable<T>`. Two bad options, imo, so it's not likely I take any action here.

sezanzeb:
Something that I would like to do quite often:

```ts
class A {
  private foo: whatever | null = null;
  
  // ... some stuff ...

  public bar() {
    z.nonNullable().parse(this.foo).whateversMethod()
  }
}
```

You could also add an assertion method there (which isn't actually typesafe at all), or a guard clause that checks for === null and throws an error conditionally, but in many cases I prefer to use zod for such things.

zod has, in many cases, become something like runtime-type-validation for me.

racinette:
@colinhacks would you consider reviving this for zod v4? There currently seems to be no way to make a type non-nullable, once it's already nullable.

Issue #3048, opened 2023-12-13
Error message based on input value


Jack-Works:
I need to generate error message based on the input value, now I have to use errorMap which is tedious.

I hope I can write code like this:

```ts
z
.string({ invalid_type_error: value => 'Expected a string, received: ' + String(value) })
.min(2, { message: value => 'Expected a string with at least 2 characters, received: ' + String(value) })
```

JacobWeisenburger:
Is this what you are looking for?

```ts
const schema = z.string( {
    errorMap: ( issue, ctx ) => {
        switch ( issue.code ) {
            case 'invalid_type':
                return { message: `Expected a string, received: ${ ctx.data }` }
            case 'too_small':
                return { message: `Expected a string with at least 2 characters, received: ${ ctx.data }` }
            default:
                return { message: ctx.defaultError }
        }
    }
} ).min( 2 )

const result1 = schema.safeParse( 42 )
!result1.success && console.log( result1.error.issues )
// [
//     {
//       code: "invalid_type",
//       expected: "string",
//       received: "number",
//       path: [],
//       message: "Expected a string, received: 42"
//     }
// ]

const result2 = schema.safeParse( 'a' )
!result2.success && console.log( result2.error.issues )
// [
//     {
//       code: "too_small",
//       minimum: 2,
//       type: "string",
//       inclusive: true,
//       exact: false,
//       message: "Expected a string with at least 2 characters, received: a",
//       path: []
//     }
// ]
```

If you found my answer satisfactory, please consider supporting me. Even a small amount is greatly appreciated. Thanks friend! 🙏
https://github.com/sponsors/JacobWeisenburger

Jack-Works:
> Is this what you are looking for?

This is possible, but annoying to write. I just want to write those error inline.

Aniket1026:
> I need to generate error message based on the input value, now I have to use errorMap which is tedious.
> 
> I hope I can write code like this:
> 
> ```ts
> z
> .string({ invalid_type_error: value => 'Expected a string, received: ' + String(value) })
> .min(2, { message: value => 'Expected a string with at least 2 characters, received: ' + String(value) })
> ```

I Agree with you as I'm also looking for exactly the same thing as having your own custom errors is super helpful. I hope they do this soon but have you found any better approach or way of doing the same thing which you've requested?

colinhacks:
[...]

PR #3152, opened 2024-01-17
Error message based on input value

joaoGabriel55:
Issue: #3048

Generate error message based on the input value
```ts
z
  .string({ invalid_type_error: value => 'Expected a string, received: ' + String(value) })
  .min(2, { message: value => 'Expected a string with at least 2 characters, received: ' + String(value) })
```

### DIFF
--- a/deno/lib/__tests__/error.test.ts
+++ b/deno/lib/__tests__/error.test.ts
@@ -516,31 +516,116 @@ test("literal bigint default error message", () => {
   }
 });
 
-test("custom error message based on input value", () => {
-  const schema = z
-    .string({
-      invalid_type_error: (value) => `Expected a string, received: ${value}`,
-    })
-    .min(2, {
-      message: (value) =>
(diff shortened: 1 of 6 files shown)


colinhacks:
Thanks Gabriel. I think if we were going to support a function here, it should conform to the `ZodErrorMap` interface. Essentially you'd be providing an error map that would be used whenever this is error happens. 

```ts
export type ZodErrorMap = (
  issue: ZodIssueOptionalMessage,
  _ctx: ErrorMapCtx
) => { message: string };
```

This is a pretty major project, since it should be supported everywhere consistently, including in all the schema constructors, not just `z.string()`. 

Also, the `ZodErrorMap` type definition will be changing (slightly) in Zod 4 😅  I think this is a great API and I'm putting this on my list of Zod 4 features, but it's probably better that I implement it. I really appreciate your work here though, and it's a great PR!

PR #3158, opened 2024-01-18
add custom error message on nativeEnum validation

### LINKED ISSUE #2919: nativeEnum invalid_type_error not returning error on wrong enum
Zod returns `Invalid enum value` although `invalid_type_error` is set.

## Steps to reproduce

```ts
const schema = z.nativeEnum({MY_ENUM:"MY_ENUM"}, {
  invalid_type_error: "this is an error",
})

console.log("result", schema.parse("AAA"));
// returns error "Invalid enum value. Expected 'MY_ENUM', received 'AAA'"
```

## Expected behavior
When the type is wrong, Zod should return whatever I give in  `invalid_type_error`.
In the example, it should return `this is an error`

If `invalid_type_error` is meant to be used for something else, I would expect the following:
1. Add in the `nativeEnum` documentation section what is `invalid_type_error` for
2. Add with JS Docs what is `invalid_type_error` for
3. Add an `invalid_enum_error` to the `params` of `nativeEnum` where we can define a custom message.

fernandollisboa:
Resolves #2919

Return `invalid_type_error` on nativeEnum validations instead of  `Invalid enum value` when the attribute is set.

```ts
const schema = z.nativeEnum({ MY_ENUM: "MY_ENUM" }, {
  invalid_type_error: "the value provided is invalid",
})

console.log("result", schema.parse("AAA"));
// returns error "the value provided is invalid"
```

### DIFF
--- a/deno/lib/__tests__/enum.test.ts
+++ b/deno/lib/__tests__/enum.test.ts
@@ -59,19 +59,27 @@ test("extract/exclude", () => {
   util.assertEqual<z.infer<typeof EmptyFoodEnum>, never>(true);
 });
 
-test('error map in extract/exclude', () => {
+test("error map in extract/exclude", () => {
   const foods = ["Pasta", "Pizza", "Tacos", "Burgers", "Salad"] as const;
-  const FoodEnum = z.enum(foods, { errorMap: () => ({ message: "This is not food!" }) });
+  const FoodEnum = z.enum(foods, {
+    errorMap: () => ({ message: "This is not food!" }),
+  });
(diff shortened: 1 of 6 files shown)


joaoGabriel55 (approved):
LGTM. Nice tests!

scarabcoder:
Is there an update on this?
Currently working with a list of US States as a native enum, having Zod throw out an error "Invalid string, expected one of <list of all 50 states>" doesn't make for the greatest user experience.

colinhacks:
My preferred solution here is the one that landed in #3169, which will land in Zod 3.23 shortly.

This is all liable to change in Zod 4 as well.

PR #3169, opened 2024-01-22
feature: add attribute message to enum validatiion

### LINKED ISSUE #3146: Enum error message cannot be customized
According to #580 , custom error message for enum should be like this.

```ts
z.enum(['aaa', 'bbb', 'ccc'], {
  errorMap: () => ({ message: 'custom error message' })
})
```

which is counterintuitive. Though enum params also provided `required_error` and `invalid_type_error`, they are not working for custom errors.

Also, I assume `required_error` and `invalid_type_error` are the same for enum, so I hope custom error message should be like:

```ts
z.enum(['aaa', 'bbb', 'ccc'], { message: 'custom error message' })
```

fernandollisboa:
Resolves #3146

Make it possible to call enum validation with a custom message using `message` attribute:

```ts
  z.enum(['aaa', 'bbb', 'ccc'], { message: 'custom error message' });
```

### DIFF
--- a/deno/lib/__tests__/enum.test.ts
+++ b/deno/lib/__tests__/enum.test.ts
@@ -59,19 +59,27 @@ test("extract/exclude", () => {
   util.assertEqual<z.infer<typeof EmptyFoodEnum>, never>(true);
 });
 
-test('error map in extract/exclude', () => {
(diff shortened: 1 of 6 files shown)


colinhacks:
This has landed in Zod 3.23.

PR #3177, opened 2024-01-24
feature: create schema field isRequired() method

### LINKED ISSUE #3103: add method to find if the given field is required

fernandollisboa:
Resolves #3103

Now it is possible to check whether a given field is required or not using `schema.shape.fieldName.isRequired()`

### DIFF
--- a/deno/lib/__tests__/object.test.ts
+++ b/deno/lib/__tests__/object.test.ts
@@ -451,3 +451,12 @@ test("passthrough index signature", () => {
   type b = z.infer<typeof b>;
   util.assertEqual<{ a: string } & { [k: string]: unknown }, b>(true);
 });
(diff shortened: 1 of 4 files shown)


colinhacks:
Thanks for your work on this Fernando. At the moment I don't think we want to include this in core - it's too niche. I left a comment on https://github.com/colinhacks/zod/issues/3103. There are also some gotchas with `.isOptional()` in that it doesn't properly handle async schemas, which is part of the reason why it's not documented, and this would represent another footgun. Closing.

PR #3178, opened 2024-01-24
feature: allow falsy error message

### LINKED ISSUE #3101: Empty error message gets evaluated as falsy and overriden
Doing this

```ts
import z from 'zod';

z.number().min(0, { message: '' }).parse(-1);
```
Results in an error of 
```
Error: [
{
"code": "too_small",
"minimum": 0,
"type": "number",
"inclusive": true,
"exact": false,
"message": "Number must be greater than or equal to 0",
"path": []
}
]
```
instead of having an empty message.

Would it be possible to explicitly check if the message is undefined rather than checking if it's falsy? After asking this on the Discord scotttrinh found the line causing this

https://github.com/colinhacks/zod/blob/3e4f71e857e75da722bd7e735b6d657a70682df2/src/helpers/parseUtil.ts#L31

fernandollisboa:
Resolves #3101

Allows setting falsy error messages on validation.

```ts
  const schema = z.string().max(1, { message: "" }).parse("aaaa")
  // returns error { message: "" }
```

### DIFF
--- a/deno/lib/__tests__/error.test.ts
+++ b/deno/lib/__tests__/error.test.ts
@@ -516,6 +516,15 @@ test("literal bigint default error message", () => {
   }
 });
 
+test.only("when the message is falsy, it is used as is provided", () => {
+  const schema = z.string().max(1, { message: "" });
+  const result = schema.safeParse("asdf");
+  expect(result.success).toEqual(false);
+  if (!result.success) {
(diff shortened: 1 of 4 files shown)


colinhacks:
Accidentally merged this with a `test.only` left in there...whoops. Keep an eye our for that in the future.

PR #3181, opened 2024-01-24
fixed Readme typo in NaNs example

RashJrEdmund:
- ### was reading through docs and saw the typo in this example.

    - see the "invalid_type_error" key-value pair below

Screenshot 2024-01-24 at 11 54 53

- changed to :
     - ### "isNaN must be 'not a number'"

### DIFF
--- a/README.md
+++ b/README.md
@@ -948,7 +948,7 @@ You can customize certain error messages when creating a nan schema.
 ```ts
 const isNaN = z.nan({
   required_error: "isNaN is required",
-  invalid_type_error: "isNaN must be a number",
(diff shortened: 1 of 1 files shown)


TahsinAyman (approved):
I think it makes sense

RashJrEdmund:
Well, I now understand what it meant.
is Nan must be not a number.
I see why I got confused.
I'll fix that right a way.

colinhacks:
Can't remember why I used NaN as an example for these lol. Guess I was feeling troll-y that day. Your change makes it a bit clearer, thanks.

RashJrEdmund:
thank you too

Issue #3196, opened 2024-01-27
"Intersection results could not be merged" does not mention the cause


JacobWeisenburger:
### Discussed in https://github.com/colinhacks/zod/discussions/3194

<div type='discussions-op-text'>

<sup>Originally posted by **JoshuaKGoldberg** January 23, 2024</sup>
Zod is great and I'm enjoying the features around intersecting/merging objects. But for `.and`, when the intersections results can't be merged for some reason, the error message thrown at runtime doesn't provide any details.

> An error occurred.
> 
> ```plaintext
> Intersection results could not be merged
> ```
> 
> Stack Trace
> 
> ```plaintext
> ZodError: [
>   {
>     "code": "invalid_intersection_types",
>     "path": [],
>     "message": "Intersection results could not be merged"
>   }
> ]
>     at get error [as error] (file:///Users/josh/repos/dot-com/node_modules/.pnpm/zod@3.22.4/node_modules/zod/lib/index.mjs:538:31)
>     at ZodIntersection.parse (file:///Users/josh/repos/dot-com/node_modules/.pnpm/zod@3.22.4/node_modules/zod/lib/index.mjs:638:22)
>     at /Users/josh/repos/dot-com/src/pages/rss.xml.ts:26:48
>     at Array.map (<anonymous>)

Not sure if this is a bug report or feature request, but: could the error message also include which field(s) are problematic?

Note that I'm not trying to report that the error is wrong. The error may be right - it just informative enough to explain why clearly.

Sorry that I'm not familiar enough with Zod to provide a standalone reproduction (I tried the Stackblitz Astro starter but couldn't get past TypeScript errors, amusingly). But here's how to get it on a real world project:

```plaintext
git clone https://github.com/JoshuaKGoldberg/dot-com
cd dot-com
git checkout 708081f # ncu-u-for-vite-5 branch
pnpm i
pnpm dev
open http://localhost:4321/rss.xml
```</div>

MiaInturi:
I also faced this problem. And i think it is a good idea add additional field for `invalid_intersection_types` code with problematic field path.

For example, for the code below, the problematic field will be `property.nested1` (because of [mergeValues](https://github.com/colinhacks/zod/blob/master/src/types.ts#L3063) function behavior)

```javascript
const schema = z
  .strictObject({
    property: z
      .strictObject({
        nested1: z.function(),
        nested2: z.number()
      })
      .optional()
  })
  .and(z.custom((value) => !(value instanceof RegExp)));

[...]

PR #3208, opened 2024-02-01
[3196] add 'mergeErrorPath' property into ZodInvalidIntersectionTypes…

MiaInturi:
PR for issue #3196 

Currently `ZodInvalidIntersectionTypesIssue` does not return enough information about occurred error. We can pass object/array of big size and depth into `z.intersection(...).parse()` function, but we will not get information what exact property is a reason of merge failure. This behavior makes it difficult to debug and find the field which cause of the error.

I want to propose additional property `mergeErrorPath` in `ZodInvalidIntersectionTypesIssue` which will contain path to field that can not be merged. This way, library users will be able to see exactly where the problem is and solve it much faster and easier.

```typescript
const schema = z
  .strictObject({
    property: z
      .strictObject({
        nested1: z.function(),
        nested2: z.number()
      })
      .optional()
  })
  .and(z.custom((value) => !(value instanceof RegExp)));

const result = schema.parse({
  property: {
    nested1: () => {},
    nested2: 1000
  }
});

// Behavior before PR
{
    code: 'invalid_intersection_types',
    path: [],
    message: 'Intersection results could not be merged'
}

// Behavior after PR
{
    code: 'invalid_intersection_types',
    path: [],
    mergeErrorPath: ['property', 'nested1']
    message: 'Intersection results could not be merged'
}
```


colinhacks:
Thanks! Great PR. I think this is a good step towards better error reporting for intersections so I'm merging it into the `v4` branch.

That said, this is likely to undergo some changes before actually landing in Zod 4. I think it would be neat if Zod reported separate issues for each merge failure, instead of eagerly aborting after the first failure. But this is a really good start.

PR #3210, opened 2024-02-02
fix(deno): re-export default from index.ts in mod.ts

chadulous:
Noticed that the deno version of zod doesn't export zod as default


colinhacks:
Good eye. This fix don't work since `mod.ts` gets generated inside `deno-build.mjs`. I'll put in a fix for this. Thanks!

Issue #2396, opened 2023-05-03
Email validation broken


timothyarmes:
Hi,

zod is rejecting valid email addresses of the form a@b.c.com (note the two dots after the @).

mauricedb:
Same here. And the email in question is actually used to send emails so definitely works. The actual email we are using is more like `abcde.abcd+1234567890123@abcdefg.com.test-google-a.com` where I replaced letters for other letters etc for privacy.

reminjp:
#2274 should solve this. but not yet merged.

cullylarson:
Another example of a failing address: `a@asdf.ka1.asdf.com`

hkroger:
Is this moving or is there a version that we can downgrade to fix this?

Dimrok:
Same for `first.last@physik.uni-muenchen.de`.

Dimrok:
To give more context the current implementation favor speed over correctness, which is debatable. 
 
> // old version: too slow, didn't support unicode

```TypeScript
// from https://stackoverflow.com/a/46181/1550155
// old version: too slow, didn't support unicode
[...]

PR #3218, opened 2024-02-06
Allow apostrophe in email validation

seanmacisaac:
Also add tests for apostrophe for both cases since the regex is different for the last character.

Not sure if the deno files are supposed to be checked in or not (since they are generated) but since they are also include them in this PR.

Fixes at least the most recent comments in https://github.com/colinhacks/zod/issues/2396


colinhacks:
I'm not particularly interested in updating the email regex.  https://github.com/colinhacks/zod/pull/2157 represents my final rule on that question.

Zod implements the "gmail standard" now, which I think is the most useful and expected for the average (Latin-alphabet-using) developer. Everything else can use `.regex()` to customize. In Zod 4 I'm planning to export some common regexes from `zod/regex` to make it easier for users who are looking to validate against a particular RFC standard. 

Apologies, there's no pleasing everyone here and imo this is the best tradeoff.

emilbryggare:
Adding a solution that I used based on @colinhacks comment so other people can solve this if they find this issue. I modified the Zod email regexp from https://github.com/colinhacks/zod/blob/76c8ace1acd9fd713f99d3dca6a3f28320972eac/deno/lib/types.ts#L567 to allow apostrophes. 

```
 email: z
    .string()
    .regex(
      /^(?!\.)(?!.*\.\.)([A-Z0-9_'+-\.]*)[A-Z0-9_'+-]@([A-Z0-9][A-Z0-9\-]*\.)+[A-Z]{2,}$/i,
      'Invalid email',
    ),
```

colinhacks:
Following up on this, you can use `.superRefine` to get a proper `invalid_string` issue instead of a generic `custom` one. cc @emilbryggare 

```ts
const emailRegex =
  /^(?!\.)(?!.*\.\.)([A-Z0-9_'+-\.]*)[A-Z0-9_'+-]@([A-Z0-9][A-Z0-9\-]*\.)+[A-Z]{2,}$/i;

const emailSchema = z.string().superRefine((data, ctx) => {
  if (!emailRegex.test(data)) {
    ctx.addIssue({
      code: z.ZodIssueCode.invalid_string,
      message: "Invalid email address",
      validation: "email",
    });
  }
});
```

PR #3238, opened 2024-02-14
docs: update link and description of conform integration

g1eny0ung:
As I was browsing through the documentation I found that the conform link was out of date (404).


colinhacks:
Thanks!

Issue #3225, opened 2024-02-09
Enum.extract() removes error messages


BrendanC23:
If I declare an enum like

```ts
onst enumErrorMap = {
    errorMap: () => ({ invalid_type_error: "Error Message" }),
};

const TestEnum = z.enum(
    [
        "value1",
        "value2",
    ],
    enumErrorMap,
);
```

The error message works properly. However, if I do `const TestEnumTwo = TestEnum.exclude(["value1"]);`, the error map is lost. It looks like the code just creates a new enum and discards the `RawCreateParams` that were passed to the original enum.

https://github.com/colinhacks/zod/blob/a5a9d31018f9c27000461529c582c50ade2d3937/src/types.ts#L4065-L4076

JacobWeisenburger:
I have confirmed this works as you have stated and I'm pretty sure this should be a bug.

```ts
const testEnum = z.enum( [ 'foo', 'bar' ], {
    errorMap: () => ( { message: 'Error Message' } ),
} )

const result1 = testEnum.safeParse( 'baz' )
!result1.success && console.log( result1.error.issues )
// [
//     {
//         received: "baz",
//         code: "invalid_enum_value",
//         options: [ "foo", "bar" ],
//         path: [],
//         message: "Error Message",
//     }
// ]

const result2 = testEnum.extract( [ 'foo' ] ).safeParse( 'baz' )
!result2.success && console.log( result2.error.issues )
// [
//     {
//         received: "baz",
//         code: "invalid_enum_value",
//         options: [ "foo" ],
//         path: [],
//         message: "Invalid enum value. Expected 'foo', received 'baz'",
//     }
// ]
```

colinhacks:
Fixed in Zod 3.23

PR #3240, opened 2024-02-14
Bugfix: Enum.extract/exclude should not remove error mapping

shaharke:
When calling .extract/exclude we should retain the original RawCreateParams passed to the enum definition. We also allow passing new RawCreateParams if that's necessary. This solves issue #3225.


colinhacks:
Agreed, love htis.

PR #3242, opened 2024-02-15
feat: add safeParseOptional to return an optional result

jonlambert:
`safeParse` is a very useful function to use when validating, however there are often circumstances in which a user might not need access to an error, and would like to use optional chaining on the result. Unfortunately, to do this one would need to narrow the result object based on the success flag.

For example, the following is impossible: 

```ts
const schema = z.object({ name: z.string() });

const getName = (data: unknown) => schema.safeParse(data).data?.name;
                                                        // ^ Property 'data' does not exist on type 'SafeParseReturnType<{ name: string; }, { name: string; }>'.
```

Instead one would need to take a more verbose approach:

```ts
const schema = z.object({ name: z.string() });

const getName = (data: unknown) => {
  const result = schema.safeParse(data);
  return result.success ? result.data.name : undefined;
};
```

Or, cast to any directly [(as is done in the Zod tests)](https://github.com/colinhacks/zod/blob/dc05fd2b17feb706263772b88fa636f0609004f1/src/__tests__/safeparse.test.ts#L16).

I tend to use a tiny util in most projects that runs the above checks. I thought it might be worthwhile opening a very basic PR to add this to Zod in case others might find this useful!

If this approach is agreed, I would be happy to expand documentation to cover this.

(ps. I did have a dig in past PRs to see if anyone has covered this already but couldn't find anything. Apologies if this is duplicated work!)

### DIFF
--- a/deno/lib/__tests__/safeparseoptional.test.ts
+++ b/deno/lib/__tests__/safeparseoptional.test.ts
@@ -1,0 +1,27 @@
+// @ts-ignore TS6133
+import { expect } from "https://deno.land/x/expect@v0.2.6/mod.ts";
+const test = Deno.test;
+
+import * as z from "../index.ts";
+const stringSchema = z.string();
+
+test("safeparseoptional fail", () => {
+  const safe = stringSchema.safeParseOptional(12);
+  expect(safe).toBeUndefined();
+});
+
+test("safeparseoptional pass", () => {
+  const safe = stringSchema.safeParseOptional("12");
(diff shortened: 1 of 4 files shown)


colinhacks:
With https://github.com/colinhacks/zod/pull/3295 merged this is now possible in `zod@beta` 

```ts
const schema = z.object({ name: z.string() });

const getName = (data: unknown) => schema.safeParse(data).data?.name;
```

Landing in Zod 3.23 in the next couple days

PR #3247, opened 2024-02-16
Add types to `ZodFirstPartySchemaTypes`

### LINKED ISSUE #3246: Type definition `ZodFirstPartySchemaTypes` is missing some first party schema types
While trying to create an object which maps from `ZodFirstPartyTypeKind` to a function that takes the associated schema as argument, I found that `ZodReadonly` and `ZodSymbol` are missing from `ZodFirstPartySchemaTypes` union. The basic premise of the mapping is as follows.

```ts
type ZodFirstPartySchemaForType<T extends ZodFirstPartyTypeKind> = ZodFirstPartySchemaTypes extends infer Schema 
  ? Schema extends { _def: { typeName: T} } 
    ? Schema 
    : never
  : never;
type ZodSchemaMap = { [key in ZodFirstPartyTypeKind]: ZodFirstPartySchemaForType<key> };
```

Based on this mapping we could then infer which types are missing:

```ts
type MissingSchemaTypes = { [key in keyof ZodSchemaMap as ZodSchemaMap[key] extends { _def: never } ? key : never ]: true }
```

image

https://github.com/colinhacks/zod/blob/76c8ace1acd9fd713f99d3dca6a3f28320972eac/src/types.ts#L4941-L4976

Could be related to #2663

MatthijsMud:
Fixes #3246

Some first party schema types were missing from the union. This pull-request adds them, along with a test that tries to keep the union and ambient type declaration in sync.

### DIFF
--- a/src/__tests__/firstpartyschematypes.test.ts
+++ b/src/__tests__/firstpartyschematypes.test.ts
@@ -1,0 +1,20 @@
+// @ts-ignore TS6133
+import { test } from "@jest/globals";
+import { util } from "../helpers/util";
+
+import { 
+  ZodFirstPartySchemaTypes,
+  ZodFirstPartyTypeKind,
+} from "..";
+
+test("Identify missing [ZodFirstPartySchemaTypes]", () => {
+  type ZodFirstPartySchemaForType<T extends ZodFirstPartyTypeKind> = ZodFirstPartySchemaTypes extends infer Schema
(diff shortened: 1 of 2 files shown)


TahsinAyman (approved):
Seems like a simple pr

colinhacks:
This has landed in Zod 3.23. 

PR #3250, opened 2024-02-16
add parent data in transform context

adamalmqvist:
Makes it easy to use parent data when using the transform method.


colinhacks:
I understand the impulse here, but Zod schemas are designed to be self-contained. The idea of a schema behaving differently based on its parent schema breaks encapsulation. If the transformation logic is conditional on the parent, then it should be defined on the parent schema.

Issue #3197, opened 2024-01-28
Bug: `z.record( z.any() )` filtering out `undefined` values


JacobWeisenburger:
`z.record( z.any() )` should allow any key and any value, but it seems to be filtering out `undefined` values.

```ts
console.log( z.object( { foo: z.any() } ).parse( { foo: undefined } ) ) // { foo: undefined }
console.log( z.record( z.any() ).parse( { foo: undefined } ) ) // {}, should be { foo: undefined }
```

kaguya3222:
I think this issue should be closed?

PR #3251, opened 2024-02-17
Bugfix: z.record().parse should not filter out undefined values

raik-casimiro:
When calling .parse on z.record() we should not filter out undefined values. Add the optional 'alwaysSet' to the z.record pairs before passing them to mergeObject to maintain undefined values.

This solves issue #3197


colinhacks:
Thanks! Great PR.

colinhacks:
This has landed in Zod 3.23.

Issue #895, opened 2022-01-29
`.pick()` mask allowing non-present props when present props are defined


prescience-data:
### Example
```typescript

// Original schema
const FooSchema = z.object({
  foo: z.string(),
  bar: z.string()
})


const CorrectlyFailingSchema = FooSchema.pick({
  notARealProperty: true // This causes TypeScript error as expected.
})


const IncorrectlyPassingSchema = FooSchema.pick({
  foo: true,
  notARealProperty: true // Once an existing property is added to the mask, any non-defined props can pass as well.
})

```

This became an issue today where after passing through a bundler (esbuild), a non-descriptive error started throwing on `e.t._parse`: 
```
TypeError: Cannot read properties of undefined
```

The issue occurred because I had removed a property `"title"` from the original schema, but not from the `pick()` mask, so `title._parse()` was throwing as `title` was now undefined.

noyan-alimov:
Ideally typescript should cause an error.
I implemented to show a helpful error message.
Example schema:
```ts
const fish = z.object({
  name: z.string(),
  age: z.number(),
  nested: z.object({}),
});
```
And then:
```ts
fish.pick({ name: true, color: true })
```
Doing this should show `color key does not exist in object schema. Available keys: name,age,nested`

prescience-data:
Additional context:

<img width="633" alt="image" src="https://user-images.githubusercontent.com/65471523/151722206-16e1dbee-8ece-43aa-8482-54cfe149de48.png">

JacobWeisenburger:
This seems answered, may I close this?

prescience-data:
> This seems answered, may I close this?

Is there an associated PR? I haven't see anything as yet, so best to keep it alive until merged imo.

JacobWeisenburger:
Ok sounds good. Just doing a little house keeping. It's hard sometimes to tell if an issue is done or not.

PR #3255, opened 2024-02-19
Restrict .pick()/.omit() mask type to only known properties

petrovmiroslav:
### Current

Both the methods .pick and .omit accept properties that don't exist in the schema.
``` 
const schema = z.object({
  existedProperty: z.string(),
});

const pick = schema.pick({
  existedProperty: true,
  NOT_existed_property: true, // there is no TypeScript error
});
```

### Fixed

Both the methods .pick and .omit now only accept known properties.
``` 
const schema = z.object({
  existedProperty: z.string(),
});

const pick = schema.pick({
  existedProperty: true,
  NOT_existed_property: true, // there is a TypeScript error
});
```

resolves https://github.com/colinhacks/zod/issues/895


colinhacks:
Thank you! Fascinating that this works...

colinhacks:
This has landed in Zod 3.23.

Issue #2725, opened 2023-09-11
Provide string-specific validation for iso8601 periods (durations) 


andoks:
Similar to the string-specific validation for [datetime](https://zod.dev/?id=iso-datetimes), maybe consider adding one for [iso8601 periods](https://en.wikipedia.org/wiki/ISO_8601#Durations), possibly called `.duration()` or `.period()`

Some references that might be relevant for a discussion:
* https://tc39.es/proposal-temporal/docs/duration.html
* https://momentjs.com/docs/#/durations/as-iso-string/

PR #3265, opened 2024-02-23
feat: Add support for ISO-8601 Durations

mastermatt:
https://en.wikipedia.org/wiki/ISO_8601#Durations
As an extension of the ISO standard, the format is also used in RFC 3339, XML Schema Part 2, TC39's Temporal proposal, and a format for JSON Schema strings since draft 2019-09.

~~Does not allow for negative values.~~

I believe this would close https://github.com/colinhacks/zod/issues/2725, however, it also referenced "Periods" which is how RFC 3339 (confusingly) refers to ISO 8601 Time Intervals and are not covered by this change. So I'll let @andoks decide if this covers their request fully.


deno/lib/types.ts
```diff
@@ -552,6 +553,9 @@ const ulidRegex = /^[0-9A-HJKMNP-TV-Z]{26}$/;
 //   /^([a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[a-f0-9]{4}-[a-f0-9]{12}|00000000-0000-0000-0000-000000000000)$/i;
 const uuidRegex =
   /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/i;
+const durationRegex =
+  /^P(?!$)(\d+Y)?(\d+M)?(\d+W)?(\d+D)?(T(?=\d)(\d+H)?(\d+M)?(\d+S)?)?$/;
```
andoks, line 557: This does not handle any fractional values, which is particularly useful to be able to use for shorter intervals, revolving around minutes, seconds and fractional seconds.

According to [wikipedia](https://en.wikipedia.org/wiki/ISO_8601#Durations)
> The smallest value used may also have a decimal fraction,[[39]](https://en.wikipedia.org/wiki/ISO_8601#cite_note-39) as in "P0.5Y" to indicate half a year. This decimal fraction may be specified with either a [comma](https://en.wikipedia.org/wiki/Comma) or a [full stop](https://en.wikipedia.org/wiki/Full_stop), as in "P0,5Y" or "P0.5Y"

But I would be happy with only allowing the seconds part to be fractional, possibly only by with a dot (although it seems like the comma is allowed according to the standard), as this is what I have seen for the most part. This would hopefully not allow anything that is not already allowed by the standard (and of course not pass everything that the standard allows for) - but will cover many useful use-cases without too much code changes

deno/lib/__tests__/string.test.ts
```diff
@@ -470,6 +470,19 @@ test("datetime parsing", () => {
   ).toThrow();
 });
 
+test("duration", () => {
+  const duration = z.string().duration();
+  expect(duration.isDuration).toEqual(true);
+
+  duration.parse("P3Y6M4DT12H30M5S");
```
andoks, line 477: Maybe add test for some more cases? Seems like the current regex does not handle various time-only periods [examples](regexr.com/7sgkp)

mastermatt:
Thanks for the feedback @andoks. I just pulled the regex used by Joi, and that was short-sighted. 
With ISO 8601-2 on the books, it makes sense to add the full functionality.
The SO link you provided was a great resource. I've updated the regex and added more test cases. It now supports positive/negative prefixes for the whole input or individual components, and decimal values on the least significant component (allows full stop and commas per the spec).

andoks:
As an additional thought:  can this end up conflicting with a possible future feature of supporting [js temporal](https://tc39.es/proposal-temporal/docs/) in the future in some way do you think?

mastermatt:
@andoks my understanding of the Temporal proposal for durations is that its string representations are based on ISO 8601 itself, not a variation e.i. the XML schema spec. Therefore if there is ever any conflict it would point to a bug in this regex.

Issue #3273, opened 2024-02-27
String length validation error messages can't distinguish between singular and plural count.


RockingThor:
<img width="742" alt="Screenshot 2024-02-27 at 9 20 51 AM" src="https://github.com/colinhacks/zod/assets/121621183/3978c27a-8bad-480f-ad7e-3fdc30f58d21">

It always uses the "(s)" universally either it is 1 or 0 or >1. 

For example, if the zod schema is like this:

```
const SignUpFormSchema = z.object({
    alphabet: z.string().min(1)
});
```

If the user tries to submit the form without inputting anything the error message will be shown as:

`String must contain at least 1 character(s)`

The error message can be improved to:

`String must contain at least 1 character`

colinhacks:
On further reflection, I'm not convinced the inconvenience of such a change for users is worth the marginal improvement to clarity here. Given Zod's current user base, many tens of thousands of tests would have to be very slightly updated to account for this change. I'm not convinced it's worthwhile. 

Zod 4 will also introduce a much easier method of customizing error messages, so hopefully fewer people will rely on the default messages too.

PR #3274, opened 2024-02-27
Sigular and plural grammatical fix 

RockingThor:
Fixed the error message on the basis of the number. Also configured the tests accordingly. #3273


deno/lib/locales/en.ts
```diff
@@ -70,11 +70,11 @@ const errorMap: ZodErrorMap = (issue, _ctx) => {
       if (issue.type === "array")
         message = `Array must contain ${
           issue.exact ? "exactly" : issue.inclusive ? `at least` : `more than`
-        } ${issue.minimum} element(s)`;
+        } ${issue.minimum} ${issue.minimum <= 1 ? "element" : "elements"}`;
```
m10rten, line 73: this is overkill, maybe do this instead:

```ts
message = `Array must contain ${
          issue.exact ? "exactly" : issue.inclusive ? `at least` : `more than`
        } ${issue.minimum} element${issue.minimum > 1 ? "s" : ""}`;
```

RockingThor: I have made the changes. Please have a look and let me know if any further changes are required.

m10rten: Looks better, it's up to maintainers to approve and merge this PR.

colinhacks:
I appreciate this, though unfortunately a lot of users' tests are written to match Zod's exact error messages, so this would be considered a breaking change. I could merge this in Zod 4, but given Zod's current level of usership, I don't think the added error clarity is worth forcing users to refactor their test suites.

PR #3275, opened 2024-02-27
Fix #3273

### LINKED ISSUE #3273: String length validation error messages can't distinguish between singular and plural count.

sgrsngh2003:
String length validation error messages can't distinguish between singular and plural count.
Fix #3273

### DIFF
--- a/deno/lib/locales/en.ts
+++ b/deno/lib/locales/en.ts
@@ -71,10 +71,18 @@ const errorMap: ZodErrorMap = (issue, _ctx) => {
         message = `Array must contain ${
           issue.exact ? "exactly" : issue.inclusive ? `at least` : `more than`
         } ${issue.minimum} element(s)`;
-      else if (issue.type === "string")
-        message = `String must contain ${
+      else if (issue.type === "string"){
+        if (issue.minimum<=1){
+          message = `String must contain ${
+            issue.exact ? "exactly" : issue.inclusive ? `at least` : `over`
+          } ${issue.minimum} character`
+        }
+        else{
+          message = `String must contain ${
           issue.exact ? "exactly" : issue.inclusive ? `at least` : `over`
         } ${issue.minimum} character(s)`;
+        }
+      }
       else if (issue.type === "number")
         message = `Number must be ${
           issue.exact
--- a/deno/lib/locales/en.ts
+++ b/deno/lib/locales/en.ts
@@ -74,7 +74,7 @@ const errorMap: ZodErrorMap = (issue, _ctx) => {
       else if (issue.type === "string"){
         message = `String must contain ${
           issue.exact ? "exactly" : issue.inclusive ? `at least` : `over`
-        } ${issue.minimum} character${issue.minimum} > 1 ? "(s)" : ""`;
+          } ${issue.minimum} character${issue.minimum>1 ? "(s)" : ""}`;  
       }
       else if (issue.type === "number")
         message = `Number must be ${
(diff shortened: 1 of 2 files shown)


src/locales/en.ts
```diff
@@ -71,10 +71,18 @@ const errorMap: ZodErrorMap = (issue, _ctx) => {
         message = `Array must contain ${
           issue.exact ? "exactly" : issue.inclusive ? `at least` : `more than`
         } ${issue.minimum} element(s)`;
-      else if (issue.type === "string")
-        message = `String must contain ${
+      else if (issue.type === "string"){
+        if (issue.minimum<=1){
```
m10rten, line 75: Add spaces between operators, variables and values.

`issue.minimum <= 1`

src/locales/en.ts
```diff
@@ -71,10 +71,18 @@ const errorMap: ZodErrorMap = (issue, _ctx) => {
...
+          message = `String must contain ${
+            issue.exact ? "exactly" : issue.inclusive ? `at least` : `over`
+          } ${issue.minimum} character`
```
m10rten, line 78: You could just do:

```ts
message = `String must contain ${
            issue.exact ? "exactly" : issue.inclusive ? `at least` : `over`
          } ${issue.minimum} character${issue.minimum > 1 ? "(s)" : ""`;
```

m10rten (changes requested):
Simple change does not need so much duplication

src/locales/en.ts
sgrsngh2003: thanks for the explanation.

sgrsngh2003: should i do the changes or its done.

m10rten: Make the change, duplication is never nice.

sgrsngh2003: so i just overwrite my code with yours??

colinhacks:
Thanks for the PR @sgrsngh2003, it's very well done. 

Unfortunately, a lot of Zod users have written tests that match against the exact error messages produced by Zod. While technically this could be merged as part of Zod 4, I'm not convinced the inconvenience of such a change for users is worth the marginal improvement to clarity here.

deno/lib/locales/en.ts
```diff
@@ -71,10 +71,11 @@ const errorMap: ZodErrorMap = (issue, _ctx) => {
         message = `Array must contain ${
           issue.exact ? "exactly" : issue.inclusive ? `at least` : `more than`
         } ${issue.minimum} element(s)`;
-      else if (issue.type === "string")
+      else if (issue.type === "string"){
         message = `String must contain ${
           issue.exact ? "exactly" : issue.inclusive ? `at least` : `over`
-        } ${issue.minimum} character(s)`;
+          } ${issue.minimum} character${issue.minimum>1 ? "(s)" : ""}`;
```
santosmarco, line 77: Why are you keeping the `(` and `)` on `(s)` if now you know it must be plural?

Maybe:
```suggestion
          } ${issue.minimum} character${issue.minimum>1 ? "s" : ""}`;  
```
?

Also, in English, `0` quantities should also be pluralized: we say `0 apples`, not `0 apple`.

Issue #3043, opened 2023-12-11
Inferred type of Object catchall with boolean is empty object 


valeneiko:
```ts
const schema = z.object({}).catchall(z.boolean());
type T = z.output<typeof schema>;
//   ^? {}
```

I am expecting `type T = Record<string, boolean>`, but inferred type is empty object instead.

JacobWeisenburger:
Are you only wanting to report the bug? or are you looking for a temp workaround?

valeneiko:
Was just reporting a bug.

I am currently using this workaround:
```ts
z.object({}).catchall(z.boolean()) as z.ZodObject<
  z.ZodRawShape,
  'strip',
  z.ZodBoolean,
  Record<string, boolean>
>
```

colinhacks:
This is intentional. Unfortunately the runtime behavior of `.catchall()` does not agree with TypeScript's interpretation of index signatures. For instance you can't declare this type:

```ts
type A = { 
  name: string;
  [k:string]: number
};
```

The `[k: string]` (or the `Record<>` type you proposed) applies to all elements of the object type, including named properties. It isn't a "catchall" in the way Zod treats it. This is why `.catchall()` doesn't change the inferred type. There's no syntax in TypeScript to represent a "catchall index signature".

colinhacks:
@valeneiko Consider `z.record` for your specific use case.

valeneiko:
@colinhacks wouldn't intersecting an index signature with an object capture the semantics of `.catchall()`? 

```tsx
type X = {
    x: number
} & {
    [key: string]: boolean
};

type Q1 = X['x'];
//   ^? type Q1 = number

type Q2 = X['y'];
//   ^? type Q2 = boolean
```

[Playground Link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAGlC8UDeAoK6oA8BcUB2ArgLYBGEATigL5QBkyaGA2gNYQi4DOw5AlngHMAurhIB7MQBsIAQzzUA3ChShIUAIoBGBLCYByTHqFKA9CYwA9APzLV0dQCYdMfSCOnz6a0A)

colinhacks:
> Nvm, just realized the type is not actually usable, just displays like that.

Indeed, this is exactly what threw me off in the first place 😅

colinhacks:
Apologies, this actually is a bug. See my reply here: https://github.com/colinhacks/zod/pull/3282

colinhacks:
[...]

PR #3282, opened 2024-02-28
Fix z.catchall() return type

kyvg:
Fixed an issue where `CatchallOutput` returned unknown for classes without additional methods, such as `ZodBoolean`, `ZodUndefined`, and `ZodNull`. This was resolved by replacing `ZodTypeAny` with `ZodFirstPartySchemaTypes`. This change addresses the behavior described in issue #3043

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -2185,13 +2185,11 @@ export type baseObjectInputType<Shape extends ZodRawShape> =
     [k in keyof Shape]: Shape[k]["_input"];
   }>;
 
-export type CatchallOutput<T extends ZodTypeAny> = ZodTypeAny extends T
-  ? unknown
-  : { [k: string]: T["_output"] };
+export type CatchallOutput<T extends ZodTypeAny> =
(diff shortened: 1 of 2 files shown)


colinhacks:
Fixed in https://github.com/colinhacks/zod/commit/4a618ef798231cdf152e55e506e7f21096d27468, thanks 👍

colinhacks:
Good catch, and a clever fix. Ultimately I want to avoid relying on ZodFirstPartySchemaTypes so that Zod is more compatible with user-defined `ZodType` subclasses. The right approach here would be to change `ZodTypeAny`, though that would potentially break user code.

```ts-diff
- type ZodTypeAny = ZodType<any, any, any>
+ type ZodTypeAny = ZodType<any, ZodTypeDef, any>
```

That said, we *can* refactor Zod's own code to stop using `ZodTypeAny`, which fixes this issue. I've done that here: https://github.com/colinhacks/zod/pull/3347

colinhacks:
Fixed in Zod 3.23

PR #3285, opened 2024-02-29
Streamline name casing in README

kristiandupont:
As per [this suggestion](https://github.com/colinhacks/zod/discussions/929#discussioncomment-2160278), I did a quick scan through the README and streamlined all name casing: schemas are camelCased, types are PascalCased.

### DIFF
--- a/README.md
+++ b/README.md
@@ -683,14 +683,14 @@ Creating an object schema
 ```ts
 import { z } from "zod";
 
-const User = z.object({
+const user = z.object({
   username: z.string(),
 });
 
-User.parse({ username: "Ludwig" });
+user.parse({ username: "Ludwig" });
 
 // extract the inferred type
-type User = z.infer<typeof User>;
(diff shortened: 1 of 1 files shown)


colinhacks:
In a lot of cases I'm partial to Pascal-casing, since schemas ultimately represent types. This also enables a pattern that I personally love where a schema and it's inferred type can share a name. They are the runtime & static representation of the same type, so this feels right to me.

```ts
const Dog = z.object({
  name: z.string(),
  age: z.number(),
});

type Dog = z.infer<typeof Dog>
```

But for utility schemas with a self-evident static type I wouldn't capitalize.

```ts
const nonemptyString = z.string().min(1)
```

That's what I do and what feels natural to me within TypeScript norms. 

I appreciate the PR, but a lot of the changes in this PR feel weird to me so I'm not going to merge. Ultimately people can do what they want and I don't think there's much value in trying to establish a dogma here.

PR #2824, opened 2023-09-29
[2609] fix ReDoS vulnerability in email regex

MacsDickinson:
Fixes #2609

The current regex used for email validation contains "catastrophic backtracking", specifically `([A-Z0-9_+-]+\.?)*`. This gets evaluated inefficiently by JS, resulting in an exponential increase in execution time for *failed* matches.

This can be replicated easily - here's execution time against `^([A-Z0-9_+-]+\.?)*[A-Z0-9_+-]@([A-Z0-9][A-Z0-9\-]*\.)+[A-Z]{2,}$`

| string | match | execution
| ----- | ------ | ----------
| aaaaaaaaaaaaaaaaaaaaaaa@test.c | false | 0.615s  |
| aaaaaaaaaaaaaaaaaaaaaaaa@test.c | false | 1.165s  |
| aaaaaaaaaaaaaaaaaaaaaaaaa@test.c | false | 2.280s  |
| aaaaaaaaaaaaaaaaaaaaaaaaaa@test.c | false | 4.507s  |
| aaaaaaaaaaaaaaaaaaaaaaaaaa@test.c | false | 8.964s  |
| aaaaaaaaaaaaaaaaaaaaaaaaaaa@test.c | false | 18.018s  |
| aaaaaaaaaaaaaaaaaaaaaaaaaaaa@test.c | false | 💀  |

This PR provides an alternative regex. Rather that matching on the `(aaaaa.)+` we instead do a negative lookahead for the presence of `..` or the email starting with a `.`. This approach isn't susceptible to ReDoS

here's execution time against `^(?!\.)(?!.*\.\.)([A-Z0-9_+-\.]*)[A-Z0-9_+-]@([A-Z0-9][A-Z0-9\-]*\.)+[A-Z]{2,}$`

| regex | email | match | execution
| ------ | ----- | ------ | ----------
| aaaaaaaaaaaaaaaaaaaaaaa@test.c | false | 0.044s  |
| aaaaaaaaaaaaaaaaaaaaaaaa@test.c | false | 0.046s  |
| aaaaaaaaaaaaaaaaaaaaaaaaa@test.c | false | 0.044s  |
| aaaaaaaaaaaaaaaaaaaaaaaaaa@test.c | false | 0.046s  |
| aaaaaaaaaaaaaaaaaaaaaaaaaaa@test.c | false | 0.045s  |
| aaaaaaaaaaaaaaaaaaaaaaaaaaaa@test.c | false | 0.045s  |
[...]

PR #3286, opened 2024-02-29
Bugfix: Commas are incorrectly allowed in email regex.

mokemoko:
In #2824, email regex changed, but it seems inadvertently allowed `,` which were previously not permitted in email. 

#### Email regex before #2824
`/^([A-Z0-9_+-]+\.?)*[A-Z0-9_+-]@([A-Z0-9][A-Z0-9\-]*\.)+[A-Z]{2,}$/i;`

#### Email regex after #2824
`/^(?!\.)(?!.*\.\.)([A-Z0-9_+-\.]*)[A-Z0-9_+-]@([A-Z0-9][A-Z0-9\-]*\.)+[A-Z]{2,}$/i;`

For the latter part `[A-Z0-9_+-\.]`, it probably wanted to add only `.` to `[A-Z0-9_+-]`.
But, the addition of `.` after the unescaped hyphen has caused it to be interpreted as a range from `+` to `.`.
Therefore, `,` contained between `+` and `.` is now permitted.

So, fix this issue with escaping hyphens simply.


colinhacks:
Thank you! Fantastic PR, apologies for the delay on such a no-brainer PR. That accidental range issue could have been a lot worse lol.

colinhacks:
This has landed in Zod 3.23.

Issue #3266, opened 2024-02-24
Different typing for SafeParseReturnType<T>


scarletquasar:
## Behavior description

Nowadays, the **safeParse** and **safeParseAsync** - functions for parsing Zod schemas - return types are based on **SafeParseReturnType<T>**, this type is an union containing the following types:

- **SafeParseSuccess\<Output\>**

  ```ts
  export declare type SafeParseSuccessOutput<Output> = {
    success: true;
    data: output;
  }
  ```

- **SafeParseError\<Input\>**

  ```ts
  export declare type SafeParseError<Input> = {
    success: false;
    error: ZodError<Input>;
  }
  ```

## Behavior problem

This kind of behavior, besids applying bad practices by design - here I'm talking about having const values true and false for success instead of a boolean - also needs an explicit declaration on what type it should be. This situation causes more complexity in the user code - complexity that should be handled by the library itself and it's against the TypeScript idea of good inference and usability.

![307189632-6373aa6c-aa30-47f2-b511-56b028f866c8](https://github.com/colinhacks/zod/assets/70824102/19410396-b002-4ea5-b398-787810eb5960)

## Solution suggestion

My solution, for that case, would be implementing a single typing for **SafeParseReturnType<T>** without unions - in a way that will not break existing code and will correctly adapt to any situation that the previously implemented unions covered. 

Code:

```ts
export declare type SafeParseResult<InputOrOutput> = {
  success: boolean,
  data?: InputOrOutput,
  error?: ZodError<InputOrOutput>
}
```

jaens:
As far as I see, the suggestion is exactly the opposite of TypeScript good practices - it incorrectly describes the return type and does not enable the standard practice of type narrowing based on eg `if (result.success) { use(result.data); }`

You are just using TypeScript wrong.

The correct expression to use in your screenshot is `error: item.success === false ? item.error : undefined`. Otherwise the code really is trying to access an undefined property (`error`  actually does **really** not exist in that case) for which TypeScript correctly raises an error.
If you object to that, I suggest not using TypeScript with the default strictness options.  Yes, it is ergonomically slightly unwieldy. Yes, every application or library WILL have the same problem and will need to use the same workaround. This is just standard TypeScript,

schicks:
I think there's a middle path here, where we keep the safety of the current type (which is the main reason I use zod) but improve the ergonomics so that it can be used more flexibly.

```typescript
export declare type SafeParseSuccessOutput<Output> = { 
    success: true; 
[...]

PR #3295, opened 2024-03-02
add never props to safe parse return types

schicks:
Possible improvement to safe parse return types as discussed in #3266.

The `?: never` properties added to the type involve no runtime overhead and support all existing patterns. However, as seen in the tests they avoid casting or intermediate variables when we attempt to access properties on the union `SafeParseReturnType`, since now both branches have the property we attempt to access. When accessing `SafeParseReturnType['data']`, we get `Output | undefined`, which is accurate to the runtime behavior. If we are in a context where we've already checked `success`, we get just `Output` as expected.

### DIFF
--- a/deno/lib/__tests__/safeparse.test.ts
+++ b/deno/lib/__tests__/safeparse.test.ts
@@ -8,13 +8,13 @@ const stringSchema = z.string();
 test("safeparse fail", () => {
   const safe = stringSchema.safeParse(12);
   expect(safe.success).toEqual(false);
-  expect((safe as any).error).toBeInstanceOf(z.ZodError);
+  expect(safe.error).toBeInstanceOf(z.ZodError);
 });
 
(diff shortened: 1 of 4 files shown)


scarletquasar (approved):

scarletquasar:
> @colinhacks does this seem like something we could add? Happy to make additional changes if necessary, but the basic functionality comes up fairly frequently when using zod in an environment where there needs to be a "default" flow (eg, react components handling network responses).

@JacobWeisenburger

JacobWeisenburger:
Seems interesting, but I am not in a position to make any decisions about these types of things. @colinhacks is the owner and would be the person to approve or not.

colinhacks:
Thanks @schicks, this is an excellent idea.

colinhacks:
This has landed in Zod 3.23.

PR #3301, opened 2024-03-05
fix: make `input` of `.required()` readonly

KATT:
More info: https://github.com/trpc/trpc/issues/5538

Unsure if this is a legit fix. I could see this working as designed.

cc @tkdodo @PeterMK85


colinhacks:
This has landed in Zod 3.23.

Issue #2724, opened 2023-09-11
wrap internal errors from `preprocess` and `refine`


Malkiz:
As [explained here](https://github.com/colinhacks/zod/issues/2401) zod intentionally does not catch errors from within `preprocess` and `refine`.

> This is behaving as intended currently. An error in a refinement is considered an application-level bug, not a validation error. The thinking is that catching this error in Zod would make it harder to debug actual real-life bugs, and you can always use try/catch. But if Zod catches these errors, there's no way to opt-out of this and allow errors to bubble up.

However, consider this schema:
```typescript
z.object({ a: z.array(z.object({ b: z.preprocess(data => { throw new Error('asdf') }, z.date()) }))}).parse({a: [{}]})
```
The error will be:
```
Uncaught Error: asdf
    at Object.transform (REPL12:1:66)
    at ZodEffects._parse (/node_modules/zod/lib/types.js:2137:38)
    at ZodObject._parse (/node_modules/zod/lib/types.js:1070:37)
    at ZodObject._parseSync (/node_modules/zod/lib/types.js:109:29)
```

This makes it difficult to know where the error came from.
Consider an object schema with many properties (that may also be nested) - if I get this error I don't know which property caused it.

In contrast, the regular zod parsing errors are amazing:
```typescript
z.object({ a: z.array(z.object({ b: z.date()})) }).parse({a: [{}]})
```
Error:
```json
[
  {
    "code": "invalid_type",
    "expected": "date",
    "received": "undefined",
    "path": [
      "a",
      0,
      "b"
    ],
    "message": "Required"
  }
]
```
It tells me exactly the path to the problem in my object.

### Proposed solution
I propose that zod will catch my internal error and wrap it in a zod error (and re-throw it) that will at least tell me where the error occurred in the object.

For example something like:
```typescript
z.object({ a: z.array(z.object({ b: z.preprocess(data => { throw new Error('asdf') }, z.date()) }))}).parse({a: [{}]})
```
Expected error:
```json
[
  {
    "code": "preprocess_error",
    "path": [
      "a",
      0,
      "b"
    ],
    "message": "Uncaught Error: asdf",
    "originalError": <the original error object>
  }
]
```

zackarydev:
But allow the behaviour of the "rethrow" to be customizable. I.e. personally I wouldn't want to use the ZodError class at all, and to use my own errors.

shaharke:
[...]

PR #3314, opened 2024-03-09
Wrap preprocess with try-catch to provide better information on error

shaharke:
Before this fix, any errors during preprocess would result in an error that doesn't provide a lot of debug information. For example, the path where the error happened was not provided by Zod.
This change will make sure that even errors during preprocess are translated to ZodError with useful metadata about the error.

Solves issue #2724

### DIFF
--- a/deno/lib/ZodError.ts
+++ b/deno/lib/ZodError.ts
@@ -18,6 +18,7 @@ export type typeToFlattenedError<T, U = string> = {
 export const ZodIssueCode = util.arrayToEnum([
   "invalid_type",
   "invalid_literal",
+  "preprocess_error",
   "custom",
   "invalid_union",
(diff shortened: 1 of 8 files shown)


colinhacks:
Zod's currently policy is to let errors in refinements, transforms, and preprocess propagate up the call stack. The goal of ZodError is to expose validation errors. In my opinion, an uncaught error inside `.preprocess()` is a bug and doesn't indicate a schema validation issue. In this scenario, I think it's actually desirable for these errors to propagate normally, for debugging reasons. Ultimately it's up to the user to catch and handle errors inside these callbacks. 


I'm open to hearing counter-arguments. If this was going to change, the behavior would have to be consistent among refine/transform/preprocess.

PR #3346, opened 2024-03-21
Support more elements in DU

colinhacks:
Supports the following as discriminator keys:

- ZodOptional
- ZodNullable
- ZodReadonly
- ZodBranded
- ZodCatch

Also fixes a bug with `z.nativeEnum` as a discriminator key.


colinhacks:
This has landed in Zod 3.23.

PR #3347, opened 2024-03-21
Fix passthrough and catchall inferred types. Remove internal usage of ZodTypeAny

colinhacks:


colinhacks:
This would be too breaky to merge in a minor release, so I'm going to rework these changes in the v4 branch.

PR #3380, opened 2024-04-04
feat: strict branding with opt-out

KATT:
when branding, it's useful to retain the branding on `_input` so you have to explicitly pass a branded value when using the schema

this reverts #1492 but with the option of opting out of branded behavior

I can make it a non-breaking change if that's preferred, so it default to non-strict mode

cc @Xetera

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -466,8 +466,16 @@ export abstract class ZodType<
     }) as any;
   }
 
-  brand<B extends string | number | symbol>(brand?: B): ZodBranded<this, B>;
-  brand<B extends string | number | symbol>(): ZodBranded<this, B> {
+  brand<B extends string | number | symbol, S extends boolean = true>(
+    brand?: B,
+    opts?: {
+      strict: S;
+    }
+  ): ZodBranded<this, B, S>;
+  brand<
+    B extends string | number | symbol,
+    S extends boolean = true
+  >(): ZodBranded<this, B, S> {
     return new ZodBranded({
       typeName: ZodFirstPartyTypeKind.ZodBranded,
       type: this,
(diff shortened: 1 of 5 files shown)


deno/lib/types.ts
```diff
@@ -4726,8 +4735,13 @@ export type BRAND<T extends string | number | symbol> = {
 
 export class ZodBranded<
   T extends ZodTypeAny,
-  B extends string | number | symbol
-> extends ZodType<T["_output"] & BRAND<B>, ZodBrandedDef<T>, T["_input"]> {
+  B extends string | number | symbol,
+  S extends boolean
```
KATT, line 4739: this could be 

```suggestion
  S extends boolean = false
```

if we don't want breaking changes

deno/lib/types.ts
```diff
@@ -466,8 +466,16 @@ export abstract class ZodType<
     }) as any;
   }
 
-  brand<B extends string | number | symbol>(brand?: B): ZodBranded<this, B>;
-  brand<B extends string | number | symbol>(): ZodBranded<this, B> {
+  brand<B extends string | number | symbol, S extends boolean = true>(
```
KATT, line 469: non-breaking:
```suggestion
  brand<B extends string | number | symbol, S extends boolean = false>(
```

deno/lib/types.ts
```diff
@@ -466,8 +466,16 @@ export abstract class ZodType<
...
+    brand?: B,
+    opts?: {
+      strict: S;
+    }
+  ): ZodBranded<this, B, S>;
+  brand<
+    B extends string | number | symbol,
+    S extends boolean = true
```
KATT, line 477: non-breaking
```suggestion
    S extends boolean = false
```

PR #3391, opened 2024-04-09
make `ZodString.date()`, `ZodString.time()`, and `ZodString.datetime()` only accept valid date and time

szamanr:
#1766 added `.date()` function but it uses a simple regex `\d{4}-\d{2}-\d{2}`, which allows many invalid dates. similar behaviour happens for `.time()` and `.datetime()`.

if we must use regex for date validation, then i propose a change that:
- rejects invalid dates such as `2024-01-32` or `2024-13-01`
- correctly handles 30th and 31st day based on the month
- is still fairly readable, as far as regexes go

one caveat is that it always accepts 29 feb, but i think that's acceptable. it's an improvement over the existing code, where non-valid dates are accepted.

you can check some test cases here: https://regexr.com/7ulgo

todo: 
- [x] add test cases
- [x] also validate time
- [x] add test cases for time


szamanr:
cc @igalklebanov @colinhacks since this hasn't been released yet, maybe we can do a quick patch before the release.

colinhacks:
I think this is a good idea, but it does introduce some inconsistency with `.datetime()`, which still doesn't enforce this. We could update the datetime logic to do the same, and I'd be okay releasing that in a minor version. Arguably it's a bugfix.

Also, your PR seems to have only updated the code in `deno/lib`? Just update `src` and `deno/lib` should get regenerated when you push.

igalklebanov:
Hey 👋

Good idea!

As @colinhacks said, all datelike stuff should be aligned.

> This extra leap day occurs in each year that is a multiple of 4, except for years evenly divisible by 100 but not by 400
https://en.wikipedia.org/wiki/Leap_year

Could check `year % 4 === 0 && (year % 100 !== 0 || year % 400 === 0)` for leap years.

szamanr:
`.datetime()` and `.date()` are using the same regex under the hood.

i added a regex for valid time, moved the code to src dir, and updated the PR title to reflect all changes.

colinhacks:
Rebased onto the changes I made in #2522

colinhacks:
Thanks @szamanr 🙌

szamanr:
> Why does it need to be a regex? Couldn't you check it easier with a Date instance or are there some problems in different runtimes?

@Shumuu how would you achieve that? i've seen some implementations using a function that checked if `new Date(dateString).getTime()` is a number, but that doesn't work properly. for example:
```
new Date("2000-02-31").getTime() // should be invalid
> 951955200000

new Date("2000-04-31").getTime() // should be invalid
> 957139200000
```

let's say we used that - in that case, we would still have to use a regex first, to rule out all sorts of strings that `Date()` constructor accepts, e.g. "123".

either way, i tend to not put trust in native Date, given that it has many issues. the best way to solve this would be to outsource the logic to a 3rd party library. e.g. date-fns has a function that would be perfect for this: `isMatch(dateString, "yyyy-MM-dd")`. but i noticed zod doesn't ship with any dependencies, so i assumed this is not accepted here.

Shumuu:
The way I solve it is creating a Date like this

`new Date("2023-02-29")` and then I check on the date instance if the day, the month and the year are equal to what I passed in. Creating an invalid Date like this will result in 2023-03-01.

colinhacks:
There's a lot of other stuff going on here. Users can specify whether to allow timezone offsets, unqualified dates, sub-second precision, etc. So we're going to be doing regex validation regardless. We *could* do all that stuff *then* try to instantiate a `Date` but I'm guessing that would be dramatically slower (I'll check)

colinhacks:
This has landed in Zod 3.23.

Issue #3383, opened 2024-04-05
z.union allow ZodType array spread as argument


FlynnHillier:
I am attempting to use a dynamically generated array of strings to define a `z.union()` that validates that a string is included within said dynamic array.

To do this I am attempting to map each element my dynamic array to a `z.literal()`, and then spreading this mapped array of literals to the union type.

```typescript
import { z } from "zod"
const TILES = ["a1","a2","a3","b1","b2","b3"]

// ERROR:
// Argument of type 'ZodLiteral<string>[]' is not assignable to parameter of type 'readonly [ZodTypeAny, ZodTypeAny, ...ZodTypeAny[]]'.
// Source provides no match for required element at position 0 in target.ts(2345)
const myUnion = z.union([
    ...TILES.map(tile => z.literal(tile))
])
```

This implementation results in typescript kicking up a fuss, as it is not sure that union has been provided any values. This is because zod's `union` type is hardcoded to have *atleast 2* ZodTypes provided to it. 
```typescript
declare const unionType: <T extends readonly [ZodTypeAny, ZodTypeAny, ...ZodTypeAny[]]>(types: T, params?: RawCreateParams) => ZodUnion<T>;
```

This means to prevent typescript from complaining, *2* literals must be hardcoded before we spread our array of literals into the union.
```typescript
const myUnion = z.union([
    z.literal("a1"),
    z.literal("a2"),
    ...TILES.map(tile => z.literal(tile))
])
```
This is evidently not ideal as we may not have two values which we always wish to be included in our union type.

Is there any way `z.union()` could be re-typed / changed in some way to better support spreading of ZodType arrays?

extradosages:
This is a good idea; one-element and even zero-element unions have reasonable conventional semantics, and can probably be easily implemented in the type system with conditionals!

colinhacks:
My goal is for `ZodUnion` to have a strongly typed tuple containing the element schemas. The current type signature is designed to prevent things like your `TILES.map()` call, because then you can't do something like this:

```ts
myUnion.options[2]; 
// should be ZodLiteral<"a3">
```

I'm looking into loosening this restriction in the next major version. This is a recurring issue for people, and it doesn't impact the inferred type so it's probably not worth the trouble.

abdirahmn1:
any updates to this?, thanks :+1:

PR #3398, opened 2024-04-14
fix(union): allow dynamic arrays of literals in types

jeferson-sb:
Addresses #3383 

## Description

Modifies the types for allowing creating z.union() for dynamic arrays spread by z.literal, I essentially removed the first two readonly ZodTypes in the array.

Note: I also tried creating a conditional return using `T extends [infer First, ...infer Rest] ? ZodUnion<T> : ZodUnion<[]>` as I first atttempt but I could not find success, if there is a better option by adding a conditional type I accept suggestions.

## New behavior

```ts
const cells = ["a1", "a2", "a3", "b1", "b2", "b3"] as const;
z.union([...cells.map((cell) => z.literal(cell))]).parse("a1");
```

### DIFF
--- a/deno/lib/__tests__/unions.test.ts
+++ b/deno/lib/__tests__/unions.test.ts
@@ -63,3 +63,16 @@ test("readonly union", async () => {
   union.parse("asdf");
   union.parse(12);
 });
+
+test("return valid over spread array of literals", () => {
+  const cells = ["a1", "a2", "a3", "b1", "b2", "b3"] as const;
+  const union = z.union([...cells.map((cell) => z.literal(cell))]);
+  union.parse("a1");
+});
+
+test("fails parse over spread array of literals for non-existent item", () => {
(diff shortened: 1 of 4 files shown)


colinhacks:
Thanks, but ZodUnion needs to track its elements as a strongly-typed tuple. Using your example:

```ts
const cells = ["a1", "a2", "a3", "b1", "b2", "b3"] as const;
const union = z.union([...cells.map((cell) => z.literal(cell))]);

union.options; 
// with your PR:
// z.ZodLiteral<"a1" | "a2" | "a3" | "b1" | "b2" | "b3">[]

// this should be
// [ ZodLiteral<"a1">, ZodLiteral<"a2">, ZodLiteral<"a3">, ZodLiteral<"b1">, ZodLiteral<"b2">, ZodLiteral<"b3"> ];
```

It's important that Zod schemas be strongly typed all the way down. Allowing arrays into ZodUnion makes it possible for type safety to lapse, as in the code example.

PR #3413, opened 2024-04-18
Faster ipv4 parsing

colinhacks:
2x performance improvement

out


colinhacks:
This has landed in Zod 3.23.

Issue #2891, opened 2023-10-21
Add a `change` function to change the existing properties of a schema


chribjel:
**Feature request:**

Similar to extend, but you are only allowed to change the existing properties of a schema, not extend it.

This is technically possible with `.extend`, but there is no typescript autocompletion for this.

```typescript
const schema = z.object({
    name: z.string(),
    age: z.number().optional(),
});

// You are only allowed to change properties that already exist in the schema you are changing.
const changedScema = schema.change({
    age: z.number(),    // remove optional
});

const changedSchema2 = schema.change({
    occupation: z.string().optional(),    // typescript error, because this property does not exist on `schema`
});
```

Q16solver:
This is very good for changing existing schemas for different use cases e.g. for my case, if I have comments for posts, under the hood they are the same schema, but have different restrictions to the schema e.g. min/max character counts etc

colinhacks:
I found a way to get key autocompletion working inside `.extend`. 

![Screenshot 2024-04-29 at 10 14 29 PM](https://github.com/colinhacks/zod/assets/3084745/de12fb77-ae2b-4712-8a22-024fad4508d9)

This doesn't ensure that you *only* specify known keys - you're still allowed to overwrite a key with a new schema.  But it does reduce the chance that you make typos. My plan is to add this to Zod 4. 

I'm not convinced its worth adding a new method just for something like `.change()`...it's nice to have in some circumstances but the subtlety may be lost on new users. I'm trying to avoid having two nearly-identical methods.

Corviel:
@colinhacks Is this present on the new zod version 4 beta?

superxiao:
This would be very useful to help bridge the type‑safety gap between Zod and Kysely - both are great tools, but they don’t integrate well.  
With something like a `change()` method, I could start with the Zod schema defined for my API, then transform specific fields to match Kysely’s DB types such as `ColumnType`, `JsonColumnType`, or `Generated`. 
That’s also helpful for handling SQLite quirks like lack of a boolean type - by converting a field from `boolean` to `1 | 0` for storage, I can preserve type‑safety and still ensure correct conversion when passing data from/to the API.

anthonyalayo:
This kinda relates to my request:
https://github.com/colinhacks/zod/issues/4057

Essentially I'd love a way to be able to update error messages for **existing** fields after the fact. I am using `drizzle-zod`, and i want to take generated zod schemas and tweak the error messages from them.

[...]

Issue #2644, opened 2023-08-10
[Feature Request] - Property mapping for smoother object chain construction - mapProps(), and mapPropsPicked()


akutruff:
For input validation, Zod sometimes feels like it's copy/paste error prone and feels "reversed."  Often, you validate inputs that are going to be sent to another model type that will be written to an API or database.  The types of the properties of the inputs and the destination type usually have the same data type, or only a few properties on the input need to be coerced, refined, or transformed before simply being copied to another type.   I would also like the some of the errors on the model type to be surfaced immediately before any other logic takes place.  

Also, autocomplete is greatly hindered as making sure the property names and property types stay in sync is difficult.

I based a lib I wrote almost entirely off Zod, and added this functionality and it's been a big help.  You can see the documentation which mirrors Zod's docs, here:  

https://github.com/kutruff/nimbit#mapprops

I'd love for this functionality to also be in Zod.  

```ts
const Address = z.object({
  street: z.string(),
  zipcode: z.string().length(5) // validate zip is 5 chars.
});

const Person = z.object({
  name: z.string(),
  age: z.number().min(0), //An existing validation on the model
  address: Address,
  title: z.string()
});

const PersonInput = z.object({
  //z.string() is copy pasted here rather than just using the model.
  name: z.string().default(''),
  //Oops. The age validation on the model isn't caught early...
  age: z.string().pipe(z.coerce.number()),
  //Need to validate sub props and coerce Address...
  //Can't reuse the Address type and we didn't get the validation. Using merge() won't guarantee prop names match.
  address: z.object({ 
    street: z.string().refine(x => x.includes('St.')),
    //expect zipcode to be a number for input but we just forgot the Address validation.
    zipcode: z.number().pipe(z.coerce.string()) 
  })
});
```

Now imagine if we had `mapProps()` that allowed us on a property by property basis to work with the first type:  

```ts
const PersonInput = Person.mapProps({
  // p is the Person.name.  It's strongly typed.
  name: p => p.default(''),
  // now we get the age validator )
  age: p => z.coerce.number().pipe(p),
  // Now we can use map props on nested objects too!
  address: p => p.mapProps({ 
    street: sub => sub.refine(x => x.includes('St.')),
    // adding pipe to reuse the length validation on Address.zipcode
    zipcode: sub => z.number().pipe(z.coerce.string()).pipe(sub) 
  })
});
```

`mapPropsPicked()` would do the same but restrict the properties would only allow the properties that were specified in the mapping.

[...]

PR #3415, opened 2024-04-18
Implement `.remap()` on `ZodObject`

colinhacks:
The ability to modify the fields of a ZodObject schema *as a function* of the current fields is a much requested feature. This PR implements a new method `.remap()` to enable this.

- https://github.com/colinhacks/zod/issues/2644
- https://github.com/colinhacks/zod/issues/2891
- https://github.com/colinhacks/zod/pull/2471

Given a simple initial schema...

```ts
const User = z.object({
  name: z.string(),
  age: z.number(),
});
```

...this method overwrites the existing fields as a function of the current `shape`.

```ts
const ModUser = User.remap((shape) => ({
  name: shape.name.optional(),
  age: shape.age.nullable(),
}));

type ModUser = z.infer<typeof ModUser>;
// { name?: string; age: number | null; }
```

Or it can use string literals to remap keys.

```ts
const ModUser = User.remap(() => ({
  name: "fullname",
  age: "newage",
}));

type ModUser = z.infer<typeof ModUser>;
// { fullname: string; newage: number; }
```

These can be mixed as well:

```ts
const ModUser = User.remap((shape) => ({
  name: "fullname",
  age: shape.age.optional(),
}));
```

For simplicity a plain object syntax is supported as well. 

```ts
const ModUser = User.remap({
  name: "fullname",
  age: User.shape.age.optional(),
});

type ModUser = z.infer<typeof ModUser>;
// { fullname: string; age?: number; }
```


iiAku:
Thanks for this contribution @colinhacks. Curious to know what `.remap` adds vs `.transform` it's not straighforward to me. Perhaps I was used to mistakenly use `.transform` as mapper and `.remap` does exactly that whereas `.transform` is more field oriented ?

colinhacks:
So I actually think the ambiguity here is kind of a fatal flaw. I also think `.remap()` would be more valuable as a post-parse transform, as @iiAku had assumed. I'm going to close for now and re-evaluate.

therealokoro:
Any update on this? I really find this useful

Psycarlo:
Any updates?

superxiao:
https://github.com/colinhacks/zod/issues/2891#issuecomment-3218099274 some use cases for this which imo make it very useful.

PR #3424, opened 2024-04-20
Zod v3.23

colinhacks:
Publish v3.23.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "zod",
-  "version": "3.22.4",
+  "version": "3.23.0",
   "author": "Colin McDonnell <colin@colinhacks.com>",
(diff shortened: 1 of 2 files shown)


samchungy:
Not sure if you noticed but I think I spotted a little mistake in your change log by the way. 

```ts
  z.object({ type: z.literal("A").optional(), value: z.number() }),
  z.object({ type: z.literal("B").nullable(), value: z.string() }),
  z.object({ type: z.literal("C").readonly(), value: z.boolean() }),
  z.object({ type: z.literal("D").readonly(), value: z.boolean() }),
  z.object({ type: z.literal("E").catch("E"), value: z.unknown() }),
]);
```

I think D is meant to be a branded example?

PR #3427, opened 2024-04-21
[2963] Fixed emails with exclamation mark failing validation #2963

### LINKED ISSUE #2963: z.string().email() fails when email contains an exclamation mark
`z.string().email()` fails with emails that contain an exclamation mark (`!`) e.g.

`a!test@googlecalender.com`

(This happened to me after upgrading from `3.21.4` to `3.22.0`.)

uzair888:
Fixes #2963


colinhacks:
Thanks, but Zod is intentionally quite strict with email addresses and intentionally implements "Gmail validation" rules. See #2157. In the next version of Zod I'm thinking about how to accommodate a wider range of preferences here.

Issue #3439, opened 2024-04-25
Default enum generic


codenothing:
If you have existing string union types, enum requires both generics to be defined in order to type guard against the input. Example:

```ts
import { z } from 'zod';

type TColors = 'Red' | 'Green' | 'Blue';

// Type error
z.enum<TColors>(["Red", "Green", "Blue"]);

// Success
z.enum<TColors, [TColors, ...TColors[]]>(["Red", "Green", "Blue"]);
```

To get around this, the  `createZodEnum`  overrides need a default generic for the second parameter
```ts
function createZodEnum<U extends string, T extends Readonly<[U, ...U[]]> = Readonly<[U, ...U[]]>>(
  values: T,
  params?: RawCreateParams
): ZodEnum<Writeable<T>>;
function createZodEnum<U extends string, T extends [U, ...U[]] = [U, ...U[]]>(
  values: T,
  params?: RawCreateParams
): ZodEnum<T>;
```

PR #3440, opened 2024-04-25
[3439] Add default value to createZodEnum generic

codenothing:
Fix for #3439

If you have existing string union types, enum requires both generics to be defined in order to type guard against the input. Example:

```ts
import { z } from 'zod';

type TColors = 'Red' | 'Green' | 'Blue';

// Type error
z.enum<TColors>(["Red", "Green", "Blue"]);

// Success
z.enum<TColors, [TColors, ...TColors[]]>(["Red", "Green", "Blue"]);
```

This PR adds a default value to the `createZodEnum` generic for the list parameter

### DIFF
--- a/deno/lib/__tests__/enum.test.ts
+++ b/deno/lib/__tests__/enum.test.ts
@@ -18,6 +18,16 @@ test("infer enum", () => {
   util.assertEqual<MyEnum, "Red" | "Green" | "Blue">(true);
 });
 
+test("typed enum input", () => {
+  type MyEnumInputs = "Red" | "Green";
+  const MyEnum = z.enum<MyEnumInputs>(["Red", "Green"]);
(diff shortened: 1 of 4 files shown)


colinhacks:
Thanks for the PR, but this is currently intentional. Something like this isn't recommended:

```ts
// Type error
z.enum<TColors>(["Red", "Green", "Blue"]);
```

It's important that ZodEnum statically infer a *tuple* (`["Red", "Blue"]`) not an array (`("Red" | "Blue")[]`). Zod's schemas are designed to be strongly typed throughout. 

To accomplish your original goal, I recommend `satisfies`

```ts
type TColors = 'Red' | 'Green' | 'Blue';

const myEnum = z.enum(["Red", "Green", "Blue"]) satisfies z.ZodType<TColors>
```

PR #3447, opened 2024-04-26
describe how one can protect from cyclical objects starting an infini…

jandockx:
Document a way to protect against cyclical objects starting an infinite loop.


colinhacks:
Thanks, but this is obscure enough that I'd rather not make a significant addition to the docs. Perhaps you can put your code in a GitHub gist and link to it?

PR #3449, opened 2024-04-27
suggestions for effect plugin

tim-smart:


plugin/effect/src/index.ts
```diff
@@ -1,21 +1,20 @@
-import { Effect } from "effect";
+import * as Effect from "effect/Effect";
```
tim-smart, line 1: alias imports are better for library code, as they cause less issues with bundlers

plugin/effect/src/index.ts
```diff
@@ -1,21 +1,20 @@
...
 import * as z from "zod";
 
 function zodEffect(this: z.ZodType, data: unknown, params?: any) {
-  return Effect.tryPromise({
-    try: async () => this.parseAsync(data, params),
-    catch(error) {
-      return error as z.ZodError;
-    },
-  });
+  return Effect.flatMap(
```
tim-smart, line 5: Effect.promise + flatMap allows us to use the safe apis

plugin/effect/src/index.ts
```diff
@@ -1,21 +1,20 @@
...
+  );
 }
 
 function zodEffectSync(this: z.ZodType, data: unknown, params?: any) {
-  return Effect.try({
-    try: () => this.parse(data, params),
-    catch(error) {
-      return error as z.ZodError;
-    },
+  return Effect.suspend(() => {
```
tim-smart, line 13: Effect.suspend allows us to wrap a side effect that returns an Effect, perfect for the sync api.

colinhacks:
Thanks!!

Issue #3378, opened 2024-04-04
[Feature] Add phone number parsing


GustavoOS:
The idea is using a zod schema like
z.string().phone()
and let parse any phone number format valid world-wide.

colinhacks:
Super hard to do correctly. If someone wants to make a specific API proposal here, I'm all ears. But this is the kind of thing that's almost impossible to make everyone (or even most people) happy.

maurer2:
Validation for E.164 phone numbers could be added via regex. It should validate most, albeit not all e164 numbers, correctly. See documentation from Twilio/Sendgrid: https://www.twilio.com/docs/glossary/what-e164#regex-matching-for-e164

Cheers

iffa:
You can achieve this easily yourself with something like the following, and using libphonenumber-js at least somewhat guarantees good results:

```typescript
import parsePhoneNumber from "libphonenumber-js";
import * as z from "zod";

/**
 * Zod schema that validates a phone number using `libphonenumber-js`.
 * Attempts to parse the provided value with a default country of `FI`.
 *
 * If the phone number is valid, the schema transforms the phone number into
 * an international format (e.g. `+358401234567`).
 */
export const zPhoneNumber = z.string().transform((value, ctx) => {
  const phoneNumber = parsePhoneNumber(value, {
    defaultCountry: "FI",
  });

  if (!phoneNumber?.isValid()) {
    ctx.addIssue({
      code: z.ZodIssueCode.custom,
      message: "Invalid phone number",
    });
    return z.NEVER;
  }

  return phoneNumber.formatInternational();
});
```

bwarner:
> [google-libphonenumber](https://github.com/ruimarinho/google-libphonenumber)

That is a huge library and will increase bundle sizes alot.

Steveb599:
has someone managed to do it?

DaiFudo:
That would be really powerful for me, but it also looks quite hard. The numbers can be different in each country.

guilleigmu:
@DaiFudo since April 2025 you can use `z.e164()`. [PR](https://github.com/colinhacks/zod/pull/3476)

MickL:
Unfortunately, the new `e164()` uses a Regex which validates a lot of phone numbers correctly but also accepts non-valid phone numbers, e.g. common leading zero after country code `+49030` should be invalid. Issue: https://github.com/colinhacks/zod/pull/5524

mondesirm:
[...]

PR #3476, opened 2024-05-08
Add E.164 format validation for strings

maurer2:
Hello, following the previous [discussion](https://github.com/colinhacks/zod/issues/3378), I added validation for the E.164 format on strings using a modified regex from this [article](https://blog.stevenlevithan.com/archives/validate-phone-number#r4-3).  I removed support for spaces as those don't seem to be [permitted](https://gotrunk.com/blog/the-e-164-format-what-is-it-and-how-to-use-it-correctly/#:~:text=How%20to%20dial%20an%20E.164%20phone%20number%20correctly). The validation could be further enhanced by prohibiting `0` as the first value since country codes start with `1`.

There is an alternative regular expression from the [Twilio docs](https://www.twilio.com/docs/glossary/what-e164#regex-matching-for-e164) for E.164 numbers. This regular expression permits very short E.164 numbers. It could be used in tandem with the existing string validations like `min` and `max`. This would give users the option to adapt the number length further, which seems to be necessary in some instances, according to this [stack overflow post](https://stackoverflow.com/questions/14894899/what-is-the-minimum-length-of-a-valid-international-phone-number).

Cheers

Thank you for your contribution to our project! Before submitting your pull request, please ensure the following:

- [x] Your code changes are well-documented.
- [x] You have tested your changes.
- [x] You have updated any relevant documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -826,6 +826,20 @@ if (!env.success) {
 
 This is recommended over using `z.string().transform(s => JSON.parse(s))`, since that will not catch parse errors, even when using `.safeParse`.
 
+### E.164 telephone numbers
+
+The z.string().e164() method can be used to validate international telephone numbers.
+
+```ts
+const e164Number = z.string().e164();
+
+e164Number.parse("+1555555"); // pass
+e164Number.parse("+155555555555555"); // pass
+
+e164Number.parse("555555555"); // fail
+e164Number.parse("+1 1555555"); // fail
+```
+
 ## Numbers
 
 You can customize certain error messages when creating a number schema.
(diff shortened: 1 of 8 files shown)


colinhacks:
This seems good to me. Is there a specific reason this is still a draft PR? My plan is to merge this into `v4`. String validations like `uuid`, etc are going to be handled quite differently in Zod 4 but I'm happy to merge this as a starting point before I start the refactor.

maurer2:
Hello, apologies for the delay. I first needed to further investigate in order to be able to add the pr description.

Cheers

maurer2:
Hello, added a section on E.164 to both Readme files.

MickL:
@maurer2 Really nice work!!! I added a new issue tho because E.164 should not allow a zero after the country code, e.g. `+49030` should be invalid but currently is valid.

Issue: https://github.com/colinhacks/zod/issues/5563

PR #3487, opened 2024-05-11
[POC] [runtime performance] bottom up lazy issues/paths construction

jussisaurio:
Hi @colinhacks ,

I made a little POC targeting Zod's runtime performance with a similar approach that has been done in other parse-dont-validate style libraries where issue paths are only allocated when they are needed (i.e. when there actually are issues in parsing). This means paths don't need to be allocated in any successful parse cases, which is the common/hot path generally.

I'm not sure if this fits at all with your plans for the `v4` version, but I did this mostly as an exercise, and if a polished version of it ends up being something that's feasible from your POV, then great!

**Main changes:**

- `_parse()` now returns either `OK(validResult)` or `INVALID(issues: IssueData[])`; i.e. issues are not added to `ctx` but are returned along with the unsuccessful status value. Parent schemas will append the child issues to their own issues array recursively
- Issue paths are lazily allocated. Parent parse functions receiving errors from children will add `path` to the child issues, i.e. if an error happens in key `foo` of an object, the parent will add `path: ["foo"]` to the child error. This happens recursively, so the parent's parent will then unshift e.g. `path: ["bar", "foo"]` later. As mentioned, this happens on demand so paths are only constructed when the parser encounters an error.
- no `ParseContext` objects are allocated during parsing inside `_parse()` since they are not necessary
- `ParseStatus` and the concept of `dirty` are removed; dirtiness is handled separately in `ZodEffects` to decide whether to continue with refinements. LMK if I've misunderstood and `dirty` is needed elsewhere
- `input` is added to `IssueData` to continue supporting error maps operating on the parse input
- Although `ctx.addIssue()` is not used in the internals, refinements continue to support the `ctx.addIssue()` API, which just pushes to the `issues` array under the hood

**Some caveats / edge cases / breaking things:**

- The code is terrible spaghetti in some spots (especially when it comes to `async` parsing, refinements and handling some edge cases like prototype pollution etc etc) because I mainly wanted to get a POC done where the test set passes and I could concretely demonstrate the performance difference
- I removed support for, and the test for `"use path in refinement context"`, since paths are now on-demand constructed, and so, not available to the refinement context. This would be a breaking change ofc, but I couldn't figure out how to make that backwards compatible
- The expected return values of a few tests (that assert on which issues are returned) changed; not sure yet if I actually broke something crucial

**Performance:**

Here's a performance comparison using the [typescript-runtime-type-benchmarks](https://github.com/moltar/typescript-runtime-type-benchmarks) library:

**colinhacks/v4:**

```
Running "parseSafe" suite...
Progress: 100%

  zod:
    718 254 ops/s, ±0.68%

Finished 1 case!
Running "parseStrict" suite...
Progress: 100%

  zod:
    661 232 ops/s, ±0.38%

Finished 1 case!
Running "assertLoose" suite...
Progress: 100%

  zod:
    653 851 ops/s, ±0.65%

Finished 1 case!
Running "assertStrict" suite...
Progress: 100%

  zod:
    664 157 ops/s, ±0.47%
```

**jussisaurio/v4-lazy-issues:**

```
Running "parseSafe" suite...
Progress: 100%

  zod:
    1 042 488 ops/s, ±7.67%   // <-- 45% faster

Finished 1 case!
Running "parseStrict" suite...
Progress: 100%

  zod:
    975 107 ops/s, ±0.13%   // <-- 46% faster

Finished 1 case!
Running "assertLoose" suite...
Progress: 100%

  zod:
    913 760 ops/s, ±3.97%   // <-- 40% faster

Finished 1 case!
Running "assertStrict" suite...
Progress: 100%

  zod:
    971 467 ops/s, ±0.24%   // <-- 46% faster
```


colinhacks:
This is great stuff! 👍 I actually have this exact kind of change implemented in a `perf` branch, and it is indeed a big part of the v4 release. In fact, the inability to do `ctx.path` in refinements is the reason I starting thinking about a new major in the first place - otherwise I would have merged this approach into Zod 3 long ago. 

Your code is currently much cleaner than mine, which is a catastrophe 🙃 But I've also done a lot of perf work on the primitives types as well. 

There's also one issue with getting rid of "dirty". You need a way to return both a value & issues for when you have   This line in ZodEffects#_parse work:

```ts
const value = inner.status === "valid" ? inner.value : input;
```

You can't assume `input` can be passed into the refinement here. The `inner` schema may have both a) transformed the data and b) thrown some non-fatal errors. In this case you will fall back to `input`, but the refinement being executed may expect a different type. 

```ts
z
    .string()
    .transform((val) => val.length)
    .refine(() => false, { message: "always fails" })
    .refine(
      (val) => {
        console.log(`val`, val, typeof val);
        return (val ^ 2) > 10;
      } // should be number but it's a string
    )
    .parse("hello")
```The inner schema may have returned a transformed value

jussisaurio:
That's awesome to hear! 🔥 Are you also addressing `parseAsync` in that branch?

colinhacks:
Indeed, I was planning to do promise detection just as you implemented in your branch, but I haven't done it yet. The goal is to detect async refinements/transforms and reflect that in the inferred type. This would then "bubble up" the schema. The goal here is to provide a single method that obviates the `parse/safeParse/parseAsync/safeParseAsync` and returns `T | ZodError` (where `T` will be a promise if any async effects exist). But this is all bound up in a huge refactor I've been sketching out.

> I tried this as well, but it didn't make any meaningful difference in my tests 🤔

IIRC I think it only makes a significant difference on primitive parsing. There's already enough overhead in ZodObject that the impact is minimal. But I'll keep experimenting.

colinhacks:
So my branch is now over a year old and the rebase onto `v4` is driving me insane. I'm gonna merge this in since it's very close perf-wise. I'll manually add some of the other optimizations on top of it, and fix the `input` issue. Jussi, you're the best. Feel free to make a new PR for promise detection too.

colinhacks:
So I'm gonna work off of this branch after all 😅 My `perf` branch is now over a year old and the rebase onto `v4` is driving me insane. I'll cherry pick some of my other optimizations into this branch since it's very close perf-wise. I'll fix the input issue and remove input from IssueData (Zod has a policy of not putting input data in issues so it doesn't get logged in sensitive environments). Jussi, you're the best. Feel free to merge your Promise detection PR into this branch too.

jussisaurio:
Awesome! 

I won't be able to work on this in recent days so in case you're actively going to be pushing to this branch, I went ahead and fixed the rebase conflicts between this branch and the promise detection branch, so feel free to cherry-pick this commit onto it if you like: https://github.com/jussisaurio/zod/pull/1/commits/0f2014063d15dc6325d859d28c2d0628f6bfb29b

Fair warning, there's a few test failures in it (because I don't unfortunately have time to fix them)

colinhacks:
@jussisaurio Awesome, just cherrypicked & fixed the tests. Gonna merge this so I can start putting up more PRs against these changes.

Try not to rewrite Zod again in the next few days — I'm switching Zod to a monorepo so everything will be moving 🙃

Issue #2256, opened 2023-03-28
bug: `z.string().url()` is incosistent across browser


ElasticBottle:
Just learned that `z.string().url()` actually fails on firefox for `https://*.example.com` while it passes for every other browser.

Not too sure if this inconsistency is intended or not.

Happy to make a PR to help resolve the inconsistency with the `new URL` being used under the hood for FF

JacobWeisenburger:
I have confirmed this. However, this is an issue with the implementation of `URL` across browsers and not necessarily a problem with `zod`.

ElasticBottle:
Yeap, totally aligned.

saw the

```
try{
   new URL()
} catch(e) {}
```

block for validating `URL`

I think we should either 

1. Document this within the docs (Am leaning towards this)
2. Do some wrapper around URLs coming from Firefox to make sure it passes
3. Do custom regex

crutchcorn:
This is also an issue [in React Native, where `new URL` does not throw correctly in its constructor](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Blob/URL.js#LL140C49-L140C49):

![A URL of "Test" being validated by React Native's URL constructor](https://user-images.githubusercontent.com/9100169/229432851-0aa4ca8e-c036-4159-8bd0-d51476c70784.png)

This appears to have impacted at least one user: https://github.com/colinhacks/zod/issues/2236 as well as someone downstream as a user of my library (which uses Zod).

While this could likely be solved by poly-filling `URL` in React Native with a non-standard (for RN) URL polyfill, this method is:

- More challenging to debug for those less familiar with the inner working of React Native/polyfills
- Potentially breaking for users with React Native if the working of the current React Native usage is required

Ideally, I actually think that a regex should potentially replace the `URL` functionality in order to keep consistency and avoid headaches for end-users.

eduardo-amaro-maciel:
Weird because I saw someone using url() normally... I don't know if you used something else, I don't remember.

toninofox:
I was looking to open another issue but I guess this can be used as well.
The following value `foo-bar: true` will be considered as a valid URL for the `z.string().url()` schema.
I had to use a custom validator using `refine` to define a regex for HTTP URLs
```
[...]

PR #3497, opened 2024-05-16
Use regex for `.url()`

peterver:
Fixes issue where url can be anything due to new URL not correctly handling certain cases.

I took this url regex from a personal lib that I have up [here](https://github.com/ValkyrieStudios/validator/blob/main/lib/functions/vUrl.ts) which was originally adopted from [In search of the perfect URL validation regex](https://mathiasbynens.be/demo/url-regex)

Should fix:
- #2256 
- #2236


monodyle:
> To clarify, I’m looking for a decent regular expression to validate URLs that were entered as user input with. I have no interest in parsing a list of URLs from a given string of text (even though some of the regexes on this page are capable of doing that). I also don’t want to allow every possible technically valid URL — quite the opposite. See [the URL Standard](http://url.spec.whatwg.org/#parsing) if you’re looking to parse URLs in the same way that browsers do.
> 
> Assume that this regex will be used for a public URL shortener written in PHP, so URLs like `http://localhost/, //foo.bar/, ://foo.bar/, data:text/plain;charset=utf-8,OHAI and tel:+1234567890` shouldn’t pass (even though they’re technically valid). Also, in this case I only want to allow the HTTP, HTTPS and FTP protocols.

This is a good point! However, I believe the valid URL should follow the standard to ensure the library is suitable for general use cases. If we need something more explicit or specific, it should be defined within the application itself.

image

In my case, the app's [custom schema](https://developer.apple.com/documentation/xcode/defining-a-custom-url-scheme-for-your-app) (deeplink `steam://run`) appears to be a valid URL for me.

The problem is the [`URL` class](https://url.spec.whatwg.org/#url) is specific for WebAPI, so the output might difference between JS runtimes. So we need to change URL validation method, but still valid with standard.

The issue is that the [`URL` class](https://url.spec.whatwg.org/#url) is specific to the WebAPI, so the output might vary between different JavaScript runtimes. Therefore, we should to change the URL validation method while ensuring it still complies with the standard.

Tooni:
@colinhacks 
what would you like to see for something like this PR to be merged?

punkpeye:
Maybe it would good time to add more maintainers.

So many people depend on this library.

Yet the progress is stagnating.

This is NOT a dig at @colinhacks. I think it is super hard to maintain open-source software at scale.

It is just the project is at scale where a larger maintainer base would help to push it forward.

colinhacks:

PR #3499, opened 2024-05-17
Switch to Biome

colinhacks:
Switches to Biome and purges eslint 🎉 The switch was remarkably painless. 

- My only nit is the requiring of an explanation when disabling a rule...feels a bit paternalistic. Opened a discussion here but the team seems set on requiring it, which is reasonable https://github.com/biomejs/biome/discussions/2907
- Can't wait for Markdown formatting so I can remove Prettier too 🙃


.prettierrc.yaml
```diff
@@ -1,4 +1,3 @@
-# .prettierrc or .prettierrc.yaml
```
vbudovski, line 1: Should this file be removed, or is prettier still needed?

biome.jsonc
```diff
@@ -0,0 +1,32 @@
...
+    "formatter": {
+      "trailingComma": "es5"
+    }
+  },
+  "linter": {
+    "enabled": true,
+    "rules": {
+      "recommended": true,
+      "suspicious": {
+        "noExplicitAny": "off"
```
vbudovski, line 21: This is a good rule to keep enabled. There are definitely some situations where `any` is the correct type to use, but they should be pretty rare, and the exceptions clearly explained when done.

biome.jsonc
```diff
@@ -0,0 +1,32 @@
...
+      },
+      "style": {
+        "noUnusedTemplateLiteral": "off",
```
vbudovski, line 24: This seems like a pretty quick fix? Were there many errors as a result of this being enabled?

deno/lib/__tests__/generics.test.ts
```diff
@@ -15,6 +15,7 @@ test("generics", () => {
         nested: schema, // as z.ZodTypeAny,
       })
       .transform((data) => {
+        // biome-ignore lint/style/noNonNullAssertion: <explanation>
```
vbudovski, line 18: Is this potentially covering up a bug? Should replace the placeholder with an explanation here.

deno/lib/__tests__/nullable.test.ts
```diff
@@ -5,7 +5,7 @@ const test = Deno.test;
 import * as z from "../index.ts";
 
 function checkErrors(a: z.ZodTypeAny, bad: any) {
-  let expected;
+  let expected: any;
```
vbudovski, line 8: This should be typed to whatever `formErrors` produces. `bad: any` should be `bad: unknown` probably.

deno/lib/__tests__/pickomit.test.ts
```diff
@@ -49,6 +49,7 @@ test("pick parse - fail", () => {
 test("omit type inference", () => {
   const nonameFish = fish.omit({ name: true });
   type nonameFish = z.infer<typeof nonameFish>;
+  // biome-ignore lint/complexity/noBannedTypes:
```
vbudovski, line 52: Do you *really* mean to use `{}`? Should add an explanation for any ignores.

deno/lib/helpers/errorUtil.ts
```diff
@@ -1,5 +1,7 @@
 export type ErrMessage = string | { message?: string };
 export const errToObj = (message?: ErrMessage) =>
   typeof message === "string" ? { message } : message || {};
+
+// biome-ignore lint/suspicious/noShadowRestrictedNames: <explanation>
 export const toString = (message?: ErrMessage): string | undefined =>
```
vbudovski, line 6: Might be worth renaming to `errToString` for symmetry, and to avoid shadowing.

package.json
```diff
@@ -12,33 +12,26 @@
...
-    "eslint-plugin-import": "^2.25.4",
-    "eslint-plugin-simple-import-sort": "^7.0.0",
-    "eslint-plugin-unused-imports": "^2.0.0",
+    "execa": "^9.1.0",
     "husky": "^7.0.4",
     "jest": "^29.3.1",
     "lint-staged": "^12.3.7",
     "mitata": "^0.1.11",
     "nodemon": "^2.0.15",
     "prettier": "^2.6.0",
```
vbudovski, line 34: Can remove this dependency too. Prettier seems to be unused.

Jayllyz: > Can remove this dependency too. Prettier seems to be unused.

Prettier still required for markdown formatting

biome.jsonc
colinhacks: Sorry, but it gets onerous to keep casting `unknowns` in `_parse` implementations and tests. It's pretty common to use `any` in library code.

biome.jsonc
colinhacks: This one I just don't like as a rule

deno/lib/helpers/errorUtil.ts
colinhacks: That'll be a breaking change so I'll put that in it's own commit for the sake of cleanliness

PR #3507, opened 2024-05-21
Fixed type Inference with `custom` and `instanceof` in `catchall`

taro-28:
Hello!

I noticed that when using `custom` or `instanceof` with `catchall`, the type inference using `infer` does not work.
I have fixed it and added tests.

# Before
```typescript
const schema = z.object({ name: z.string() }).catchall(z.instanceof(Date));

// { name: string }
type Schema = z.infer<typeof schema>;
```

# After
```typescript
const schema = z.object({ name: z.string() }).catchall(z.instanceof(Date));

// { name: string } & { [k: string]: Date };
type Schema = z.infer<typeof schema>;
```

### DIFF
--- a/deno/lib/__tests__/object.test.ts
+++ b/deno/lib/__tests__/object.test.ts
@@ -193,6 +193,16 @@ test("test catchall parsing", async () => {
   expect(result2.success).toEqual(false);
 });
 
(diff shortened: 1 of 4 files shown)


colinhacks:

PR #3513, opened 2024-05-22
fix: ipv6 regex validation

### LINKED ISSUE #3512: IPv6 validation error
IPv6 validation fails when the last two parts of the IPv6 are compressed, both when one or both are omitted. It does not happen if it is any other part. I have checked the regex you use in https://github.com/colinhacks/zod/blob/1247caf58d1c189155bd04c068be1c758ab77c1c/src/types.ts#L613 and I have also checked it in external tools and the zod behavior is reproduced.

```ts
const ip = '2001:4888:50:ff00:500:d::';
const schema = z.string().ip();
const { success } = schema.safeParse(ip); // false
```

```ts
const ip = '2001:4888:50:ff00:500:d:0:0';
const schema = z.string().ip();
const { success } = schema.safeParse(ip); // true
```

```ts
const ip = '2001:4888:50:ff00::d:0:0';
const schema = z.string().ip();
const { success } = schema.safeParse(ip); // true
```

avegao:
IPv6 validation fails when the last two parts of the IPv6 are compressed, both when one or both are omitted. It does not happen if it is any other part.

I search a better regex for IPv6 and I found this: https://stackoverflow.com/a/17871737

I fix regex expression in standard and Deno src files and added tests.

It fixes #3512


avegao:
Hello team,

This is very annoying for my use case, could you please review this? 🙏 

Thanks

abnerwei:
same problem. ipv6  `::` ending

tgomesfork (approved):

holmok-bestow:
Is there any chance of this getting merged and released before v4?

Issue #3486, opened 2024-05-11
multipleOf acts wrong starting with 7 decimal places


OlliL:
If you want to enforce a max amount of decimal places you can use `multipleOf()`. If you want to enforce a max amount of 3 decimal places you can use `multipleOf(0.001)` which works perfectly fine. 
- 2.234 / 0.001 = 2234 :white_check_mark:
- 2.2345 / 0.001 = 2234.5 :bangbang:

So while this works up to 6 digits (`multipleOf(0.000001)`) fine, it stops working with 7 digits and above. To make it parse with 7 decimal places and above, the number to parse must contain exactly the amount of decimal places as specified in multpleOf.A small showcase:

```
  const number3 = 5.123;
  const number6 = 5.123456;
  const number7 = 5.1234567;
  const number8 = 5.12345678;
  let schema6 = z.number().multipleOf(0.000001);
  let schema7 = z.number().multipleOf(0.0000001);
  console.log(schema6.parse(number3));  // OK
  console.log(schema6.parse(number6));  // OK
  console.log(schema6.parse(number7));  // Error --> OK
  console.log(schema6.parse(number8));  // Error --> OK
  console.log(schema7.parse(number3));  // Error --> Bug
  console.log(schema7.parse(number6));  // Error --> Bug
  console.log(schema7.parse(number7));  // OK
  console.log(schema7.parse(number8));  // Error --> OK
```

The bug happens because of how JavaScript handles floats. Zods internal function floatSafeRemainder tries to parse its parameter. While it works fine with 0.000001, it stops working fine with 7 and more digits because then step becomes 1e-7 instead of 0.0000001 and the parsing logic stops working.

![grafik](https://github.com/colinhacks/zod/assets/4688427/7955417b-d1a9-4898-ba24-cd025fcd70f8)

A possible fix could be to account for that exponential representation of complex numbers could be the following drop-in replacement logic for stepDecCount:

```
    let stepDecCount = (step.toString().split(".")[1] || "").length;
    if(stepDecCount === 0 && /\d?e-\d?/.test(step)) {
      stepDecCount = step.toString().match(/\d?e-(\d?)/)[1];
    }
```
Seems to work fine for me.....

dellamora:
I would like to work on this. Can you assign me this issue ^^

OlliL:
Any news about getting this into main?

dellamora:
> Any news about getting this into main?

nothing yet

OlliL:
[...]

PR #3514, opened 2024-05-22
 Fix for `multipleOf` issue with 7 decimal places 

dellamora:
This pull request addresses the issue  [#3486](https://github.com/colinhacks/zod/issues/3486#event-12839437351) where `multipleOf()` was behaving incorrectly when dealing with numbers having 7 or more decimal places.

Changes:

- Modified the `stepDecCount` variable declaration from `const` to `let` to allow its value to be updated later in the code.
- Added a conditional check to test if the step value is in scientific notation (e.g., 1e-7). If it is, the code extracts the exponent part and uses it to update `stepDecCount`.
- This change ensures that `stepDecCount` correctly reflects the number of decimal places in `step`, **even when `step` is in scientific notation.**
- This fix should resolve the issue and make `multipleOf()` function as expected for numbers with 7 or more decimal places.


colinhacks:

PR #3518, opened 2024-05-23
Adding support for `.example()` to document schema examples

erunion:
I've added a new `.example()` method that, like `.describe()`, will allow consumers to document example values and representations of their schemas.

```ts
const documentedString = z
  .string()
  .url()
  .example("https://example.com")
  .example("https://example.local");

documentedString.examples; // ["https://example.com", "https://example.local"]
```

Like the JSON Schema [`examples`](https://json-schema.org/draft/2020-12/json-schema-validation#name-examples) array this currently has no type or shape restrictions. It may at some point be nice to restrict content being supplied to `.example()` to match the schema type (eg. only strings can be supplied on `z.string()`) but I've not (yet) done that here.

This work is being done to resolve https://github.com/colinhacks/zod/discussions/2266, and should, with a followup change to `zod-to-json-schema`[^1], allow Zod consumers to document schema examples within their generated JSON Schema representations.

[^1]: I plan on submitting a change to support `examples` to `zod-to-json-schema` after this work is in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@
     - [Async transforms](#async-transforms)
   - [`.default`](#default)
   - [`.describe`](#describe)
+  - [`.example`](#example)
   - [`.catch`](#catch)
   - [`.optional`](#optional)
   - [`.nullable`](#nullable)
@@ -2322,6 +2323,28 @@ documentedString.description; // A useful bit of text…
 ```
 
 This can be useful for documenting a field, for example in a JSON Schema using a library like [`zod-to-json-schema`](https://github.com/StefanTerdell/zod-to-json-schema)).
+
+### `.example`
+
+Use `.example()` to add an `examples` property to the resulting schema.
+
+```ts
+const documentedString = z.string().url().example("https://example.com");
+documentedString.examples; // ["https://example.com"]
+```
+
+Subsequent calls of `.example()` will add more examples to your schema.
+
+```ts
+const documentedString = z
+  .string()
+  .url()
+  .example("https://example.com")
+  .example("https://example.local");
+documentedString.examples; // ["https://example.com", "https://example.local"]
+```
+
+Like with `.describe()` this can be useful for documenting field examples, such as when using a JSON Schema library like [`zod-to-json-schema`](https://github.com/StefanTerdell/zod-to-json-schema)).
 
 ### `.catch`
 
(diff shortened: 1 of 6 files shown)


samchungy:
I feel that having an `any` type makes it pretty unusable. It would make it impossible to catch any drift from the schema which is a huge red flag.

The biggest challenge will be whether the input or output should be used in the type definition. Or both?

erunion:
@samchungy What if instead `.example()` is moved from being a global like `.describe()` and instead exists only on primitives?

maryam-saeidi:
Any update on this feature?

> The biggest challenge will be whether the input or output should be used in the type definition.
In which scenario might we want to use the output type? It seems to me the example is for input type, so we can start by using only input.

Btw, we can also provide an example directly to a type definition function, like the following, right?
```ts
z.string({
  description: 'An identifier for the resource.',
  example: '9c235211-6834-11ea-a78c-6feb38a34414',
})
```
Maybe it is worth mentioning it in the documentation.

samchungy:
> In which scenario might we want to use the output type?

In my case, when you're representing a response type. Eg this is the type that it should be after a transform.

samchungy:
> > In my case, when you're representing a response type. Eg this is the type that it should be after a transform.
> 
> Since this PR's main focus is related to [json-schema example](https://json-schema.org/draft/2020-12/json-schema-validation#name-examples) and the OpenAPI Specification, input type seems more relevant. Shall we start with that?
> 
> I think if we remove `any` usage and use the input type, this PR should be ready for review, and hopefully, it will be merged soon. 🤞🏻

You can use a JSON Schema to represent a response type too

maryam-saeidi:
> You can use a JSON Schema to represent a response type too

🤔 Do you have an example?

erunion:
So what should I do here? Move `.example()` to only exist on primitives? Something else?

maryam-saeidi:
I like the idea of first covering primitive types and avoiding `any`.

@colinhacks, do you have any input regarding this feature and the criteria that need to be covered in order to merge this PR?

samchungy:
> > You can use a JSON Schema to represent a response type too
> 
> 🤔 Do you have an example?

The simplest example is any open api schema you can describe a request and response type.

erunion:
Closing this out because this is now supported by the new `.meta()` work in v4! Great work Zod team.

https://v4.zod.dev/v4#metadata

Issue #3529, opened 2024-05-25
Descriptions from root schema carry over to schemas created with a chained `.array()`


EthanBehrends:
Related to changes made in #1756.

The above PR addresses an issue where descriptions would not carry over to schemas created using the chaining helper methods, like `.optional()`.

```typescript
// Before the above changes
z.string().describe("a description").description // "a description"
z.string().describe("a description").optional().description // undefined

// After
z.string().describe("a description").description // "a description"
z.string().describe("a description").optional().description // "a description"
```

The changes to fix this issue were also applied to `.array()`, which results in some unexpected behavior.

```typescript
// With this bug,
z.string().describe("a description").array()
// is equivalent to:
z.string().describe("a description").array().describe("a description")
// but this works as expected
z.array(z.string().describe("a description"))
```

The fix was super simple, I'll open a PR shortly!

PR #3530, opened 2024-05-25
Remove createParams cascade from .array()

EthanBehrends:
See [this issue](https://github.com/colinhacks/zod/issues/3529) for more context.

Basically, the description for a type was getting unexpectedly duplicated when using `.array()`. For example:
```typescript
z.string().describe("a description").array()

// ^ was functionally the same as v

z.array(
  z.string().describe("a description")
).describe("a description")
```

The issue was not present for `z.array()`
```typescript
z.array(z.string().describe("a description")) // works as expected
```

This unexpected behavior was introduced in a fix trying to ensure that something like `z.string().describe("a description").optional()` persisted the description after being wrapped by `ZodOptional`.

While this behavior makes sense for types like `ZodOptional` or `ZodNullable`, `ZodArray` is a completely separate type.

Other types that do this cascading `createParams`:
- `optional`
- `nullable`
- `promise`
- `or`
- `and`
- `transform`
- `default`
- `brand`
- `catch`

I think the logic makes sense for all of these (maybe `or` and `and` are also exceptions? not sure)

Added a test for this as well :)

### DIFF
--- a/deno/lib/__tests__/description.test.ts
+++ b/deno/lib/__tests__/description.test.ts
@@ -26,3 +26,11 @@ test("description should carry over to chained schemas", () => {
     description
   );
 });
+
+test("description should not carry over to chained array schema", () => {
+  const schema = z.string().describe(description)
+
+  expect(schema.description).toEqual(description);
+  expect(schema.array().description).toEqual(undefined);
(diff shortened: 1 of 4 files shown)


colinhacks:
Yikes, good catch!

PR #3533, opened 2024-05-26
Add method to preserve readonly props after call to extend

llowrey:
The only way to have object properties marked as `readonly` is to call the `readonly()` method, but this results in a `ZodReadonly` object which cannot be extended via the `extend()` method. The `merge()` method does not preserve the `readonly` markers. The `objectUtil.mergeShapes()` sort of achieves the desired result, however the resulting object is not compatible with `ZodSchema<T>` and thus can't be passed as a parameter to a method that expects `ZodSchema<T>`

This change adds the `readonlyProps()` method to `ZodObject` which results in a projection of the `ZodObject` with all fields marked `readonly`.

Additionally, the handling of the `Output` type parameter was changed so that when using `z.infer` the resulting type to has the `readonly` markers preserved.

The `readonlyProps` method could be extended in the future to allow for a list of properties to be specified which would allow only specific props to be marked `readonly`. However, it's easy enough to use `extend()` that this would be only a convenience.

Many many many alternative approaches were considered, such as overloading the `readonly()` method to return a `ZodReadonlyObject` class which would have an `extend()` method. IMO, this is not a good solution because the semantics of the `readonly()` are to freeze, and render readonly, the value, not the property assignment.

This solution works very well for my usecase where a common base object contains `readonly` properties.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1383,6 +1383,35 @@ person.parse({
 
 Using `.catchall()` obviates `.passthrough()` , `.strip()` , or `.strict()`. All keys are now considered "known".
 
+### `.readonlyProps`
+
+You can use the `.readonlyProps` method to mark all of the properties as readonly. When the `.extend` method is called on the resulting `ZodObject` the readonly decoration of the properties will be preserved.
+
+````ts
+const base = z
+  .object({
+    id: z.string(),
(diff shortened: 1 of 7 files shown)


colinhacks:

PR #3535, opened 2024-05-29
Allow creation of discriminated unions with a readonly array of options

haines:
If I extract the array of options out of the `z.discriminatedUnion` call, type-checking fails because an array is not assignable to a non-empty tuple:

```ts
const options = [
  z.object({ type: z.literal("x"), val: z.literal(1) }),
  z.object({ type: z.literal("y"), val: z.literal(2) }),
];

```

```
TS2345: Argument of type '(ZodObject<{ type: ZodLiteral<"x">; val: ZodLiteral<1>; }, "strip", ZodTypeAny, { type: "x"; val: 1; }, { type: "x"; val: 1; }> | ZodObject<{ type: ZodLiteral<...>; val: ZodLiteral<...>; }, "strip", ZodTypeAny, { ...; }, { ...; }>)[]' is not assignable to parameter of type '[ZodDiscriminatedUnionOption<"type">, ...ZodDiscriminatedUnionOption<"type">[]]'.
  Source provides no match for required element at position 0 in target.
```

However, if I add `as const` to the `options` definition to make it a readonly tuple, now type-checking fails because the options are not mutable:

```
TS2345: Argument of type 'readonly [ZodObject<{ type: ZodLiteral<"x">; val: ZodLiteral<1>; }, "strip", ZodTypeAny, { type: "x"; val: 1; }, { type: "x"; val: 1; }>, ZodObject<{ type: ZodLiteral<...>; val: ZodLiteral<...>; }, "strip", ZodTypeAny, { ...; }, { ...; }>]' is not assignable to parameter of type '[ZodDiscriminatedUnionOption<"type">, ...ZodDiscriminatedUnionOption<"type">[]]'.
  The type 'readonly [ZodObject<{ type: ZodLiteral<"x">; val: ZodLiteral<1>; }, "strip", ZodTypeAny, { type: "x"; val: 1; }, { type: "x"; val: 1; }>, ZodObject<{ type: ZodLiteral<...>; val: ZodLiteral<...>; }, "strip", ZodTypeAny, { ...; }, { ...; }>]' is 'readonly' and cannot be assigned to the mutable type '[ZodDiscriminatedUnionOption<"type">, ...ZodDiscriminatedUnionOption<"type">[]]'.
```

This PR makes it possible to do this with `as const` by making the generic constraint readonly.

### DIFF
--- a/deno/lib/__tests__/discriminated-unions.test.ts
+++ b/deno/lib/__tests__/discriminated-unions.test.ts
@@ -308,3 +308,14 @@ test("optional and nullable", () => {
   if (value.key === "b") value.b;
   if (value.key === null) value.b;
 });
+
+test("readonly array of options", () => {
+  const options = [
+    z.object({ type: z.literal("x"), val: z.literal(1) }),
+    z.object({ type: z.literal("y"), val: z.literal(2) }),
+  ] as const;
+
+  expect(
+    z.discriminatedUnion("type", options).parse({ type: "x", val: 1 })
+  ).toEqual({ type: "x", val: 1 });
+});
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -3101,7 +3101,7 @@ export type ZodDiscriminatedUnionOption<Discriminator extends string> =
 
 export interface ZodDiscriminatedUnionDef<
   Discriminator extends string,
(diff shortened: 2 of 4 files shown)


haines:
@colinhacks could you please take a look?

Issue #2060, opened 2023-02-19
Lazy Parsing (Feature request)


bluepnume:
This is something I've often felt would be useful in zod.

### Use case

- I'm calling an external api
- I want to parse the data I get back, to ensure nothing unexpected happens when I consume the data
- Certain properties I need 100% of the time, others I only need in special cases.

### Example

Here's a contrived example:

```typescript
const productSchema = z.object({
  id: z.string(),
  name: z.string(),
  isAvailableInBlue: z.boolean()
});

const offerProduct = async () => {
    const res = await fetch('https://some-api/product');
    const json = await res.json();
    const product = productSchema.parse(json);
    
    if (customerRequestedProductInBlue()) {
        if (!product.isAvailableInBlue) {
          console.log(`Sorry, ${ product.name } is not available in blue`);
          return;
        }
    }
    
    console.log(`Would you like a ${ product.name } today?`)
}
```

In this case:
- I need to be 100% sure `product.name` is present, and is a string
- I only care about `product.isAvailableInBlue` if the customer actually requested a blue product

So far so good. But let's say the api maintainer changes `isAvailableInBlue` to be an integer `0` or `1` -- which is obviously a backwards-incompatible change, but that happens.

Now: my code will error out 100% of the time when I call `productSchema.parse(json)`

### The problem

In cases like this, where I have no control over the external API, obviously it's annoying when a type changes like this, and I want my code to fail rather than assuming `product.isAvailableInBlue` is a boolean.

BUT, ideally I'd like it to only fail at runtime when I try to access `product.isAvailableInBlue`. That way so long as the customer is not requesting a blue product, my code will continue to run.

That way, I still have a bug I need to go fix, but the damage is much more minimal than `productSchema.parse(json)` failing 100% of the time. Most of my customers can still complete a checkout.

### A proposed solution

What I'd love to see is something like `productSchema.lazyParse()`:

```typescript
const productSchema = z.object({
  id: z.string(),
  name: z.string(),
  isAvailableInBlue: z.boolean()
});

const offerProduct = async () => {
    const res = await fetch('https://some-api/product');
    const json = await res.json();
    const product = productSchema.lazyParse(json);
    
    if (customerRequestedProductInBlue()) {
        if (!product.isAvailableInBlue) {
          console.log(`Sorry, ${ product.name } is not available in blue`);
          return;
        }
    }
    
    console.log(`Would you like a ${ product.name } today?`)
}
```

[...]

PR #3550, opened 2024-06-06
Propose feature`.defer()` to support lazy parsing

SnowSuno:
I propose a new feature `defer` to support lazy parsing, which is requested on #2060.

I think this feature would have pretty much use cases such as:
- Only parse the needed fields and prevent unused fields breaking the application. 
- Defer parsing of expensive large schemas.
- And more...

## What it does

You can mark a subpart of a schema as 'deferred' using `.defer()`

The parsing of a schema which is marked 'deferred', is skipped during the parsing, and **outputs a function that can be called to parse the input laterwise** where the schema is actually used.

```typescript
const schema = z.object({
  foo: z.string(),
  bar: z.string().defer(),
})

const parsed = schema.parse({ foo: "foo", bar: "bar" });  // bar is not parsed here

parsed.foo
parsed.bar() // bar is parsed here
```

## Interface
```typescript
// standalone
z.defer(z.string());

// method
z.string().defer();

// can be combined with other effects
z.string().transform(val => val.length).defer();
```

## Additional helper methods

Since the realworld usage would require scenarios like 'defer all props of an object', I also propose a helper method `.deferProps()` in `ZodObject`.

(Just like `.partial()` in `.optional()`)

> I don't think `.deferProps()` is the best naming of the functionality. I'm open for suggestions.

```typescript
z.object({
  foo: z.number(),
  bar: z.string(),
}).deferProps();

// same as 
z.object({
  foo: z.number().defer(),
  bar: z.string().defer(),
});
```

As `.partial()` accepts a mask to mark the fields to be optional. `.deferProps()` also does.
```typescript
z.object({ ... }).deferProps({ foo: true })  // defer prop 'foo' only
```

## WIP

The docs and more exhaustive test cases is a work in process. I plan to work on it after discussions about this new feature request.

### DIFF
--- a/deno/lib/__tests__/defer.test.ts
+++ b/deno/lib/__tests__/defer.test.ts
@@ -1,0 +1,110 @@
+// @ts-ignore TS6133
+import { expect } from "https://deno.land/x/expect@v0.2.6/mod.ts";
+const test = Deno.test;
+
+import { util } from "../helpers/util.ts";
+import * as z from "../index.ts";
+
+test("parse defer primitives", () => {
+  const schema = z.number().defer();
+
+  expect(() => schema.parse("not a number")).not.toThrow();
+  expect(() => schema.parse("not a number")()).toThrow();
+
+  expect(() => schema.parse(42)).not.toThrow();
+  expect(schema.parse(42)()).toEqual(42);
+});
+
+test("defer type inference", () => {
+  const schema = z
+    .object({
+      a: z.string(),
+      b: z.number().nullable(),
+    })
+    .partial();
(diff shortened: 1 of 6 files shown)


colinhacks:

bluepnume:
@SnowSuno this is awesome and would love it if you could re-target Zod 4.

Would you consider using `Object.defineProperty` rather than turning the property into a function, and parse on the initial property get?

That way the parse returns something of the same shape as the original output and we only get run-time errors when we access an invalid property rather than parse-time errors.

PR #3580, opened 2024-06-17
fix: adds ` | never` to `objectOutputType` as a signal to the LSP tha…

ahrjarrett:
This small PR updates the type of `objectOutputType` to fully evaluate the type.

This only seems to occur with `catchall` schemas, and only sometimes.

**Before:**

Screenshot 2024-06-17 at 12 12 45 PM

You can see in the screenshot that the output type includes the type constructor that created it (`objectOutputType`).

The fix for this when it occurs is to unify the result that `objectOutputType` returns with `never`.

The union of any type with the bottom (`never`) is the first type (` | never` is an identity operation), so this is a no-op in every context except for the LSP, which interprets a union as a signal that it needs to evaluate the node.

**After:**

Screenshot 2024-06-17 at 12 14 26 PM

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -2330,11 +2330,13 @@ export type objectOutputType<
   Shape extends ZodRawShape,
   Catchall extends ZodTypeAny,
   UnknownKeys extends UnknownKeysParam = UnknownKeysParam
-> = objectUtil.flatten<
-  objectUtil.addQuestionMarks<baseObjectOutputType<Shape>>
-> &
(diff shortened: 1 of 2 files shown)


ahrjarrett:
@colinhacks any reason not to merge this? Would be great to remove the patch my team is using so we can see the evaluated types for `catchall` and `passthrough` schemas.

Before: 
Screenshot 2024-07-12 at 12 57 38 PM

After:
Screenshot 2024-07-12 at 12 58 56 PM

colinhacks:

PR #3600, opened 2024-06-27
Fixed typo in README

shadone:


colinhacks:
This copy was just replaced with a banner image, but thanks!

PR #3610, opened 2024-07-02
Switch to `isolatedDeclarations`

colinhacks:
- Adds explicit types for all exports
- Lays the groundwork for publishing to JSR later
- Lays the groundwork for faster builds later (still using tsc for now)
- Prevents accidental `.d.ts` blowups from inferred recursive types


packages/zod/src/helpers/util.ts
```diff
@@ -38,11 +40,11 @@ export const getValidEnumValues = (obj: any) => {
     filtered[k] = obj[k];
   }
   return objectValues(filtered);
-};
+}
 
-export const objectValues = (obj: any) => {
+export function objectValues(obj: any): any {
```
alii, line 45: Just out of pure curiosity, what's the reason for switching to functions here? Is it a requirement of JSR's "fast types" (or isolatedDeclarations)?

colinhacks: Nope, just slowly migrating more arrow functions to `function` because I think it's prettier

alii: Agree it's prettier🔥🔥🔥. Think at the time i must have just been curious if/how it would affect isolatedDeclarations

PR #3615, opened 2024-07-03
fix(ZodDefault): apply default even if inner parse returns undefined

### LINKED ISSUE #3614: .default() should apply default if wrapped schema transforms to undefined
This may seem strange, but it's because I've been dealing with AWS CloudFormation parameters, where you can only have empty strings, not null or undefined values...

I was trying to do

```ts
// helper used in place of `.optional()` for my schema for parsing CloudFormation metadata
function optional<T extends z.ZodTypeAny>(schema: T) {
  return z
    .union([z.literal('').transform(() => undefined), schema.optional()])
    .optional()
}

const parameter = optional(z.string()).default('test')
```

I expected `parameter.parse('')` to be `'test'` but instead it's `undefined`.  This violates the output type of `.default('test')`, so I think even if the input is defined, `ZodDefault` should check that the parse output is too.

jedwards1211:
fix #3614

Note: I can't see any reasonable way to make
```ts
    z.union([
      z.literal("").transform(async () => undefined),
      z.string().optional(),
    ])
    .default("inner")
    .default("outer")
    .parse("")
```
return `"outer"`, it returns `"inner"`.

### DIFF
--- a/deno/lib/ZodError.ts
+++ b/deno/lib/ZodError.ts
@@ -27,6 +27,7 @@ export const ZodIssueCode = util.arrayToEnum([
   "invalid_return_type",
   "invalid_date",
   "invalid_string",
+  "failed_default",
   "too_small",
   "too_big",
   "invalid_intersection_types",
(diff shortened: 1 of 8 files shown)


colinhacks:

PR #3624, opened 2024-07-05
Support for all types of discriminators in `z.discriminatedUnion`

RichardCPoint:
Currently in discriminated unions, the discriminator key must be one of a limited number of literal types (see [`getDiscriminator`](https://github.com/colinhacks/zod/blob/4641f434f3bb3dab1bb8cb07f44dd2693c72e35e/src/types.ts#L3060C1-L3091C3)).

This change allows *any* Zod type to be used for the discriminator key

This resolves the following issue:
* #1075

and also covers other use-cases, e.g. to allow an "anything else" case, where the final discriminator key accepts any value other than the known enum values from the other cases:
* #3654 

Async parsing is supported (and covered by a unit test), with validation of the keys happening in parallel.

At parsing time, if the object being parsed matches more than one discriminator key, the object is considered invalid (previously this could always be checked at type-construction time).

## Performance

There is only a performance hit if some of the discriminator keys are not one of the literal types already recognized (in which case all unrecognized keys must be checked one by one). Existing use-cases will have the same performance as before.

### DIFF
--- a/deno/lib/ZodError.ts
+++ b/deno/lib/ZodError.ts
@@ -21,6 +21,7 @@ export const ZodIssueCode = util.arrayToEnum([
   "custom",
   "invalid_union",
   "invalid_union_discriminator",
+  "ambiguous_union_discriminator",
   "invalid_enum_value",
   "unrecognized_keys",
   "invalid_arguments",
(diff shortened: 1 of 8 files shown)


RichardCPoint:
Closing, as it has been made clear that no further changes will be accepted for `z.discriminatedUnion`.

PR #3629, opened 2024-07-09
Add typed variant of parse, parseAsync, safeParse, and safeParseAsync

### LINKED ISSUE #1748: `parse.typed(...)` or some kind of typed parse method
It'd be nice to have zod help the developer max out their chances of parsing a valid object, when it's manually constructed. Example:

```ts
const User = z.object({
  name: z.string().regex(/^\w+ \w+$/),
  age: z.number(),
})

User.parse({
  name: 'Bob',
  birthdate: new Date('1970-01-01'),
})
```

The above code compiles, but we have enough information to _know_ that it'll throw a ZodError at runtime. It'd be nice if there was a method like

```ts
User.parse.typed({
  name: 'Bob',
  birthdate: new Date('1970-01-01'),
})
```

Where the expected input type was `Input` (from the `ZodType<Output, Def, Input>`) rather than `unknown`. So the above code would error because the developer used `birthdate` instead of `age`.

This would be especially useful for types with a custom `.refine` or `.regex` method or similar where it could be misleading to use _just_ a type declaration:

```ts
const user: z.infer<typeof User> = {
  name: "Robert'; -- DROP TABLE Students; --"
  age: 40,
}
```

The above will compile, and implies that the `const user` is a validated `User` instance, but it isn't because `name` doesn't match the regex. The current options are to do the above, which is safe-looking at compile time, but unsafe at runtime, or to use `.parse(...)` which is unsafe at compile time and safe-ish at runtime (but will throw errors).

wbobeirne:
Closes #1748. I didn't add any tests since this is build time and not run time, and it's just a wrapper around the existing methods, but if there are some tests you would find useful just let me know and I'll add 'em!

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,10 @@
   - [`.parseAsync`](#parseasync)
   - [`.safeParse`](#safeparse)
   - [`.safeParseAsync`](#safeparseasync)
+  - [`.typedParse`](#typedparse)
+  - [`.typedSafeParse`](#typedsafeparse)
+  - [`.typedParseAsync`](#typedparseasync)
+  - [`.typedSafeParseAsync`](#typedsafeparseasync)
   - [`.refine`](#refine)
     - [Arguments](#arguments)
     - [Customize error path](#customize-error-path)
@@ -2055,6 +2059,27 @@ For convenience, this has been aliased to `.spa`:
 ```ts
 await stringSchema.spa("billie");
 ```
+
+### `.typedParse`
+
+A typed version of `.parse` that type-checks the input
+
+```ts
+await stringSchema.typedParse("typed"); // No problem.
+await stringSchema.typedParse(12); // Argument of type 'number' is not assignable to parameter of type 'string'.
+```
+
+### `.typedSafeParse`
+
+A typed version of `.safeParse` that type-checks the input
+
+### `.typedParseAsync`
+
+A typed version of `.parseAsync` that type-checks the input
+
+### `.typedSafeParseAsync`
+
+A typed version of `.safeParseAsync` that type-checks the input
 
 ### `.refine`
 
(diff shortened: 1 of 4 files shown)


mmkal:
A thought - I know this is closer to what I originally proposed two years ago but what about making parse/safeParse/parseAsync generic, taking an options typearg object, instead? Would avoid expanding the API surface just for the sake of types:

```diff
-  parse(data: unknown, params?: Partial<ParseParams>): Output {
+  parse<Options extends {strict: boolean} = {strict: false}>(data: Options['strict'] extends true ? Input : unknown, params?: Partial<ParseParams>): Output {
```

Usage:

```ts
const S = z.string()

S.parse<{strict: true}>('x') // ok
S.parse<{strict: true}>(123) // ts error
S.parse(123) // no ts error
```ts

colinhacks:
My preference for this would be a single `.assertType(input: Input)` method that returns `void`...basically just there as a utility for type checking. I'm trying to get rid of methods not add more 🙃 Open to suggestions on naming. Something like this will land in Zod 4.

wbobeirne:
Personally I don't love that because it puts the onus on the developer remembering to make the call, and comes with some runtime overhead.

Right now parsing a schema silently succeeds even if it's been statically defined in a way that would always fail at runtime. I quite like mmkal's suggestion with my suggested addition of having a way to override the default, because then I could opt in to forcing my team to consider whether or not they're parsing the schema with trusted input or not, instead of just assuming it's untrusted. This is especially valuable if part of why you use Zod schemas is also the transformation and coercion, so you're using them a lot in business logic with trusted inputs.

If you're sold on that approach, I can rework my PR to use generics on the existing methods instead.

wbobeirne:
Closing based on https://github.com/colinhacks/zod/issues/1748#issuecomment-2967769964

PR #3653, opened 2024-07-19
Switch to `tshy`

colinhacks:
- Builds all packages with `tshy`
- Removes 4 unnecessary tsconfigs
- Switches to `"type": "module"` and adds necessary extensions to all imports
- Simplifies configs across the board


package.json
```diff
@@ -60,5 +60,8 @@
     "bench": "pnpm run build:cjs && tsx benchmarks/index.ts",
     "prepare": "husky install",
     "publish:jsr": "jsr publish --dry-run"
+  },
+  "dependencies": {
+    "tshy": "^3.0.2"
```
jakebailey, line 65: Was this intended to be a devDep? (it's the monorepo root so not harmful of course)

colinhacks: Fixed! Thanks

PR #3705, opened 2024-08-12
Update README.md

jubar:
In the customize error path section, we should use `!==` instead of `===` comparison.


colinhacks:
It's correct as is - `.refine()` will throw an error when the result is falsy.

PR #3712, opened 2024-08-16
Add support for `base64url` strings

### LINKED ISSUE #3711: Add support for `base64url` strings
The `.base64()` function validates strings that contain binary information encoded in Base 64. A modification of Base 64 that can safely be used in URLs and filenames is given by the `base64url` standard specified in [RFC 4648 § 5](https://datatracker.ietf.org/doc/html/rfc4648#section-5). As summarized [here](https://base64.guru/standards/base64url), 

> it uses the same algorithm as the main standard, but differs in the following:
> - Replaces “+” by “-” (minus)
> - Replaces “/” by “_” (underline)
> - Does not require a padding character
> - Forbids line separators

I propose a method `.base64url()` that is similar to `.base64()`, but instead matches against a regular expression that describes `base64url`-encoded values.

marvinruder:
Fixes #3711

### DIFF
--- a/src/__tests__/string.test.ts
+++ b/src/__tests__/string.test.ts
@@ -164,89 +164,85 @@ test("email validations", () => {
   ).toBe(true);
 });
 
-test("base64 validations", () => {
-  const validBase64Strings = [
-    "SGVsbG8gV29ybGQ=", // "Hello World"
-    "VGhpcyBpcyBhbiBlbmNvZGVkIHN0cmluZw==", // "This is an encoded string"
-    "TWFueSBoYW5kcyBtYWtlIGxpZ2h0IHdvcms=", // "Many hands make light work"
-    "UGF0aWVuY2UgaXMgdGhlIGtleSB0byBzdWNjZXNz", // "Patience is the key to success"
-    "QmFzZTY0IGVuY29kaW5nIGlzIGZ1bg==", // "Base64 encoding is fun"
-    "MTIzNDU2Nzg5MA==", // "1234567890"
-    "YWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXo=", // "abcdefghijklmnopqrstuvwxyz"
-    "QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVo=", // "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
-    "ISIkJSMmJyonKCk=", // "!\"#$%&'()*"
-    "", // Empty string is technically valid base64
-    "w7/Dv8O+w74K", // ÿÿþþ
-  ];
+const validBase64Strings = [
+  "SGVsbG8gV29ybGQ=", // "Hello World"
+  "VGhpcyBpcyBhbiBlbmNvZGVkIHN0cmluZw==", // "This is an encoded string"
+  "TWFueSBoYW5kcyBtYWtlIGxpZ2h0IHdvcms=", // "Many hands make light work"
+  "UGF0aWVuY2UgaXMgdGhlIGtleSB0byBzdWNjZXNz", // "Patience is the key to success"
+  "QmFzZTY0IGVuY29kaW5nIGlzIGZ1bg==", // "Base64 encoding is fun"
+  "MTIzNDU2Nzg5MA==", // "1234567890"
+  "YWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXo=", // "abcdefghijklmnopqrstuvwxyz"
+  "QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVo=", // "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+  "ISIkJSMmJyonKCk=", // "!\"#$%&'()*"
+  "", // Empty string is technically valid base64
+  "w7/Dv8O+w74K", // ÿÿþþ
+];
 
-  for (const str of validBase64Strings) {
-    expect(str + z.string().base64().safeParse(str).success).toBe(str + "true");
-  }
-
-  const invalidBase64Strings = [
-    "12345", // Not padded correctly, not a multiple of 4 characters
-    "12345===", // Not padded correctly
-    "SGVsbG8gV29ybGQ", // Missing padding
-    "VGhpcyBpcyBhbiBlbmNvZGVkIHN0cmluZw", // Missing padding
-    "!UGF0aWVuY2UgaXMgdGhlIGtleSB0byBzdWNjZXNz", // Invalid character '!'
-    "?QmFzZTY0IGVuY29kaW5nIGlzIGZ1bg==", // Invalid character '?'
-    ".MTIzND2Nzg5MC4=", // Invalid character '.'
-    "QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVo", // Missing padding
-    "w7_Dv8O-w74K", // Has - and _ characters (is base64url)
-  ];
-
-  for (const str of invalidBase64Strings) {
-    expect(str + z.string().base64().safeParse(str).success).toBe(
-      str + "false"
-    );
-  }
+test.each(validBase64Strings)("base64 should accept %s", (str: string) => {
+  expect(z.string().base64().safeParse(str).success).toBe(true);
 });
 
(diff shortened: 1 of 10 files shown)


marvinruder:
The new method should probably also be added to https://github.com/colinhacks/zod/blob/main/README.md#strings, but I didn’t want to predict the version number under which this feature may be released.

src/types.ts
```diff
@@ -616,6 +617,10 @@ const ipv6Regex =
 const base64Regex =
   /^([0-9a-zA-Z+/]{4})*(([0-9a-zA-Z+/]{2}==)|([0-9a-zA-Z+/]{3}=))?$/;
 
+// https://base64.guru/standards/base64url
+const base64urlRegex =
+  /^([0-9a-zA-Z-_]{4})*(([0-9a-zA-Z-_]{2}(==)?)|([0-9a-zA-Z-_]{3}(=)?))?$/;
```
rotu, line 622: According to the spec, excess pad characters may be ignored:
https://datatracker.ietf.org/doc/html/rfc4648#section-3.3
Why not use the simpler regex `/^[0-9a-zA-Z-_=]*$/`?

src/types.ts
```diff
@@ -1200,6 +1218,9 @@ export class ZodString extends ZodType<string, ZodStringDef, string> {
   get isBase64() {
     return !!this._def.checks.find((ch) => ch.kind === "base64");
   }
+  get isBase64url() {
```
rotu, line 1221: This (and `base64url()`) above could use a comment. If you haven't heard of `base64url` before, you might assume this is some sort of URL (`e.g. `data:base64,abc`) rather than a text encoding of binary data.

src/__tests__/string.test.ts
```diff
@@ -199,6 +202,52 @@ test("base64 validations", () => {
...
+    "ISIkJSMmJyonKCk", // "!\"#$%&'()*"
+    "ISIkJSMmJyonKCk=", // "!\"#$%&'()*" with padding
+    "", // Empty string is technically valid base64url
+    "w7_Dv8O-w74K", // ÿÿþþ
+    "123456",
+  ];
+
+  for (const str of validBase64urlStrings) {
+    expect(str + z.string().base64url().safeParse(str).success).toBe(
+      str + "true"
```
rotu, line 231: what's with this weird concatenation?

rotu:
I like the idea, but I'm not sure I like the implementation. If the idea is "parse, don't validate", then maybe it would be more on-brand to provide a transform to get at the represented binary data, rather than just validate the string unchanged. (of course the same goes for the existing `z.string().base64()` method)

src/__tests__/string.test.ts
marvinruder: I took that part from the `base64` tests above, it probably helps with debugging tests since you can directly see which string is failing.

marvinruder:
> I like the spirit, but I'm not sure I like the implementation. If the idea is "parse, don't validate", then maybe it would be more on-brand to provide a transform to get at the represented binary data, rather than just validate the string unchanged. (of course the same goes for the existing `z.string().base64()` method)

Parsing the encoded data into a buffer would be an interesting opt-in feature given that people are only interested in the binary data. My main use case for `base64url` at the moment is [Web Authentication](https://www.w3.org/TR/webauthn-2/), and [a popular library implementing that standard](https://www.npmjs.com/package/@simplewebauthn/server) expects base64url-encoded strings in many of its interfaces, so I am only interested in validation, as the library already does the parsing and expects unparsed strings. It would be great if Zod would (continue to) offer validation-only processing of base64(url)-encoded strings.

src/types.ts
marvinruder: This regex would allow `=` at any position in the string, which is not valid `base64url`. Even a regex allowing `=` only at its end would allow invalid strings like `ABCDE`; here, the characters `ABCD` encode three bytes of information, and the `E` does not hold enough information for a full octet. The corresponding `base64` expression would be `ABCDE===` with three padding characters, which is not allowed either.

The existing `base64Regex` is rather advanced and validates things like “only multiples of four characters allowed” and “at most two padding characters allowed”. As `base64url` is derived from `base64` with the addition “padding can be skipped”, my `base64urlRegex` is derived from `base64Regex` by making padding characters optional, which retains the validations for the aforementioned edge cases.

src/types.ts
marvinruder: Agreed, added in 05c4922fa36a225b5671d9b4c5d160d5ef563cd1.

src/__tests__/string.test.ts
rotu: Weird. I don't insist you change it, but you can get better DX with:
```js
const validBase64urlStrings = [
  "SGVsbG8gV29ybGQ", // "Hello World"
  "SGVsbG8gV29ybGQ=", // "Hello World" with padding
  // ...
];
test.each(validBase64urlStrings)(
  "base64url should accept %p",
  (str: string) => {
    expect(z.string().base64url().safeParse(str).success).toBe(true);
  }
);
```

marvinruder: That works much better, I changed the test cases for `base64()` as well.

src/types.ts
rotu: > This regex would allow = at any position in the string, which is not valid base64url

By my reading of section 3.3, it is valid. The spec allows extra padding as valid input to a decoder:

> Furthermore, such specifications MAY ignore the pad character, "=", treating it as non-alphabet data, if it is present before the end of the encoded data.  If more than the allowed number of pad characters is found at the end of the string (e.g., a base 64 string terminated with "==="), the excess pad characters MAY also be ignored.

On the other hand, by section 3.2, an encoder *must* include the padding:

> Implementations MUST include appropriate pad characters at the end of encoded data unless the specification referring to this document explicitly states otherwise.

So I guess the question is should this err on the side of leniency or strictness?

marvinruder: Okay, I didn’t catch that after my first reading of the spec. But if I understand it correctly, the “specifications” in

> Furthermore, such specifications MAY ignore the pad character, "=", …

refer to the sentence at the beginning of the paragraph

> …, unless the specification referring to this document explicitly states otherwise.

meaning that a specification based on base64 like MIME or PEM can decide to ignore pad characters before the end of the encoded data, just as MIME ignores CRLFs instead of rejecting input containing those non-alphabet characters. Section 5 of the RFC, specifying base64url, neither explicitly states that non-alphabet characters or padding characters before the end of the encoded data may be ignored (it only allows “skipping the padding” while referring to section 3.2, which described padding at the end of the encoded data), nor does it explicitly reference section 3.3. In my understanding, this would not allow padding characters before the end of the encoded data.

Also, JavaScript runtimes like Node.js or Bun do not simply ignore those padding characters and instead terminate decoding when reaching them:

```js
// Node.js:
> Buffer.from("SGVsbG8gV29ybGQ=", "base64url")
<Buffer 48 65 6c 6c 6f 20 57 6f 72 6c 64>
> Buffer.from("SGVsbG8gV29ybGQ", "base64url")
<Buffer 48 65 6c 6c 6f 20 57 6f 72 6c 64>
> Buffer.from("SGVs=bG8gV29ybGQ", "base64url")
<Buffer 48 65 6c>

// Bun:
> Buffer.from("SGVsbG8gV29ybGQ=", "base64url")
Buffer(11) [Uint8Array] [
   72, 101, 108, 108,
  111,  32,  87, 111,
  114, 108, 100
]
> Buffer.from("SGVsbG8gV29ybGQ", "base64url")
Buffer(11) [Uint8Array] [
   72, 101, 108, 108,
  111,  32,  87, 111,
  114, 108, 100
]
> Buffer.from("SGVs=bG8gV29ybGQ", "base64url")
Buffer(3) [Uint8Array] [ 72, 101, 108 ]
```

So perhaps I am wrong in interpreting the specification, but if so, runtimes aren’t better at that.

rotu: > Also, JavaScript runtimes like Node.js or Bun do not simply ignore those padding characters and instead terminate decoding when reaching them

Oof! That's an excellent observation and a great reason to reject `=` in the middle! Also it seems like the trailing `=` is not handled particularly well, so I think your regex hits the perfect balance.

rotu:
lgtm!

PR #3727, opened 2024-08-26
fix: dirty union bug

mmkal:
bug repro: https://stackblitz.com/edit/stackblitz-starters-swdj3e?file=index.test.js&startScript=test

Basically, zod v3 wrongly reports a "normal" error rather than a union error for a union of strict objects (and presumably other unions that set status=dirty:

```ts
import { test, expect } from 'vitest';
import { z } from 'zod';

test('hmm', () => {
  const union = z.union([
    z.object({ x: z.number() }).strict(),
    z.object({ y: z.number() }).strict(),
  ]);

  expect(union.safeParse({ x: 1, y: 1 })).toMatchInlineSnapshot(`
    {
      "error": [ZodError: [
      {
        "code": "unrecognized_keys",
        "keys": [
          "y"
        ],
        "path": [],
        "message": "Unrecognized key(s) in object: 'y'"
      }
    ]],
      "success": false,
    }
  `);
  expect(union.safeParse({ x: 1, y: 1 })).toMatchObject({
    success: false,
    error: {
      issues: [
        {
          code: 'invalid_union',
          unionErrors: [expect.any(z.ZodError), expect.any(z.ZodError)],
        },
      ],
    },
  });
});
```

Note: this does not repro in v4. I'm not sure if there's anything going into zod v3 still but opening this against main in hopes that it can get fixed in v3.

### DIFF
--- a/src/__tests__/unions.test.ts
+++ b/src/__tests__/unions.test.ts
@@ -62,3 +62,23 @@ test("readonly union", async () => {
   union.parse("asdf");
   union.parse(12);
 });
+
+test("union of strict objects", async () => {
+  // ensure bug in 3.23.8 is fixed, where "dirty" parse failures would mean reporting the first error, rather than a proper `invalid_union` with `unionErrors`.
+  // Demo of bug: https://stackblitz.com/edit/stackblitz-starters-swdj3e?file=index.test.js&startScript=test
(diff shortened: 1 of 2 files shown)


mmkal:
v4 is life

PR #3788, opened 2024-10-09
add readonly to BRAND-ed internal type structure

### LINKED ISSUE #3787: Request: make BRAND<T> readonly to support @typescript-eslint/prefer-readonly-parameter-types
Currently, BRAND is defined as follows:

```ts
export type BRAND<T extends string | number | symbol> = {
  [BRAND]: { [k in T]: true };
};
```

https://github.com/LumaKernel/zod/blob/3032e240a0c227692bb96eedf240ed493c53f54c/src/types.ts#L4904-L4906

Here is the suggestion to update this to following readonly added type.

```ts
export type BRAND<T extends string | number | symbol> = {
  readonly [BRAND]: { readonly [k in T]: true };
};
```

This can support the [`@typescript-eslint/prefer-readonly-parameter-types`](https://typescript-eslint.io/rules/prefer-readonly-parameter-types/https://typescript-eslint.io/rules/prefer-readonly-parameter-types/) rule.

This can be now confirmed as following with the repository: TO-BE-ADDED

```sh
npx eslint .

/<proj>/int.ts
  8:22  error  Parameter should be a read only type  @typescript-eslint/prefer-readonly-parameter-types

/<proj>/span.ts
  15:23  error  Parameter should be a read only type  @typescript-eslint/prefer-readonly-parameter-types

✖ 2 problems (2 errors, 0 warnings)

```

And I already confirmed that, if I replace the `BRAND` definition in `node_modules/` as above, this lint error is resolved for `span.ts` (using `z.object()` base).

Unfortunately, for the limitation of the eslint rule, the branded types for literal types like `number & { readonly some: null }` is not treated as readonly by the rule. That's why, `int.ts` (using `z.number()` base) in the repo cannot be resolved by this change, but it's still happy for the pattern of `span.ts`, it's mainly used in my project.

LumaKernel:
Resolves https://github.com/colinhacks/zod/issues/3787

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -4902,7 +4902,7 @@ export interface ZodBrandedDef<T extends ZodTypeAny> extends ZodTypeDef {
 
 export const BRAND: unique symbol = Symbol("zod_brand");
 export type BRAND<T extends string | number | symbol> = {
-  [BRAND]: { [k in T]: true };
+  readonly [BRAND]: { readonly [k in T]: true };
 };
 
 export class ZodBranded<


colinhacks:

PR #3796, opened 2024-10-12
Input contravariance

### LINKED ISSUE #3795: ZodType is not contravariant to the input
## Background

[Contravariance](https://www.typescriptlang.org/docs/handbook/2/generics.html#variance-annotations) is a property associated of generic types on their type parameters where the generic type `G<A>` extends `G<B>` if `B` extends `A`. It is usually associated with type parameters used as inputs.

Each `ZodType` is composed of an input and an output type. Consider, for instance, the following Zod schema:

```ts
const s = z.object({
    a: z.string().default("foo")
});
```

Its input type has `a` as an optional field, but it is required on the output type:

```ts
type A = z.input<typeof s>; // { a?: string | undefined; }
type B = z.output<typeof s>; // {a: string}
```

## Description

Given that the input and output types are type parameters to `ZodType`, one might expect `ZodType` to be covariant on the output type and contravariant on the input type. While the former is true, the latter, unfortunately, is not.

## Example
```ts
function f<S extends z.ZodType<string, any, any>>(_: S) {}
f(z.literal("a")); // Works fine

function g<S extends z.ZodType<any, any, "a">>(_: S) {}
g(z.string()); // Errors out :(
```

## Expected behaviour

`ZodType` should be contravariant with respect to the input type

## Actual behaviour

`ZodType` is not contravariant with respect to the input type

mstniy:
Fixes https://github.com/colinhacks/zod/issues/3795.

Breaking changes:
- Turns `safeParse`, `safeParseAsync`, `spa`, and `catch` into independent functions taking schemas as parameters, instead of being methods on `ZodType`.

### DIFF
--- a/deno/lib/__tests__/all-errors.test.ts
+++ b/deno/lib/__tests__/all-errors.test.ts
@@ -60,7 +60,7 @@ test("form errors type inference", () => {
 });
 
 test(".flatten() type assertion", () => {
-  const parsed = Test.safeParse({}) as z.SafeParseError<void>;
+  const parsed = z.safeParse(Test, {}) as z.SafeParseError<void>;
   const validFlattenedErrors: TestFlattenedErrors = parsed.error.flatten(
     () => ({ message: "", code: 0 })
   );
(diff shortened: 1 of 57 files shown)


colinhacks:

PR #3810, opened 2024-10-19
fix record type with branded keys

netanelgilad:
A fix for record types with branded keys being inferred as `Partial`s.
Should fix the case described in https://github.com/colinhacks/zod/discussions/2069

From what I understood the ternary in `RecordType` verifies that the key is not a literal or an enum, by checking that any string / number / symbol is assignable to `K` and so it's not a literal.
I extended the ternary to take into account the cases where `K` is branded, and for that I introduced `AnyBrand` that will allow me to test for any string / number / symbol _with any brand_ is assignable to `K`.


colinhacks:

PR #3817, opened 2024-10-24
Fix: Optional object properties are not equivalent to required properties with undefined values

MrPossumz:
The object input and output logic currently used in `z.object()` converts all keys with a potential value of undefined to optional even if they are required to be present on the object.

Current Behavior:
```ts
const obj = z.object({
  a: z.string(),
  b: z.union([z.string(), z.undefined()]),
});

obj._output; // { a: string; b?: string | undefined; }
```
This is different than the type determined by passing the object's type directly into `ZodType`, which **does** respect the difference between an absent key versus a key present with an undefined value.

Expected Behavior:
```ts
type ObjectWithKey = ZodType<{ a: string; b: number | undefined }>;
// ObjectWithKey output and input = { a: string; b: number | undefined }

type ObjectWithOptionalKey = ZodType<{ a: string; b?: number | undefined }>;
// ObjectWithKey output and input = { a: string; b?: number | undefined }
```

This PR addresses that by correctly differentiating the optional keys from required keys that may have a value of undefined. It also accounts for the difference in input/out for the zod default type and behavior.

All tests related to the code changes are passing but I did notice that a test for the Ip string type was failing on the `main` branch.


andersk:
Your modified types don’t match the behavior of Zod at runtime (which you haven’t changed):

```js
> obj.parse({ a: 'foo' })
{ a: 'foo' }
```

colinhacks:

PR #3819, opened 2024-10-26
add `isZodError` to avoid peer dependency problem

xieyuheng:
- with the help of `instanceofZodError = true` in `ZodError`
- example of peer dependency problem: https://github.com/colinhacks/zod/discussions/2415


xieyuheng:
nevermind, i will just use:

```typescript
import { ZodError } from "zod"

export function isZodError(error: unknown): error is ZodError {
  if (!(error instanceof Error)) return false

  if (error instanceof ZodError) return true
  if (error.constructor.name === "ZodError") return true
  if ("issues" in error && error.issues instanceof Array) return true

  return false
}
```

xieyuheng:
> You can simply use
> 
> ```ts
> import { ZodError } from 'zod';
> 
> export const isZodError = (value: unknown): value is ZodError => value instanceof ZodError;
> ```
> 
> @xieyuheng

no i can not, there is peer dependency problem.

TasseDeCafe:
I'm having this problem too inside my Turbo Repo monorepo. This check returns false even though the error is an instance of ZodError: "error instanceof ZodError".

xieyuheng:
@TasseDeCafe have you tried my workaround above?

TasseDeCafe:
@xieyuheng, I did, and it does work work, thank you.

I'm just not sure what the exact origin of this problem is. I understand that this is caused by having multiple instances of zod in my monorepo, but I'm not sure if I should be worried about that or not.

xieyuheng:
@TasseDeCafe it would be better if zod can provide a reliable API for testing `ZodError` class across different (future) versions of zod.
but do not worry too much, nothing is perfect.

clicktodev:
will this be revisited?

PR #3820, opened 2024-10-27
feat: z.string.cidr() - support CIDR notation

wataryooou:
This PR supported CIDR notation. (This PR doesn't support IP address range.)
FYI: https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing

```ts
const cidr = z.string().cidr();
cidr.parse("192.168.0.0/24"); // pass
cidr.parse("2001:db8::/32"); // pass
cidr.parse("192.168.1.1/33"); // fail
cidr.parse("2001:db8::1/129"); // fail
```


colinhacks:
Is there are argument for making this `z.string().cidr()`?

wataryooou:
@colinhacks 

> Is there are argument for making this `z.string().cidr()`?

That's a great idea 👍 
I've made the changes accordingly. Could you review it once more?

gdostie (approved):

PR #3822, opened 2024-10-28
fix: bigint coerce crash

### LINKED ISSUE #3821: Bigint with coerce throws exception in safeParse and safeParseAsync
Hi,

I noticed a bug while using safeParse and safeParseAsync. Those methods IMO shouldn't throw any errors. Instead, they should hold error in returned object. However, currently this does not work in case of coerced big integers.

Here's an example snippet:
```
const { z } = require("zod");
const schema = z.object({
  data: z.coerce.bigint(),
});
const example = { data: undefined };
try {
  const result = schema.safeParse(example);
  console.log("RESULT ok?: ", result.success);
} catch (err) {
  console.error("Catched: ", err);
}
```

Excepted output
`RESULT ok?: false`

**Actual output**

```
Catched:  TypeError: Cannot convert undefined to a BigInt
    at BigInt (<anonymous>)
    at ZodBigInt._parse (/home/aleksander/projects/zod-bigint-error/node_modules/zod/lib/types.js:1205:26)
    at ZodObject._parse (/home/aleksander/projects/zod-bigint-error/node_modules/zod/lib/types.js:1864:37)
    at ZodObject._parseSync (/home/aleksander/projects/zod-bigint-error/node_modules/zod/lib/types.js:146:29)
    at ZodObject.safeParse (/home/aleksander/projects/zod-bigint-error/node_modules/zod/lib/types.js:176:29)
    at Object.<anonymous> (/home/aleksander/projects/zod-bigint-error/index.js:10:25)
    at Module._compile (node:internal/modules/cjs/loader:1469:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1548:10)
    at Module.load (node:internal/modules/cjs/loader:1288:32)
    at Module._load (node:internal/modules/cjs/loader:1104:12)
```

kodemon:
Discovered when parse is executed through `isOptional` check on `ZodBigInt` which adds a `undefined` safeParse execution to look for errors, it crashes when attempting to add `undefined` to BigInt as its an invalid value.

My suggested fix here is to just try catch the instantiation and cast the coerced value to `undefined` if it throws and error. This way all parser checks will fail on invalid values but not crash on internal checks which provides undefined as the input value.

closed #3821

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -1549,7 +1549,11 @@ export interface ZodBigIntDef extends ZodTypeDef {
 export class ZodBigInt extends ZodType<bigint, ZodBigIntDef, bigint> {
   _parse(input: ParseInput): ParseReturnType<bigint> {
     if (this._def.coerce) {
-      input.data = BigInt(input.data);
+      try {
+        input.data = BigInt(input.data);
+      } catch {
+        input.data = undefined;
+      }
     }
     const parsedType = this._getType(input);
     if (parsedType !== ZodParsedType.bigint) {
--- a/src/types.ts
+++ b/src/types.ts
@@ -1549,7 +1549,11 @@ export interface ZodBigIntDef extends ZodTypeDef {
 export class ZodBigInt extends ZodType<bigint, ZodBigIntDef, bigint> {
   _parse(input: ParseInput): ParseReturnType<bigint> {
     if (this._def.coerce) {
-      input.data = BigInt(input.data);
+      try {
+        input.data = BigInt(input.data);
+      } catch {
+        input.data = undefined;
+      }
     }
     const parsedType = this._getType(input);
     if (parsedType !== ZodParsedType.bigint) {


deno/lib/types.ts
```diff
@@ -1549,7 +1549,11 @@ export interface ZodBigIntDef extends ZodTypeDef {
 export class ZodBigInt extends ZodType<bigint, ZodBigIntDef, bigint> {
   _parse(input: ParseInput): ParseReturnType<bigint> {
     if (this._def.coerce) {
-      input.data = BigInt(input.data);
+      try {
+        input.data = BigInt(input.data);
+      } catch {
+        input.data = undefined;
```
hdsuperman, line 1555: Maybe here should return INVALID instate of `undefined`

src/types.ts
```diff
@@ -1549,7 +1549,11 @@ export interface ZodBigIntDef extends ZodTypeDef {
 export class ZodBigInt extends ZodType<bigint, ZodBigIntDef, bigint> {
   _parse(input: ParseInput): ParseReturnType<bigint> {
     if (this._def.coerce) {
-      input.data = BigInt(input.data);
+      try {
+        input.data = BigInt(input.data);
+      } catch {
+        input.data = undefined;
```
hdsuperman, line 1555: Maybe here should return INVALID instate of undefined

kodemon: That works and looks more correct 👍🏻

colinhacks:
Love it, thanks!

PR #3841, opened 2024-11-07
fix(types): fix type cache crash

odbol:
There's some kind of weird race condition or something where the strict null check fails but then returns an undefined object. 

See https://github.com/firebase/genkit/issues/972 for details on how to reproduce.

```
  name: 'TypeError',
  message: "Cannot destructure property 'shape' of 'this._getCached(...)' as it is undefined.",
  stack: "TypeError: Cannot destructure property 'shape' of 'this._getCached(...)' as it is undefined.\n" +
    '    at ZodObject._parse (/Users/fuego/Documents/development/gemineye-server/node_modules/zod/lib/types.js:1848:17)\n' +
    '    at ZodObject._parseSync (/Users/fuego/Documents/development/gemineye-server/node_modules/zod/lib/types.js:146:29)\n' +
    '    at /Users/fuego/Documents/development/gemineye-server/node_modules/zod/lib/types.js:1711:29\n' +
    '    at Array.map (<anonymous>)\n' +
    '    at ZodArray._parse (/Users/fuego/Documents/development/gemineye-server/node_modules/zod/lib/types.js:1710:38)\n' +
    '    at ZodObject._parse (/Users/fuego/Documents/development/gemineye-server/node_modules/zod/lib/types.js:1864:37)\n' +
    '    at ZodObject._parseSync (/Users/fuego/Documents/development/gemineye-server/node_modules/zod/lib/types.js:146:29)\n' +
    '    at ZodObject.safeParse (/Users/fuego/Documents/development/gemineye-server/node_modules/zod/lib/types.js:176:29)\n' +
    '    at ZodObject.parse (/Users/fuego/Documents/development/gemineye-server/node_modules/zod/lib/types.js:157:29)\n' +
    '    at /Users/fuego/Documents/development/gemineye-server/node_modules/@genkit-ai/flow/lib/flow.js:522:55',
```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -2421,7 +2421,7 @@ export class ZodObject<
   private _cached: { shape: T; keys: string[] } | null = null;
 
   _getCached(): { shape: T; keys: string[] } {
-    if (this._cached !== null) return this._cached;
+    if (this._cached) return this._cached;
     const shape = this._def.shape();
     const keys = util.objectKeys(shape);
     return (this._cached = { shape, keys });


src/types.ts
```diff
@@ -2421,7 +2421,7 @@ export class ZodObject<
   private _cached: { shape: T; keys: string[] } | null = null;
 
   _getCached(): { shape: T; keys: string[] } {
-    if (this._cached !== null) return this._cached;
+    if (this._cached) return this._cached;
```
m10rten, line 2424: As you can see the type above is {shape:T;keys:string[]} or null, with a default to null.

So this questions as to why the null check is not correct?

odbol: Yeah it's definitely weird. Seems like it shouldn't happen, at least according to Typescript. Is it possible that somehow shape and keys could end up undefined? Would that cause _cached to be undefined?

I'm not sure how this is happening, if it's some kind of node.js interpreter error or race condition, or some weird edge case that Genkit just ran across... but this patch does fix it at least, and seems still correct. What do you think we should do?

odbol:
Hey @m10rten, any thoughts on next steps here? This bug is blocking our launch, and I'm not sure how else to fix it...

m10rten:
> Hey @m10rten, any thoughts on next steps here? This bug is blocking our launch, and I'm not sure how else to fix it...

Hi, suggest to use a type cast for the time being as when this gets merged you wont be able to use it till its released.

odbol:
> > Hey @m10rten, any thoughts on next steps here? This bug is blocking our launch, and I'm not sure how else to fix it...
> 
> Hi, suggest to use a type cast for the time being as when this gets merged you wont be able to use it till its released.

hmmm... I'm not sure how to use a typecast in this situation. The use of Zod is pretty heavily buried into the Genkit library I'm using; I don't think their API lets us get around that. 

But are you're saying that this fix does have a chance of getting merged? Any ETA on when the next release will be?

colinhacks:

PR #3859, opened 2024-11-19
Simplifying installation guide now that Deno supports installing from npm

philhawksworth:
Since Deno now supports installing directly from npm, no differentiation is required between npm/bun and deno. We can streamline the install guide.

This PR updates the readme to:
- remove the dedicated section on importing from deno.land as this is no longer required (in English and Simplified Chinese files)
- add `deno add` example (in English and Simplified Chinese files)
- add `deno add` example for canary

This update has been tested against the provided basic usage examples to validate that it works as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,7 @@
     - [Utilities for Zod](#utilities-for-zod)
 - [Installation](#installation)
   - [Requirements](#requirements)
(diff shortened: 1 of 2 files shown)


philhawksworth:
The deploy preview looks good to me. Do you think I could get a review on this update to the docs, @colinhacks ?

colinhacks:
Thanks Phil!

Issue #3860, opened 2024-11-19
RFC/Feature Request: Add a type-safe convert/input output mapper method for compile time input validation


oberbeck:
**RFC/Feature Request**: Add a type-safe `convert`/`input output mapper` method for compile time input validation

**Summary**

Introduce a new method, ie. `convert`, for Zod schemas that provides type-safe input validation and transformation. This method would enforce that the provided input matches the schema’s defined input type (`T["_input"]`) at compile time, ensuring a stricter contract compared to `parse`.

**Motivation**

Zod’s `parse` and `safeParse` methods currently accept `unknown` as input. While this is intentional for unstructured or untrusted data validation, there are common use cases where input data is already structured (e.g., from a typed API response, database query, or other controlled sources). For such cases, allowing type-safe input validation without relying on runtime type checking enhances both safety and developer experience.

The proposed method aligns with Zod’s TypeScript-first philosophy by:

1. Enforcing input types at compile time, reducing reliance on runtime checks.
2. Providing a clear distinction between structured and unstructured input handling.
3. Supporting scenarios where transformations (via `.transform()`) are applied, while maintaining type safety.

**Proposed API**

Add a method `convert` to Zod schemas with the following behavior:

```typescript
schema.convert(input: T["_input"]): T["_output"];
```

- **Input Type (`T["_input"]`)**: Enforces that the input matches the schema's input type at compile time.
- **Output Type (`T["_output"]`)**: Returns the validated and transformed data, adhering to the schema's transformation logic.

  **Example Usage**

https://github.com/colinhacks/zod/blob/dad88200ec45b673d33eae442dff40d805d8add9/src/__tests__/convert.test-d.ts#L21-L59

**Comparison with Existing Methods**

1. **`parse`**:

   - Accepts `unknown` input and validates it.
   - Use case: General-purpose validation of unstructured data.

2. **`convert`** (Proposed):
   - Enforces that the input matches the schema’s input type at compile time.
   - Use case:
     - Validating already-structured data with guaranteed input types, while benefiting from transformations and Zod’s validation.
     - Using zod as typesafe converter ie. between well defined input and output types.

**Potential Concerns**

- **Confusion with Existing Methods**:
  - I am uncertain about how to name the method in order to clearly differentiate itself from `parse`, `safeParse`, `transform`,... to avoid ambiguity. Suggested name: `convert`, `inputToOutput`, `parseConvert`?

**Example implementation**

https://github.com/colinhacks/zod/pull/3861

[...]

PR #3861, opened 2024-11-19
Feat/add convert

oberbeck:
Example implementation which fixes https://github.com/colinhacks/zod/issues/3860


colinhacks:
Went a different way with this, but thanks for the PR/concept! This is now achievable with the new `.decode()` methods in Zod 4.1, which were added in conjunction with the Codec API: https://colinhacks.com/essays/introducing-zod-codecs

PR #3866, opened 2024-11-21
feat(v3): Added file type

obha:
This new type initially checks and validates a file object against size and MIME type.


deno/lib/README.md
```diff
@@ -1925,6 +1926,34 @@ TestSchema.parse(new Test()); // passes
...
+const imageFile = z.file({
+  required_error: "file is required",
+  invalid_type_error: "This object must be a file",
+});
+```
+
+You can validate the file's minimum and maximum size.
+
+```ts
+const imageFile = z.file().size({ min: 100000, max: 200000 });
```
m10rten: Would be good to specify in this documentation what this size is, bytes most likely.

obha: Correct, it's in bytes

colinhacks:

PR #3880, opened 2024-12-04
Avoid stackoverflows and memory issues in zod while processing large strings

klh:
This commit will:

* Add a generalized helper for chunkProcessing in parseUtil
* Utilize it for base64 string verification

NOTES: there are sure to be other cases when handling large strings that can benefit from this, but i think maintainers should decide where and when.


klh:
bump on this?

PR #3881, opened 2024-12-04
feat: Add support for ISO 4217 Currencies

volcanoxp:
https://es.wikipedia.org/wiki/ISO_4217

Add support for ISO 4217 Currencies


ykhedher:
Why was this closed ?

PR #3910, opened 2024-12-13
Update type in test for upcoming intersection equality behavior change

MichaelMitchell-at:
If https://github.com/microsoft/TypeScript/pull/60726 is merged, then the type `{ name: string } & { [z.BRAND]: { superschema: true } }` will become identical to `{ name: string ; [z.BRAND]: { superschema: true } }`, which is not identical to `{ name: string } & z.BRAND<'superschema'>` because ` z.BRAND<'superschema'>` cannot be "trivially merged" into the object literal type.


colinhacks:

PR #3914, opened 2024-12-17
fix: string().email() does not validate successfully where emails have Ampersand (&) in them #3913

### LINKED ISSUE #3913: z.string().email() does not validate successfully where emails have Ampersand (&) in them
z.string().email() fails validation when an email with an Ampersand (&) is used.

Example:
o&leary@example.com

This is allowed in RFFC 5332 as well.

SunilChackoPS:
Hello Team,
                 Thank you for the wonderful library. I am  a first timer for contribution to zod.
                  The email address validation changed with following criteria.
                   
A) Where & Can Appear (Local Part Only)
B) The local part (before @) may include & anywhere, provided it is not:
     a) At the start of the local part.
     b) the end of the local part.
     c) Consecutive with itself (&&).
This means & can appear in the middle of the local part multiple times, as long as the other constraints are respected.

The test case files are updated along with 'types.ts'
image

Please find the attached cmd output while executing yarn play after the change.

It fixes #3913


andersk:
> provided it is not:
> a) At the start of the local part.
> b) the end of the local part.
> c) Consecutive with itself (&&).

What’s the justification for these restrictions? I don’t see such restrictions in RFC 5322.

SunilChackoPS:
> > provided it is not:
> > a) At the start of the local part.
> > b) the end of the local part.
> > c) Consecutive with itself (&&).
> 
> What’s the justification for these restrictions? I don’t see such restrictions in RFC 5322.

Basically, based on practical implementation considerations and user experience concerns.
some email processing tools  misinterpret these as separators or special characters, leading to unexpected behavior.(Allowing & at the start or end of the local part)
Yes the RFC allows multiple special characters in sequence, in practice, email addresses with consecutive '&' are rare (could easily be misinterpreted or mishandled by downstream systems)
Also, this is  balancing RFC flexibility with real-world usability

Still, If the maintainer believes these restrictions are unnecessary or overly conservative, I’m happy to revisit and refine them based on additional guidance!

BigBrainAFK:
> A) Where & Can Appear (Local Part Only)

This is not part of RFC5322. The RFC allows the dot-atom token (which later resolves to allow ampersand) in both local and domain.

colinhacks:

PR #3921, opened 2024-12-20
Add `zod-struct` to utilities for Zod

reesericci:
`zod-struct` is a utility to create simple structs using Zod schemas.


StefanTerdell:
Heh, neat! Looks alot like a prototype I made about a year go (maybe there's something in it you'll find useful): https://github.com/stefanterdell/zonk

glhf

PR #3926, opened 2024-12-27
fix: replace es2019 class fields with local members to address #3611

### LINKED ISSUE #3611: Enum parsing error: TypeError: Cannot read private member from an object whose class did not declare it (codesandbox repro included)
Reproduction: https://codesandbox.io/p/sandbox/zod-test-d6y6ts

I recently upgraded to 3.23.x from 3.22.4 and starting getting this error: 
```
TypeError: Cannot read private member from an object whose class did not declare it.
``` 

After some digging, I narrowed it down to ZodEnums parsing under specific circumstances. It happens when a ZodEnum has been deep cloned (in this case with lodash) and is then added as value of a ZodObject.

See the reproduction for details and some cases that do and don't work.  It's the parsing on the last line that causes the error. Comment it out, and the error goes away.

Edit the package.json to 3.22.4 and the error also disappears. 

I think this might be a related issue: https://github.com/colinhacks/zod/issues/3520

Thanks!

alon-gb:
Fixes #3611

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -4342,14 +4342,19 @@ function createZodEnum(
   });
 }
 
+const lookupSymbol = Symbol("lookup");
 export class ZodEnum<T extends [string, ...string[]]> extends ZodType<
   T[number],
   ZodEnumDef<T>,
   T[number]
 > {
-  #cache: Set<T[number]> | undefined;
-
+  private [lookupSymbol]: Set<T[number]> | undefined;
   _parse(input: ParseInput): ParseReturnType<this["_output"]> {
+    let lookup = this[lookupSymbol];
+    if (!lookup) {
+      lookup = new Set(this._def.values);
+      this[lookupSymbol] = lookup;
+    }
     if (typeof input.data !== "string") {
       const ctx = this._getOrReturnCtx(input);
       const expectedValues = this._def.values;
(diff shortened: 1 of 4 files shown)


deno/lib/types.ts
```diff
@@ -4342,14 +4342,20 @@ function createZodEnum(
...
   ZodEnumDef<T>,
   T[number]
 > {
-  #cache: Set<T[number]> | undefined;
-
+  private [lookupSymbol]: Set<T[number]> | undefined;
   _parse(input: ParseInput): ParseReturnType<this["_output"]> {
+    let lookup = this[lookupSymbol];
+    if (!lookup) {
+      console.log("setting lookup");
```
nadavye, line 4355: Remove the consolle.

src/types.ts
```diff
@@ -4342,14 +4342,20 @@ function createZodEnum(
...
   ZodEnumDef<T>,
   T[number]
 > {
-  #cache: Set<T[number]> | undefined;
-
+  private [lookupSymbol]: Set<T[number]> | undefined;
   _parse(input: ParseInput): ParseReturnType<this["_output"]> {
+    let lookup = this[lookupSymbol];
+    if (!lookup) {
+      console.log("setting lookup");
```
nadavye, line 4355: Remove the console.

nadavye (approved):
Please see my comments

src/types.ts
alon-gb: Removed. Thanks.

deno/lib/types.ts
alon-gb: Removed. Thanks.

PR #3999, opened 2025-02-20
Add `.extend` to Zod enums

jeremy-code:
- Add `.extend` method to Enum class in Zod to allow extending enums with additional values

---

Essentially, adds functionality equivalent to the following.

```ts
const SalmonAndTrout = z.enum(["Salmon", "Trout"]);
const FishEnum = z.enum([...SalmonAndTrout.options, "Tuna"]);

const SalmonAndTrout = z.enum(["Salmon", "Trout"]);
const FishEnum = SalmonAndTrout.extend(["Tuna"]);
```

I think it would be useful to have alongside `.extract` and `.exclude`. The behavior is also equivalent to `z.object`'s `.extend`. I couldn't find any issues on this or similar, hence the PR. Let me know what you think.

## Summary by CodeRabbit

- **New Features**
  - Introduced a method to extend existing enum definitions with additional values, enhancing flexibility in creating comprehensive enum types.

- **Tests**
  - Added test cases to ensure the new enum extension functionality operates correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1167,6 +1167,15 @@ You can also retrieve the list of options as a tuple with the `.options` propert
 FishEnum.options; // ["Salmon", "Tuna", "Trout"];
 ```
 
+** `.extend()` **
+
+You can create supersets of a Zod enum with the `.extend` method.
+
+```ts
+const SalmonAndTrout = z.enum(["Salmon", "Trout"]);
+const FishEnum = SalmonAndTrout.extend(["Tuna"]);
+```
+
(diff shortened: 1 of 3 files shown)


jeremy-code:
Sorry! Just renamed `.extends` to `.extend`. I think I mixed up the JS/TS keyword in my head and thought the .object method was `.extends` but it is actually `.extend`, which aligns with `.extract` and `.exclude` better.

colinhacks:

Issue #3954, opened 2025-01-22
Add support for custom nanoid sizes


Jodenee:
## Problem
Nanoids can vary in length, while they are mainly 21 characters in length some may be larger. For example the npm package 'nanoid' allows developers to generate ids at any length so it would be nice to have that built into zod.

## Ideal solution

```ts
const goodNanoid = "MJBTltG6zIHelSq42cz51j_lESP7YSD_lLoOG4y_hsikIfCI403gLw6oJnreQW7u"
const badNanoid = "xv02pdYYi8Tb1SF1WjfSe" // bad because the nanoid's length is expected to be 64 characters
const schema = z.string().nanoid(64)

const goodNanoidResult = await schema.safeParseAsync(goodNanoid)
const badNanoidResult = await schema.safeParseAsync(badNanoid)

console.log(goodNanoidResult.success) // should be true
console.log(badNanoidResult.success) // should be false
```

## Work around

```ts
const goodNanoid = "MJBTltG6zIHelSq42cz51j_lESP7YSD_lLoOG4y_hsikIfCI403gLw6oJnreQW7u"
const badNanoid = "xv02pdYYi8Tb1SF1WjfSe" // bad because the nanoid's length is expected to be 64 characters
// regex taken from https://github.com/colinhacks/zod/blob/f7ad26147ba291cb3fb257545972a8e00e767470/src/types.ts#L643C21-L643C40
const schema = z.string().regex(/^[a-z0-9_-]{64}$/i) // if you wish to check for a higher or lower size nanoid just change the 64 with the size of your choice

const goodNanoidResult = await schema.safeParseAsync(goodNanoid)
const badNanoidResult = await schema.safeParseAsync(badNanoid)

console.log(goodNanoidResult.success) // true
console.log(badNanoidResult.success) // false
```

If anymore information is needed please let me know.

colinhacks:
I like this and am definitely open to a PR

j0hnm4r5:
Nanoids can also have a [custom alphabet](https://github.com/ai/nanoid?tab=readme-ov-file#custom-alphabet-or-size).

shi-rudo:
> * Please let me know if this feature is still relevant for your use case with the latest version of Zod by commenting here.

For my part, I can say that this is still relevant.

colinhacks:
Landed in #4004 — `z.nanoid({ length: 64 })`, and the same params object works on `zod/mini`. Will go out in the next release.

[...]

PR #4004, opened 2025-02-22
Add Allow arbitrary character length in nanoid

oimo23:
I hope you don't mind, but I've gone ahead and implemented it.
https://github.com/colinhacks/zod/issues/3954.

## Changes Introduced:
・Allow arbitrary character length in nanoid.
・Since the existing implementation assumed the first argument to nanoid would be a custom error message, I made sure not to break that behavior.
・I added an interface for arguments such as min, max, and length.
```
// Both approaches work

// Added in this PR
nanoid(64, { message: "Must be 64 characters long" });

// Current Version
nanoid("Custom Error");
```

## Issue Reference:
Closes https://github.com/colinhacks/zod/issues/3954.

## Summary by CodeRabbit

- **New Features**
	- Enhanced identifier validation to allow users to specify a custom length for generated IDs.
	- Improved error feedback by providing tailored messages when an identifier does not meet the expected length.

- **Tests**
	- Added a new test scenario to ensure the custom-length identifier functionality and error handling work as intended.


Kang-Jh:
Any update?

deno/lib/types.ts
```diff
@@ -1114,8 +1119,30 @@ export class ZodString extends ZodType<string, ZodStringDef, string> {
   uuid(message?: errorUtil.ErrMessage) {
     return this._addCheck({ kind: "uuid", ...errorUtil.errToObj(message) });
   }
-  nanoid(message?: errorUtil.ErrMessage) {
-    return this._addCheck({ kind: "nanoid", ...errorUtil.errToObj(message) });
+  nanoid(
+    arg1?: number | string | { length?: number; message?: string },
```
johnjensenish: Accepting an object may be a good idea, but it's not implemented it so let's remove the type
```suggestion
    arg1?: number | string
```

src/types.ts
```diff
@@ -640,7 +640,11 @@ const ulidRegex = /^[0-9A-HJKMNP-TV-Z]{26}$/i;
 //   /^([a-f0-9]{8}-[a-f0-9]{4}-[1-5][a-f0-9]{3}-[a-f0-9]{4}-[a-f0-9]{12}|00000000-0000-0000-0000-000000000000)$/i;
 const uuidRegex =
   /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/i;
-const nanoidRegex = /^[a-z0-9_-]{21}$/i;
+
+function nanoidRegex(length?: number): RegExp {
```
johnjensenish: `check.value` has type `number` so you don't need to handle undefined.
```suggestion
function nanoidRegex(length: number): RegExp {
```

johnjensenish (changes requested):
Thanks for the PR @oimo23 

I wanted to provide some suggestions that would resolve the feedback given by the bot. I'm hoping to move this feature along.

johnjensenish:
I think if we adopt the changes I suggest then this PR will resolve #3954

oimo23:
@johnjensenish 
Thank you. I've implemented the suggested changes. 

It took a few extra tweaks of my own to get all the tests passing, but we're green now. Back to you for review.

oimo23:
@johnjensenish 
Hi, Please take a look if you have time.

johnjensenish (approved):
@oimo23 These changes now look satisfactory to me. I should clarify that I'm not a maintainer, just looking to help.

johnjensenish:
@colinhacks Think this is ready for you to take a look at.

colinhacks:
Picking this up. I've reworked the implementation — the length rides on the params object (`z.nanoid({ length: 64 })`) rather than a positional argument, matching how `z.uuid({ version })` already works and reaching `z.nanoid()` and `zod/mini` instead of only the deprecated `z.string().nanoid()`. The default keeps its shared regex so an ordinary `z.nanoid()` costs nothing extra, and a length that isn't a positive integer now throws instead of quietly building a regex that matches literal braces.

Dropped the Zod 3 half — v3 is in maintenance mode and takes security and correctness fixes only. Merging this once CI is green.

PR #4030, opened 2025-03-16
Fix union type inference in object properties

Kunal-Darekar:
#fixes : #2654 
/claim #2954 
This PR fixes an issue where union types were being inferred differently depending on whether they were used directly or as part of an object property.
Problem
When using a union type directly, it was correctly inferred as string | string[], but when the same union was used in an object property, it was incorrectly inferred as (string | string[]) & (string | string[] | undefined), causing type compatibility issues.
Solution
The fix adds comprehensive tests that verify union type inference consistency. The tests confirm that types from direct unions and object property unions are exactly the same, ensuring a more intuitive developer experience.
Testing
Added comprehensive tests in src/__tests__/union-inference.test.ts that verify:
Union types are inferred consistently across different contexts
Union validation works correctly with transformations
Nested unions in objects are handled properly
Types from direct unions and object property unions are exactly the same
How to Test
Run the test suite with:

```
yarn test src/__tests__/union-inference.test.ts


```
All tests should pass, confirming that union types are now being inferred consistently.
.

## Summary by CodeRabbit

- **Chores**
  - Created a new configuration file `.coderabbit.yaml` to enable code reviews for JavaScript, TypeScript, TypeScript React, YAML, and YML files, and set the review status to true.


colinhacks:

PR #4054, opened 2025-03-27
feat: `z.envbool()` - coerce string inputs to boolean

### LINKED ISSUE #3906: RFC: `z.envbool()`
An customizable, first-class way to coerce string inputs to `boolean`, inspired by common environment variable conventions. 

```
z.envbool()
```

Default behavior:

- Returns `true` for `["true", "1", "yes", "on"]` (case insensitive)
- Returns `false` for `["false", "0", "no", "off"]` (case insensitive)
- Throws otherwise

The choice to throw on unrecognized values is intentional. This API will be used for parsing potentially sensitive environment variables, so this API intentionally avoids any non-explicit defaults (e.g. defaulting to `false`) to avoid potentially unexpected behavior.

To specify a value to be returned when an uexpected input is encountered, users can use the pre-existing `.catch()` method.

```ts
z.envbool().catch(false)
```

To customize the set of truthy and falsy strings:

```ts
z.envbool({
  true: ["yes", "on", "enabled", "y"],
  false: ["no", "off", "disabled", "n"]
})
```

To customize the set of truthy and falsy strings:

```ts
z.envbool({
  true: ["yes", "on", "enabled", "y"],
  false: ["no", "off", "disabled", "n"]
})
```

To make it case sensitive:

```ts
z.envbool({
  case: "sensistive" // default "insensitive"
})
```

Jay-Karia:
## **`z.envbool()`**

Closes #3906

### Usage

Using default truthy and falsy values
```ts
const envbool = z.envbool();
envbool.parse("1") // true
envbool.parse("on") // true
envbool.parse("ON") // true

envbool.parse("0") // false
envbool.parse("off") // false
envbool.parse("OFF") // false

envbool.parse("foo") // throws Error
```

Custom values
```ts
const envbool = z.envbool({
  true: ["always"],
  false: ["never"],
  case: "sensitive"
})

envbool.parse("always") // true
envbool.parse("on") // true

envbool.parse("never") // false
envbool.parse("off") // false

envbool.parse("Always") // throws Error
envbool.parse("Off") // throws Error
```

### Additional functions

```ts
const envbool = z.envbool();

envbool._truthyValues; // Default truthy values
envbool._falsyValues; // Default falsy values
envbool._caseSensitivity; // Default case sensitivity
```

### Default Values

```ts
truthyValues = ["true",  "1",  "on",  "yes",  "y",  "enabled"];
falsyValues = ["false",  "0",  "off", "no",  "n",  "disabled"];
caseSensitivity = "insensitive"
```

## Summary by CodeRabbit

- **New Features**
  - Added an environment boolean validation option that allows the parsing of boolean-like string values with both default and customizable configurations, including control over case sensitivity.

- **Tests**
  - Introduced comprehensive tests to ensure the new environment boolean validation behaves correctly under various configurations and accurately distinguishes valid from invalid boolean representations.


Jay-Karia:
@colinhacks, Do I need to fix the Vercel CI, before pushing ?

colinhacks:
This feature has been implemented in the Zod 4 beta, and I'm not planning to backport it to Zod 3 for compatibility reasons I'm afraid. But thanks for the PR, it was really well done.

PR #4065, opened 2025-04-02
Add runtypes to zod codemod

tadhglewis:
https://github.com/tadhglewis/runzod

**_Very_ hacked together and needs additional testing before I recommend others to use.** Drafting this to gauge interest.

We extensively use runtypes across hundreds of repos and we're looking at migrating to zod, so we should be able to test this internally quite well.


colinhacks:
The README has been cut down significantly in favor of the new docs site. The ecosystem page has moved here: https://github.com/colinhacks/zod/blob/v4/packages/docs/content/ecosystem.mdx

If you finalize work on this, feel free to open a new PR against that file :)

PR #4068, opened 2025-04-03
Fixed missing CIDR suffix in CIDR version example in README.md.

wrnotc:
## Summary by CodeRabbit

- **Documentation**
  - Updated usage examples to clarify the correct CIDR notation by including subnet masks for both IPv4 and IPv6 validations.


colinhacks:
These docs have been reworked since this PR. But thanks!

Issue #4089, opened 2025-04-11
v4: toJSONSchema union oneOf issue


samchungy:
Howdy Colin, loving exploring through v4 so far. I was comparing some of the output of `.toJSONSchema` with my own library https://github.com/samchungy/zod-openapi so I can delete some of my generation logic.

I love how you've baked in component registry into Zod. 🙌 I can delete so much of that logic which I was doing. Thank you.

Feedback:

1. `z.union()` currently generates a [`oneOf`](https://json-schema.org/understanding-json-schema/reference/combining#oneOf) schema.

This is technically incorrect as Zod proceeds as soon as a single schema matches so it should be [`anyOf`](https://json-schema.org/understanding-json-schema/reference/combining#anyOf). Given we now have a `discriminatedUnion` method which no longer requires a discriminator, should we map `.union()` to `anyOf` and `.discriminatedUnion` to `oneOf`? Or should we add something to `.union()` which does a similar only one check?

2. `.optional()` generates `null` which is incorrect

```ts
const foo = z.string().optional();

const object = z.object({ a: foo });

console.log(
  JSON.stringify(
    z.toJSONSchema(object, {
      reused: "ref",
    }),
    null,
    2
  )
);
```

```json
{
  "type": "object",
  "properties": {
    "a": {
      "oneOf": [
        {
          "type": "string"
        },
        {
          "type": "null"
        }
      ]
    }
  },
  "required": []
}
```

Would you be open to me raising some PRs against v4 to fix some behaviours as I encounter them?

DASPRiD:
I agree with the `anyOf` there, see also the [Swagger documentation](https://swagger.io/docs/specification/v3_0/data-models/oneof-anyof-allof-not/#anyof).

Scrolling down a little bit to the "Difference Between anyOf and oneOf" section shows why `anyOf` is the correct way here.

colinhacks:
Both fixed in the latest version of `@zod/core` and associated betas

hc-12:
I'm using the latest version (v4.0.5) and noticed that `discriminatedUnion` doesn't appear to implement `oneOf`. Has this feature been added yet?

Teablack:
Fix was released in 4.1.13

PR #4090, opened 2025-04-11
`.toJSONSchema()` Fix optional behaviour

samchungy:
https://github.com/colinhacks/zod/issues/4089

`.optional()` and `.nonoptional()` shouldn't actually change our type as optionality and nullability are different things.

I've changed it to just pipe through our inner type.

### DIFF
--- a/packages/core/src/to-json-schema.ts
+++ b/packages/core/src/to-json-schema.ts
@@ -408,8 +408,7 @@ export class JSONSchemaGenerator {
       }
       case "optional": {
         const inner = this.process(def.innerType, params);
-        const json: JSONSchema.BaseSchema = _json as any;
-        json!.oneOf = [inner, { type: "null" }];
+        Object.assign(_json, inner);
         break;
       }
       case "nullable": {
@@ -420,7 +419,6 @@ export class JSONSchemaGenerator {
       case "nonoptional": {
         const inner = this.process(def.innerType, params);
         Object.assign(_json, inner);
-        _json.not = { type: "null" };
         break;
       }
       case "success": {
--- a/packages/zod/tests/json-schema.test.ts
+++ b/packages/zod/tests/json-schema.test.ts
@@ -573,6 +573,7 @@ describe("toJSONSchema", () => {
     const schema = z.object({
       required: z.string(),
       optional: z.string().optional(),
+      nonoptional: z.string().optional().nonoptional(),
     });
 
     const result = toJSONSchema(schema);
@@ -580,15 +581,11 @@ describe("toJSONSchema", () => {
     expect(result).toMatchInlineSnapshot(`
       {
         "properties": {
+          "nonoptional": {
+            "type": "string",
+          },
           "optional": {
-            "oneOf": [
-              {
-                "type": "string",
-              },
-              {
-                "type": "null",
-              },
-            ],
+            "type": "string",
           },
           "required": {
             "type": "string",
@@ -596,6 +593,7 @@ describe("toJSONSchema", () => {
         },
         "required": [
           "required",
+          "nonoptional",
         ],
         "type": "object",
       }


colinhacks:
This was intentional because it feels like a more expected behavior to me. In general I think mapping `undefined` -> `null` is a more reasonable default behavior than just ignoring it. I'm open to being convinced otherwise. I'd also be happy supporting this change behind an `optional` param. Curious to hear others chime in here.

dswbx:
> This was intentional because it feels like a more expected behavior to me. In general I think mapping `undefined` -> `null` is a more reasonable default behavior than just ignoring it. I'm open to being convinced otherwise. I'd also be happy supporting this change behind an `optional` param. Curious to hear others chime in here.

While it's probably working most of the time, defining a property with `null` is technically still defined. The properties describe what type to expect, and dealing with `null` and `undefined` used to be tedious in JS environments. The produced JSON Schema becomes also harder to read, and if you e.g. rely on form generators (like I do), then you need to add special logic to handle these unnecessary cases :/

For this reason, I've also opened a PR (https://github.com/colinhacks/zod/pull/4124) which treats all "to be undefined" types (`optional`, `nullable` and `nonoptional`) the same, merging with inner and rely on `required` property + simplifying `additionalProperties` for strict objects/interfaces.

samchungy:
My users typically have OpenAPI spec generated from Zod types to represent how to call the API. They will then instead use the Zod Types to actually parse the request.

If the JSON Schema was suggesting that `null` could be accepted but the Zod Type wouldn't accept it I think users would get frustrated. I think that given we have `.nullable()` and `.nullish()` which actually accept `null` we should keep that `null` behaviour behind those modifiers?

mdoi2:
I prefer the behavior presented by samchungy.
I am also concerned that with the current behavior, the value according to the json schema and the result of the zod parsing are misaligned.

```ts
import * as z from 'zod';

const schema = z.object({
  value: z.string().optional(),
});

console.dir(z.toJSONSchema(schema), { depth: null });
// OUTPUT:
// {
//   type: 'object',
//   properties: {
//     value: { oneOf: [ { type: 'string' }, { type: 'null' } ] }
//   },
//   required: []
// }

console.log(schema.safeParse({ value: null }).success);
// OUTPUT: false

console.dir(z.toJSONSchema(schema.shape.value), { depth: null });
// OUTPUT: { oneOf: [ { type: 'string' }, { type: 'null' } ] }

console.log(schema.shape.value.safeParse({ value: null }).success);
// OUTPUT: false
```

mattpocock:
Here's the scenario I'm scared about:

1. Pass an instruction to a LLM (for structured outputs etc) via JSON schema, telling it it can return `null`.
2. The LLM returns `null`
3. The zod parsing fails because the LLM returned `null`

Is this a danger?

Either way, confusing `null` and `undefined` is never a good plan.

igalklebanov:
Hey 👋

Agree with OP, explicit and 1:1 > convenient.

I'd intuitively expect `optional()` (key optional) to remove the prop name from "required" props list of the object type.
I'd expect `nullable()` to havw the current behavior - `oneOf` with `null`.
I'd expect `nullish()` to be a combination of both.
I'd expect value optional to behave similarly to key optional as I'd expect it to be omitted when serialized.

muningis:
Gotta agree with @samchungy.

`{"oneOf": [{ type: "string" }, { "type": "null" }] }` allows `string | null`, however `z.string().optional()` will fail on `null` [0], that's an issue by itself, because JSON Schema and Zod Schema won't have 1:1 behaviour anymore. And as Matt mentioned, mix-matching `null` and `undefined` ain't good (I personally consider that borderline grey area). 


[0] [Playground link](https://zod-playground.vercel.app?appdata=N4IgzgxgFgpgtgQxALhALwHQHsBGArGCAFwApgAdAOwAJrKE4ZlrMwiAnAS0oHMSBKbAAcinLPQA2AgDRUAvvypUQ0kADcEEgK4wwKANohg9Rs0paJEuSAC6qtTHZgxlFCADMGAEwAWbyDkgA)

ghardin1314:
> Here's the scenario I'm scared about:
> 
> 1. Pass an instruction to a LLM (for structured outputs etc) via JSON schema, telling it it can return `null`.
> 2. The LLM returns `null`
> 3. The zod parsing fails because the LLM returned `null`
> 
> Is this a danger?
> 
> Either way, confusing `null` and `undefined` is never a good plan.

A part from the theory here (which I agree with), a practical issue in todays LLM world is that many models only support a partial JSONSchema spec. I have had many, many issues trying to generate structured outputs using a union type using `oneOf`. 

Obviously this may change in the future with better models but I think making `null` the default behavior here would cause serious headaches trying to adopt `v4` into ai workflows

PR #4091, opened 2025-04-11
`.toJSONSchema()`: Prefer `const` over `enum` for single literal

samchungy:
https://json-schema.org/understanding-json-schema/reference/const


colinhacks:
Thanks Sam!

PR #4107, opened 2025-04-11
Update benchmark instructions to include switching to the v4 branch

morinokami:
Added `git switch v4` to the benchmark instructions, as the `bench` command doesn’t work unless you're on the `v4` branch.


colinhacks:
TIL about `git switch`...guess I'm old. Thanks!

PR #4111, opened 2025-04-11
Fix improper function definition in migration guide

jacksonblankenship:
I noticed a function definition improperly using `const` in place of `function` when working through the migration guide.


colinhacks:
Thank you!!

PR #4117, opened 2025-04-12
fix(docs): enum typo

zirkelc:


colinhacks:
This is correct as-is actually but I see how it's confusing. The behavior is designed to align with these TypeScript built-ins:

https://www.typescriptlang.org/docs/handbook/utility-types.html#excludeuniontype-excludedmembers
https://www.typescriptlang.org/docs/handbook/utility-types.html#extracttype-union

zirkelc:
Sorry, should have checked more carefully. It's not a typo, the examples are mixed up. 

The first example is under the "exclude" heading but used the "extract" example. See this screenshot:

image

zirkelc:
I didn't check it carefully, it's not a typo but the examples are mixed up:

image

I now switched the code examples, I think that's correct now.

colinhacks:
Oh 😅 That's been wrong for a long time. Good eye!

PR #4119, opened 2025-04-12
Fix params option to work with regex method

chimame:
## Overview

While checking the operation of v4, I noticed that the second argument option of the regex method does not work.


colinhacks:
Whoa good catch! Thanks!

Issue #4120, opened 2025-04-12
v4: `toJSONSchema` `pattern` not included for `.regex` and misc


li-kai:
`.regex`, `.starts_with`, `.includes` and similar functions use a regex pattern underneath. When generating json schema, [`pattern` is not included if `format`](https://github.com/colinhacks/zod/blob/35edafea7986307c2342c691e4d15d17cd60ff6f/packages/core/src/to-json-schema.ts#L169-L173) is also present.

Expected:
```ts
expect(toJSONSchema(z.string().regex(/asdf/))).toMatchInlineSnapshot(`
  {
    "format": "regex",
    "pattern": "asdf",
    "type": "string",
  }
`);
```

Received:
```ts
expect(toJSONSchema(z.string().regex(/asdf/))).toMatchInlineSnapshot(`
  {
    "format": "regex",
    "type": "string",
  }
`);
```

There's a few solutions depending on how strict we abide to the specs (from least to most):
1. Only add `pattern` for `regex` format
1. Add both `format` and `pattern` regardless of `format`, supposedly [the best practice](https://www.learnjsonschema.com/2020-12/format-annotation/format/). (Implemented in #4121)
1. Add `pattern` and remove non-standard `format`s, which differ by json schema draft target.
1. Add `pattern`, remove non-spec compliant `format`s, which is almost everything since zod regexes are stricter than the ones in RFCs.

PR #4121, opened 2025-04-12
Add pattern to json schema if present

li-kai:
Addresses https://github.com/colinhacks/zod/issues/4120 by implementing approach 2: adding both `format` and `pattern` regardless of `format`, supposedly [the best practice](https://www.learnjsonschema.com/2020-12/format-annotation/format/). This does increase the schema payload size, which should be largely irrelevant for most purposes of json schema.

### DIFF
--- a/packages/core/src/to-json-schema.ts
+++ b/packages/core/src/to-json-schema.ts
@@ -168,7 +168,8 @@ export class JSONSchemaGenerator {
         // custom pattern overrides format
         if (format) {
           json.format = formatMap[format] ?? format;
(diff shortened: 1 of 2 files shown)


colinhacks:
Ah, yeah, this is great. The current behavior is a relic of an older implementation where I checked the format against a set of known JSON Schema formats (datetime, email, etc) and used only `format` in those cases. Then fell back to `pattern` for everything else.

Eventually I decided to always add `format` but forgot to refactor those if statements.

Great catch!

Issue #4123, opened 2025-04-12
v4: `toJSONSchema` should not use union for optional, and set false to additionalProperties


dswbx:
When defining optional and required properties, the generated JSON Schema should adhere to the spec, which says to use the `required` property. Marking a property as optional with a union allowing `null` is prone to errors since it could still be defined, but just null. 

Additionally, if you define a `strictObject` or `strictInterface`, the produced `additionalProperties` should simply use `false`. While `{ "not": {} }` leads to the same result, it's harder to read and unnecessary. 

```ts
const schema = z.strictObject({
   name: z.string().describe("The name of the person"),
   age: z.number().optional(),
});
```

Expected:
```json
{
  "type": "object",
  "properties": {
    "name": {
      "description": "The name of the person",
      "type": "string",
    },
    "age": {
      "type": "number",
    }
  },
  "required": [ "name" ],
  "additionalProperties": false
}
```

Actual:
```json
{
  "type": "object",
  "properties": {
    "name": {
      "description": "The name of the person",
      "type": "string",
    },
    "age": {
      "oneOf": [
        {
          "type": "number",
        }, {
          "type": "null",
        }
      ],
    },
  },
  "required": [ "name" ],
  "additionalProperties": {
    "not": {},
  },
}
```

colinhacks:
Both fixed in latest betas

PR #4124, opened 2025-04-12
v4: `.toJSONSchema`: Enforce spec-based optional logic (`required`) and simplify `additionalProperties` for strict objects

dswbx:
Addresses https://github.com/colinhacks/zod/issues/4123 by changing the following:
* skipping `optional`, `nullable` and `nonoptional` and rely on the `required` property instead to prevent union (`oneOf`) types
  * there is an admittedly ambiguous definition at [10.3.2.1](https://json-schema.org/draft/2020-12/json-schema-core#section-10.3.2.1), but it also says that properties that exist must validate against its schema. Since JSON Schema is environment independent, they didn't cover the JS aspect of it
  * while using the `required` property should be the preferred way, technically, using an array of types would also be allowed, e.g. `{ "type": ["string", "null"] }` [7.6.1](https://json-schema.org/draft/2020-12/json-schema-core#section-7.6.1)
  
* checking `object`'s `additionalProperties` for `never` behavior and applies `false` instead of `{ "not": {} }`
  * see [4.3.2](https://json-schema.org/draft/2020-12/json-schema-core#name-boolean-json-schemas) *"Using the boolean values ensures that the intent is clear to both human readers and implementations."*

### DIFF
--- a/packages/core/src/to-json-schema.ts
+++ b/packages/core/src/to-json-schema.ts
@@ -289,6 +289,11 @@ export class JSONSchemaGenerator {
             ...params,
             path: [...params.path, "additionalProperties"],
           });
+
+          // if "additionalProperties" is `{"not": {}}`, then it should be `false`
+          if (JSON.stringify(json.additionalProperties) === '{"not":{}}') {
+            json.additionalProperties = false;
+          }
         }
 
         break;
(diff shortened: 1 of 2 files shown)


colinhacks:
Thanks! The optionality stuff has been updated already in https://github.com/colinhacks/zod/pull/4090

The `JSON.stringify` approach in `additionalProperties` seems a little fragile to me, and I'm not sure the complexity is worth it, given that the resulting JSON Schemas are equivalent.

Appreciate the PR though!

PR #4127, opened 2025-04-12
Add Spanish locale (i18n)

ByteEVM:
Initial base for Spanish (`es`) locale — enhancing Zod v4 [Internationalization](https://v4.zod.dev/error-customization#internationalization)


colinhacks:
Beautiful, thanks! If you have any additional tweaks, be sure to land those in the next couple weeks before the first stable release Zod 4 - at that point these message formats will be locked. Thanks again!

PR #4137, opened 2025-04-13
fix (v4): check for null prototype in parsedType function

Bram-dc:
When passing a Object.create(null) object (a " null -prototype object") the parseType function errors because it cannot read the name from the constructor.

Fastify test suite uses these null prototype objects for some reason.


colinhacks:
Thanks! Good call, great PR.

Bram-dc:
I think however this kind of logic should not be in en.ts, a language implementation should not be able to crash to code. I do not think localization should be implemented in the current manner.

Issue #4134, opened 2025-04-13
v4: toJSONSchema `.default()` `.catch()` input values considered non optional


samchungy:
Something tricky to deal with in my library when generating JSON Schema from Zod Types was figuring out whether to generate an `input` (request) type or a `output` (response) type. I assume that was the intention of the `pipes` parameter.

The current `required` implementation only considers the `output` type.

https://github.com/colinhacks/zod/blob/2ade678ffc5fbe609d92537f3910f91f15d77725/packages/core/src/util.ts#L538

I considered changing `optional`

https://github.com/colinhacks/zod/blob/2ade678ffc5fbe609d92537f3910f91f15d77725/packages/core/src/schemas.ts#L1480

to a function instead of a getter which takes an optional `output` defaulted `pipes` value but that lead me down a rabbit hole... Keen to help you implement this but I understand that this might require quite a bit of tinkering.

```ts
console.log(
  JSON.stringify(
    z.toJSONSchema(
      z.object({
        jobId: z.string().default("foo"), // or .catch()
      }),
      {
        reused: "ref",
        pipes: "input",
      }
    ),
    null,
    2
  )
);
```

```json
{
  "type": "object",
  "properties": {
    "jobId": {
      "type": "string",
      "default": "foo"
    }
  },
  "required": [
    "jobId"
  ]
}
```

In this scenario `jobId` should not be marked as required.

colinhacks:
The `pipes` param has been generalized to `io`. As you pointed out, input & output types diverge for a number of different schema types (not just pipes). `ZodDefault` is now reflected as optional with `io: "input"`. 

```ts
import * as z from "zod";

console.log(
  JSON.stringify(
    z.toJSONSchema(
      z.object({
        jobId: z.string().default("foo"), // or .catch()
      }),
      {
        reused: "ref",
        io: "input",
      }
    ),
    null,
    2
  )
);

/**
 * {
  "type": "object",
  "properties": {
    "jobId": {
      "type": "string",
      "default": "foo"
    }
  },
  "required": []
}
 */
```

I think `.catch()` is a little trickier. Arguably the equivalent JSON Schema for `ZodCatch` is `{}` (anything) but in practice I don't think it makes sense to throw away all the type information here. The `.catch()` functionality is Zod-specific and it's safe to assume that in most cases you'll want the input to conform to the schema instead of just defaulting to the catch fallback.

PR #4074, opened 2025-04-08
Zod 4

colinhacks:
<p align="center" >
 
</p>

<p align="center">
  Announcement post: <a href="https://v4.zod.dev/v4">v4.zod.dev/v4</a><br>
  Release notes / breaking changes: <a href="https://v4.zod.dev/v4/changelog">v4.zod.dev/v4/changelog</a>
</p>

## Summary by CodeRabbit

- **New Features**
  - Launched a modern documentation site with comprehensive guides, API references, error handling, and migration resources for Zod 4.
  - Added extensive benchmarking tools and scripts comparing Zod versions and alternative validation libraries.
  - Implemented new configuration files for workspace, formatting, linting, and editor support.
  - Introduced new schema methods and validation features, including template literals, advanced string formats, and improved error customization.
  - Added rich ecosystem resource listings with live GitHub star counts and support indicators.
  - Provided new React components for documentation layout, sidebar, headings, featured sponsors, and navigation enhancements.
  - Enhanced CI/CD workflows with updated branch triggers and package manager usage.

- **Bug Fixes**
  - Updated workflow triggers and package manager commands to ensure compatibility with the latest tooling and branch naming conventions.

- **Documentation**
  - Replaced legacy documentation with new, structured content covering usage, migration, metadata, ecosystem, and library author guidance.
  - Added detailed API documentation and migration guides for Zod 4, with improved navigation and metadata.
  - Introduced new error formatting and customization documentation.
  - Added comprehensive docs on JSON Schema conversion, metadata registries, and parsing strategies.

- **Chores**
  - Migrated the project to a pnpm monorepo structure with workspace and package management enhancements.
  - Removed deprecated configuration files, test suites, and legacy documentation.
  - Updated copyright and license information.
  - Switched package manager from Yarn to pnpm across CI and development scripts.

- **Refactor**
  - Transitioned from legacy tools (Jest, Babel, ESLint, Prettier) to Biome, Vitest, and pnpm.
  - Streamlined and modernized build, test, and linting scripts.
  - Removed or replaced outdated or redundant files and scripts.
  - Simplified internal module exports and reorganized source structure.

- **Style**
  - Introduced new formatting and linting standards across the codebase.

- **Tests**
  - Removed legacy test suites and introduced new benchmarking and validation tests for current and future development.

PR #4153, opened 2025-04-15
Set `qin` property on `ZodOptional`

### LINKED ISSUE #4152: v4: `qin` prop discrepancy in `ZodOptional`
Version `4.0.0-beta.20250414T061543` and `@zod/core@0.5.0`
The type of the `qin` property does not match the actual implementation, @colinhacks 

Image

RobinTail:
Fixes #4152 

## Summary by CodeRabbit

- **Refactor**
  - Updated internal handling of optional schema properties to improve consistency. No visible changes to end-user functionality.

### DIFF
--- a/packages/core/src/schemas.ts
+++ b/packages/core/src/schemas.ts
@@ -2983,7 +2983,7 @@ export const $ZodOptional: core.$constructor<$ZodOptional> = /*@__PURE__*/ core.
   "$ZodOptional",
   (inst, def) => {
     $ZodType.init(inst, def);
-    // inst._zod.qin = "true";
+    inst._zod.qin = "true";
     inst._zod.qout = "true";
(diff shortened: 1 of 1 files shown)


RobinTail:
@coderabbitai review

RobinTail:
targeting #4074

colinhacks:
Yeah, I commented this out because I actually never access `.qin` (as you probably noticed. Basically `qin` is a "virtual property"  only there so `ZodObject` can handle it in the static domain. Whereas `qout` is referenced at runtime in the implementation of ZodObjectLike.

I considered prefixing all "virtual" properties with `_` but it seems a little redundant with `_zod`. I think I'll just rely on JSDoc to point people in the right direction.

samchungy:
Do we need to access qin for scenarios like this?

https://github.com/colinhacks/zod/issues/4134

RobinTail:
fixed by removal in @zod/core 0.11.0

PR #4150, opened 2025-04-14
Optimize `extendShape`

Andarist:
It has been proposed [here](https://github.com/microsoft/TypeScript/pull/61505#issuecomment-2800012653) and, in fact, it cuts [`AnswerOverflow/apps/dashboard`](https://github.com/AnswerOverflow/AnswerOverflow/tree/1ab687793a598789a0c86b4ca5f17ee6f3a256aa/apps/dashboard)'s instantiations from 13672403 to 3538464 (74% reduction) and check time from 6.5s to 4s (38% reduction).

## Summary by CodeRabbit

- **Refactor**
	- Improved the efficiency of type-level merging for object shapes, resulting in faster and more direct merging when object keys do not overlap. This change affects only type definitions and does not impact runtime behavior.

PR #4155, opened 2025-04-15
Refactor `Extend` to preserve declaration links

Andarist:
This targets v4 and it's basically the same change as in https://github.com/colinhacks/zod/pull/4150 (but that targets v3).

The proposed version is better because it preserves declaration links and that helps in-IDE navigation. To get proper declaration links you need to iterate over a single "modifier object" and `[P in keyof T | keyof U]` breaks that (it sources keys from 2 objects, so the compiler can't decide from which one the declaration, and modifiers, should be picked from). See this [TS playground](https://www.typescriptlang.org/play/?ssl=3&ssc=23&pln=3&pc=47#code/C4TwDgpgBAthBOBzCBGAPAFQDRQKoD4oBeKAeRgEthMcBrCEAewDMoMoAyKep1gzvAG4AUKEiwEyAEw08hEjxZsBivlAgAPYBAB2AEwDOUHRABuCKAH5lXXFABcZStWzcGS9l1VyBuEWOg4JAgAZll+EgBvKABtAAUoCh03XmUAHxSlXABdRwTNbX0jbztrXHjshzZ4lXdWDEqAXxFRcGgAMUZGYihI4SgBqGYuxwNgeCTEYUbhVvEAIQBDeB6+wagAI2XHHQBXGA2EadmAehOoRG7gbr0IAGMAG2XF4ApGZO0xgznoeAgDXYPYAoHpBZDoTqMHBLeD4GIAIi28HhlTOUEYtB+UD+AKBUlBkggMkh0OWcMRyxRUDRGKxOMBwBCBOCYRJUBh5KRVLROkYkCAA)

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -70,6 +70,7 @@
       ".next",
       "packages/bench",
       "packages/treeshake",
+      "packages/tsc",
       ".source"
     ]
   }
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -70,7 +70,6 @@
       ".next",
       "packages/bench",
       "packages/treeshake",
-      "packages/tsc",
       ".source"
     ]
   }
--- a/packages/core/src/util.ts
+++ b/packages/core/src/util.ts
@@ -124,19 +124,16 @@ export type NoNeverKeys<T> = {
 export type NoNever<T> = Identity<{
   [k in NoNeverKeys<T>]: k extends keyof T ? T[k] : never;
 }>;
-export type ExtendShape<A extends object, B extends object> = Identity<
+export type ExtendShape<A extends object, B extends object> = Flatten<
   // fast path when there is no keys overlap
-  keyof A & keyof B extends never
-    ? A & B
-    : {
-        [K in keyof A as K extends keyof B ? never : K]: A[K];
-      } & {
-        [K in keyof B]: B[K];
-      }
+  keyof A & keyof B extends never ? A & B
+  : {
+      [K in keyof A as K extends keyof B ? never : K]: A[K];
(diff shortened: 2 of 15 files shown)


Andarist:
@colinhacks this should pass CI now, could you re-approve the workflow run?

colinhacks:
Love this!!

Rebased to fix CI. 

Cleaned up the utils — the `Identity`/`Flatten` were in there to force TS to resolve the the type. With the old implementation of ExtendShape (without `Identity`) I was seeing unresolved types in my return types, e.g.:

Screenshot 2025-04-16 at 16 49 49

But it seems Identity/Flatten aren't necessary with the new and improved version. Very nice 😊

Andarist:
> Unfortunately there seems to be no escaping some kind of exponential explosion with chained called to .extend() that I can find. (Mateusz, this comment is mostly for posterity, not trying to nerd snipe you here. 🙃)

I think this is what I was trying to reproduce [here](https://github.com/colinhacks/zod/pull/4150#issuecomment-2812337535). Maybe v3 and v4 behave differently for some reason here? Or maybe there is some crucial difference between what you are generating and the stuff I prepared? 

I have checked this bench briefly and from what I see:
- the current answer for Zod 4 is 38
- with the build from [here](https://github.com/microsoft/TypeScript/pull/61535#issuecomment-2810186183) it's 46

So it's higher - but not by a lot. But it gives up based on `instantiationDepth`'s limit (100) and not on the `instantiationCount`'s count limit (5000000). Despite being able to check more "layers" it's also substantially faster (total time 2.90s -> total time 0.52s).

PR #4173, opened 2025-04-16
fix(docs): wrong comment for object document

shivamvijaywargi:
It looks like the comments were interchanged. To my understanding, obj: A will pass because the property in the interface A is optional, whereas obj: B will fail because it is not optional.


colinhacks:
Good catch! I fixed this in a slightly different way here: https://github.com/colinhacks/zod/commit/c8e72e8f0f8b1f5962954c035b52e46b2a57ef4e

colinhacks:
@coderabbitai review

PR #4182, opened 2025-04-16
v4: toJSONSchema add reused option ref-registry

MaciejDabek:
I wanted to propose an additional option for `toJSONSchema` parameter `reused` which would be `"ref-registry"`. It would be in the middle between `"inline"` and `"ref"`, so for every seen schema:
- would be handled as `"inline"` if metadata does not contain `id` 
- would be handled as `"ref"` if metadata do contain `id`

It can be useful for any JSON Schema to language converter that will produce types for every schema with `id` in metadata. There is also a big point for readability because we have refs only to proper chunks of full schema. 

I added full test for reused parameter, but let me know if you prefer any change in docs or tests.


colinhacks:
@coderabbitai review

MaciejDabek:
@colinhacks What are your thoughts on the proposal?

I updated test, to use `test` instead of `it` as changed in b9d781b.

colinhacks:
Honestly I thought this was the default behavior 😅 I even wrote as much in the docs
Screenshot 2025-04-17 at 15 09 47

I implemented this initially then must have broken it when implementing `external`. 

I'll fix this and make it the default behavior, since I think it's a no-brainer. Perhaps later I'll add a flag to disable this if people ask for it.

MaciejDabek:
Ok 😄  I saw the comment and I tried it, but it only uses `id` for schemas seen multiple times when reused is set to `ref`. 

If this is intended, it works great for my use case 😄  

But looking from the perspective of possible use cases, three options for `reused` can be the sweet spot:
- `inline` - works especially well for smaller schemas
- `ref` - probably produces the smallest JSON Schema because everything possible is reused
- `ref-registry` - gives you the option described above

If it will be only depending on `id` being in metadata. Then user will be blocked from possibly reusing some schemas and generating different JSON Schema for different use cases.

jeremyjacob:
This would be great to have as an option, thank you MaciejDabek. Not everything needs to be a ref and having many __schemaX definitions can make the spec difficult to read.

Super nice to have JSONSchema functionality native nonetheless.

colinhacks:
Whether a schema gets extracted into a ref based on the existence of an `id` is an orthogonal consideration to `reused`. In other words, a schema with an `id` property in metadata will be extracted regardless of whether it's reused.

If you want to disable `id`-based ref extraction, you can pass an empty registry into `metadata`.

```ts
const str = z.string().meta({ id: "foo" });
const obj = z.object({ a: str, b: str });
z.toJSONSchema(obj, { reused: "ref", metadata: z.registry() });
```

Unless I'm mistaken, this accounts for all the use cases being described? @MaciejDabek

MaciejDabek:
@colinhacks Yes, it works, but it maybe a little confusing.

Currently it works like `reused`:
- `inline` _without registry_ - everything is inline (as you expect)
- `inline` _with registry_ - everything with ID is extracted to defs and rest is inline (maybe confusing that it is inline)
- `ref` _without registry_ - everything that is seen multiple times is extracted to anonymous defs
- `ref` _with registry_ - everything with ID is extracted to def but also everything that is seen multiple times is extracted to anonymous defs

So @jeremyjacob the case that you are interested in is `inline` _with registry_.

PR #4185, opened 2025-04-16
Languages: add Ukrainian language

MrBarracuda:


packages/core/src/locales/ua.ts
```diff
@@ -0,0 +1,126 @@
...
+  e164: "номер E.164",
+  jwt: "JWT",
+  template_literal: "вхідні дані",
+};
+
+const error: errors.$ZodErrorMap = (issue) => {
+  switch (issue.code) {
+    case "invalid_type":
+      return `Неправильні вхідні дані: очікується ${issue.expected}, отримано ${parsedType(issue.input)}`;
+    // return `Неправильні вхідні дані: очікується ${issue.expected}, отримано ${util.getParsedType(issue.input)}`;
```
VladSez, line 76: Do we need this? 😅

MrBarracuda: > Do we need this commented out code? =)

image
I just followed the en.ts file, which also has this commented line of code

colinhacks:
Thank you!!

PR #4198, opened 2025-04-16
fix(v4): Change name of catch schema

chimame:
## Overview

`catch` is a reserved word, so if you try to use it as shown below, an error will occur. Therefore, we added an underscore to it, just like `default`.

image


colinhacks:
So, this is intentional. You'll see a lot of APIs that share names with reserved words:

```
z.undefined
z.null
z.void
z.interface
z.enum
z.default
z.catch
z.instanceof
```

That's because Zod is meant to be consumed with the `z` variable. There's little reason to define individual imports these days, all major bundlers are good enough to treeshake properly with `import * as z`.

If you want to stick with individual named imports, use `import {catch as _catch}`

PR #4200, opened 2025-04-16
feat(lang): Add ota locale in Latin script

kadimelifba:
Created translation for Ottoman Turkish language. For now, it is in Latin script. In the future, I am planning to use Ottoman Script.


colinhacks:
Thank you!!

PR #4235, opened 2025-04-19
fix(locale): show correct prefix instead of [object Object] in AZ and ES language

AliReza-Kamkar:
Fix issue of showing [object Object] in startsWith error for AZ and ES languages

image


colinhacks:
Fixed in `@zod/core@0.8.1` and latest betas

AliReza-Kamkar:
@colinhacks 
Glad it’s fixed😃
Appreciate the quick update. closing this PR then.

Issue #4168, opened 2025-04-15
[i18n] Request for internationalization (locales)


colinhacks:
Zod 4 has a more mature system for internationalizaton. Here is an example of loading the Spanish locale:

```ts
import * as z from "zod";

z.config(z.locales.es());

const result = z.string().safeParse(123);
result.error!.issues[0].message;
// => Entrada inválida: se esperaba string, recibido número
```

I'm putting out an open call for locales to the Zod ecosystem. I think a human touch is important to make error messages friendly and useful. If you plan to contribute a new locale, please leave a comment below so others know it is in progress! This is to avoid redundant PRs. 

## Instructions
  
Refer to this pull request for an example of creating a locale: https://github.com/colinhacks/zod/pull/4158/files

To summarize:

### 1 Create a new file under `packages/core/src/locales/<code>.ts`

Use the language code for your language. Dialects (`pt-BR`) are also encouraged. 

Using the English locale as a starting point, translate all error messages: https://github.com/colinhacks/zod/blob/v4/packages/core/src/locales/en.ts

### 2 Export your dialect from `packages/core/locales.ts`

```ts
// other languages...
import zz from "./locales/zz.js";

export { 
  // ... other languages
  zz
};
```

### 3 Add to docs

Add your new language variant to the [`Error customization` page](https://github.com/colinhacks/zod/blob/v4/packages/docs/content/error-customization.mdx) under the "Locales" section. Keep the list in alphabetical order.

### Submit the PR

---

Thank you for your help on this!

helmerdavila:
Adding portuguese

https://github.com/colinhacks/zod/pull/4171

helmerdavila:
Adding french

https://github.com/colinhacks/zod/pull/4172

Abdalrhman-Almarakeby:
Added Arabic #4176.

TheOnlyTails:
Added Hebrew #4183.

szarbartosz:
Added Polish #4184

Heenkkk:
Added Finnish #4188

GrahamQuan:
Add Simplified Chinese #4189

hoangtrung99:
@colinhacks 
Hi sir, I have created a PR to add Vietnamese, please check it.
[#4191](https://github.com/colinhacks/zod/pull/4191)

filipditrich:
Added Czech language in https://github.com/colinhacks/zod/pull/4192. Hope it helps!

[...]

PR #4237, opened 2025-04-19
feat(lang): Added initial `Tamil` locale

sasivarnan:
Add initial `Tamil` locale as part of #4168


packages/core/src/locales.ts
```diff
@@ -1,5 +1,6 @@
 import az from "./locales/az.js";
 import en from "./locales/en.js";
 import es from "./locales/es.js";
+import ta from "./locales/ta.js";
 
-export { az, es, en };
+export { az, es, en, ta };
```
sasivarnan, line 6: This can be probably written as following 

```js
export { default as az } from "./locales/az.js";
export { default as en } from "./locales/en.js";
export { default as es } from "./locales/es.js";
export { default as ta } from "./locales/ta.js";
```

This way, upon merging one locale PR, another locale PR doesn't need to resolve conflict on the `export` before mering.

@colinhacks Please share your thoughts on this.

colinhacks: Would've been smart...I've been resolving these conflicts for an hour now 😅

colinhacks:
Thank you!!

PR #4241, opened 2025-04-20
fix(docs): Typo in `z.object()` vs `z.interface()`

juangiordana:
Update documentation to clarify that optional properties in `z.interface()` are defined with a `?` suffix, not a prefix.


colinhacks:
Very good catch!

Issue #4240, opened 2025-04-20
Error in Enum docs for `.exclude()` / `.extract()`


juangiordana:
Error in Enum docs for `.exclude()` / `.extract()`.

The code samples in `.exclude()` belong to  `.extract()` and viceversa.

https://github.com/colinhacks/zod/blob/354c8d6e49deb09f42482d8db1211889192d901e/packages/docs/content/api.mdx?plain=1#L616-L653

PR #4242, opened 2025-04-20
fix(docs): Swap Enum `.exclude() / .extract()` code samples.

juangiordana:
The code samples in `.exclude()` belong to `.extract()` and vice versa.


juangiordana:
Closes #4240

PR #4252, opened 2025-04-21
fix(docs): Grammatical error in Refinements

juangiordana:
Correct grammatical error in Refinements


colinhacks:
You are a god of copyediting

Issue #4238, opened 2025-04-19
Infer with key optional properties on an extended interface are not represented correctly


justin-prather:
possibly related to #4082 

When trying to get the TS type from an extended z.interface(), any key optional types in an extension of the original interface are not represented properly. 

```
// top level interface

const t = z.interface({
    'defaultValue?': z.number(),
});

// z.infer<typeof t> correctly inferred to:
// type T = {
//    defaultValue?: number | undefined;
// }

// extended interface
const t2 = t.extend({
    'alias?': z.string(),
});

// z.infer<typeof t2> incorrectly inferred to 
// type T2 = {
//     alias?: unknown;
//     defaultValue?: number | undefined;
//     "alias?": string;
// }

SeeJayy:
+1 on this, it works fine in generating the correct JSON schema but doesn't infer correctly.

akselil:
This occurs also with "zod": "4.0.0-beta.20250430T185432" / "typescript": "5.8.3",

colinhacks:
ZodInterface has been dropped for a variety of reasons, including the complexity of type inference with all those question marks 🫠 https://github.com/colinhacks/zod/pull/4271/files

PR #4257, opened 2025-04-22
#4238 Fix extended interfaces causing optional keys to break types

SeeJayy:
#4238 

Use util.CleanInterfaceShape when merging interfaces to stop extended interfaces with optional keys causing a duplicate key. Included optional tests to show it is fixed


colinhacks:
Thanks! Good fix, good PR. Unfortunately no longer applicable after https://github.com/colinhacks/zod/pull/4271. Pretty happy to be rid of `z.interface` to avoid all these type shenanigans around `.shape`.

PR #4266, opened 2025-04-23
fix(transform): abort further processing when inner value is invalid

### LINKED ISSUE #2113: refine() function gets called even when parser already failed
Using version 3.20.6

In certain scenarios, refine gets called even when the parser has already failed.   It seems like including a z.custom() can trigger this.

What's worse is that the type flowing into the refine's function (x) is different than what typescript thinks it is

Here's a minimal example:
```Typescript
const example1 = z
	.custom<number>(x => typeof x === 'number')
	.transform(x => String(x))
	.refine(x => {
		console.log(typeof x); // prints 'Object'
		console.log("I get called even though I shouldn't!!!");
		return true;
	})
	.safeParse({}); //will fail because it is not a number

console.log(example1.success); // false (like it should be)
```
And here's a more complex example with the same issue
```Typescript
const example2 = z
	.union([z.number(), z.string(), z.custom<number>(x => typeof x === 'number')])
	.refine(() => true)
	.pipe(z.any())
	.or(z.date())
	.pipe(z.any())
	.refine(x => {
		console.log("I get called even though I shouldn't!!!!!!!!!");
		return true;
	})
	.safeParse(true); // will fail immediately because it's a bool
```

As a work-around, using `.transform(x,ctx)=>...) ` approach to validate seems to work.

### LINKED ISSUE #2192: refine() function gets called even when parser already failed within translate()/refine()/regex()

### LINKED ISSUE #2243: when validation fails, transform is skipped but superRefine not
When some validation fails, it looks like transform is skipped but a future superRefine is not skipped, then superRefine tends to fail.
I checked the documentation but I didn't found something related to this behavior :(

```js
const schema = z
  .object({
    foo: z.string().email(),
  })
  .transform((arg, ctx) => {
    console.log("transform", arg);
    return { bar: arg.foo };
  })
  .superRefine((arg, ctx) => {
    console.log("superRefine", arg); // typescript always expects 'bar' attribute

    // This line tends to fail when transform is skipped
    if (arg.bar.length >= 35) {
      ctx.addIssue({
        code: z.ZodIssueCode.custom,
        message: "It hould be less than 35.",
        path: ["bar"],
      });
    }
  });

console.log(schema.safeParse({ foo: "test@example.com" }));
console.log(schema.safeParse({ foo: "averybigemailfortesting@example.com" }));

// This fails on email validation, transform is skipped but superRefine not
console.log(schema.safeParse({ foo: "testexample.com" }));
```

image

DASPRiD:
Previously when an inner value was invalid, transforms would not run, but subsequent refinements would still run with the non-transformed value, which in most cases would violate the typings.

This fix returns `INVALID` from the transform in case of a failed inner validation and stop outer refinements from running with the wrong types.

Closes #2243
Closes #2192
Closes #2113

## Summary by CodeRabbit

- **Tests**
	- Added a new test to verify behavior when a validation transform is skipped and an additional refinement is applied.
- **Bug Fixes**
	- Improved handling of invalid parse results during schema transformations to ensure consistent error reporting.

### DIFF
--- a/deno/lib/__tests__/refine.test.ts
+++ b/deno/lib/__tests__/refine.test.ts
@@ -305,3 +305,19 @@ test("fatal superRefine", () => {
   expect(result.success).toEqual(false);
   if (!result.success) expect(result.error.issues.length).toEqual(1);
 });
+
+test("superRefine after skipped transform", () => {
+  const schema = z
+    .string()
+    .regex(/^\d+$/)
+    .transform((val) => Number(val))
+    .superRefine((val) => {
+      if (typeof val !== "number") {
+        throw new Error("Called without transform");
+      }
+    });
+
+  const result = schema.safeParse("");
+
+  expect(result.success).toEqual(false);
+});
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -4724,7 +4724,7 @@ export class ZodEffects<
           parent: ctx,
         });
 
-        if (!isValid(base)) return base;
+        if (!isValid(base)) return INVALID;
 
         const result = effect.transform(base.value, checkCtx);
         if (result instanceof Promise) {
@@ -4738,7 +4738,7 @@ export class ZodEffects<
         return this._def.schema
           ._parseAsync({ data: ctx.data, path: ctx.path, parent: ctx })
           .then((base) => {
-            if (!isValid(base)) return base;
+            if (!isValid(base)) return INVALID;
 
             return Promise.resolve(effect.transform(base.value, checkCtx)).then(
               (result) => ({ status: status.value, value: result })
--- a/src/__tests__/refine.test.ts
+++ b/src/__tests__/refine.test.ts
@@ -304,3 +304,19 @@ test("fatal superRefine", () => {
   expect(result.success).toEqual(false);
   if (!result.success) expect(result.error.issues.length).toEqual(1);
 });
+
+test("superRefine after skipped transform", () => {
+  const schema = z
+    .string()
+    .regex(/^\d+$/)
+    .transform((val) => Number(val))
+    .superRefine((val) => {
+      if (typeof val !== "number") {
+        throw new Error("Called without transform");
+      }
+    });
+
+  const result = schema.safeParse("");
+
+  expect(result.success).toEqual(false);
+});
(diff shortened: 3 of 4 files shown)


colinhacks:
Thanks for the excellent PR and apologies it took so long. The codebase has changed under this PR so I ported this fix here: https://github.com/colinhacks/zod/pull/4554

PR #4271, opened 2025-04-24
Support recursive types in `z.object()`, drop `z..interface()`

colinhacks:
> ⚠️ This PR contains breaking changes. `z.interface()` has been removed in its entirety.

! I finally figured out a way to support cyclical type inference in `z.object()`. I've been trying to get this working for years and thought it was impossible. 

Screenshot 2025-04-29 at 15 08 52

This new capability negates the primarily raison d'être of `z.interface()`. Thus this PR also:

- Removes `z.interface()`. 
- Removes `core.ZodObjectLike` (an interface that served as a shared parent for ZodObject and ZodInterface)
- Removes `optional: string[]` from `ZodObjectDef`. Optionality information is stored on the elements inside `shape`.

This is a major breaking change but Zod 4 is still in beta for two more weeks. I'm loathe to do a change of this magnitude but in general consolidating on a single object API is a huge win, so it's worth it.


colinhacks:
@coderabbitai review

apfelbox:
`.interface()` also solved some other issues, namely the problem with `type?:` vs. `type: ...|undefined` among other things. Do these things work now with `.object()` as well? Because when switching back from `.interface()` (where everything worked) to `.object()` I now get a ton of errors in TypeScript.

PR #4272, opened 2025-04-24
.catchall() method should accept undefined

called-d:
The .strip() method has been deprecated. Currently, loose/strict object schema cannot be made strip (default) afterwards.

```ts
const schema = z.looseObject( ... )

const stripSchema = schema.strip() // deprecated
const stripSchema2 = schema.catchall() // set strip (default); Expected 1 arguments, but got 0.ts(2554)
```

.catchall() method accepts undefined, which can be used in replacement of the .strip() method

```ts
schema.catchall()            // = .strip()
schema.catchall(z.never())   // = .strict()
schema.catchall(z.unknown()) // = .loose()
```


colinhacks:
Not a fan of that from an API perspective. The `.strip()` method is discouraged but it won't be removed, so feel free to continue using it if you like.

PR #3918, opened 2024-12-19
`recordWithBrandedKeys` should not be `Partial`

AlexErrant:
I am attempting to fix a regression. Here's a timeline:

1. [This PR](https://github.com/colinhacks/zod/pull/2097) describes an issue and is merged. Note that it primarily adds
```
  : K extends BRAND<string | number | symbol>
  ? Record<K, V>
```

2. For reasons that are unclear to me [this commit](https://github.com/colinhacks/zod/commit/defdab9c163d9a994f927a0644ab4ec172513fcb) reverses the order of the above and uses `[]` syntax, causing a regression:

```
  : [BRAND<string | number | symbol>] extends [K]
  ? Record<K, V>
```

3. My PR switches the order back to 1, keeps the `[]` syntax, and adds a test to prevent regressions:

```
  : [K] extends [BRAND<string | number | symbol>]
  ? Record<K, V>
```

Issue #992, opened 2022-03-02
Typing for record became too loose after addition of Partial


hsarokina-ita:
This issue is related to: https://github.com/colinhacks/zod/issues/751

After updating zod version from v3.12.0 to v3.13.3 we started to receive the following TS error: `error TS2322: Type 'Partial<Record<string, string>>' is not assignable to type 'Record<string, string>'.
  'string' index signatures are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.
`

We are using zod `record` with string keys and values. I think that returning Partial for string records is a bug, because we are sure, that all keys that record has are checked by zod and contain non-`undefined` value, but TS says that they can contain `undefined` that forces us to do extra checks in the later code to convince TS that all values are strings.

I've created a repo with simple reproduction to illustrate such use case better:
https://github.com/hsarokina-ita/zod-partial-record
`safeParse` returned `success: false` for object with `undefined` values, this means that zod checked for it to be defined. But at the same time I can assign record with undefined value to typing returned by parser.

amuttsch:
I have a similar issue with a string key, but my value is not a `string` but a `z.union` of some objects. I don't want my record value to be optional, this seems to be contradictionay to the "everything is required until `.optional()` is called".

colinhacks:
This was changes to more accurately reflect the type Zod actually validates. Zod can't do exhaustiveness checking - for instance it doesn't/can't verify that all elements of the enum show up as keys:

```ts
z.record(z.enum(['a', 'b', 'c']), z.string());
// => Record<'a' | 'b' | 'c', string>
```

This is inaccurate - Zod merely validates each key against the key schema, but can't guarantee that all three keys exist in the object. So a more accurate type is `Partial<Record<'a' | 'b' | 'c', string>>`.

But your right that wrapping in `Partial` was too permissive. Just updated the typings: 

```ts
z.record(z.string, z.string);
// Record<string,string>;

z.record(z.enum(['a', 'b', 'c']), z.string());
// => Partial<Record<'a' | 'b' | 'c', string>>

```

Merged in 3.13.3.

alexanderhorner:
This still seems to be happening when using custom string literal types.
```ts
export const UUIDSchema = z.string().uuid() as z.ZodType<UUID>;
````

When defining a record with UUIDSchema as key, it will be Partial.

AlexErrant:
[...]

PR #4277, opened 2025-04-24
[v4 beta] Properly support branded keys in records

jthemphill:
This should resolve the longstanding discussion:

* https://github.com/colinhacks/zod/discussions/2069

I'm running into this problem:

```ts
const stringRecord = z.record(z.string(), z.number());

// ✅ Record<string, number>. This is correct.
type stringRecord = z.infer<typeof stringRecord>

const brandedStringRecord = z.record(z.string().brand("SomeBrand"), z.number());

// ❌ Partial<Record<string & $brand<"SomeBrand">, number>>.
// But this is wrong - a branded string should be treated the same way as a string.
type brandedStringRecord = z.infer<typeof brandedStringRecord>;
```

Colin explains the rationale for sometimes returning `Partial<Record<K, V>>` in https://github.com/colinhacks/zod/issues/992#issuecomment-1057290762. But that logic doesn't apply to brands - this is just an oversight.

---

@AlexErrant noticed the same problem, and submitted a PR into Zod v3, which this PR is based on:

* https://github.com/colinhacks/zod/pull/3918

I'm going with a different approach than that PR takes, though.

Today, the code says that the record is _not_ partial if one of these is true:

1. All strings are valid keys (`string extends K`)
2. All numbers are valid keys (`number extends K`)
3. All symbols are valid keys (`symbol extends K`)

This PR adds three more checks. The record is _also_ not partial if one of these is true:

4. All branded strings are valid keys (`string & $brand<never> extends K`)
5. All branded numbers are valid keys (`number & $brand<never> extends K`)
6. All branded symbols are valid keys (`symbol & $brand<never> extends K`)

When my change didn't work, I found that the problem was because `$brand<T>` uses `T` in an index key signature. This meant that `brand<never>` resolved to `{}`, which is not really correct from a type theory perspective.

So... I'm also changing the type implementation of `$brand` in this PR. I think this is an acceptable change to make while Zod v4 is still in beta.

Let me know what you think!

### DIFF
--- a/packages/core/src/core.ts
+++ b/packages/core/src/core.ts
@@ -48,7 +48,7 @@ export /*@__NO_SIDE_EFFECTS__*/ function $constructor<T extends ZodTrait, D = T[
 //////////////////////////////   UTILITIES   ///////////////////////////////////////
 export const $brand: unique symbol = Symbol("zod_brand");
 export type $brand<T extends string | number | symbol = string | number | symbol> = {
-  [$brand]: { [k in T]: true };
+  [$brand]: T;
 };
 
 export class $ZodAsyncError extends Error {
--- a/packages/core/src/schemas.ts
+++ b/packages/core/src/schemas.ts
@@ -2426,7 +2426,13 @@ type $InferZodRecordOutput<
       ? Record<Key["_zod"]["output"], Value["_zod"]["output"]>
       : symbol extends Key["_zod"]["output"]
         ? Record<Key["_zod"]["output"], Value["_zod"]["output"]>
-        : Partial<Record<Key["_zod"]["output"], Value["_zod"]["output"]>>
+        : string & core.$brand<never> extends Key["_zod"]["output"]
+          ? Record<Key["_zod"]["output"], Value["_zod"]["output"]>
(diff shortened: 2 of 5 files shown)


colinhacks:
I may be losing my mind, but this appears to already by working with Zod 4. Can you verify this?

Screenshot 2025-05-02 at 18 56 59

jthemphill:
> I may be losing my mind, but this appears to already be working with Zod 4. Can you verify this?

Hi! I would define this as exactly the problem I am trying to solve. I believe the correct type should be

```
type brandedStringRecord = {
    [x: string & z.core.$brand<"SomeBrand">]: number;
}
```

As your screenshot shows, Zod 3 and Zod 4 are both making the record partial, and they weaken the value type into `number | undefined`, because a branded string is not a string. This is the correct behavior if the key is an *enum*, for [reasons you've stated before](https://github.com/colinhacks/zod/issues/992#issuecomment-1057290762). But branded strings are not enums!

jthemphill:
This PR has morphed into this issue:

* https://github.com/colinhacks/zod/issues/4464

PR #4289, opened 2025-04-28
fix: `strict()` marks as deprecated in v4

soartec-lab:
The following commit removed the deprecated mark from `strict()`:
https://github.com/colinhacks/zod/pull/4074/commits/4d269a2f379914ab5b8be7f7a33e959cf2e2d53d

However, in the migration guide for zod 4, this is deprecated:

https://v4.zod.dev/v4/changelog#deprecates-strict

I marked `strict()` as deprecated to fix this inconsistency.

### DIFF
--- a/packages/zod/src/schemas.ts
+++ b/packages/zod/src/schemas.ts
@@ -1266,7 +1266,7 @@ export interface ZodObject<
 
   loose(): ZodObject<Shape, Record<string, unknown>>;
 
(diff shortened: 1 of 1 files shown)


colinhacks:
I unmarked these as deprecated on purpose because in practice they will never be removed. Forgot to update the migration post too. Thanks!

soartec-lab:
Okey, I'll make a PR to update the docs later

PR #4298, opened 2025-05-01
docs: in v4, `strict()` is not deprecated

soartec-lab:
The following commit removed the deprecated mark from strict():

https://github.com/colinhacks/zod/commit/4d269a2f379914ab5b8be7f7a33e959cf2e2d53d

And this was an omission from the guide, so I fixed it.

https://github.com/colinhacks/zod/pull/4289#issuecomment-2842957366


colinhacks:
Very true. Though I'm not actually removing these methods (and never will in `zod`) I still want to steer people towards the top-level functions and highlight their existence. Instead of removing the section I hijacked the PR and tweaked the working. Good call nonetheless!

PR #4302, opened 2025-05-01
Add z.looseObject

marco-carvalho:
Since `z.strictObject` exist, it should be available a `z.looseObject`.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new schema creation function that allows objects to accept unknown keys without validation errors.
- **Tests**
  - Added tests to confirm that objects created with the new schema function correctly allow extra properties.

### DIFF
--- a/src/__tests__/object.test.ts
+++ b/src/__tests__/object.test.ts
@@ -317,6 +317,18 @@ test("strictcreate", async () => {
   expect(asyncResult.success).toEqual(false);
 });
 
+test("looseObject", async () => {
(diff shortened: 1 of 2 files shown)


colinhacks:
Thanks for this, but I'm not really looking to add any new APIs to Zod 3. If people want this, they should upgrade to Zod 4 imo. Adding it to the v3 docs would also just confuse people who may be on older v3 versions, so it would require appropriate disclaimers...not worth the effort.

Issue #3636, opened 2024-07-11
datetime validation fails if seconds are missing


u2ix:
Have an issue with a valid datetime value, as produced by the browser in a datetime picker field, is not accepted by the `datetime()` validation. After some digging I figured that when the seconds `:00` are added the validation passes.

But according to ISO 8601, just the hour is required for a valid time. Also as the timestamp is produced like this `"2024-07-18T11:00"`, I feel it would be really great if this is accepted by the library as also the Date constructor doesn't have any issue with this.

For now I had to put in a quite ugly `preprocess` to make the validation pass on the value provided by the browser:
```
z.preprocess(input => `${input}:00`,
            z.string().datetime({ local: true }))
```

sachix1001:
I have the same issue.

itsthekeming:
Likewise. I'm using [js-joda](https://js-joda.github.io/js-joda/) to handle dates, and its `LocalDateTime.toString()` implementation [follows the spec](https://github.com/js-joda/js-joda/issues/205).

anthonyhagi:
I've noticed the same issue with `z.string().time()` where it requires `HH:MM:SS` to actually validate. The `<input type='time' />` input by default uses `HH:MM` as mentioned in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time)

Hopefully this can be fixed sometime soon

u2ix:
@fvilers But a ISO time format is also valid without the seconds, and it's the way that browsers return the date in a valid ISO format, but the validation in the library tells us it's wrong. Going for a regex here, instead of fixing a obvious bug in this library, that many people have (see all the up-votes), seems to be kinda hack or a workaround, but not solving the problem.

fvilers:
@u2ix As I'm reading https://en.wikipedia.org/wiki/ISO_8601#Times you may be right. It looks like the seconds may be omitted. I might create a PR to add this feature (an approval from the maintainer(s) would be great). Using a regex is not hacky as it's what is used behind the scenes by the `time()` validator. And as I see things, there is no bug  in the current implementation as the documentation clearly states that the seconds are mandatory (`z.string().time(); // ISO time format (HH:mm:ss[.SSSSSS])`)

ronnyandre:
Pleeease fix this annoying issue (not bug).. I think a modern library should handle without seconds.

stevemk42:
Same here. input with datetime-local type fails zod validation. 

[...]

PR #4315, opened 2025-05-03
Allow HH:MM format in `z.string().datetime()` and `z.string().time()`

timorthi:
Addresses #3636 

This updates the time regex to be in line with the ISO8601 spec that allows seconds to be omitted.

These strings are also `Date`-compatible: 
```javascript
> new Date('2024-01-01T00:00Z')
2024-01-01T00:00:00.000Z
> new Date('2024-01-01T00:00+0500')
2023-12-31T19:00:00.000Z
> new Date('2024-01-01T00:00')
2024-01-01T08:00:00.000Z
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@
 <!-- <hr/> -->
 <br/>
 <br/>
- 
+
 ## Table of contents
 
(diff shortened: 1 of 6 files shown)


colinhacks:
Thanks, great PR.

colinhacks:
Landed in 3.24.4

Issue #4329, opened 2025-05-06
v4: using `z.input<..>` with union over transformed values reflects output, not input values (@zod/core 0.11.1+ regression)


rikbrown:
This feels pretty broken and is a recent regression in the beta.

repro:

```ts
const numberSchemaTransformToString = z.number().transform((n) => `_${n}`)
const booleanSchemaTransformToNumber = z.boolean().transform((b) => (b ? 1 : 0))

const unionSchema = z.union([numberSchemaTransformToString, booleanSchemaTransformToNumber])
type UnionInput = z.input<typeof unionSchema> // string | 1 | 0 but should be number | boolean

type NumberInput = z.input<typeof numberSchemaTransformToString> // works as you'd expect
```

rikbrown:
This regression occurred in 0.11.2 (working in 0.11.1) cc @colinhacks 

Based on commit titles alone it's probably related to this:
[Fix union inferred type issue](https://github.com/colinhacks/zod/commit/8cdf540cf4e27117b0aa75ef0c9389f9ce908beb)

rikbrown:
<img width="1098" alt="Image" src="https://github.com/user-attachments/assets/09ad3170-04a9-4d95-b637-5ace6efb3e0b" />

naively I'd say this is a typo

rikbrown:
yes, patching that fixed it.

https://github.com/colinhacks/zod/pull/4330

PR #4330, opened 2025-05-06
fix(v4): use input, not output, for input of `$ZodUnionInternals`

rikbrown:
fixes https://github.com/colinhacks/zod/issues/4329

tested with a local patched version so far.
feel free to reject this and fix it a different way, or if I am not following the pr process correctly for zod.


colinhacks:
Good catch 🫠

PR #4335, opened 2025-05-08
fix(docs): edited the link to the stars icon

Olexandr88:


colinhacks:
I think I prefer this link to the GitHub home page. thanks though 👍

Olexandr88:
Знімок екрана 2025-05-28 о 11 22 55
this link is also given here. So I think it is more correct to leave the link on github, but of course it is up to you to decide

Issue #4049, opened 2025-03-25
z.string().jwt() failure - headers do not require a typ


codyfrisch:
Currently the .jwt validation code returns false if there is no decoded.typ in the header portion of the jwt. However the typ is not strictly mandatory.

https://www.rfc-editor.org/rfc/rfc7519#section-5.1

Note "Use of this Header Parameter is OPTIONAL."

This means VALID jwts are failing.

colinhacks:
Fixed in latest versions of Zod 3 and Zod 4

PR #4350, opened 2025-05-11
Remove check for JWT typ header

scottanderson:
Per RFC 7519, JWT implementations are required to ignore the typ header:
- The `typ` header will typically not be used when it is already known that the object is a JWT.
- This parameter is ignored by JWT implementations.
- Use of this Header Parameter is OPTIONAL.

> [5.1](https://datatracker.ietf.org/doc/html/rfc7519#section-5.1).  "typ" (Type) Header Parameter
>
>    The "typ" (type) Header Parameter defined by [[JWS](https://datatracker.ietf.org/doc/html/rfc7519#ref-JWS)] and [[JWE](https://datatracker.ietf.org/doc/html/rfc7519#ref-JWE)] is used
>    by JWT applications to declare the media type [[IANA.MediaTypes](https://datatracker.ietf.org/doc/html/rfc7519#ref-IANA.MediaTypes)] of
>    this complete JWT.  This is intended for use by the JWT application
>    when values that are not JWTs could also be present in an application
>    data structure that can contain a JWT object; the application can use
>    this value to disambiguate among the different kinds of objects that
>    might be present.  It will typically not be used by applications when
>    it is already known that the object is a JWT.  This parameter is
>    ignored by JWT implementations; any processing of this parameter is
>    performed by the JWT application.  If present, it is RECOMMENDED that
>    its value be "JWT" to indicate that this object is a JWT.  While
>    media type names are not case sensitive, it is RECOMMENDED that "JWT"
>    always be spelled using uppercase characters for compatibility with
>    legacy implementations.  Use of this Header Parameter is OPTIONAL.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved JWT validation to be less restrictive, allowing tokens without the `typ` property in their header.

### DIFF
--- a/deno/lib/types.ts
+++ b/deno/lib/types.ts
@@ -752,7 +752,7 @@ function isValidJWT(jwt: string, alg?: string): boolean {
       .padEnd(header.length + ((4 - (header.length % 4)) % 4), "=");
     const decoded = JSON.parse(atob(base64));
     if (typeof decoded !== "object" || decoded === null) return false;
-    if (!decoded.typ || !decoded.alg) return false;
+    if (!decoded.alg) return false;
     if (alg && decoded.alg !== alg) return false;
     return true;
   } catch {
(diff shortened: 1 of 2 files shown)


scottanderson:
Fixes #4049

colinhacks:
Thanks for this PR. This'll land in Zod 3.25. I've implemented this fix over in the `v4` branch which is where active development is now happening (including for Zod 3): https://github.com/colinhacks/zod/pull/4404

Issue #3982, opened 2025-02-12
[Feature Request] Add parent field to `RefinementCtx`


Faithfinder:
I'm looking to transform an object like this:

```ts
const Raw = {
  value1: 1,
  value1_prev: 2,
  value2: 3,
  value2_prev: 4,
}
```
into an object like this:
```ts
const Parsed = {
  value1: {
    current: 1,
    prev: 2,
  },
  value2: {
    current: 3,
    prev: 4,
  }
}
```

It would be nice to define it via `value1Schema.transform()`, instead of a giant `transform` on the parent object, however I would need access to the parent object on the field to achieve that. Is that possible?

callmeyaz:
I just started evaluating Zod today and I am surprised that parent object is not available in the Zod.RefinementCtx. This feature is must as there are times when validation of a value is dependent on the value of its siblings.

A very common use case is using "a value is only required when another optional value is not provided" condition in addIssue() function.

colinhacks:
Sorry, this will never happen. Litigated extensively elsewhere. 

@dosu explain

PR #4352, opened 2025-05-12
Added Parent property to RefinementCtx

callmeyaz:
There is a very important use case mentioned here: https://github.com/colinhacks/zod/issues/3982 where we need to evaluate a value only when another optional value is set.


callmeyaz:
@colinhacks I am working on a extension for Zod and this change is much needed for my work. Can you please review and share if you have any concern. 

Much appreciated!

PR #4354, opened 2025-05-12
fix(core): `z.enum` filters out members whose key maps to an unrelated numeric value

### LINKED ISSUE #4353: [v4]: `z.enum({ A: 1, B: "A" })` does not accept `"A"`
Hello!

It looks like `z.enum` can handle object enums with key/value pairs that are the same "diagonally" (e.g. `{ A: "B", B: "A" }`) when the values are all strings, but behaves differently when one of the values is a number.

[StackBlitz](https://stackblitz.com/edit/vitest-dev-vitest-srmahjsw?file=test%2Fenum.test.ts)

Pasted from the sandbox:

```typescript
// `z.enum` behaves as expected with a matching key/value when values are strings
const schema_01 = z.enum({
  A: "B",
  B: "A",
})

schema_01.parse("A") 
// => "A" ✅

// `z.enum` behavior does not match with a matching key/value when one of those values is a number
const schema_02 = z.enum({
  A: 1,
  B: "A",
})

schema_02.parse("A") 
// 🚫 [ { "code": "invalid_value", "message": "Invalid input: expected 1", "path": [], "values": [ 1 ] } ]
```

It looks like the value is being filtered out [here](https://github.com/colinhacks/zod/blob/v4/packages/core/src/schemas.ts#L2627-L2630):

```typescript
// packages/core/src/schemas.ts
export const $ZodEnum: core.$constructor<$ZodEnum> = {
  // ...

  const values = Object.entries(def.entries)
    // remove reverse mappings
    .filter(([k, _]) => {
      return typeof def.entries[def.entries[k]] !== "number";
    })
    .map(([_, v]) => v);
  
  // ...
}
```

I've got the project pulled down locally, I'll see if I can push up a change. I'm not totally sure what the intended behavior is here though, so as a first iteration I'll probably just remove it and see what feedback I get in review.

ahrjarrett:
Closes #4353 

Looking for feedback on the intended behavior here.

At first I tried to be clever and only remove the value if the match went both ways (e.g. `{ a: 1, 1: 'a' }`, but that caused tests in `packages/zod/tests/discriminated-unions.test.ts` to fail, so I pushed up the least clever thing I could think of, to get more info before spending more time on it.


colinhacks:
Fixed in https://github.com/colinhacks/zod/commit/6d3aa58325b46e42e01213ab231b66aa9ba62f72

We need to do this weird filtering in order to handle native TypeScript `enum`s. Just needed to get a little more clever with it. Good catch!

PR #4368, opened 2025-05-15
Add `jitless` flag to avoid `eval`/`new Function`

colinhacks:
Run this before any Zod code:

```ts
import { z } from "zod/v4"

z.config({ jitless: true })
```

This short-circuits all code paths that would call `eval`/`new Function`. 

Some runtimes (Next middleware) seem to detect usage of these APIs during build-time. Fixing that will require something more involved.


robfig:
LGTM

PR #4374, opened 2025-05-15
fix: use JSON.stringify to escape toDotPath keys w/ quotes

mmkal:
This should also correctly handle even weirder stuff like newlines in proeprty keys. JSON.stringify should do all the right escaping - if it's a valid JSON string, it's a valid key.

Note: one existing test was updated:

```ts
expect(z.core.toDotPath(["data[0]", "value"])).toMatchInlineSnapshot(`"data[0].value"`); // <- before
expect(z.core.toDotPath(["data[0]", "value"])).toMatchInlineSnapshot(`"["data[0]"].value"`); // <- after
```

This looks correct to me: Previously it was implying a shape of
`{data: [{value: '...'}]}` when actually it should be a shape of
`{'data[0]': {value: '...'}}`

### DIFF
--- a/packages/zod/src/v4/classic/tests/error-utils.test.ts
+++ b/packages/zod/src/v4/classic/tests/error-utils.test.ts
@@ -399,6 +399,14 @@ test("z.toDotPath", () => {
     `"user.$special["Symbol(#symbol)"]"`
   );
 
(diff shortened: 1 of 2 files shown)


colinhacks:
Definitely better, thanks!

PR #4379, opened 2025-05-16
Allow generic on coerce functions

colinhacks:
Support an optional generic parameter on `z.coerce.string()`, etc.

```ts
const a = z.coerce.string<any>();
type Ain = z.input<typeof a>; // any

const b = z.coerce.string();
type Bin = z.input<typeof a>; // unknown
```

This also lets TypeScript set the input type appropriately in inference-constrained environments.

```
import { z } from "zod/v4";

z.pipe(z.string(), z.coerce.number());
```


erlandsona:
Alright 🤘🙌💪💯

PR #4382, opened 2025-05-16
Fix: $ZodCatch schema catchValue handling on validation error

chimame:
## Overview

Fixed a bug in handling the error value of `catchValue` in the pipe schema.

## Details

The value defined in catch is not set in the following schema.

```ts
const stringWithDefault = z
  .pipe(
    z.transform((v) => (v === "none" ? undefined : v)),
    z.string()
  )
  .catch("default");

expect(stringWithDefault.parse(undefined)).toBe("default"); // <-- The value is set to `undefined`.
```


colinhacks:
Yikes. Good catch, thanks for the PR!

PR #4416, opened 2025-05-19
fix: no json-schema default for piped type w/ input=io

mmkal:
this was a minor bug (I believe) in `toJsonSchema`. Basically this:

```ts
const StringLength = z.string().transform(s => s.length).default(5)
z.toJSONSchema(StringLength, { io: 'input' })
```

should result in `{type: 'string'}` but instead it's `{type: 'string', default: 5}` which doesn't make sense - the default of `5` is internal to zod's piping process, when the user asks for input: 'io', they shouldn't care that zod will apply a default when it transforms the string into a number.

Ideally there would be some way of at least documenting "btw this is going to be transformed, and when it's transformed if you pass undefined it'll default to 5" but I think when passing `input: 'io'`, that transformation is essentially private. Someone reading the json schema should probably know it's optional somehow though?

there was actually a test that caught this, previously we were ending up with nonsenical defaults like the below (see test diff)

```
{
  "$schema": "https://json-schema.org/draft-2020-12/schema",
  "default": 1234,
  "type": "string",
}
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Adjusted JSON schema generation to include default values only in output schemas when appropriate, improving accuracy for schemas with default values.

- **Tests**
  - Updated test cases to align with the revised handling of default values in JSON schema output.

### DIFF
--- a/packages/zod/src/v4/classic/tests/json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/json-schema.test.ts
@@ -1643,7 +1643,6 @@ test("defaults/prefaults", () => {
   expect(toJSONSchema(c, { io: "input" })).toMatchInlineSnapshot(`
     {
       "$schema": "https://json-schema.org/draft-2020-12/schema",
-      "default": 1234,
       "type": "string",
     }
   `);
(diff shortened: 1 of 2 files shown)


mmkal:
Ah got it. Yeah it occurred to me after this PR it'd bubble up already but hadn't explicitly checked. Yeah R and E and then what are the letters are there?

PR #4420, opened 2025-05-19
Fix copy/paste v4 docs typo

transitive-bullshit:
loving zod v4 :) just a small doc fix

## Summary by CodeRabbit

- **Documentation**
  - Updated documentation to clarify that the `inferSchema` function now only accepts string schemas, not object schemas.


colinhacks:
Thanks!

Issue #4418, opened 2025-05-19
ZodError no longer has a message


kibertoad:
3.25.3, v4 mode.

```ts
z.string().parse({})
```

Previously the following regex for an error would pass - `/Expected string, received number/`

But now error thrown has `message` set to `""` for whatever reason (blank string).

Is there a reason for this behaviour? It is hard to see it as an improvement, immediate visibility of issues has suffered.

colinhacks:
Yeah sorry, this change was too much breakage for not enough benefit. Fixed in latest. 

https://github.com/colinhacks/zod/pull/4424

kibertoad:
Thank you for addressing it so swiftly!

PR #4424, opened 2025-05-19
Revert `Error` inheritance changes

colinhacks:
This change was too breaky. While avoiding `Error` speeds up the error path for `.safeParse` by around 2.4x there are other ways to expose this behavior would without breaking everyone's snapshot tests. There's no obvious way to get `.toThrowErrorMatchingInlineSnapshot` to play nice with an error that doesn't extend `Error`.


colinhacks:
Closes https://github.com/colinhacks/zod/issues/4418

Issue #4429, opened 2025-05-20
[v4]: Enum doesn't have `type` in toJSONSchema


legendhimself:
Enum should have number or string as a type in the json schema.

For native enums, it should also include `x-enumNames`

```ts
z.object({
	id: z.string(),
	role: z.enum(['Admin', 'Member', 'Owner']),
	joinedAt: z.string(),
	name: z
		.object({
			firstName: z.string(),
			lastName: z.string().optional(),
		})
		.optional(),
	avatar: z.string().optional(),
	mobileNumber: z.string().optional(),
});
```

For this schema it generates
```json
{
        "type": "object",
        "properties": {
          "id": {
            "type": "string"
          },
          "role": {
            "enum": [
              "Admin",
              "Member",
              "Owner"
            ]
          },
          "joinedAt": {
            "type": "string"
          },
          "name": {
            "type": "object",
            "properties": {
              "firstName": {
                "type": "string"
              },
              "lastName": {
                "type": "string"
              }
            },
            "required": [
              "firstName"
            ]
          },
          "avatar": {
            "type": "string"
          },
          "mobileNumber": {
            "type": "string"
          }
        },
        "required": [
          "id",
          "role",
          "joinedAt"
        ]
      }
```

PR #4430, opened 2025-05-20
fix: enum type in json schema

imranbarbhuiya:
fixes: #4429

Since native enum can have number values too so I added a check to see typeof first value. Only first value check if enough as enums generate value first then name

### DIFF
--- a/packages/zod/src/v4/classic/tests/json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/json-schema.test.ts
@@ -597,6 +597,12 @@ describe("toJSONSchema", () => {
           "b",
           "c",
         ],
+        "type": "string",
+        "x-enumNames": [
+          "a",
+          "b",
+          "c",
+        ],
       }
     `);
   });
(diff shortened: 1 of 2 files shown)


jeremyfiel:
declaring the `type` keyword with an enum is redundant and usually wrong.  the enum itself declares the types. 

An enum can have any number of primitive types in the same definition.  by declaring `type`,  you're incorrectly constraining the property to that `type`

take this example: 

```json
{
  "type": "string",
  "enum": [
    1,
    3
  ]
}
```

If you pass the number value `1` as a data instance, it fails for `expected type: string` but, clearly the enum allows `1` or `3`

imranbarbhuiya:
Oh I see. I was expecting it to match [openapi schema](https://swagger.io/docs/specification/v3_0/data-models/enums/) too while this fn gives json schema but it's a valid openapi schema too so all the fields it provides works with openapi schema except enum due to missing `type`. Do you think keeping it will be invalid or just redundant? so it solves openapi schema too

jeremyfiel:
Your link to Swagger documentation is slightly misleading and instead I recommend you consult the official OpenAPI Specification. 

https://spec.openapis.org/oas/v3.0.4#json-schema-keywords  
This definition refers to the official JSON Schema Specification below

https://json-schema.org/draft-04/draft-fge-json-schema-validation-00#rfc.section.5.5.1

the `type` keyword is not required by OAS or JSON Schema. It's not necessary to keep it because it will constrain an enum definition with multiple types, incorrectly.

PR #4436, opened 2025-05-20
Human friendly error messages

RobinTail:
In v4 I suggest to make `ZodError::message` more readable and human friendly by utilizing the already implemented `prettifyError()` in its getter.

image

In order to preserve the observability of `issues` that used to be serialized into snapshots before, I added a custom serializer into vitest configuration.

Related to #4418 and #4424 

I also think that those symbols like ❌ and ➡️ are too opinionated, it should be the responsibility of an UI to decide if such icons are required and/or which ones to use. But that could be another PR.

## Summary by CodeRabbit

- **New Features**
  - Introduced improved error snapshot formatting for validation errors, providing clearer and more structured error messages in test outputs.

- **Bug Fixes**
  - Enhanced error message readability by switching to a human-friendly, prettified format instead of raw JSON.

- **Chores**
  - Updated test configuration to use a custom snapshot serializer for consistent error representation across all tests.
  - Removed deprecated function and updated example code to compare error outputs between two library versions.


packages/zod/src/v4/core/errors.ts
```diff
@@ -193,7 +193,7 @@ const initializer = (inst: $ZodError, def: $ZodIssue[]): void => {
   // inst.message = JSON.stringify(def, util.jsonStringifyReplacer, 2);
   Object.defineProperty(inst, "message", {
     get() {
-      return JSON.stringify(def, util.jsonStringifyReplacer, 2);
+      return prettifyError(inst);
```
RobinTail, line 196: this 👐🏽

colinhacks:
I'm afraid I'm avoiding this intentionally to avoid pulling `prettifyError` code into the bundle for Zod Mini. It's a lot of code for a small win, in my opinion. I'm not sure I like it more than the JSON.stringify version actually, since the alignment of the multi-line message gets messed up by Node's default error printer:
 
Screenshot 2025-05-27 at 19 14 27

Issue #4148, opened 2025-04-14
v4: Error - "Cannot call a class as a function" in React Native.


mcculloughrt-svh:
Building a schema with zod v4 (for example simply `const str = z.string()`, any schema construction method will cause the error) in an Expo project throws a "Cannot call a class as a function" error. 

I was able to confirm a minimal repro by using `npx create-expo-app@latest`, adding zod v4, and adding the above string parse line to the home component function body, and nothing else. 

There's no indication of deeper error source. No linter error occurs, and it's such a basic functionality that I know works in other environments I suspect something fishy is happening with the Metro bundling process and the $constructor function from @zod/core.

chungweileong94:
This seems like a dual package hazard issue, and since RN 0.79 the `unstable_enablePackageExports` option is enable by default https://github.com/facebook/metro/pull/1448.

I created a repro with Expo SDK 53, which came with RN 0.79. As soon as you import zod, it starts throwing the error. If you try to opt-out of `unstable_enablePackageExports`, it will not be able to resolve some of the modules.
https://github.com/chungweileong94/RN-zod-v4-bug

rikbrown:
Is there any progress or other workarounds for this? We're also seeing this issue and I'm surprised no one else is hitting it with React Native.

chungweileong94:
> Is there any progress or other workarounds for this? We're also seeing this issue and I'm surprised no one else is hitting it with React Native.

So far based on what I'm noticing is that you will hit the (similar) issue if you are using:
- Zod v3 + RN with `unstable_enablePackageExports`
- Zod v4 + RN

Both of these combinations are considered pretty new, `unstable_enablePackageExports` only got enabled by default in the recent RN/Expo release, and usually RN dev don't upgrade that often, so it makes sense that not many people are hitting this issue just yet. As for Zod v4, it's beta🤷

For now, my workaround is to stay on Zod v3, and make sure the `unstable_enablePackageExports` is off.

joshuayoes:
I'm also experiencing this. I tried several combinations of package exports but none of them seemed to resolve my issue. I tried:
- Adding a `react-native` export key for all fields to the commonjs build
- Removing all the export keys to force it to go to commonjs
- `config.resolver.unstable_enablePackageExports = false` reveals there is an issue with parsing `@zod/core`.
- I also tried an option where I resolved only the source development build, but the `.js` extensions in the the source files in `@zod/core` could not be resolved by Metro.

jcppman:
[...]

PR #4466, opened 2025-05-21
fix: React Native / Hermes issues

boorad:
#4148 shows some of the errors folks are seeing when using Zod v4 in a React Native / Expo app.  This PR may not be fully merged, but showcases the issue inside of an app, where we are guaranteed to use the Hermes engine.

I have added a `safeCreateSchema()` function to `util.ts` that safely creates schema classes unlike `$constructor` in `core.ts`.  I have only patched `string` to use this new util, and you can see that in the last commit.  

The problem is that this fix, as it stands now, is super-invasive, and requires changing all of the places where Zod calls `new Class(def)`.  These calls are all in `schemas.ts` and `api.ts`.

I've spent the morning trying to get a patch to `$constructor` that makes Hermes happy, keeps the surface area of this change small, and still passes the Zod test suite.  But I think it's time to get @colinhacks brain on this.

----
Here's what LLMs say about this issue:

Why does Hermes not like new Class and give the "Cannot call a class as a function" error?
Hermes is stricter than V8 about ES6 class semantics. Specifically:

- It enforces that ES6 classes must be called with new. If you call a class constructor as a plain function (without new), Hermes throws "Cannot call a class as a function".
- If you try to "fake" class instantiation (e.g., by using Object.create/Class.apply/Class.call), or if the prototype chain is not set up exactly as Hermes expects, you can also get this error.
- Hermes also does not tolerate certain patterns that V8/VMS allow, such as "re-instantiating" a class instance by calling its constructor manually, or using proxies to intercept the new operator.

In Zod’s case:
If your $constructor or any Zod schema class is instantiated in a way that:

- Calls the constructor function directly (not with new), or
- Uses a pattern that breaks the prototype chain or confuses the [[Construct]]/[[Call]] distinction, Hermes will throw this error, even if V8 allows it.

----

Here's the test app with only `string` patched.  `object` still exhibits the issue.

simulator_screenshot_89D37CB4-7ACB-4E58-AF9F-0F573026EF4E

## Summary by CodeRabbit

- **New Features**
  - Introduced a new Expo-based mobile app for testing Hermes compatibility with Zod, featuring themed UI, tab navigation, and a suite of interactive tests.
  - Added custom components for collapsible sections, external links, parallax scroll views, themed text and views, animated icons, and adaptive tab bar backgrounds.
  - Implemented dynamic theming with light and dark mode support.

- **Chores**
  - Added project configuration files including TypeScript, ESLint, Metro bundler, and package management.
  - Provided scripts for project reset and starter templates.
  - Included documentation and setup instructions for developers.

- **Bug Fixes**
  - Improved schema instantiation logic in Zod to enhance compatibility with the Hermes JavaScript engine.


boorad:
fixed in #4482

colinhacks:
Fix landed in stable: `zod@3.25.24`

PR #4482, opened 2025-05-22
fix: rewrite $constructor to be compatible with Hermes

### LINKED ISSUE #4148: v4: Error - "Cannot call a class as a function" in React Native.

gdorsi:
fixes #4148

This PR fixes the compat issues with Hermes by using a function as a class wrapper in $constructor

Tested with the repro provided in https://github.com/colinhacks/zod/pull/4466

EDIT: We are using this fork in [jazz](https://github.com/garden-co/jazz) and can confirm that it solved our issues with React Native

## Summary by CodeRabbit

- **Refactor**
  - Improved the internal initialization and construction process for certain components, resulting in a more modular and streamlined structure. No changes to the public interface or user-facing functionality.

### DIFF
--- a/packages/zod/src/v4/core/core.ts
+++ b/packages/zod/src/v4/core/core.ts
@@ -14,42 +14,52 @@ export /*@__NO_SIDE_EFFECTS__*/ function $constructor<T extends ZodTrait, D = T[
   initializer: (inst: T, def: D) => void,
   params?: { Parent?: typeof Class }
 ): $constructor<T, D> {
+  function init(inst: T, def: D) {
+    Object.defineProperty(inst, "_zod", {
+      value: inst._zod ?? {},
+      enumerable: false,
+    });
+    // inst._zod ??= {} as any;a
+    inst._zod.traits ??= new Set();
+    // const seen = inst._zod.traits.has(name);
+
+    inst._zod.traits.add(name);
+    initializer(inst, def);
+    // support prototype modifications
+    for (const k in _.prototype) {
+      Object.defineProperty(inst, k, { value: (_.prototype as any)[k].bind(inst) });
+    }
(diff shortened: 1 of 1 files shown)


packages/zod/src/v4/core/core.ts
```diff
@@ -14,42 +14,52 @@ export /*@__NO_SIDE_EFFECTS__*/ function $constructor<T extends ZodTrait, D = T[
...
-      return inst?._zod?.traits?.has(name);
+  const Parent = params?.Parent ?? Object;
+  class Definition extends Parent {}
+
+  function _(this: any, def: D) {
+    if (params?.Parent) {
+      return construct(new Definition(), def);
     }
+
+    return construct(this, def);
```
gdorsi, line 53: Using the if here to reuse the instance when the parent is not defined.

Can be simplified in doing always `return construct(new Definition(), def);` didn't perf-test the difference honestly.

colinhacks:
Landed in zod@beta, can you try & report back?

rdnsan:
> Landed in zod@beta, can you try & report back?

it works for me. using `zod@3.25.24-beta.0`

colinhacks:
Landed in stable: `zod@3.25.24`

PR #4518, opened 2025-05-25
v4: Add `discriminator` to Union Def type

samchungy:
I need to access the `discriminator` field from within toJSONSchema to generate some OpenAPI specific discriminator fields and I noticed this wasn't available on the type definition

## Summary by CodeRabbit

- **New Features**
  - Added support for an optional discriminator property in union schema definitions.

### DIFF
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -1852,6 +1852,7 @@ export type $InferUnionInput<T extends $ZodType> = T extends any ? core.input<T>
 export interface $ZodUnionDef<Options extends readonly $ZodType[] = readonly $ZodType[]> extends $ZodTypeDef {
   type: "union";
   options: Options;
+  discriminator?: string;
 }
 
 export interface $ZodUnionInternals<T extends readonly $ZodType[] = readonly $ZodType[]>


jeremyfiel:
Couple details here with the `discriminator` keyword and OpenAPI and JSON Schema history

OAS 3.0.x added this keyword to help with code generators and discriminated unions but this keyword is NOT official JSON Schema and is not supported in any way in JSON Schema draft-04, which OAS 3.0.x is based upon. 

The upgrade to OAS 3.1.x moves to JSON Schema draft 2020-12, but this keyword is still not added to the spec, as it's only defined by OpenAPI and indeed was deprecated in OAS 3.1.x in favor of native JSON Schema. The tagged union can be achieved with `if, then` keyword conditional statements. 

imo, adding `discriminator` to the `toJsonSchema` definition is inherently wrong because it was never part of the specification.

samchungy:
> Couple details here with the `discriminator` keyword and OpenAPI and JSON Schema history
> 
> OAS 3.0.x added this keyword to help with code generators and discriminated unions but this keyword is NOT official JSON Schema and is not supported in any way in JSON Schema draft-04, which OAS 3.0.x is based upon.
> 
> The upgrade to OAS 3.1.x moves to JSON Schema draft 2020-12, but this keyword is still not added to the spec, as it's only defined by OpenAPI and indeed was deprecated in OAS 3.1.x in favor of native JSON Schema. The tagged union can be achieved with `if, then` keyword conditional statements.
> 
> imo, adding `discriminator` to the `toJsonSchema` definition is inherently wrong because it was never part of the specification.

To be clear, I'm not proposing we map it by default if that is what you're concerned about, just exposing the value which is already present via the TypeScript type so libraries like mine can go in an declare an override to use it in a discriminator and/or if/then statement 😃

jeremyfiel:
i was more concerned with exposing a keyword that isn't supported by JSON Schema in the type definition. I think it misrepresents JSON Schema support of this keyword.

samchungy:
That's fair enough, I appreciate the context though as it's very relevant for me.

It still seems like it's available in [3.1x](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.1.md#discriminator-object) for now and has been postedponed [to be deprecated in 4.x](https://github.com/OAI/OpenAPI-Specification/issues/2143#issuecomment-2070156342)?

By the way, this definition isn't just used by the JSON Schema function, it's used by any library trying to introspect the schema and is part of the recommended way to [walk](https://zod.dev/packages/core?id=internals) the schema.

jeremyfiel:
Ah, you're right with the delayed deprecation.  👍 

I'm not sure where else it would make sense to implement support for it, but I still don't agree with using toJsonSchema..  haha

colinhacks:
The `discriminator` key is available for `$ZodDiscriminatedUnion` instances. This is a subclass of `$ZodUnion` and includes all the discrimination logic. It isn't included in the `$ZodTypes` union as it's a bit of special case - you'll have to do a special check within your `"union"` handler for it 👍

colinhacks:
cc @samchungy this should make introspection cleaner re: https://github.com/colinhacks/zod/pull/4518

PR #4542, opened 2025-05-28
fix: basics.mdx error typo

akuul:
`if` clause was using `error` naming, but binding was called `err`. Renamed to `error` to follow more of Zod error naming


colinhacks:
Good eye. Thanks!

PR #4557, opened 2025-05-29
Remove defaults from JSON Schema for transforming schemas

colinhacks:


colinhacks:
cc @mmkal @samchungy — landed in `zod@3.25.35`

PR #4559, opened 2025-05-29
feat: Configurable references for `toJSONSchema()`

### LINKED ISSUE #4281: v4: toJSONSchema: inability to customize references
Even when using an external registry, external storage for the definitions and `uri` function returning a string

```ts
{
  external: { registry: z.registry(), defs: {}, uri: () => "TEST" }
}
```

the references produced by `toJSONSchema` looks this way:

```json
{
  "$ref": "TEST#/$defs/schema0"
}
```

- what if I wanna give those references a naming of my choice (`TEST` instead of `schema0`) ? 
- if the `external/defs` is used then why `$ref` still refers via the "document root" (`#`) ? There is even no `$defs` in that case.
- what if the produced JSON Schema is not the final document itself, but a much deeper part of an actual document (OpenAPI), so that the "document root" is beyond the awareness of `toJSONSchema` and therefore `external/defs` is what I wanna use to manage it myself ?

### Suggestion

I suggest `uri` (or similar hook) to be able to replace the whole `$ref` string, instead of prefixing it, so that I could give it a name and adjust its path:

```
#/components/schemas/MyFancySchema
```

RobinTail:
Closes #4281 

This should simplify using `toJSONSchema()` for making tools and libraries that utilize the result as a part of a larger document having its own system of references.

The problem it solves:
- self-references: `{ref: "#"}` — in OpenAPI it points to the document root, not to a schema inside of it
- provides a way to define specify references inside another document;

What's changed
- Introducing `internal` property (as an opposite to `external`) — the naming is quesitonable, but the idea is that it's only applicable when `external` is not set;
  - there is a `makeRef()` property inside of it used to produce both `defId` (optional) and `ref`;
  - the default logic retained as a fallback when `internal` is not defined;
  - library authors can customize that function in order to make references relative to (for example) an OpenAPI document containing the subject-schema, such as: `#/components/schemas/Schema(N)`, returning also `Schema(N)` as the `defId`:
    - thus, the result would contain `$defs` having the chosen naming and the whole object then could be placed into the final document.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -1954,13 +1954,13 @@ test("use output type for preprocess", () => {
   `);
 });
 
-describe("internal parameter", () => {
-  const oasDoc = new Map<z.core.$ZodType, { name: string; depiction: z.core.JSONSchema.BaseSchema }>();
-  const makeRef = (zodSchema: z.core.$ZodType, jsonSchema: z.core.JSONSchema.BaseSchema) => {
+describe("internal parameter: custom makeRef() function", () => {
+  const oasDoc = new Map<z.core.$ZodType, string>();
+  const makeRef = (zodSchema: z.core.$ZodType) => {
     if (!oasDoc.has(zodSchema)) {
-      oasDoc.set(zodSchema, { name: `Schema${oasDoc.size + 1}`, depiction: jsonSchema });
+      oasDoc.set(zodSchema, `Schema${oasDoc.size + 1}`);
     }
-    const { name } = oasDoc.get(zodSchema)!;
+    const name = oasDoc.get(zodSchema)!;
     return { defId: name, ref: `#/components/schemas/${name}` };
   };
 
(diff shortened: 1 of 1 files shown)


packages/zod/src/v4/classic/tests/to-json-schema.test.ts
```diff
@@ -1954,6 +1954,215 @@ test("use output type for preprocess", () => {
...
+      subcategories: Category[];
+    }
+
+    const categorySchema: z.ZodType<Category> = z.object({
+      name: z.string(),
+      subcategories: z.array(z.lazy(() => categorySchema)),
+    });
+
+    const { $defs, ...rest } = z.toJSONSchema(categorySchema, { internal: { makeRef } });
+    expect($defs).toMatchInlineSnapshot(`
```
RobinTail, line 1979: this could go directly to OpenAPI document / components / schemas

packages/zod/src/v4/core/to-json-schema.ts
```diff
@@ -560,12 +567,23 @@ export class JSONSchemaGenerator {
...
       // unrepresentable: _params?.unrepresentable ?? "throw",
       // uri: _params?.uri ?? ((id) => `${id}`),
       external: _params?.external ?? undefined,
+      internal: _params?.internal ?? {
+        makeRef: (_, jsonSchema, isSelf) => {
+          const uriPrefix = `#`;
+          if (isSelf) return { ref: uriPrefix };
+          const defUriPrefix = `${uriPrefix}/${defsSegment}/`;
+          const defId = jsonSchema.id ?? `__schema${this.counter++}`;
+          return { defId, ref: defUriPrefix + defId };
```
RobinTail, line 584: the default logic retained

Issue #4326, opened 2025-05-05
v4: `z.object().check()` types the `value` property as `any`


bryanforbes:
Using `@zod/mini` version `4.0.0-beta.20250505T195954`, the the following code has incorrect type information for the check function:

```typescript
const Thing = z.object({
  name: z.string(),
  age: z.number()
}).check(({ value, issues }) => {
});

type ThingOutput = z.infer<typeof Thing> // { name: string; age: number }
type ThingInput = z.input<typeof Thing> // { name: string; age: number }
```

<img width="265" alt="Image" src="https://github.com/user-attachments/assets/85f9ed00-4b7d-4cfe-b6ca-763cc99b753d" />

With the previous beta, `z.interface()` would give the correct type for `value`.

bryanforbes:
@rphlmr Thank you for fixing that as well as providing a work-around!

bryanforbes:
For anyone else experiencing this, I adapted the solution @rphlmr gave to my application. I have one schema module and added this to the top of that module:

```typescript
import * as z from '@zod/mini';

// TODO: remove after https://github.com/colinhacks/zod/pull/4568 lands
declare module '@zod/mini' {
  export interface ZodMiniType {
    check(
      ...checks: (z.core.CheckFn<z.core.output<this>> | z.core.$ZodCheck<z.core.output<this>>)[]
    ): this;
  }
}
```

PR #4568, opened 2025-05-30
Fix type inference in ZodMiniType check method

rphlmr:
Closes: https://github.com/colinhacks/zod/issues/4326

`.check()` on a `z.object` now correctly infers `value`.

**Before:**
```ts
const Thing = z.object({
  name: z.string(),
  age: z.number()
}).check(({ value, issues }) => {
  value // => any
});
```

**After:**
```ts
const Thing = z.object({
  name: z.string(),
  age: z.number()
}).check(({ value, issues }) => {
  value // => { name: string; age: number }
});
```

### DIFF
--- a/packages/docs/content/error-formatting.mdx
+++ b/packages/docs/content/error-formatting.mdx
@@ -150,7 +150,6 @@ This returns the following string:
 
 ## `z.flattenError()`
 
-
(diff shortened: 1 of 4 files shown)


colinhacks:
Nice catch, thanks!!

PR #4584, opened 2025-05-31
minor grammar typo fix

sculli-net:
Noticed a tiny typo. 

Changed 
> If you only care about validating inputs, _omit the `output` field is optional._

to

> If you only care about validating inputs, _**you can** omit the `output` field **as it** is optional._

Tried to keep your prose style intact.


colinhacks:
Thanks!

PR #4587, opened 2025-06-01
fix(docs): remove z.literal(Symbol)

jeremy-code:
Zod 4 (colinhacks#4074)

- Remove `z.literal(Symbol("terrific"));` from docs in api.mdx

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -89,7 +89,6 @@ const tuna = z.literal("tuna");
 const twelve = z.literal(12);
 const twobig = z.literal(2n);
 const tru = z.literal(true);
-const terrific = z.literal(Symbol("terrific"));
 ```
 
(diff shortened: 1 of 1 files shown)


jeremy-code:
I believe this is correct because #4074 is anything to go by, the docs should default to Zod v4 (hence [metadata.mdx](https://github.com/colinhacks/zod/blob/main/packages/docs/content/metadata.mdx)). Per [changelog](https://zod.dev/v4/changelog?id=zliteral), 

> [z.literal()](https://zod.dev/v4/changelog?id=zliteral)
> [drops symbol support](https://zod.dev/v4/changelog?id=drops-symbol-support)
> Symbols aren't considered literal values, nor can they be simply compared with ===. This was an oversight in Zod 3.

colinhacks:
Ah good catch!

PR #4589, opened 2025-06-01
fix: allow coerce date in to-json-schema

imranbarbhuiya:
We often need to take date in apis so we use `z.coerce.date()` to allow string input which gets converted to date. So `coerce` can be allowed as date-time string without any pattern

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -58,6 +58,13 @@ describe("toJSONSchema", () => {
         "type": "string",
       }
     `);
+    expect(z.toJSONSchema(z.coerce.date())).toMatchInlineSnapshot(`
+      {
+        "$schema": "https://json-schema.org/draft/2020-12/schema",
+        "format": "date-time",
+        "type": "string",
+      }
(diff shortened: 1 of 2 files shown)


colinhacks:
I'm against this, as you should be using `z.iso.datetime()` to achieve this. The JSON Schema should reflect the same type as the Zod validator. `z.date()` doesn't validate datetime strings, it validates `Date` instances (which have no JSON equivalent). `z.coerce.date()` technically accepts almost anything including numbers (timestamps), non-ISO date strings, and other dates. 

I'm commited to keeping `z.toJSONSchema()` strict when it comes to unrepresentable types, especially where's already an alternative (`z.iso.date()`) that's exactly what you're looking for.

imranbarbhuiya:
Will you accept a pr to allow `unrepresentable` to be a fn (or another option name) which returns a custom value for cases where I cant update the source of the schema, I can instead pass a custom fn to to-json and make it work. This isnt just for date but in other cases too. Currently, it either throws or makes it any where the context gets lost and we wont know the type so can't override after the json gets generated. Allowing an fn can solve other user's custom requirements too. If yes, should I just make `unrepresentable` fn or add a new option and what should I name that option?

Issue #4249, opened 2025-04-20
V4: toJSONSchema z.literal is missing "type"


marclp-es:
I am using toJSONSchema to generate an openapi schema for gemini and it is complaining that "type" is missing when using z.literal() compared to using v3 with zod-to-json-schema, I am not sure if this is intended, an oversight, or google's APIs are very picky. Here is an example of a simple json schema produced in v3 vs the same in v4 using z.toJSONSchema.

V3:
`
{
  "type": "object",
  "properties": {
    "test": {
      "type": "string",
      "const": "test"
    }
  },
  "required": [
    "test"
  ],
  "additionalProperties": false,
  "$schema": "http://json-schema.org/draft-07/schema#"
}
`

V4:
`
{
  "type": "object",
  "properties": {
    "test": {
      "const": "test"
    }
  },
  "required": [
    "test"
  ]
}
`

samchungy:
`const` is technical valid by itself so it seems Google is picky. In my own library I made the decision to always include `type` as a number of different libraries which rely on spec aren't always compliant. But to be fair to them, the spec can be iffy at times.

PR #4577, opened 2025-05-30
fix(JSONSchema): add type for enum

imranbarbhuiya:
Adds `type` for enum in `toJSONSchema` to make it more swagger/openapi friendly.
Redo of #4430 without x-enumNames
Fixes: #4429

ref: https://x.com/colinhacks/status/1928526844397043787

PR #4590, opened 2025-06-01
fix: add type to literal enum

imranbarbhuiya:
followup of: #4577
fixes: #4249
probably also fixes: https://github.com/vercel/ai/pull/6421#issuecomment-2925378022

Since google and other packages like swagger expects a `type` in it so adding support for it will be better I think


colinhacks:
Thanks! Landed in 3.25.49.

PR #4591, opened 2025-06-01
fix(v4): reflect inclusive boundaries in minLength/maxLength issue objects

wancup:
This updates the `minLength` and `maxLength` errors in v4 to include the `inclusive` property. Currently, I believe validations like `z.string().max()` use inclusive boundaries (https://github.com/colinhacks/zod/blob/50e9afb5606018ecacd6c64aee90f4f8006d851b/packages/zod/src/v4/core/checks.ts#L619), but the default error messages do not seem to reflect this behavior.
In v3, validations like `z.string().max()` also used inclusive boundaries, and the issue objects correctly indicated this. 

If this change was intentional as part of the migration to v4, or if I have misunderstood the current behavior, please feel free to close this pull request!


colinhacks:
Good catch, thanks!

PR #4602, opened 2025-06-02
v4(toJSONSchema): Add `includes` position check

samchungy:
Adds an extra regex check marker to includes when `position` isa added


colinhacks:
Ooo nice.

colinhacks:
Landed in `zod@3.25.49`

PR #4617, opened 2025-06-04
fix(v4): add exact to length check issue

IRelaxxx:
Add the ```exact: true``` property to the issues generated by the `length` check.

Test case:

```typescript
import z3 from "zod/v3";
import z4 from "zod/v4";

const schema3 = z3.string();
const schema4 = z4.string();
const mess3 = schema3.length(5).safeParse("aabcdef");
const mess4 = schema4.length(5).safeParse("abcdef");

console.log("mess3", mess3.error?.issues[0]);
console.log("mess4", mess4.error?.issues[0]);
```

Results in this on main:
```
mess3 {
  code: 'too_big',
  maximum: 5,
  type: 'string',
  inclusive: true,
  exact: true,
  message: 'String must contain exactly 5 character(s)',
  path: []
}
mess4 {
  origin: 'string',
  code: 'too_big',
  maximum: 5,
  path: [],
  message: 'Too big: expected string to have <5 characters'
}
```
With this PR:
```
mess3 {
  code: 'too_big',
  maximum: 5,
  type: 'string',
  inclusive: true,
  exact: true,
  message: 'String must contain exactly 5 character(s)',
  path: []
}
mess4 {
  origin: 'string',
  code: 'too_big',
  maximum: 5,
  exact: true,
  path: [],
  message: 'Too big: expected string to have <5 characters'
}
```

This PR does not change the default error messages but exposes the ```exact``` property for custom error message generation.

### DIFF
--- a/packages/zod/src/v4/classic/tests/array.test.ts
+++ b/packages/zod/src/v4/classic/tests/array.test.ts
@@ -49,6 +49,7 @@ test("array length", async () => {
     [
       {
         "code": "too_small",
+        "exact": true,
         "message": "Too small: expected array to have >2 items",
         "minimum": 2,
         "origin": "array",
@@ -63,6 +64,7 @@ test("array length", async () => {
     [
       {
         "code": "too_big",
+        "exact": true,
         "maximum": 2,
         "message": "Too big: expected array to have <2 items",
         "origin": "array",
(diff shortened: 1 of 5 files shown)


colinhacks:
This PR does change the default error messages—any changes to the `en.ts` locale changes default messages for Zod. This would qualify as a breaking change (though it's borderline) and I don't think it's worth it in this case. I would accept a PR adding the `exact` flag to these issue types (which seems like it was probably your intention).

IRelaxxx:
I have reverted the changes to the error messages. This PR now only adds the exact flag.

colinhacks:
Thanks Alexander 👍👍

Issue #4625, opened 2025-06-05
`.optional()` breaks type inference for mutual recursive types in v4


romandecker:
When implementing a mutual recursive type as suggested in the official docs for v4, making one part of the recursion optional will break type inference:

TS Playground link: https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgLzgXzgMyhEcDkyEAJgPQBuALPgNwBQdApgB6SxwDGEAdgM7y8ANhBgBlDgAtGIAIZwAvCjpw4AOggAjAFaMOMABQJlKuEJEBhEowBcSkydX8owbgHN9ASgA0De68bwGsIcANa8nojG9lABAK5Q3CiqMlBQMgCe+kEQoeJSsh709mjGaB4MdFx88DK2AIIKcNyxgoJwMrzt3On0TKzQ8FX8cNm5ktJyisjq2roGRiqjIZbENklOLu7exv4CwjDhHpH2pKRwMTDxidMpaZlmYuMFNPavr6dwAO4QLcRf0CEvCNYvBPowvjJuPAAAYPXjQuAwCAjcEQMAwYA8GSCKLnOIJJK3DL6B55CYedTozHcbGeF7vM4uTCMKCdWH7eHtTqxbghbgQT6JDSMYSfXEfC5XJJojFYwT6G6pYmkp4yDyFFQfDgyWK8RidOQwdJgcEsnBQUo+Mq9PpsQY8YZgGT+MmyRrIYwzHR6Qy4uG2RV3En7V1qnwqa0MIbwGD6+CKJ0u1WqJ2sxj6ZqtQp0WP8RwcugfAB6AH4gA

```ts
import { z } from 'zod/v4';

export const slotSchema = z
  .object({
    slotCode: z
      .string(),

    get blocks() {
      return z.array(blockSchema);
    }
  })


const a: A = null as any;

export const blockSchema = z.object({
  blockCode: z.string(),
  get slots() {
    // return z.array(slotSchema);              // would work, but we want `slots` to be optional
    return z.array(slotSchema).optional();      // infers `slots` as unknown below
    // return z.optional(z.array(slotSchema));  // causes a type error
  },
});


export const pageSchema = z
  .object({
    slots: z.array(slotSchema),
  });

const test = pageSchema.parse(null);
test.slots
// ^?
```

Removing the `.optional()` will actually correctly infer the type (but I want `slots` to be optional).

gislerro:
If you add

```ts
[...]

Issue #4592, opened 2025-06-01
Broken output type for recursive property


RobinTail:
Somewhere between 3.25.42 and 3.25.45 the output type of recursive property was broken.

In the following code sample:

```ts
const feature = z.object({
  title: z.string(),
  get features() {
    return z.array(feature).optional();
  },
});

const output = z.object({
  id: z.int().nonnegative(),
  name: z.string(),
  features: feature.array(), // <—
});

type Output = z.output<typeof output>;
```

The type of the `features` property becomes "never":

![Image](https://github.com/user-attachments/assets/1a1b046d-ea23-446d-865e-8f37e0e691f7)

CC @colinhacks

RobinTail:
Not fixed in 3.25.46

RobinTail:
Interestingly, when I do `features: feature.array().optional()` it fixes the type.
And even `feature.array().optional().nonoptional()` too, though, `.nonoptional()` seems to be doing nothing, because the prop remains optional.
So, I presume it's somehow related to the array schema.

colinhacks:
It seems this never actually worked? In earlier 3.25.x versions this field resolved to `unknown`. Now it resolves to `never`, which I'm still looking into.  Can you confirm that this type was resolving properly before? 

This is hitting against the limits of what the cyclical inference can do (generally multiple function/method calls will break things). It may actually be preferable for this to be `never` as a signal that the schema declaration needs to be reworked. But ideally this would cause a circularity error (never thought I'd say that) instead of failing silently

RobinTail:
@colinhacks ,

1. It seems that it used to be resolved to `unknown`, but there was no issue because it used to be assignable with a sample data.
2. It resolves correctly (in both old and new versions) when I add `.optional()` which is confusing in terms of limitations you mention.

<img width="509" alt="Image" src="https://github.com/user-attachments/assets/cfcafdaa-76e7-44f6-a54a-bb3cc41a32a7" />

If there is such a limitation, why it depends on optionality?
And if it's really beyond the designed capability, how would you suggest to rework the schema?

Because I'd say that it's just a little more advanced than the one in the documentation: https://zod.dev/api?id=recursive-objects

RobinTail:
I found a workaround that somehow works.

```ts
const feature = z.object({
  title: z.string(),
  get features() {
    return z.array(feature).optional();
  },
});

const output = z.object({
  id: z.int().nonnegative(),
  name: z.string(),
[...]

PR #4630, opened 2025-06-05
Fix json

colinhacks:
Fixes https://github.com/colinhacks/zod/issues/4625
Properly fixes https://github.com/colinhacks/zod/issues/4592


colinhacks:
Landed in 3.25.52

PR #4643, opened 2025-06-07
chore: allow readonly arrays in `z.literal`

vladfrangu:
This PR fixes a type-error when using `z.literal` with a `readonly` / `const` array 

See https://github.com/discordjs/discord.js/pull/10922/files#diff-59f89a435004f01781c0f55a059480576a62675bd42e3d68e3bef6106be51fa9R38-R41


colinhacks:
Thanks! Good catch.

PR #4657, opened 2025-06-09
fix: zodMiniPartial can't use record type

### LINKED ISSUE #4614: Zod v4 mini's z.partial does not accept record schemas
I started experimenting locally with writing a TypeScript to Zod v4 mini schema code generator in my spare time and most of it seemed to go very smoothly but I hit a snag in the process when I noticed that `z.partial` doesn't accept `z.ZodMiniRecord` as input even when we aren't trying to use a mask.

There is a partialRecord, but it's quite awkward to work with for these purposes where you don't know whether the schema input is a record or object especially if you want the types to be correct and narrowed. See workaround below.

If you have something like `type Y = Partial<X>`. In typescript's Partial,`X` can be a record or an object, so in my opinion the natural analog would seem to be `const YSchema = z.partial(xSchema)` generally but at least in the case of zod v4 mini, that wouldn't work unless X is an object rather than a record.

Support for masks in partials for objects does complicate this story a bit (since an analog for "infinite" type keys like string for records doesn't make sense given the lack of subtraction type support in typescript), but I don't think we need to extend the support of masks to record schema partials

I'm wondering if we ca expand the logic in z.partial for v4 mini to work with records when you aren't using a masking parameter and tweak associated function declarations.

Current workaround to produce a function partial built around z.partial and z.recordPartial:
```ts
import { z } from "zod/v4-mini";
import { util } from "zod/v4/core";


type PartialSchemaType<
  T extends z.ZodMiniRecord | z.ZodMiniObject,
  M extends T extends z.ZodMiniObject ? util.Mask<keyof T["shape"]> : undefined
> = T extends z.ZodMiniRecord<infer KeyType, infer ValueType>
  ? z.ZodMiniRecord<z.ZodMiniUnion<[KeyType, z.ZodMiniNever]>, ValueType>
  : T extends z.ZodMiniObject
  ? M extends undefined
    ? z.ZodMiniObject<
        {
          [k in keyof T["shape"]]: z.ZodMiniOptional<T["shape"][k]>;
        },
        T["_zod"]["config"]
      >
    : z.ZodMiniObject<
        {
          [k in keyof T["shape"]]: k extends keyof M
            ? z.ZodMiniOptional<T["shape"][k]>
            : T["shape"][k];
        },
        T["_zod"]["config"]
      >
  : never;

const partial = <
  T extends z.ZodMiniRecord | z.ZodMiniObject,
  M extends T extends z.ZodMiniObject ? util.Mask<keyof T["shape"]> : never
>(
  schema: T,
  mask?: M
): PartialSchemaType<T, M> => {
  switch (schema.def.type) {
    case "object": {
      return (
        mask
          ? z.partial(schema as z.ZodMiniObject, mask)
          : z.partial(schema as z.ZodMiniObject)
      ) as PartialSchemaType<T, M>;
    }
    case "record": {
      const recordDef = (schema as z.ZodMiniRecord)._zod.def;
      return z.partialRecord(
        recordDef.keyType,
        recordDef.valueType
      ) as PartialSchemaType<T, M>;
    }
  }
};
```

The above workaround is of course less than ideal to require for code generation purposes since we neither want to inline so much stuff nor require a runtime dependency for the generated code on another library which provides the above if it can be avoided.

As a general user who has gotten quite used to writing `Partial<Record<string, string>>` and the like, I think it's nice in any case to have that same composability and intuitive 1:1 mapping where we could write this as `z.partial(z.record(z.string(), z.string()))` and so on

lucasporrini:
## 🔧 Description
This PR adds support for using `z.partial()` with record schemas in Zod v4 mini, making it consistent with TypeScript's `Partial<Record<...>>` behavior.

## 🐛 Context
The original issue (#4614) reported that `z.partial()` in Zod v4 mini doesn't accept record schemas as input, even when not using a mask. This creates an inconsistency with TypeScript's `Partial` type, which works with both objects and records.

## ✅ What this PR does
- Adds support for `z.partial()` with record schemas
- Implements proper type inference for partial records
- Maintains backward compatibility with existing object partial functionality
- Uses `z.partialRecord` internally for record schemas

## 🧪 Tests
Added comprehensive tests in `packages/zod/src/v4/mini/tests/partial.test.ts` that verify:
- Basic record partial functionality
- Type inference for string-keyed records
- Type inference for enum-keyed records
- Validation of valid and invalid data
- Empty object handling
- Error cases for invalid types

##📎 Related issue
Fixes #4614

### DIFF
--- a/packages/zod/src/v4/mini/schemas.ts
+++ b/packages/zod/src/v4/mini/schemas.ts
@@ -792,14 +792,6 @@ export function omit<T extends ZodMiniObject, const M extends util.Mask<keyof T[
   return util.omit(schema, mask);
 }
 
-export function partial<T extends ZodMiniObject>(
-  schema: T
-): ZodMiniObject<
-  {
-    [k in keyof T["shape"]]: ZodMiniOptional<T["shape"][k]>;
-  },
-  T["_zod"]["config"]
->;
 export function partial<T extends ZodMiniObject, M extends util.Mask<keyof T["shape"]>>(
   schema: T,
   mask: M
@@ -809,8 +801,33 @@ export function partial<T extends ZodMiniObject, M extends util.Mask<keyof T["sh
   },
   T["_zod"]["config"]
 >;
-export function partial(schema: ZodMiniObject, mask?: object) {
-  return util.partial(ZodMiniOptional, schema, mask);
+
+export function partial<T extends ZodMiniObject>(
+  schema: T
+): ZodMiniObject<
+  {
+    [k in keyof T["shape"]]: ZodMiniOptional<T["shape"][k]>;
+  },
+  T["_zod"]["config"]
+>;
+
+export function partial<Key extends core.$ZodRecordKey, Value extends SomeType>(
+  schema: ZodMiniRecord<Key, Value>
+): ZodMiniRecord<ZodMiniUnion<[Key, ZodMiniNever]>, Value>;
+
+export function partial<T extends ZodMiniObject | ZodMiniRecord>(schema: T, mask?: any): any {
+  if (schema._zod.def.type === "record") {
+    const recordDef = schema._zod.def;
+    return record(union([recordDef.keyType, never()]), recordDef.valueType);
+  }
+  const { type, shape, ...config } = schema._zod.def;
+  return new ZodMiniObject({
+    type: "object",
+    shape: mask
+      ? Object.fromEntries(Object.entries(shape).map(([k, v]) => [k, k in mask ? optional(v) : v]))
+      : Object.fromEntries(Object.entries(shape).map(([k, v]) => [k, optional(v)])),
+    ...util.normalizeParams(config),
+  }) as any;
 }
 
 export type RequiredInterfaceShape<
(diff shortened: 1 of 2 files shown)


colinhacks:
It's a nice idea and a good PR, but I don't think this makes sense for Zod Mini. With the new implementation, any call to `z.partial()` will now pull ZodUnion, ZodRecord, and ZodNever into the bundle. These are a couple comparatively hefty classes in terms of bundle size.  I suspect usage of `z.partial()` is more commonplace than either unions or records, so in a lot of cases this change will means these classes will get pulled in despite being unused.

I do think this change would be well-aligned with Zod's aim to mirror TypeScript 1:1, but sometimes the design goals of Zod Mini conflict with that. It's why I still recommend regular Zod in almost all cases.

Not as elegant but this is the recommended approach here:

```ts
const myRecord = z.record(z.string(), z.string());

z.partialRecord(myRecord.def.keyType, myRecord.def.valueType);
```

I'd be open to a PR that adds support for this though:

```ts
const myRecord = z.record(z.string(), z.string());

z.partialRecord(myRecord);
```

lucasporrini:
Sure, I understand that the objective is to keep zod mini as light as possible.
I saw that issue so I was looking to help this project.
Thank you for your feedback and your package, I'll see if I can help on any other issue.

PR #4667, opened 2025-06-10
feat: allow setting transform input and output JSON schema

DASPRiD:
So far there was no way to set the JSON schema for a transform, even though a `transform` could represent valid JSON values in either input or output, in contrast to other types like e.g. `bigint`.

While there is a `_zod.toJSONSchema`, that one has a) no typed return value and b) only serves as a base for an actually generated schema.

This PR adds an optional params object to `transform`, both classic and miniy, which accepts an optional `toJSONSchema(io: "input" | "output")`. It can then decide whether to return `undefined` or an actual JSON schema object. In case `toJSONSchema` is not defined or the function returns `undefined`, the previous default behavior stays in place and an exception is thrown.

I'm open to other ideas to make this happen, but this seemed the cleanest considering this is an edge case.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1586,11 +1586,13 @@ export const ZodTransform: core.$constructor<ZodTransform> = /*@__PURE__*/ core.
 );
 
 export function transform<I = unknown, O = I>(
-  fn: (input: I, ctx: core.ParsePayload) => O
+  fn: (input: I, ctx: core.ParsePayload) => O,
+  params?: core.$ZodTransformParams
 ): ZodTransform<Awaited<O>, I> {
   return new ZodTransform({
     type: "transform",
     transform: fn as any,
+    toJSONSchema: params?.toJSONSchema,
   }) as any;
 }
 
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -256,6 +256,32 @@ describe("toJSONSchema", () => {
     expect(() => z.toJSONSchema(dynamicCatchSchema)).toThrow("Dynamic catch values are not supported in JSON Schema");
   });
 
+  test("override for transform", () => {
+    const noSchemaForIo = z.transform(() => true, {
+      toJSONSchema: () => undefined,
+    });
+    const schemaForInput = z.transform(() => true, {
+      toJSONSchema: (io) => (io === "input" ? { type: "string" } : undefined),
+    });
+    const schemaForOutput = z.transform(() => true, {
+      toJSONSchema: (io) => (io === "output" ? { type: "string" } : undefined),
+    });
+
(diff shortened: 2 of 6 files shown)


colinhacks:
The Zod-native way of handling this is to `.pipe()` into a new Zod schema. This way the transform itself is sandwiched between two introspectable Zod schemas and everything is convertible to JSON Schema. If you have a strong case for this approach over Pipe let me know but I'm dubious.  (Good PR though)

DASPRiD:
Here's my case, which is not working right now :)

https://github.com/DASPRiD/zod-joda/blob/17bb680509d0f2f21f4904589b4f5a18204a5fbb/src/index.ts#L21-L49

I'd like to accept both JsJoda instance as well as strings. Thus I need a `transform` as the base. For now I "fixed" it in the a fix release by disallowing JsJoda instances and only accepting strings, but really this is a regression on my end which I'd like to rectify:

https://github.com/DASPRiD/zod-joda/blob/main/src/index.ts#L20-L38

Apart from that, overriding the `format` property for a JSON schema is a PITA right now, as you can see there, would be nice if there was a cleaner way which doesn't involve overriding _zod internals.

colinhacks:
I see. The return type of your schemas is a JsJoda instance but you want them to be represented as formatted strings in JSON Schema. Makes sense.

I think you're right...if `_zod.toJSONSchema` is set it should totally override the current logic. I'm going to make this change. As far as I know you're the only person using this functionality currently lol. 

When I added `._zod.toJSONSchema` I was hoping it would be used in custom types, just like this. That's why it isn't explicitly marked with `@internal` like some other fields. You should feel comfortable setting it here. There's not really an obvious place to document this pattern because it's quite an advanced pattern, but you're using this as intended. 


---

Another approach to consider re: customizing JSON Schema is just to add all the fields you want as metadata. For better or worse, all metadata fields are blindly written into the JSON Schema currently. 

```ts
const inputSchema = string().meta({ example: example(params), format: schemaFormat });
```

But this only works if `z.toJSONSchema()` is pulling metadata from the global registry. This is the default behavior but it can be overridden: https://zod.dev/json-schema?id=metadata

colinhacks:
> I think you're right...if _zod.toJSONSchema is set it should totally override the current logic. I'm going to make this change. As far as I know you're the only person using this functionality currently lol.

Implemented in ffc41bd9 and landed in 3.25.59

DASPRiD:
@colinhacks That mostly does the trick… but only mostly. If you do this, the `toJSONSchema` gets forgotten, which is really bad:

```ts
test("custom toJSONSchema", () => {
  const schemaA = z.instanceof(Date);
  schemaA._zod.toJSONSchema = () => ({
    type: "string",
    format: "date-time",
  });
  const schema = schemaA.refine(() => true);

  expect(z.toJSONSchema(schema)).toMatchInlineSnapshot(`
    {
      "format": "date-time",
      "type": "string",
    }
  `);
});
```

Also, could we add the `io` parameter to the `_zod.toJSONSchema` function? That would be totally BC, and the function could then decide whether it has a schema for the given `io`.`

PR #4679, opened 2025-06-11
Docs - Use consistent variable names for IP examples

lb7:
Just a small tweak to the variable names to make the example consistent


colinhacks:
Thanks!

PR #4682, opened 2025-06-12
chore: update husky from v7 to v9

noritaka1166:
update husky from v7 to v9
- <https://github.com/typicode/husky/releases>


colinhacks:
Love it

PR #4684, opened 2025-06-12
Fixed allowsEval for RN

V1RE:
In React Native navigator is defined but doesn't contain the userAgent property, thus throwing an error when trying to call String#includes().

By first verifying the type of navigator.userAgent to be a string, it doesnt crash.

image


colinhacks:
Sorry, just seeing this. Already pushed a fix. Sorry for the inconvenience with this one. I'm working on better tests for Hermes. Fixed in 3.25.64.

PR #4689, opened 2025-06-12
docs: added v4 support for `regle` library

victorgarciaesgi:
Hey!

Regle supports Zod v4 as it relies on Standard Schema. I added unit tests specifically for it 👍


colinhacks:
I just removed the `v4` flag & associated column from the table. It was vestigial—anything on the v4 Ecosystem page is expected to have v4 support — the old Ecosystem listing is still available at v3.zod.dev for anyone who needs it.  👍

PR #4701, opened 2025-06-13
fix #4700: Use array as argument of templateLiteral

### LINKED ISSUE #4700: Incorrect use of z.templateLiteral() in docs
Here is link to the part containing the issue in the doc. [https://zod.dev/api#:~:text=To%20define%20a%20template%20literal%20schema%3A](url).

`const schema = z.templateLiteral("hello, ", z.string(), "!");` doesn't validate `hello, ${string}!`, but validates just `hello, `, since argument is expected to be an array.

martin-ragan:
#4700


colinhacks:
Oops, thanks!

PR #4707, opened 2025-06-13
feat(v4): Port v3 `discriminator` type inference

ByteEVM:
The type for `discriminator` i.e., `_zod.def.discriminator` was fixed as **string**.
Now the actual value is inferred.

The `Discriminator` generic was added in 2nd position compared to v3 — making it safe to update if your codebase already defined `Options` type.

Example:
```ts
const TestDiscriminator = z.discriminatedUnion("kind", [
  z.object({
    kind: z.literal("1st").optional(),
    a: z.string(),
  }),
  z.object({
    kind: z.literal("2nd"),
    b: z.string(),
  }),
]);

type ExtractedDiscriminator = typeof TestDiscriminator._zod.def.discriminator;
// returns `kind`
```

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1202,10 +1202,12 @@ export function union<const T extends readonly core.SomeType[]>(
 }
 
 // ZodDiscriminatedUnion
-export interface ZodDiscriminatedUnion<Options extends readonly core.SomeType[] = readonly core.$ZodType[]>
-  extends ZodUnion<Options>,
-    core.$ZodDiscriminatedUnion<Options> {
-  _zod: core.$ZodDiscriminatedUnionInternals<Options>;
+export interface ZodDiscriminatedUnion<
+  Options extends readonly core.SomeType[] = readonly core.$ZodType[],
+  Discriminator extends string = string,
+> extends ZodUnion<Options>,
+    core.$ZodDiscriminatedUnion<Options, Discriminator> {
+  _zod: core.$ZodDiscriminatedUnionInternals<Options, Discriminator>;
 }
 export const ZodDiscriminatedUnion: core.$constructor<ZodDiscriminatedUnion> = /*@__PURE__*/ core.$constructor(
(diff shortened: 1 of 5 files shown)


colinhacks:
Why do you want the discriminator to be statically knowable?

ByteEVM:
> Why do you want the discriminator to be statically knowable?

I'd say it's a _nice-to-have_, as it was already present in v3 and some validations based on this behavior were lost.

Like validating whether an union is correct at compile time, or another example:

> Determining the exact values of all discriminator variants ahead of time

With this PR:

```ts
import { z } from "zod/v4";

const Test = z.discriminatedUnion("kind", [
  z.object({
    kind: z.literal("a"),
  }),
  z.object({
    kind: z.literal("b"),
  }),
]);

type Key = typeof Test._zod.def.discriminator;
//   ^? type Key = "kind"

type Values = z.input<typeof Test>[Key];
//   ^? type Values = "a" | "b"
```

Without this PR (_may break_):

```ts
import { z } from "zod/v4";

const Test = z.discriminatedUnion("kind", [
  z.object({
    kind: z.literal("a"),
  }),
  z.object({
    kind: z.literal("b"),
  }),
]);

// An error-prone utility to extract discriminator
type DiscriminatorKey<T extends z.ZodDiscriminatedUnion> =
  T["options"][number] extends z.ZodObject<infer Shape>
    ? {
        [K in keyof Shape]: Shape[K] extends z.ZodLiteral<string> ? K : never;
      }[keyof Shape]
    : never;

type Key = DiscriminatorKey<typeof Test>
//   ^? type Key = "kind"

type Values = z.input<typeof Test>[Key]
//   ^? type Values = "a" | "b"
```

Another ex;
You may want to _partially validate_ one of its variants (using `.pick`), with a variant being selected by default.
Say a multi-page form, where you validate each page individually — and have different forms for some customers.
TS could help with validation in such case with proper inference of discriminator key and values.

ByteEVM:
Looks like this PR was re-implemented by @colinhacks, at #5044 

good to see inference is back!

ByteEVM:
Resolved—

PR #4708, opened 2025-06-14
chore: remove unused octokit

noritaka1166:
remove unused octokit
- <https://github.com/octokit/octokit.js>


colinhacks:
lol thanks

Issue #4727, opened 2025-06-16
ZodMiniObject.shape is declared in types but undefined at runtime


andersk:
```ts
import { z } from "zod/v4-mini";

const schema = z.object({ a: z.string() });
const shape = schema.shape; // TS infers { a: z.ZodMiniString<string> }
console.log(shape) // undefined
shape.a; // TS accepts this, but it throws TypeError at runtime
```

https://github.com/colinhacks/zod/blob/8c74035ee6bcf27ab4b4398708b379713cc3eeee/packages/zod/src/v4/mini/schemas.ts#L694-L702

This is a landmine when migrating from v4 to v4-mini.

andersk:
Not mentioning the bot to avoid wasting GPU power on a nonsense argument with an AI—but when there’s a mismatch between TypeScript and runtime, that’s a bug, not a known limitation. It’s fine if it’s not going to work at runtime (we can replace `.shape` with `.def.shape`), but then it should be removed from the TypeScript types (so we can find all the places where that replacement is needed).

colinhacks:
Fixed in 3.25.67, thanks!

PR #4733, opened 2025-06-17
Add Zod Mini example for destructuring extension

alloy-d:
Hi there! Thanks for Zod. It's been a joy to use.

No huge change here, I just stumbled upon this example in the docs that was missing a Zod Mini version.


colinhacks:
Good PR, bad timing. As of [yesterday](https://github.com/colinhacks/zod/issues/4727) the original code works for both Zod and Zod Mini, so I'm going to leave it as is. Thanks though!

alloy-d:
Oh, wonderful! Thank you for pointing me to that issue, too—a tiny part of me was wondering why it had type-checked fine without the `.def` in the first place. 😄

PR #4736, opened 2025-06-17
fix: use static components

anubra266:
Updates Accordion and Tabs to use static components, so content can be crawled by LLM

### DIFF
--- a/packages/docs/app/(doc)/[[...slug]]/page.tsx
+++ b/packages/docs/app/(doc)/[[...slug]]/page.tsx
@@ -1,15 +1,10 @@
-import { Heading } from "@/components/heading";
-import { Tabs, Tab } from "@/components/tabs";
 import { Accordion, Accordions } from "@/components/accordion";
+import { Heading } from "@/components/heading";
+import { Tab, Tabs } from "@/components/tabs";
 import { source } from "@/loaders/source";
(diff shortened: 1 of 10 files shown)


colinhacks:
Working in dev but the build is failing for me. But Zod now has [zod.dev/llms.txt](zod.dev/llms.txt) and [zod.dev/llms-full.txt](zod.dev/llms-full.txt) which is presumably better.

colinhacks:
Sorry, re-opening because I found a fix locally. I'm not opposed to this, and I appreciate the PR. But I'm trying to make sure I understand what's happening here. Where is this documented on the Fumadocs docs? Does this make all Accordions and Tabs fully SSRd + hydrated? Weird this isn't the default in Fumadocs. Does this mean all the Callouts are also not scrapable? Perhaps those should get the same treatment? cc @anubra266

anubra266:
Yeah @colinhacks it's not the default behaviour in fumadocs see https://github.com/fuma-nama/fumadocs/issues/1881

colinhacks:
Hm there are issues with this breaking formatting of Accordions (no border) and for some reason they can't be closed after being opened. I don't really have bandwidth to look into this. Let's just say this is problem is solved with llms.txt.

PR #4740, opened 2025-06-18
Allow JWT extension typ values

hitsthings:
We use `JWT+s2s` for our service-to-service JWT tokens and `JWT+at`, `JWT+aut`, etc for other forms of tokens. Since updating to the latest v3, those tokens don't pass validation.

NOTE: I just did this in the GitHub UI. I should write a test for it before merging. Opening a PR now for a sanity check though.

### DIFF
--- a/packages/zod/src/v3/types.ts
+++ b/packages/zod/src/v3/types.ts
@@ -706,7 +706,7 @@ function isValidJWT(jwt: string, alg?: string): boolean {
       .padEnd(header.length + ((4 - (header.length % 4)) % 4), "=");
     const decoded = JSON.parse(atob(base64));
     if (typeof decoded !== "object" || decoded === null) return false;
-    if ("typ" in decoded && decoded?.typ !== "JWT") return false;
+    if ("typ" in decoded && !decoded.typ?.startsWith("JWT")) return false;
     if (!decoded.alg) return false;
     if (alg && decoded.alg !== alg) return false;
     return true;


packages/zod/src/v3/types.ts
```diff
@@ -706,7 +706,7 @@ function isValidJWT(jwt: string, alg?: string): boolean {
       .padEnd(header.length + ((4 - (header.length % 4)) % 4), "=");
     const decoded = JSON.parse(atob(base64));
     if (typeof decoded !== "object" || decoded === null) return false;
-    if ("typ" in decoded && decoded?.typ !== "JWT") return false;
+    if ("typ" in decoded && !decoded.typ?.startsWith("JWT")) return false;
```
hitsthings, line 709: If you prefer I can do
```suggestion
    if ("typ" in decoded && decoded.typ !== "JWT" && !decoded.type?.startsWith("JWT+")) return false;
```
But I'd imagine other people could be using `JWT-at` or `JWT/at` or whatever.

colinhacks:
I think Zod should stay stricter for now, as in a typical use case people just want to validate vanilla JWTs. Perhaps down the line as other variants get more popular, there will be more demand for a more generalized API here. Closing, but thanks for the PR 👍

fZlatar:
> I think Zod should stay stricter for now, as in a typical use case people just want to validate vanilla JWTs. Perhaps down the line as other variants get more popular, there will be more demand for a more generalized API here. Closing, but thanks for the PR 👍

I was also checking on this and in my opinion typ should not be checked at all if there is no typ option in schema to pass. I think that this is a design flaw and should be fixed because right now it is useless for any serious use case (for example access token). The startsWith solution is also completely wrong since the standard for access tokens is "at+jwt". In my opinion if there is no option to pass for typ just remove this check and everything works.

TimoGlastra:
There is quite a lot of official specifications that require the `typ` value to be more specific to prevent ambiguity 

Section 3.11 of RFC 8725 JSON Web Token Best Current Practices (co-written by one of the the original authors of the JWT spec), reccomends the usage of explicit typing: https://www.rfc-editor.org/rfc/rfc8725.html#name-use-explicit-typing

Specifically it ends with:
> Explicit typing is RECOMMENDED for new uses of JWTs.

I think by default the zod library should not limit the `typ` and that options can extend the allowed `typ` values specific to your use case. 

I think "I think Zod should stay stricter for now, as in a typical use case people just want to validate vanilla JWTs" is a bit of a dated answer.

PR #4752, opened 2025-06-21
chore: improve library authors doc

lawvs:
Correct a typo in the documentation and introduce an example function to detect Zod schema versions.


colinhacks:
I think this is kind of confusing actually, and the `schema is z3.ZodTypeAny` type predicate isn't correct. I think people will figure it out.

Issue #4768, opened 2025-06-25
v4(toJSONSchema): .catch() should be marked as optional?


samchungy:
Currently piping toJSONSchema against my suite of existing unit tests and this raised some errors.

Possibly related to: https://github.com/colinhacks/zod/issues/3824

```ts
import z from 'zod/v4';

const catchSchema = z.object({
  a: z.string().catch('a'),
});

console.dir(
  z.toJSONSchema(catchSchema, {
    io: 'input',
  }),
  { depth: null },
);
```

Results in:

```ts
{
  '$schema': 'https://json-schema.org/draft/2020-12/schema',
  type: 'object',
  properties: { a: { default: 'a', type: 'string' } },
  required: [ 'a' ]
}
```

But catchSchema.parse({}) passes.

samchungy:
Another one to think about:

```ts
import z from 'zod/v4';

const catchSchema = z.object({
  a: z.never(),
});

console.dir(
  z.toJSONSchema(catchSchema, {
    io: 'input',
  }),
  { depth: null },
);
```

renders:

```ts
{
  '$schema': 'https://json-schema.org/draft/2020-12/schema',
  type: 'object',
  properties: { a: { not: {} } },
  required: [ 'a' ]
}
```

We can't actually send `a` here

samchungy:
Another one to think about:

```ts
z.union([z.string(), z.number().optional()])
z.union([z.string(), z.number().default(2)])
```

Should this mark the schema as optional?

samchungy:
I'm not sure how we should treat a `z.custom()`. In my lib we would essentially run a .parse() on `undefined` to determine if it was optional or not.

PR #4769, opened 2025-06-25
Fix optionality of schemas

samchungy:
Resolves https://github.com/colinhacks/zod/issues/4768

Changes:
- Marks `.catch()` schemas as input optional.
- Marks `.never()` schemas as input and output optional.
- Marks `.undefined()` schemas as input and output optional.
- Changes `.undefined()` output to align with `.never()`. At the moment it outputs `null` which isn't strictly correct. The Zod Schema would never be able to accept `null`. We essentially want them to send us nothing right? As `undefined` doesn't exist we can treat it the same as `.never()`
- Marks `z.union()` schemas with optional members as `input` and `output` optional.

### DIFF
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -1379,9 +1379,6 @@ export interface $ZodNever extends $ZodType {
 
 export const $ZodNever: core.$constructor<$ZodNever> = /*@__PURE__*/ core.$constructor("$ZodNever", (inst, def) => {
   $ZodType.init(inst, def);
-
-  inst._zod.optin = "optional";
-  inst._zod.optout = "optional";
   inst._zod.parse = (payload, _ctx) => {
     payload.issues.push({
       expected: "never",
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -1241,6 +1241,8 @@ export const $ZodUndefined: core.$constructor<$ZodUndefined> = /*@__PURE__*/ cor
     $ZodType.init(inst, def);
     inst._zod.pattern = regexes.undefined;
     inst._zod.values = new Set([undefined]);
+    inst._zod.optin = "optional";
+    inst._zod.optout = "optional";
 
     inst._zod.parse = (payload, _ctx) => {
       const { value: input } = payload;
@@ -1378,6 +1380,8 @@ export interface $ZodNever extends $ZodType {
 export const $ZodNever: core.$constructor<$ZodNever> = /*@__PURE__*/ core.$constructor("$ZodNever", (inst, def) => {
   $ZodType.init(inst, def);
 
+  inst._zod.optin = "optional";
+  inst._zod.optout = "optional";
   inst._zod.parse = (payload, _ctx) => {
     payload.issues.push({
       expected: "never",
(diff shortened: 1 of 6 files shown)


packages/zod/src/v4/core/schemas.ts
```diff
@@ -1378,6 +1380,8 @@ export interface $ZodNever extends $ZodType {
 export const $ZodNever: core.$constructor<$ZodNever> = /*@__PURE__*/ core.$constructor("$ZodNever", (inst, def) => {
   $ZodType.init(inst, def);
 
+  inst._zod.optin = "optional";
```
colinhacks: My current mental model:

- `optin` should be "optional" if `undefined` will pass validation
- `optout` should be "optional" if the schema might return `undefined` (with exceptions for unknowable stuff like transforms)

So it seems like ZodNever should be false for both. I read [this comment](https://github.com/colinhacks/zod/issues/4768#issuecomment-3003422511) and still don't understand the rationale. What am I missing?

colinhacks:
The key reason why things are the they way the are is that Zod is trying to distinguish between these two types:

```ts
type A = { a?: string }
type B = { a: string | undefined }
```

I was trying to avoid marking a type as "runtime-optional" (`inst._zod.optin = "optional"`) unless it was also marked as "statically optional" (`optin: "optional"` in the internals interface). But I suppose it's not a big deal. 

I rolled back the changes to ZodNever because I'm still dubious, but I wanted to get this merged. We can continue that discussion if you like, but I think I'm unlikely to be convinced.

packages/zod/src/v4/core/schemas.ts
samchungy:
```ts
import Ajv from 'ajv';

const ajv = new Ajv();

const jsonSchema = {
  type: 'object',
  properties: {
    a: {
      not: {},
    },
    b: {
      type: 'string',
    },
  },
  required: ['a', 'b'],
  additionalProperties: false,
};

console.log(
  ajv.validate(jsonSchema, {
    b: 'test',
  }),
);

// false;
```

Just trying to placate validators to allow parsing those JSON Schemas with valid objects. This should pass right?

samchungy: When `a` isn't required:

```ts
console.log(
  ajv.validate(jsonSchema, {
    a: 'bar',
    b: 'test',
  }),
);

// false ✅ 

console.log(
  ajv.validate(jsonSchema, {
    b: 'test',
  }),
);

// true ✅ 

```

colinhacks: Yeah I'm not a fan. This doesn't work in Zod

```ts
z.object({
  a: z.never(),
}).parse({}); // ❌ 
```

And it doesn't work in TypeScript:

Screenshot 2025-07-03 at 17 10 05


The only reason to use `z.never()` is if you really just want validation to always fail, so making this change to get AJV to pass validation isn't compelling imo. This goes against Zod's internal concept of `optin` and `optout`. 

Of course there's no reason why `z.toJSONSchema` needs to use those flags. It can use an orthogonal mechanism to determine "JSON Schema" optionality. It would take some reworking but it wouldn't be super hard to implement a `isOptional` utility; you could use `isTransforming` as a starting point. Probably not worth it but if we keep encountering tension between the "Zod way" and the "JSON Schema way" then that's always an option.

samchungy: Ah I see, cool. I guess the workaround here is you can always go z.never().optional() if you need the behaviour. Sounds good to me.

PR #4391, opened 2025-05-17
feat: add z.looseEnum for flexible enum validation

Lavdewangan1:
# Add `z.looseEnum()` for permissive enum validation

## Fixes #4362  - Request for permissive enum validation

Introduces a new `z.looseEnum()` validator that provides more flexible input validation while maintaining type safety.

### Key Features
- ✅ Preserves TypeScript type hints for preferred values
- ✅ Accepts any string input (unlike strict `z.enum()`)
- 🔗 Seamless integration with other Zod validators
- 🏗 Designed for permissive request validation

## Usage

```typescript
// Basic usage
const StatusSchema = z.looseEnum(['pending', 'completed']);
StatusSchema.parse('pending');  // valid
StatusSchema.parse('other');    // also valid
StatusSchema.parse(123);        // throws ZodError

// With object schemas
const RequestSchema = z.object({
  status: z.looseEnum(['on', 'off']),
  id: z.string()
});

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

- **New Features**
  - Introduced a new schema type, `looseEnum`, which validates any string input, allowing both predefined enum values and arbitrary strings.
- **Bug Fixes**
  - None.
- **Documentation**
  - Updated sponsor link in the README file.
- **Tests**
  - Added comprehensive tests for the new `looseEnum` schema, including integration and edge cases.
  - Suppressed TypeScript warnings for unused imports in test files.
- **Chores**
  - Upgraded development dependencies and added new ones in the project configuration.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

PR #4785, opened 2025-06-30
z.looseEnum() for more permissive enum validation

marco-carvalho:


samchungy:
Just putting this comment from the last PR here for visibility: https://github.com/colinhacks/zod/pull/4391#issuecomment-2921619876

marco-carvalho:
hi @colinhacks WDYT about this PR?

colinhacks:
Hesitant to add a new top-level type for a quite niche use case. I evaluated ways of folding this into $ZodEnum but it was just hairy enough to dissuade me. 

Zod 4 already lets you further refine the inferred type of `z.string()` with a hand-written generic. Given the relative obscurity of this use case, I think this is more than adequate:

Image

PR #4811, opened 2025-07-03
Add JSON Schema draft-04 output

colinhacks:
## Summary
- support `draft-4` in `z.toJSONSchema`
- generate boolean exclusives and enum literals for draft-04
- omit `propertyNames` for records when targeting draft-04
- document new option
- test draft-04 behaviour
- include draft-4 schema in type definitions

## Testing
- `pnpm test`
- `pnpm build`

------
https://chatgpt.com/codex/tasks/task_e_686628b0349c832f9665ce89343e959b

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -540,9 +540,7 @@ describe("toJSONSchema", () => {
   });
 
   test("number constraints draft-4", () => {
-    expect(
-      z.toJSONSchema(z.number().gt(5).lt(10), { target: "draft-4" })
-    ).toMatchInlineSnapshot(`
+    expect(z.toJSONSchema(z.number().gt(5).lt(10), { target: "draft-4" })).toMatchInlineSnapshot(`
       {
         "$schema": "http://json-schema.org/draft-04/schema#",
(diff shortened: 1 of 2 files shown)


kibertoad:
@colinhacks This is awesome, thank you!

Can this also handle the case of `age: z.optional(z.nullable(z.coerce.number()))`?

In OpenAPI 3.1 this is represented as a union of null and number, which is what zod does now.
However, in OpenAPI 3.0 type "null" is not a thing, property "nullable = true" should be used instead.

colinhacks:
> property "nullable = true" should be used instead 

If this was going to happen it would have to be another `"target"`, probably called `"openapi-3.0"`. Hopefully that'll be the last one... 🫠

kibertoad:
@colinhacks gotcha! bah, yeah, wish OA was more aligned with JSON Schema 😅

Doradas0:
Hey @colinhacks, do you have any estimate when this might get merged. I'm really looking forward to using v4 but this is currently a key blocker for using our schemas in our AWS API Gateway data models.
Has anyone else figured out a way to use the `override` method to make the schema compatible with `draft-4`

colinhacks:
Landed in [4.0.14](https://www.npmjs.com/package/zod/v/4.0.14)

PR #4816, opened 2025-07-03
feat: redo destructuring

### LINKED ISSUE #4773: v4 error invalid redefinition of lexical identifier when using in QuickJS / embedded JS runtimes
Hi,

I'm using Zod v4 in a JavaScript environment based on [rquickjs](https://github.com/DelSkayn/rquickjs).

When I destructure parameters like `key` or `collection` in my evaluated function:

```
const { key, collection } = input;
const parsed = MyZodSchema.parse(input);
```

I receive the error:

```
  → Error: invalid redefinition of lexical identifier
    at <input>:11:52
    at Function (native)
    at compile (@junobuild/sputnik/dev:26:1)
```

This doesn't happen with Zod v3, but does with Zod v4, maybe because v4 uses internal `const` bindings during validation, which collide with my own in strict runtimes like QuickJS?

Is it possible to isolate these bindings better?

peterpeterparker:
# Motivation

The commit dfda9931e00faaba1a7fd453330902ce16af45b9 aimed to minimize the use of destructuring patterns in order to resolve #4773. After debugging (way) much further, I finally realized that the issue wasn't related to Zod, but was actually a side effect of a bug in my project, triggered by Zod v4 starting to generate random constants at runtime.

As a result, I decided to submit this PR to reintroduce the destructuring pattern, which was the preferred style in the first place.

# Notes

The commit dfda9931e00faaba1a7fd453330902ce16af45b9 appeared to contain various changes that were not all related to this concern. Therefore, I manually cherry-picked what seemed relevant to the destructuring topic.

### DIFF
--- a/packages/zod/src/v4/core/api.ts
+++ b/packages/zod/src/v4/core/api.ts
@@ -1508,11 +1508,11 @@ export function _stringbool(
   schemas.$ZodPipe<schemas.$ZodString, schemas.$ZodTransform<boolean, string>>,
   schemas.$ZodBoolean<boolean>
 > {
-  const params = util.normalizeParams(_params);
+  const { case: _case, error, truthy, falsy } = util.normalizeParams(_params);
 
-  let truthyArray = params.truthy ?? ["true", "1", "yes", "on", "y", "enabled"];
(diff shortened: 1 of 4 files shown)


colinhacks:
Very kind of you, but I'm not opinionated on this. Happy to keep it as is!

PR #4837, opened 2025-07-07
fix: cleanup _idmap of $ZodRegistry

safareli:


safareli:
shouldn't `  _idmap: Map<string, Schema> = new Map();` be WeakMap too? otherwise presense of id in meta results in strong pointer of Schema to be created which means it will not be removed from the `_map` ever even if there are no strong pointers to Schema from outside of this class thus creating mem leak potential.

colinhacks:
You're correct. Unfortunately registries need to be iterable for `z.toJSONSchema()` as implemented (this is why CI is failing), and the fact that it's using `WeakRef` for the schema map is a bit of a vestige of an older implementation. I think I'll be switching to a plain `Map` and adding a `.clear()` method. It's not ideal, but honestly I think a true WeakMap implementation would also cause a lot of mysterious issues for people who don't understand how they work. Regardless, merging this fix 👍 Thank you.

PR #4876, opened 2025-07-12
Include export to `zod/core` + update documentation and tests

alexmarqs:
With the release of Zod ^4.0.0, we now have clean exports for modules like `zod/mini`, which is great. I think it makes sense to also add a similar export for the core package (`zod/core`).  WDYT?

I've opened a quick PR to introduce this—curious to hear your thoughts! If you notice any missing doc updates or anything else that needs tweaking, feel free to point it out and I’ll fix it. 

Thanks!

CC: @colinhacks

### DIFF
--- a/packages/docs/content/error-customization.mdx
+++ b/packages/docs/content/error-customization.mdx
@@ -220,7 +220,7 @@ result.error.issues;
 
 The `iss` object is a [discriminated union](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions) of all possible issue types. Use the `code` property to discriminate between them.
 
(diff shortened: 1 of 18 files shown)


colinhacks:
I thought about this, but intentionally left it out.

For regular users, it adds clutter to the autocomplete:
Screenshot 2025-07-23 at 15 20 37

And tooling authors should be using the "permalink" at `"zod/v4/core"`, per https://zod.dev/library-authors. This more future-proof in case I ever do a Zod 5.

Good PR though!

alexmarqs:
got it! thanks for the explanation @colinhacks 🤝

PR #4911, opened 2025-07-17
Add missing field to $ZodIssueInvalidType

mjomble:
The `received` field is present on some invalid_type issues, but gave a TypeScript error when I tried to use it.

PS. It seems to me like the `input` field should be marked as optional, since it's only present if the `reportInput` flag is set. Based on the TS typings, I assumed it's always present and tried to use it, only to find out it can be undefined.
But that's a slightly bigger change and should get its own PR.


colinhacks:
I added that as a bit of additional metadata that may be helpful when debugging a validation error. But I don't consider it part of the "public API" and I don't want people to start depending on it. I also add a `note` field to a number of errors—same thing. Just for aesthetics/debugging, not something you should rely on.

You're right about `input`, I'm looking into it.

PR #4919, opened 2025-07-18
Doc tweak - spread vs destructuring

mjomble:
The documentation is referring to the spread syntax as "destructuring".

Although the features are closely related and look very similar, there is a difference:

[Destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring) involves extracting a "smaller" value from a "larger" value (e.g. `const { id, ...fields } = data`)
[Spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) involves building a "larger" value from "smaller" values (e.g. `const data = { id, ...fields }`)


colinhacks:
Oops good call!

PR #4920, opened 2025-07-18
feat: Icelandic translation

ekh-anadeainc:
Hi, my colleagues from Iceland have provided a translation of the errors into Icelandic.


colinhacks:
Awesome, thanks!

PR #4941, opened 2025-07-22
patch

### LINKED ISSUE #4880: `File` cannot be named
```
Exported variable 'orpc' has or is using name 'File' from external module ".../node_modules/.pnpm/zod@4.0.5/node_modules/zod/v4/core/schemas" but cannot be named.ts(4023)
```

I guess something has not exported in latest version "v4.0.5"

### LINKED ISSUE #4882: [v4] `keyType` of `partialRecord` is not typed correctly
You can access the key schema of a Zod `record` or `partialRecord` with `.keyType` in v4. If the key schema is a Zod enum, you can access `enum` to extract an enum-like object and `options` to extract the string values like so:

```javascript
const myRecord = z.record(z.enum(["a", "b", "c"]), z.string());

console.log(myRecord.keyType.enum);
// { a: "a", b: "b", c: "c" }
console.log(myRecord.keyType.options);
// [ "a", "b", "c" ]
```

Zod's exported types seem to indicate that the same values are in place for an equivalent `partialRecord`. However, unexpected values are produced instead.

```javascript
const myPartialRecord = z.partialRecord(z.enum(["a", "b", "c"]), z.string());

console.log(myPartialRecord.keyType.enum);
// undefined
console.log(myPartialRecord.keyType.options);
// [ ZodEnum, ZodNever ]
```

The TypeScript definition for `keyType` of a `partialRecord` should be updated if this is functioning as intended.

### LINKED ISSUE #4883: [v4] Breaking change in behavior of optional object properties from v3
In Zod v4, when object properties with `.catch()` or `.default()` are made optional (either with `.optional()` on the property or `.partial()` on the entire object), the caught values are always returned, even when they're not present in the input. This diverges from behavior in v3 where these properties were wholly ignored.

```javascript
const mySchema = z.object({
  first: z.string().catch("placeholder").optional(),
  second: z.string().default("placeholder").optional(),
});

console.log(mySchema.parse({}));
// v4: { first: "placeholder", second: "placeholder" }
// v3: {}
```

I don't believe this was mentioned in the migration guide, causing an unexpected breaking change in my codebase.

### LINKED ISSUE #4929: Zod v4: Metadata type signigure to allow all valid object keys
The current metadata type in zod v4 is:

```ts
type MetadataType = Record<string, unknown> | undefined;
```

This is not assignable to types like this:

```ts
interface MyMetadata {
  a: number
}
```

`MyMetadata` could have keys of any type, not just strings. So, TypeScript will throw an error. To fix this, metadata types have to be passed like this:

```ts
interface MyMetadata {
  a: number;
  [key: string]: unkown;
}
```

This is cumbersome and makes it hard to use utility types like `Omit`.

I'm proposing to widen the metadata type to:

```ts
type MetadataType = object | undefined;
```

### LINKED ISSUE #4933: Node types again
```
E:/Projects/inNENU/node_modules/.pnpm/zod@4.0.5/node_modules/zod/v4/core/checks.d.cts(257,55): error TS2304: Cannot find name 'File'.
```

Image

For detailed reasons, see: https://github.com/colinhacks/zod/issues/4786

colinhacks:
- **Fix File reference issues. Fixes #4880. Fixes #4933.**
- **Relax metadata constraint to object. Fixes #4929**
- **Clean up partialRecord implementation. Closes #4882.**
- **Fixes catch handling in object schemas. Fixes #4883.**
- **Add treeshake case**
- **Document optionality changes**


ahrjarrett:
Hey @colinhacks, I have a PR up that fixes this issue: #4895

I had another PR regarding prototype pollution in the fast-pass schemas that got merged into the v4 branch*.

Would you be open to merging my change into main proper so my contribution will count this time?

I'm happy to add a changeset ~or transfer the tests you've added here to my branch~ (done), if that would make things easier.

PR #4955, opened 2025-07-23
docs: fix refine() when() example copy-pasted comment

FilipLeonard:
Fix refine() when() example copy-pasted comment.


colinhacks:
Oops. Thanks!

Issue #4956, opened 2025-07-23
Feature Request: add regexes for md5, sha1, sha256, etc.


codyfrisch:
I realized they are simple regexes but having them built in would be rather convenient, md5, sha1, sha256, sha512, in hex and base64 would useful. Just to save time and also make the schema more obvious what it is.

zaewc:
Totally agree! I’ve often thought it would be great to have this built into Zod as well. Would it be alright if I took a shot at adding this?

codyfrisch:
Closed by #5049

PR #4963, opened 2025-07-24
feat(v4): add hash format validation (MD5, SHA-1, SHA-256, SHA-384, SHA-512)

976520:
- Hash format validation, Base64 variants
- Core API functions: `_md5()`, `_sha1()`, `_sha256()`, `_sha384()`, `_sha512()`
- Mini API support
- Eng locale messages

## Usage
```typescript
import * as z from "zod/v4";

const md5Schema = z.md5();
md5Schema.parse("5d41402abc4b2a76b9719d911017c592");

const sha256Base64Schema = z.sha256Base64();
sha256Base64Schema.parse("XohImNooBHFR0OVvjcYpJ3NgPQ1qq73WKhHvch0VQtg="); 
```

## Issue
https://github.com/colinhacks/zod/issues/4956

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -309,6 +309,26 @@ export interface ZodString extends _ZodString<core.$ZodStringInternals<string>> 
   cidrv6(params?: string | core.$ZodCheckCIDRv6Params): this;
   /** @deprecated Use `z.e164()` instead. */
   e164(params?: string | core.$ZodCheckE164Params): this;
+  /** @deprecated Use `z.md5()` instead. */
+  md5(params?: string | core.$ZodCheckMD5Params): this;
+  /** @deprecated Use `z.md5Base64()` instead. */
+  md5Base64(params?: string | core.$ZodCheckMD5Base64Params): this;
+  /** @deprecated Use `z.sha1()` instead. */
+  sha1(params?: string | core.$ZodCheckSHA1Params): this;
+  /** @deprecated Use `z.sha1Base64()` instead. */
+  sha1Base64(params?: string | core.$ZodCheckSHA1Base64Params): this;
+  /** @deprecated Use `z.sha256()` instead. */
+  sha256(params?: string | core.$ZodCheckSHA256Params): this;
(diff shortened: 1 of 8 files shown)


packages/zod/src/v4/classic/schemas.ts
```diff
@@ -309,6 +309,26 @@ export interface ZodString extends _ZodString<core.$ZodStringInternals<string>>
...
+  /** @deprecated Use `z.sha256Base64()` instead. */
+  sha256Base64(params?: string | core.$ZodCheckSHA256Base64Params): this;
+  /** @deprecated Use `z.sha384()` instead. */
+  sha384(params?: string | core.$ZodCheckSHA384Params): this;
+  /** @deprecated Use `z.sha384Base64()` instead. */
+  sha384Base64(params?: string | core.$ZodCheckSHA384Base64Params): this;
+  /** @deprecated Use `z.sha512()` instead. */
+  sha512(params?: string | core.$ZodCheckSHA512Params): this;
+  /** @deprecated Use `z.sha512Base64()` instead. */
+  sha512Base64(params?: string | core.$ZodCheckSHA512Base64Params): this;
```
lyricnz: IMHO no need to add functions which are deprecated from day 1

zaewc: At first, I was a bit concerned that removing the deprecated chaining methods like `z.string().md5()` might break the hash validation features. After taking a closer look, I realized those chaining methods aren’t necessary at all direct API calls like `z.md5()` work perfectly fine on their own

[42a79b7](https://github.com/colinhacks/zod/pull/4963/commits/42a79b7b60fc2c0a892adde91394030fff94e91b)

So I followed your advice, removed the deprecated methods, and updated the test code to use the direct API approach instead
Thanks so much for reviewing this 😘 this is a feature I really need, so I’d be very grateful if it could be merged...

lyricnz: I'm just a user like you, but I'll review the PR nevertheless.

colinhacks:
Thanks for the PR! This was independently implemented and available as z.hash()

https://zod.dev/api?id=hashes

PR #4967, opened 2025-07-24
Update docs z.enum with object literal example

perspectivezoom:
From the tests, it looks like `z.enum` supports object literal comprehension out-of-the-box:

https://github.com/colinhacks/zod/blob/3048d14bc7b803507302b8ee3cd1eeeffbd27396/packages/zod/src/v4/classic/tests/enum.test.ts#L12-L24

But it doesn't look like the rest of the world realizes it. Googling for "zod z.enum object literal" leads me to stackoverflow discussions ([1](https://stackoverflow.com/questions/73825273/creating-a-zod-enum-from-an-object), [2](https://stackoverflow.com/questions/75217817/zod-enum-and-object-as-const)) that keep trying to use `Object.values` or other stuff to get this working. I don't really understand. Cursor is yelling at me for no good reason:

Screenshot 2025-07-24 at 10 17 05 AM

Updates the documentation with an object literal example, in the hopes that this knowledge propagates and people stop trying to hack around something that works out-of-the-box natively.

Note that [valibot's enum docs](https://valibot.dev/guides/enums/) call this out explicitly and provide an example for each:

> Since TypeScript enums are transpiled to JavaScript objects by the TypeScript compiler, you can use the `enum` schema function for both. Just pass your enumerated data type as the first argument to the schema function.

Screenshot 2025-07-24 at 9 30 25 AM

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -695,7 +695,7 @@ FishEnum.parse("Swordfish"); // => ❌
   ```
 </Callout>
 
-Object literals are supported.
+Enum-like object literals (`{ [key: string]: string | number }`) are supported.
 
 ```ts
 const Fish = {
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -700,7 +700,7 @@ Object literals are supported.
 ```ts
 const Fish = {
   Salmon: "Salmon",
-  Tuna: "Tuna,
+  Tuna: "Tuna
 } as const
 
 const FishEnum = z.enum(Fish)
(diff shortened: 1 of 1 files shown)


anton89:
it should be nativeEnum and not enum that works, just pointing out the mistake

perspectivezoom:
> it should be nativeEnum and not enum that works, just pointing out the mistake

@anton89 `nativeEnum` was deprecated and rolled into `enum` as part of v4. It will work with both, but `enum` is preferred.
https://zod.dev/v4/changelog?id=znativeenum-deprecated
Screenshot 2025-08-13 at 10 26 12 AM

PR #4998, opened 2025-07-28
Include the received value in discriminated union errors

xdumaine:
Discriminated union errors should be readable similar to enum errors.


tomas-phillips-95 (approved):

colinhacks:
Thanks for the PR but Zod 3 is now in maintenance mode. Only critical bug fixes will be merged. Moreover, changes to error messages are also considered breaking changes so this wouldn't be mergable in a `v3.x` version anyway. Apologies for the inconvenience here, it's long past time I update the CONTRIBUTING.md guide to reflect the current state of affairs.

PR #5006, opened 2025-07-29
add support for `hhmm` offset to iso.datetime

vpontis:
- ISO 8601:2019 allows both extended (+hh:mm) and basic (+hhmm) timezone offset formats
- Many systems (Java SimpleDateFormat, databases) generate the format without a colon
- Zod currently only supports the extended format with colons
- Added test case documenting this limitation with links to ISO standards

All existing tests pass. The new test case demonstrates that both formats are now accepted:

```
  // Basic format (without colon)
  datetime.parse("2022-10-04T05:15:00.000+0000")  // ✅ now works
  datetime.parse("2022-10-04T05:15:00.000-0500")  // ✅ now works

  // Extended format (with colon)
  datetime.parse("2022-10-04T05:15:00.000+00:00") // ✅ still works
  datetime.parse("2022-10-04T05:15:00.000-05:00") // ✅ still works
```


colinhacks:
My issue with this is that it's a slippery slope towards supporting the full "basic" format in which all the punctuation is optional. Supporting some bits and not others feels icky to me, this should really be all or nothing. There are also a lot of cases (probably a majority) where you want to be strict about extended-mode format because your db/tooling expects it: https://chatgpt.com/share/688925db-2280-800f-9fc9-54a258991f50 (including OffsetDateTime.parse(...) in Java, unless that's a hallucination) 

I'd be open to supporting a "full basic" variant behind a `basic: true` flag (analogous to the existing local/offset flags).

vpontis:
Hmm, yea it's a bit confusing. This is part of the ISO spec but it's not obvious to me what's basic vs extended in the ISO spec. For example [this doc from NASA](https://fits.gsfc.nasa.gov/iso-time.html) mentions `hhmm`. 

When we upgraded Zod that has [this PR](https://github.com/colinhacks/zod/pull/4680) we had a couple people complain who are sending `hhmm` for timezone. So it seems relatively common. 

But if you still don't want to support it, that's ok. We'll just figure out a workaround for our usage.

colinhacks:
RFC 3339 represents the stricter version of ISO datetimes intended as a true standard for internet protocols, that' what Zod sticks to. @dosu learn this

vpontis:
@colinhacks thank you for clarifying!

PR #5007, opened 2025-07-29
feat(currency): add currency validation support

ykhedher:
- Introduced `z.currency()` method for validating ISO 4217 currency codes.
- Added documentation for currency validation, including examples and case sensitivity.
- Implemented currency validation logic in core and mini schemas.
- Added a comprehensive set of ISO 4217 currency codes for validation.

### Related to #3881


ykhedher:
cc @colinhacks

colinhacks:
I appreciate the concept, but I don't want Zod to get into the business of maintaining big lists of strings. In general, they should be sourced from somewhere else and passed into `z.enum()`.

PR #5013, opened 2025-07-30
feat: Filter out unrepresentable schemas from unions

Bram-dc:
I needed this to properly create openapi types for example my date schema:
```
export const zodISODateTime = () => {
	return z
		.union([z.date(), z.iso.datetime()])
		.overwrite((v) => {
			if (typeof v === 'string') {
				return new Date(v)
			}
			return v
		})
		.pipe(z.date())
}
```

`z.date()` is not representable, but I still want it to work internally. The resulting json schema should skip it.

I was thinking we should not create an extra unrepresentable option "filter" but instead include it in the "any" option, because this is also done for `z.literal()`

### DIFF
--- a/packages/docs/content/json-schema.mdx
+++ b/packages/docs/content/json-schema.mdx
@@ -224,7 +224,8 @@ interface ToJSONSchemaParams {
 
   /** How to handle unrepresentable types.
    * - `"throw"` — Default. Unrepresentable types throw an error
-   * - `"any"` — Unrepresentable types become `{}` */
+   * - `"any"` — Unrepresentable types become `{}`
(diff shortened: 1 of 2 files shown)


Bram-dc:
> I was thinking we should not create an extra unrepresentable option "filter" but instead include it in the "any" option, because this is also done for `z.literal()`

In this PR I **did** create an extra option. @colinhacks what are your thoughts?

colinhacks:
Thanks for the contribution, but this is very a specific behavior better implemented with `override`. Stick with `unrepresentable: "any"` and apply custom filtering logic inside `override` to filter out empty options.

PR #5018, opened 2025-07-31
Fix recursive object initialization errors with check() and other methods

yz89122:
## Summary

This PR fixes `ReferenceError: Cannot access before initialization` errors that occur when using methods like `.check()`, `.catchall()`, `.passthrough()`, etc. on recursive object schemas.

## Problem

When defining recursive objects and using certain methods, the object spread syntax (`{...shape}`) causes initialization errors:

```typescript
const Category = z.object({
  id: z.string(),
  get subcategories() {
    return z.array(Category).optional();
  },
}).check((ctx) => {
  // Custom validation logic
});
// ReferenceError: Cannot access 'Category' before initialization
```

The issue occurs because object spread syntax immediately evaluates all properties, including getters. When the getter references a recursive type that's still being initialized, it triggers a ReferenceError.

## Solution

Replace object spread with a new `util.objectClone()` function that uses `Object.create()` with `Object.getOwnPropertyDescriptors()` to properly clone objects while preserving property descriptors (including getters) without evaluating them immediately.

This approach:
- Preserves all property descriptors exactly as they are
- Defers getter evaluation until after initialization is complete
- Maintains the same behavior for non-recursive objects
- Works with the existing shape getter pattern

## Changes

- Added `util.objectClone()` function that properly clones objects with descriptors
- Updated shape getter implementations in `object()`, `strictObject()`, and `looseObject()` to use `objectClone()` instead of object spread
- Added test case demonstrating recursive objects with `.check()` method

## Testing

All existing tests pass, plus new test case for recursive objects with `.check()` method.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1149,7 +1149,7 @@ export function object<T extends core.$ZodLooseShape = Partial<Record<never, cor
   const def: core.$ZodObjectDef = {
     type: "object",
     get shape() {
-      util.assignProp(this, "shape", { ...shape });
+      util.assignProp(this, "shape", shape ? util.objectClone(shape) : {});
       return this.shape;
     },
     ...util.normalizeParams(params),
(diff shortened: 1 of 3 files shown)


yz89122:
I ran the tests with the latest `typescript` (`5.9.2`) on the `main` branch, and they failed in the same way. As you can see the GitHub Actions of the last merged commit https://github.com/colinhacks/zod/actions/runs/16682585688/job/47224521400.

cc @colinhacks 

An issue https://github.com/colinhacks/zod/issues/5035 and a PR https://github.com/colinhacks/zod/pull/5033 created.

PR #5028, opened 2025-08-02
chore: upgrade vitest to v3

abnud11:
This PR upgrades vitest to v3

In v3, the workspace file is deprecated and `test.projects` is recommended instead, so this PR removes the workspace file and uses `test.projects` file instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "tslib": "^2.8.1",
     "tsx": "^4.19.3",
     "typescript": "^5.5.4",
-    "vitest": "^2.1.9",
+    "vitest": "^3.2.4",
     "zod": "workspace:*",
     "zod3": "npm:zod@~3.24.0",
     "zshy": "^0.3.3"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -93,8 +93,8 @@ importers:
         specifier: ^5.5.4
         version: 5.5.4
       vitest:
-        specifier: ^2.1.9
-        version: 2.1.9(@types/node@22.13.13)(lightningcss@1.29.3)(terser@5.39.0)
+        specifier: ^3.2.4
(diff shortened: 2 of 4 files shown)


yz89122:
~~It seems like vitest v3 will not run typecheck without an explicit `--typecheck`~~

vitest v2 with `typescript@latest` (`5.9.2`) will report some type errors (`TS2615`) as you can see in recently merged commit https://github.com/colinhacks/zod/commit/f2949a81a06fe197c53e47c1fab024cebbd7f1f1.

abnud11:
@yz89122 Added, thanks for the note.

yz89122:
> @yz89122 Added, thanks for the note.

Sorry, I was incorrect about the `--typecheck` options. It seems like that we should [define `projects` in `/vitest.config.ts`](https://vitest.dev/guide/projects.html#defining-projects) instead of `/vitest.root.mts`.

```ts
// /vitest.config.ts
import { defineConfig } from "vitest/config";

export default defineConfig({
  test: {
    projects: ["packages/*"],
  },
});
```

abnud11:
@yz89122 Done

ahmad-habannakeh-midani-silver:
@yz89122 I merged from upstream main, now tests fail on my machine. Even the main branch without my changes fails the tests on my machine.

But I see them succeeding in the CI environment. How is that possible? What happened? Is there something I should do?

abnud11:
@yz89122 I merged from upstream main, now tests fail on my machine. Even the main branch without my changes fails the tests on my machine.

But I see them succeeding in the CI environment. How is that possible? What happened? Is there something I should do?

abnud11:
@yz89122 I fixed merge conflicts

abnud11:
@yz89122 I fixed merge conflicts again. Please have a look at the PR.

PR #5090, opened 2025-08-12
#5088: improved validation messages for "invalid type" (de)

### LINKED ISSUE #5088: validation messages for "invalid type" (de)
```
z.config(de());
z.string().safeParse(null);
```
produces "Ungültige Eingabe: erwartet string, erhalten null" which is confusing for end-users. It would be better to say "Wert erforderlich" (i.e. "value required").

sbublava459:
Provides improved German translations for null / undefined. Fixes #5088.


colinhacks:
It's a reasonable idea, but this is better implemented with a global error map as needed for your application. Zod provides a semi-standardized `invalid_type` error message. Also, changing default error messages is a breaking change so Zod's error messages are functionally locked at this point.

PR #5093, opened 2025-08-12
fix(v4): make size-based Set/File min/max inclusive; update snapshots

### LINKED ISSUE #5091: min() / max() for sets are off by one
```
console.log(z.set(z.string()).min(1).safeParse(new Set()).error);
```
prints:

```
ZodError: [
  {
    "origin": "set",
    "code": "too_small",
    "minimum": 1,
    "path": [],
    "message": "Too small: expected set to have >1 items"
  }
]
```
This is because "inclusive" is set to true, which seems wrong (and is not the case for arrays).

eslamgmalelsayed:
Set and File min()/max() now emit inclusive: true
Aligns messages with Arrays (>= for min, <= for max)
Updates tests (inline snapshots)
Closes #5091


colinhacks:
Ah, very nice. Good catch.

PR #5098, opened 2025-08-14
fix(record): handle non-function constructor field in isPlainObject

### LINKED ISSUE #5066: "Invalid input: expected record, received object" for plain object with non-function `constructor` field
With the upgrade to Zod4 I started getting errors thrown on an existing config file, within a part matching `z.record(z.string(), z.string())`, with the message:

`"Invalid input: expected record, received object"`

Digging through existing Issues, it sounded like Zod4 was doing a POJO check that resulted in the parsed object being identified as a non-POJO, making it an invalid Zod Record. It took me a while to figure out *why*, though, until I noticed that one of the arbitrary keys in the end-user-created-config was `"constructor"`. Sure enough, removing that key resolved the issue.

On the one hand I understand that having a field called `constructor` is non-ideal, however it's totally valid JavaScript, having `{"constructor": "some string"}` matches the schema, and I can't control what field names users will provide.

My workaround is to replace `z.record(z.string(),z.string())` with:

```ts
const permissiveStringRecord = z.object({}).catchall(z.string());
type PermissiveStringRecord = z.infer<typeof permissiveStringRecord>; // {[x: string]: string;}
```

jiji-hoon96:
## Problem
z.record() schema fails to parse plain objects with non-function constructor field, throwing "Invalid input: expected record, received object" error.

## Solution
Add type check in isPlainObject function to treat objects with non-function constructor as plain objects.

## Tests
- Added comprehensive test cases for constructor field scenarios
- All existing 2827 tests pass
- No regressions

Closes #5066

### DIFF
--- a/packages/zod/src/v4/classic/tests/index.test.ts
+++ b/packages/zod/src/v4/classic/tests/index.test.ts
@@ -786,6 +786,37 @@ test("isPlainObject", () => {
   expect(z.core.util.isPlainObject("string")).toEqual(false);
   expect(z.core.util.isPlainObject(123)).toEqual(false);
   expect(z.core.util.isPlainObject(Symbol())).toEqual(false);
+  expect(z.core.util.isPlainObject({ constructor: "string" })).toEqual(true);
+  expect(z.core.util.isPlainObject({ constructor: 123 })).toEqual(true);
+  expect(z.core.util.isPlainObject({ constructor: null })).toEqual(true);
+  expect(z.core.util.isPlainObject({ constructor: undefined })).toEqual(true);
+  expect(z.core.util.isPlainObject({ constructor: true })).toEqual(true);
+  expect(z.core.util.isPlainObject({ constructor: {} })).toEqual(true);
+  expect(z.core.util.isPlainObject({ constructor: [] })).toEqual(true);
+});
+
+test("shallowClone with constructor field", () => {
+  const objWithConstructor = { constructor: "string", key: "value" };
+  const cloned = z.core.util.shallowClone(objWithConstructor);
+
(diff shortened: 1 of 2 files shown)


packages/zod/src/v4/core/util.ts
```diff
@@ -377,6 +377,8 @@ export function isPlainObject(o: any): o is Record<PropertyKey, unknown> {
   const ctor = o.constructor;
   if (ctor === undefined) return true;
 
+  if (typeof ctor !== "function") return true;
```
adam-coster, line 380: Line 380 could replace line 378, instead of having both, since this check includes the case where `ctor === undefined` (since in that case `typeof ctor` is `"undefined"`, which does not equal `"function"`). But the extra clarity doesn't hurt anything.

jiji-hoon96: You're suggestion is technically correct - line 380 (typeof ctor !== "function") would handle both cases since typeof undefined returns "undefined", not "function".
However, I'd keep both lines for clarity. The explicit undefined check makes the intent more obvious and doesn't hurt performance.
It's a style preference, and the current approach is more readable.

jiji-hoon96:
@colinhacks I believe this issue needs to be addressed. I would appreciate it if you could review the PR. 🙏

flbn:
any updates on this?

jiji-hoon96:
There are no additional changes, and we are awaiting the maintainer's approval.

flbn:
@colinhacks there's some other people with interest, would love to help out and get this merged if this is the desired fix.

flbn:
^^^ following up again, working around this isn't great

Issue #4977, opened 2025-07-25
Add `.with()`


colinhacks:
An escape hatch API. Calling `schema.with<T>()` or `schema.with(T)` will return `typeof schema & { _zod: T }`. Basically a way to override any of the schemas internals, including:

Input & output types #4032 

```ts
z.string().check(z.lowercase()).with<{ input: Lowercase<string>, output: Lowercase<string> }>
```

Brands: #4764 

```ts
const USD = z.string().with<{
  input: string & z.$brand<"USD">;
  output: string & z.$brand<"USD">;
}>();

type USDIn = z.input<typeof USD>; // string & z.core.$brand<"USD">
type USDOut = z.output<typeof USD>; // string & z.core.$brand<"USD">
```

Statically known "tags" for versioning, etc. #4319

```ts
z.object({
  name: z.string().with({ version: 2 })  // ZodString & $with<{ version: 2}> (also available at runtime in .def.with)
})
```

Subtypes: `string & {}`

Others:
- Named elements in tuple: https://github.com/colinhacks/zod/issues/4481
- Loose enum (`string & {}`): https://github.com/colinhacks/zod/pull/4785/files

jedwards1211:
@colinhacks I think intersecting with a complex generic class instance leads to bad performance when TS measures variance.  IIRC I had tried something similar years ago and ran into TS compiler performance problems before I settled on the approach of creating wrapper classes that extend `ZodType` (with Zod 3) to add custom fields because it has better performance.

The issue was something like, in many cases when comparing class instance types, TS can skip a deep structural comparison, but if any arbitrary properties are intersected on, it always has to do the complete deep comparison.

jiji-hoon96:
Hello, @colinhacks  I think this feature would be useful, so I would like to contribute to it. Would it be okay if I worked on it and submitted a PR?

PR #5099, opened 2025-08-14
feat: add WithOptions and WithSchema types for .with() method

jiji-hoon96:
## Description
Implements `.with()` method to override input/output types while preserving validation logic.

## Usage
```typescript
// Basic usage
const schema = z.string().with<string, "CustomString">();

#4977


colinhacks:
Closing — `.with()` already exists as an alias for `.check()`, so the name is taken. Overriding the inferred input/output types is a different feature and would need a different name and a separate discussion. Thanks for putting it together.

PR #4622, opened 2025-06-05
misc 1

colinhacks:
- **WIP**
- **Add**

PR #5115, opened 2025-08-19
fix: Support overriding `$defs` / `definition`

### LINKED ISSUE #5103: Override isn't called in all instances of the schema
If a schema is referenced multiple times, and the second usage of the schema isn't the same, the override is not processed.

```ts
const fooSchema = z.string().meta({
  id: 'ORIGINAL',
});

const mySchemaCorrect = z.object({
  foo: fooSchema,
  foo2: fooSchema,
});

const mySchemaIncorrect = z.object({
  foo: fooSchema.describe('A foo string'),
  foo2: fooSchema,
});

const parseSchema = (schema: ZodType<any>) => {
  return z.toJSONSchema(schema, {
    target: 'draft-7',
    unrepresentable: 'throw',
    cycles: 'throw',
    reused: 'inline',
    io: 'input',
    override(ctx) {
      if ('id' in ctx.jsonSchema) {
        ctx.jsonSchema.id = 'CHANGED';
      }
    },
  });
};

const jsonSchemaCorrect = parseSchema(mySchemaCorrect);
/*
{
  '$schema': 'http://json-schema.org/draft-07/schema#',
  type: 'object',
  properties: {
    foo: { '$ref': '#/definitions/ORIGINAL' },
    foo2: { '$ref': '#/definitions/ORIGINAL' }
  },
  required: [ 'foo', 'foo2' ],
  definitions: { ORIGINAL: { id: 'CHANGED', type: 'string' } }
}
*/
const jsonSchemaIncorrect = parseSchema(mySchemaIncorrect);
/*
{
  '$schema': 'http://json-schema.org/draft-07/schema#',
  type: 'object',
  properties: {
    foo: {
      description: 'A foo string',
      allOf: [ { '$ref': '#/definitions/ORIGINAL' } ]
    },
    foo2: { '$ref': '#/definitions/ORIGINAL' }
  },
  required: [ 'foo', 'foo2' ],
  definitions: { ORIGINAL: { id: 'ORIGINAL', type: 'string' } }
}
*/
```

This prevents me from commenting usages of the schema. I needed this to delete the `id` property from the JSON Schema, using:

```
if ('id' in ctx.jsonSchema) {
    ctx.jsonSchema.id = undefined;
}
```

reteps:
Closes #5103

The previous behavior had this logic:

- If schema has parent, override isn’t called on child schema
- A .describe or .meta will set a parent

However, this logic considers the concept of a parent based on the Zod schema;  which in certain cases makes it impossible to override a schema with no parent in the outputted JSON Schema. This is a bug fix to support overriding schemas at the root level. An alternative fix would be smarter detection of what counts as a parent schema: considering the concept of 'parent' in a JSON Schema `$ref` as well.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -1535,6 +1535,52 @@ test("describe with id", () => {
   `);
 });
 
+test("override with id", () => {
+  const jobId = z.string().meta({ id: "jobId" });
+
+  const a = z.z.toJSONSchema(
+    z.object({
+      current: jobId.describe("Current job"),
+      previous: jobId.describe("Previous job"),
+    }),
+    {
+      override(ctx) {
+        if (ctx.path.join("/") === "$defs/jobId") {
+          ctx.jsonSchema.id = "jobIdChanged";
+        }
+      },
+    }
+  );
+
+  expect(a).toMatchInlineSnapshot(`
+    {
+      "$defs": {
+        "jobId": {
+          "id": "jobIdChanged",
+          "type": "string",
+        },
+      },
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "additionalProperties": false,
+      "properties": {
+        "current": {
+          "$ref": "#/$defs/jobId",
+          "description": "Current job",
+        },
+        "previous": {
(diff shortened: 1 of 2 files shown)


reteps:
Related: #4622 #5533

colinhacks:
Fixed in #5578 👍

PR #5119, opened 2025-08-20
feat(string): add .slug() convenience validator

rishi111-code:
Adds `z.string().slug(message?)` as a convenience method for validating lowercase, URL-safe slugs.

## Rationale
Zod provides built-in helpers like `.email()` / `.url()` but lacks a first-class `.slug()`. Slug validation is common across CMS, blog platforms, and REST resources. This PR wraps `.regex()` with a canonical pattern to improve discoverability and DX without expanding internal check kinds.

## Implementation
- Uses `/^[a-z0-9]+(?:-[a-z0-9]+)*$/` (lowercase letters, digits, single hyphens between segments).
- No leading/trailing hyphen, no consecutive hyphens.
- Accepts optional custom error message.
- Adds tests in `src/__tests__/string.slug.test.ts`.

## Alternatives
- As a helper export (`isSlug`) if adding to `ZodString` is undesirable.

## Notes
- Keeps bundle impact minimal (reuses `.regex()`).

### DIFF
--- a/packages/zod/src/v3/helpers/slug.ts
+++ b/packages/zod/src/v3/helpers/slug.ts
@@ -1,0 +1,1 @@
+export const isSlug = (value: string) => /^[a-z0-9]+(?:-[a-z0-9]+)*$/.test(value);
--- a/packages/zod/src/v3/types.ts
+++ b/packages/zod/src/v3/types.ts
@@ -854,7 +854,19 @@ export class ZodString extends ZodType<string, ZodStringDef, string> {
           });
           status.dirty();
         }
-      } else if (check.kind === "cuid2") {
+      } else if (check.kind === "slug") {
+        const slugRegex = /^[a-z0-9]+(?:-[a-z0-9]+)*$/;
+        if (!slugRegex.test(input.data)) {
+          ctx = this._getOrReturnCtx(input, ctx);
+          addIssueToContext(ctx, {
+            validation: "slug",
+            code: ZodIssueCode.invalid_string,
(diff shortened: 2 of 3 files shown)


packages/zod/src/v3/helpers/slug.ts
```diff
@@ -0,0 +1 @@
+export const isSlug = (value: string) => /^[a-z0-9]+(?:-[a-z0-9]+)*$/.test(value);
```
imranbarbhuiya, line 1: You arent using this fn and zod doesnt have any other helper fns like this so ig this should be removed and the ZodString one is the way to go

packages/zod/src/v3/types.ts
```diff
@@ -854,7 +854,19 @@ export class ZodString extends ZodType<string, ZodStringDef, string> {
           });
           status.dirty();
         }
-      } else if (check.kind === "cuid2") {
+      } else if (check.kind === "slug") {
+        const slugRegex = /^[a-z0-9]+(?:-[a-z0-9]+)*$/;
```
imranbarbhuiya, line 858: For consistency, I would put this regex in 600 line with other regexes

imranbarbhuiya (changes requested):

colinhacks:
Zod 3 is feature-frozen but I've implemented `slugify` here as an *overwrite mutation* (similar to trim, toLowerCase, etc). 3de39eea6f7ed286ae182093d0c91f3a6fdcca06

Landed in 4.2

Issue #4646, opened 2025-06-07
Feature: Add "json" schema primitive


dbartholomae:
Hi there! Great work with zod and v4!
Now that it's easier to tree-shake zod thanks to v4-mini, would you be open to add more commonly-used schemas as primitives?
Specifically, I'm thinking about json:
```
// Json types from https://github.com/sindresorhus/type-fest/blob/041e67e1f180b1d6adddca57f9429789c16edf81/source/json-value.d.ts#L31 inlined here to avoid adding a dependency
export type JsonObject = {[Key in string]: JsonValue} & {[Key in string]?: JsonValue | undefined};
export type JsonArray = JsonValue[] | readonly JsonValue[];
export type JsonPrimitive = string | number | boolean | null;
export type JsonValue = JsonPrimitive | JsonObject | JsonArray;

export function json() {
  return z.string().transform((str, ctx): JsonValue => {
    try {
      return JSON.parse(str) as JsonValue;
    } catch {
      ctx.addIssue({ code: "custom", message: "Invalid JSON", input: str });
      return z.NEVER;
    }
  });
}
```

This is such a common use case that it would be awesome to have it as part of the library.

If you're in principle open to adding this, but don't have the time due to the current changes coming with v4, I'm also very happy to provide a PR.

coderatchet:
Hey there, you can fine an existing JSON utility defined here https://zod.dev/api#json hope it's what you're after!

dbartholomae:
Thanks! Not sure how I overlooked that. But it looks like this expects an object as input, while the type I'm suggesting is expecting a string as input and is then JSON.parsing it.
Maybe adding `coerce.json()` would be the solution then?

coderatchet:
I don't think there is currently a json parser for strings, but I'm sure you could probably wrap the native JSON.parse method to achieve what you wanted, if so, you would probably not need a specific json parser anyway. Good luck

dbartholomae:
Yes, JSON.parse works, see the code example above. This issue is just about making a common util available directly in zod.

colinhacks:
While it seems reasonable that Zod could have a `z.json` that automatically calls `JSON.parse()` during Zod's own `parse()` call, ultimately 

1) This ship has sailed somewhat since `z.json()` already expects a JSON-serializable *object* (this was probably an oversight in retrospect)
2) This is achievable with transforms/pipe as others have said.

[...]

PR #5120, opened 2025-08-20
added z.jsonString() schema primitive

Sachinkry:
Implements the `z.jsonString()` schema primitive for validating and parsing JSON strings, addressing issue #4646.

Basic JSON Syntax Validation
```typescript
const schema = z.jsonString();
schema.parse('{"name": "Alice", "age": 30}'); // ✅ returns original string
schema.parse("invalid json"); // ❌ throws error
```

JSON Parsing with Schema Validation
```typescript
const userSchema = z.jsonString(z.object({
  name: z.string(),
  age: z.number(),
}));
userSchema.parse('{"name": "Alice", "age": 30}'); // ✅ returns { name: "Alice", age: 30 }
```

Details
- Implement z.jsonString() to validate JSON strings and optionally parse+validate against an inner schema.
- Adds core $ZodJSONString, public APIs (classic/mini), and pipe-based parsing to inner schemas.
- Emits invalid_format on JSON syntax errors; returns original string when no inner schema is provided.
- Includes focused tests (10) and documentation updates (API + v4 changelog).


colinhacks:
Thanks for the PR here. This is one of the use cases addressed by Zod's new(-ish) codecs feature, which I think obviates the need for this PR somewhat. https://zod.dev/codecs

I think the idea of a `z.jsonString()` is still quite reasonable, but it's hard to imagine why you'd do that without eventually parsing that into usable JSON data (in which case you're double-parsing and should probably be using a codec). So I'm a little concerned about encouraging bad practice here. It's also each enough to do with z.stringFormat:

```ts
const A = z.stringFormat("json_string", (val) => {
  try {
    JSON.parse(val);
    return true;
  } catch {
    return false;
  }
});
```

PR #5052, opened 2025-08-06
feat: add openapi target to json schema

colinhacks:
## Summary
- add an `openapi-3.0` target to `z.toJSONSchema`
- support nullable schemas and unions using OpenAPI `nullable`
- document the new target

## Testing
- `pnpm --filter zod test`

------
https://chatgpt.com/codex/tasks/task_e_6892b797d7f4832f9b435a56a2693a6b

Issue #3978, opened 2025-02-09
Compile-time type checking `parse`


CaptainStiggz:
I'm sure this has been raised before, but I couldn't find a satisfying answer. Since the input argument of `parse(input: unknown)`, is of type `unknown`, the `parse` method can accept anything and guarantee the type of the output. However, in many cases, this seems to sidestep the whole point of typescript, turning what would be compile-time errors into runtime errors. Consider this example:

```typescript
import { z } from 'zod'

const ExampleSchema = z.object({
  name: z.string(),
})

type Example = z.infer<typeof ExampleSchema>

export function exampleRaw(): Example {
  return {
    // typescript compile-time error
    wrongKey: 'Foo',
  }
}

export function exampleParse(): Example {
  // no compile-time error
  return ExampleSchema.parse({
    wrongKey: 'Foo',
  })
}
```

In this case above, we'd be better of without zod at all, since adding the `parse` method obfuscates the compile-time error. In this simple example, raw typescript is preferable, but one might imagine wanting to use some features of zod, like `transform` or the fact that `passthrough` is false by default to have more precise control over which properties come out of the function.

We can build a function like this: which I think would do everything we want, getting us both the power of `zod` and the power of standard compile-time type checking. 

```typescript
export function typedParse<T extends z.input<typeof ExampleSchema>>(input: T): Example {
  return ExampleSchema.parse(input)
}

export function exampleTypedParse(): Example {
  // typescript compile-time error
  return typedParse({
    wrongKey: 'Foo',
  })
}
```

The question is, why isn't this the default behavior of `parse`? Or why isn't there at least some sort of a `typedParse` function built into the library?

colinhacks:
This is under consideration for Zod 4. It isn't the default behavior for parse because in general inputs are `unknown` - that's why they're getting parsed.

colinhacks:
This is now achievable with the new `.decode()` methods in Zod 4.1, which were added in conjunction with the Codec API: https://colinhacks.com/essays/introducing-zod-codecs

Issue #1892, opened 2023-01-16
`.parse()` Typescript linter accepts anything (string, number...)


sneko:
Hi,

I didn't find someone having my issue so I'm probably doing something wrong but... I can't figure the solution out.

Let's say I have:
```ts
const TestSchema = z.object({
  email: z.string(),
});

TestSchema.parse('hello');
```

The last line will not throw an error with the Typescript whereas when hovering the parse function I can see:
```ts
(method) ZodType<{ email: string; }, ZodObjectDef<{ email: ZodString; }, "strip", ZodTypeAny>, { email: string; }>.parse(data: unknown, params?: Partial<z.ParseParams> | undefined): {
    email: string;
}
```

**Note that I use `strict: true` in my `tsconfig.json` (and at runtime, `zod` throws an error obviously).

A "workaround" would be to always do this:
```ts
const TestSchema = z.object({
  email: z.string(),
});
type TestSchemaType = z.infer<typeof TestSchema>;

TestSchema.parse(<TestSchemaType>'hello');
```
... but it's pretty verbose and annoying.

Any idea what's the issue to get the linter working normally?

Thank you,

JacobWeisenburger:
> Any idea what's the issue to get the linter working normally?

What do you mean by 'normally'?

---

> but I don't get why this is not used by default?

https://github.com/colinhacks/zod/issues/1748#issuecomment-1364573095

---

> ```ts
> const typedParse = <Schema extends z.ZodSchema> (
>   schema: Schema, data: z.infer<Schema>
> ): z.infer<Schema> => schema.parse( data )
> 
> typedParse(ATest, 'hello');
> ```

Another way you can solve this problem:

```ts
type TypedParseSchema<Schema extends z.ZodSchema> = Omit<Schema, 'parse'> & {
    parse ( data: z.infer<Schema> ): z.infer<Schema>
}
const useTypedParse = <Schema extends z.ZodSchema>
    ( schema: Schema ): TypedParseSchema<Schema> => schema

const schemaWithTypedParse = useTypedParse( z.object( {
    email: z.string(),
} ) )

schemaWithTypedParse.parse( 'hello' )
//                          ^^^^^^^
// Argument of type 'string' is not assignable to parameter of type '{ email: string; }'.
```

Would this work for you?

PR #5130, opened 2025-08-21
Zod 4.1

### LINKED ISSUE #5108: not clear from documents that v4 requires z.literal key for discriminated union
This might be a small thing, but 

Can I suggest that the documentation at https://zod.dev/api note more clearly that the discriminated union key must be a literal if that is the case? I recognize that it is a literal in the example, but I had wrestled with it for some time until I realized it could not be a string. 

Thanks!

(also, RIP Terence Stamp - General Zod, who passed today at 87)

### LINKED ISSUE #5116: 4.0.16 removed the only way to get a typed array of keys from an object schema in zod/mini
- From https://github.com/colinhacks/zod/pull/5045#issuecomment-3189995841

In 4.0.15, one could do this to get a typed array of keys from an object schema:

```ts
import * as z from "zod/mini";

const Settings = z.object({
  foo: z.string(),
  bar: z.string(),
  baz: z.string(),
  quux: z.string(),
});

const allKeys = z.keyof(Settings).def.values;
// allKeys has type ("foo" | "bar" | "baz" | "quux")[]

function showSettings(settings: z.infer<typeof Settings>) {
  for (const key of allKeys) console.log(key, settings[key]);
}
```

However, this broke in 4.0.16 with #5045.

One can write `Object.keys(z.keyof(Settings).def.entries)` or `Object.keys(Settings.shape)`, but these have the weaker type `string[]` (see https://github.com/microsoft/TypeScript/pull/12253#issuecomment-263132208), which turns `settings[key]` into an error with the TypeScript `strict` flag.

The full `zod` API has `z.keyof(Settings).options` of type `("foo" | "bar" | "baz" | "quux")[]`, but this is not available in `zod/mini`.

### LINKED ISSUE #5127: Stricter UUID validation should allow sentinel value of max UUID as valid.
With zod v4 validation of uuid was [updated](
https://zod.dev/v4/changelog?id=stricter-uuid) and now corresponds stricter RFC spec. But RFC 9562 spec defines 2 special UUIDs - Nil and [Max (RFC)](https://www.rfc-editor.org/rfc/rfc9562#name-max-uuid) which are `00000000-0000-0000-0000-000000000000` and `ffffffff-ffff-ffff-ffff-ffffffffffff` correspondingly.

Zod validation of `z.uuid()` supports Nil UUID fine but treats [Max UUID as invalid](https://zod-playground.vercel.app?appdata=N4IgzgxgFgpgtgQxALhALwHQFcsEsAmAFAJQgA0IAbggDZYxgoDaIA5AGadecC03fXAb24iurEAF0KlGACcwuAPYA7FCAAsGAAwYAjAHZyIXGABai-AFlcy3Cna0wMAL5A). 

This does not look correct.

Spec outlines it as max uuid can be used as a special value:

> A Max UUID value can be used as a sentinel value in situations where a 128-bit UUID is required, but a concept such as "end of UUID list" needs to be expressed and is reserved for such use as needed for implementation-specific situations. Note that the Max UUID value falls within the range of the "yet-to-be defined" future UUID variant as per the last row of [Table 1](https://www.rfc-editor.org/rfc/rfc9562#table1) rather than the variant defined by this document.

And although max uuid falls into "Reserved for future definition" category of a spec (because of `F` in variant field) - it also serves as a valid sentinel value in current spec and should be allowed.

For anyone who needs a quick fix - switching from z.uuid() to z.guid() as doc suggesting, allows max uuid value, [as expected](https://zod-playground.vercel.app?appdata=N4IgzgxgFgpgtgQxALhALwHQHMCuBLAEwAoBKEAGhADcEAbHGMFAbRAHIAzL7rgWh-7dBfHqO5sQAXUpUYAJzB4A9gDsUIACwYADBgCMAdgog8YAFpKCAWTwq8KDnTAwAvkA).

This issue may have a broader impact as there are multiple packages that generate code from openapi definitions and use zod for schema validation - so `z.uuid()` is not so easily replaced with `z.guid()` in code-generated models.

colinhacks:
# Zod v4.1 

The first minor version since the introduction of Zod 4 back in May. This version contains a number of features that barely missed the cut. With Zod 4 stable and widely adopted, there's now more time to resume feature development. 
 
## Codecs

This is the flagship feature of this release. Codecs are a new API & schema type that encapsulates a _bi-directional transformation_. It's a huge missing piece in Zod that's finally filled, and it unlocks some totally new ways to use Zod. 

```ts
const stringToDate = z.codec(
  z.iso.datetime(),  // input schema: ISO date string
  z.date(),          // output schema: Date object
  {
    decode: (isoString) => new Date(isoString), 
    encode: (date) => date.toISOString(),
  }
);
```

New top-level functions are added for processing inputs in the _forward_ direction ("decoding") and _backward_ direction ("encoding").

```ts
z.decode(stringToDate, "2025-08-21T20:59:45.500Z")
// => Date

z.encode(stringToDate, new Date())
// => "2025-08-21T20:59:45.500Z"
```

> For bundle size reasons, these are only available as top-level functions instead of methods. 

To parse data in the "forward" direction (AKA _decode_), you can continue using the existing `.parse` method/function. Alternatively you can use the new top-level `z.decode` function:

```ts
stringToDate.parse("2024-01-15T10:30:00.000Z"); // method form
z.parse(stringToDate, "2024-01-15T10:30:00.000Z"); // functional form
// => Date

z.decode(stringToDate, "2024-01-15T10:30:00.000Z")
// => Date
```

While `z.decode()` behaves identically to `z.parse()` in the example above, its inputs are *strongly typed* . That is, it expects an input of type `string`, whereas `.parse()` accepts `unknown`. 

```ts
z.parse(stringToDate, Symbol('not-a-string')); // functional form
// => fails at runtime, but no TypeScript error

z.decode(stringToDate, Symbol("not-a-string"));
//                     ^ ❌ Argument of type 'symbol' is not assignable to parameter of type 'Date'. ts(2345)
```

> This is a highly requested feature unto itself: 
> 
> - #3860
> - #1748
> - #3978
> - #1892

To parse data in the "backward direction" (AKA _encode_), use the new top-level `z.encode()` function.

```ts
z.encode(stringToDate, new Date())
// => "2025-08-21T20:59:45.500Z"
```

You can use any Zod schema with `z.encode()`. The vast majority of Zod schemas are *non-transforming* (the input and output types are identical) so `z.decode()` and `z.encode()` behave identically. Only certain schema types change their behavior:

- **Codecs** — runs from `B->A` and executes the `encode` transform during encoding
- **Pipes** — these execute `B->A` instead of `A->B`
- **Defaults and prefaults** — Only applied in the forward direction
- **Catch** — Only applied in the forward direction

> **Note** — To avoid increasing bundle size unnecessarily, there are no method equivalents of `z.decode()` and `z.encode()`. These are top-level functions only.

The usual _async_ and _safe_ variants exist as well:

```ts
// decode functions
z.decode(stringToDate, "2024-01-15T10:30:00.000Z")
await z.decodeAsync(stringToDate, "2024-01-15T10:30:00.000Z")
z.safeDecode(stringToDate, "2024-01-15T10:30:00.000Z")
await z.safeDecodeAsync(stringToDate, "2024-01-15T10:30:00.000Z")

// encode functions
z.encode(stringToDate, new Date())
await z.encodeAsync(stringToDate, new Date())
z.safeEncode(stringToDate, new Date())
await z.safeEncodeAsync(stringToDate, new Date())
```

Below are some "worked examples" for some commonly-needed codecs. These examples are all tested internally for correctness. Just copy/paste them into your project as needed. There is a more comprehensive set available at [zod.dev/codecs](https://zod.dev/codecs).

### `stringToBigInt`

Converts `bigint` into a serializable form.

```
const stringToBigInt = z.codec(z.string(), z.bigint(), {
  decode: (str) => BigInt(str),
  encode: (bigint) => bigint.toString(),
});

z.decode(stringToBigInt, "12345");  // => 12345n
z.encode(stringToBigInt, 12345n);   // => "12345"
```
 
### `json`

Parses/stringifies JSON data.

```ts
const jsonCodec = z.codec(z.string(), z.json(), {
  decode: (jsonString, ctx) => {
    try {
      return JSON.parse(jsonString);
    } catch (err: any) {
      ctx.issues.push({
        code: "invalid_format",
        format: "json_string",
        input: jsonString,
        message: err.message,
      });
      return z.NEVER;
    }
  },
  encode: (value) => JSON.stringify(value),
});
```

To further validate the data, `.pipe()` the result of this codec into another schema.

```ts
const Params = z.object({ name: z.string(), age: z.number() });
const JsonToParams = jsonCodec.pipe(Params);

z.decode(JsonToParams, '{"name":"Alice","age":30}');  // => { name: "Alice", age: 30 }
z.encode(JsonToParams, { name: "Bob", age: 25 });     // => '{"name":"Bob","age":25}'
```

### Further reading

**For more examples and a technical breakdown of how encoding works, [read the full docs →](https://zod.dev/codecs)** That page contains implementations for several other commonly-needed codecs:

- [**`stringToNumber`**](/codecs#stringtonumber)
- [**`stringToInt`**](/codecs#stringtoint)
- [**`stringToBigInt`**](/codecs#stringtobigint)
- [**`numberToBigInt`**](/codecs#numbertobigint)
- [**`isoDatetimeToDate`**](/codecs#isodatetimetodate)
- [**`epochSecondsToDate`**](/codecs#epochsecondstodate)
- [**`epochMillisToDate`**](/codecs#epochmillistodate)
- [**`jsonCodec`**](/codecs#jsoncodec)
- [**`utf8ToBytes`**](/codecs#utf8tobytes)
- [**`bytesToUtf8`**](/codecs#bytestoutf8)
- [**`base64ToBytes`**](/codecs#base64tobytes)
- [**`base64urlToBytes`**](/codecs#base64urltobytes)
- [**`hexToBytes`**](/codecs#hextobytes)
- [**`stringToURL`**](/codecs#stringtourl)
- [**`stringToHttpURL`**](/codecs#stringtohttpurl)
- [**`uriComponent`**](/codecs#uricomponent)
- [**`stringToBoolean`**](/codecs#stringtoboolean)

## `.safeExtend()`

The existing way to add additional fields to an object is to use `.extend()`. 

```ts
const A = z.object({ a: z.string() })
const B = A.extend({ b: z.string() })
```

Unfortunately this is a bit of a misnomer, as it allows you to overwrite existing fields. This means the result of `.extend()` may not literally `extend` the original type (in the TypeScript sense).

```ts
const A = z.object({ a: z.string() }) // { a: string }
const B = A.extend({ a: z.number() }) // { a: number }
```

To enforce true `extends` logic, Zod 4.1 introduces a new `.safeExtend()` method. This statically enforces that the newly added properties conform to the existing ones.

```ts
z.object({ a: z.string() }).safeExtend({ a: z.number().min(5) }); // ✅
z.object({ a: z.string() }).safeExtend({ a: z.any() }); // ✅
z.object({ a: z.string() }).safeExtend({ a: z.number() });
//                                       ^  ❌ ZodNumber is not assignable 
```

Importantly, this new API allows you to safely extend objects containing refinements.

```ts
const AB = z.object({ a: z.string(), b: z.string() }).refine(val => val.a === val.b);
const ABC = AB.safeExtend({ c: z.string() });
// ABC includes the refinements defined on AB
// a regular `.extend()` would have thrown an error
```

## `z.hash()`

A new top-level string format for validating hashes produced using various common algorithms & encodings.

```ts
const md5Schema = z.hash("md5");          
// => ZodCustomStringFormat<"md5_hex">

const sha256Base64 = z.hash("sha256", { enc: "base64" }); 
// => ZodCustomStringFormat<"sha256_base64">
```

The following hash algorithms and encodings are supported. Each cell provides information about the expected number of characters/padding.

| Algorithm / Encoding | `"hex"`  | `"base64"`  | `"base64url"`  |
|-----------|-----------|--------------|-----------------|
| `"md5"`   | 32        | 24 (22 + "==") | 22              |
| `"sha1"`  | 40        | 28 (27 + "=")  | 27              |
| `"sha256"`| 64        | 44 (43 + "=")  | 43              |
| `"sha384"`| 96        | 64 (no padding)| 64              |
| `"sha512"`| 128       | 88 (86 + "==") | 86              |

## `z.hex()`

To validate hexadecimal strings of any length.

```ts
const hexSchema = z.hex();

hexSchema.parse("123abc");    // ✅ "123abc"
hexSchema.parse("DEADBEEF");  // ✅ "DEADBEEF" 
hexSchema.parse("xyz");       // ❌ ZodError
```

---

## Commits

- edd4feaa - Closes #5127
- 5d4a3159 - Closes #5116
- f3f09556 - Closes #5108
- 0114d5bd - #5122
- 3b077c3a - #5121
- 1e06af8e - #5113
- b01b6f39 — #5052
- 571ab0c3 — #5051
- d3ea1119 — #5049
- b8e3f877 — #4865

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -2437,7 +2437,7 @@ const UniqueStringArray = z.array(z.string()).check((ctx) => {
 
 ## Codecs
 
-> **New** — Introduced in `zod@4.1`. Refer to the dedicated [Codecs](/codecs) page for more information.
+> **New** — Introduced in Zod 4.1. Refer to the dedicated [Codecs](/codecs) page for more information.
 
 Codecs are a special kind of schema that implement *bidirectional transformations* between two other schemas.
 


samchungy:
>z.object({ a: z.string() }).safeExtend({ a: z.number().min(5) }); // ✅

Should this be:

```ts
z.object({ a: z.string() }).safeExtend({ a: z.string().min(5) });
```

instead?

packages/docs/content/api.mdx
```diff
@@ -1141,10 +1206,52 @@ This approach has a few advantages:
...
+3. It's more `tsc`-efficient — the `.extend()` method can be expensive on large schemas, and due to [a TypeScript limitation](https://github.com/microsoft/TypeScript/pull/61505) it gets quadratically more expensive when calls are chained
 4. If you wish, you can change the strictness level of the resulting schema by using `z.strictObject()` or `z.looseObject()`
 </Callout>
 
+### `.safeExtend()`
+
+The `.safeExtend()` method works similarly to `.extend()`, but it won't let you overwrite an existing properly with a non-assignable schema. In other words, the result of `.safeExtend()` will have an inferred type that [`extends`](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#conditional-type-constraints) the original (in the TypeScript sense).
+
+```ts
+z.object({ a: z.string() }).safeExtend({ a: z.number().min(5) }); // ✅
```
samchungy:
```suggestion
z.object({ a: z.string() }).safeExtend({ a: z.string().min(5) }); // ✅
```

colinhacks: Yep thanks!

PR #5146, opened 2025-08-24
fix(v4): toJSONSchema - add missing oneOf inside items in tuple conversion

### LINKED ISSUE #5143: [v4] toJSONSchema - `openapi-3.0` - wrong tuple output
Given the following code:

```ts
z.tuple([z.string(), z.number()])
```

the output of

```ts
z.toJSONSchema(zodSchema, { target: "openapi-3.0" });
```

is

```ts
{
  "type": "array",
  "items": [
    {
      "type": "string"
    },
    {
      "type": "number"
    }
  ]
}
```

when it should be something like

```ts
{
  type: "array",
  items: {
    oneOf: [
      { type: "string" },
      { type: "string" },
    ],
  },
  minItems: 2,
  maxItems: 2,
}
```

You can test the behaviour in this repository: 
https://github.com/marcalexiei/zod-playground/tree/main/examples/openapi-30.

marcalexiei:
- really closes #5143
- fixes oversight of #5145

---

In #5145 I forgot that `items` shouldn't be an array...
but an object with a `oneOf` property which is the array containing the items:

image

Sorry about the fuss 😅

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -696,7 +696,7 @@ describe("toJSONSchema", () => {
     expect(z.toJSONSchema(schema, { target: "openapi-3.0" })).toMatchInlineSnapshot(`
       {
         "items": {
-          "oneOf": [
+          "anyOf": [
             {
               "type": "string",
             },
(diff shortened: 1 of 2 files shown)


marcalexiei:
I have pushed a commit with a test for draft-7 but https://github.com/colinhacks/zod/pull/5152 address it better so I removed it.

colinhacks:
Thanks for all your work on these! Your PRs have all landed in 4.1.3 👍

PR #5148, opened 2025-08-25
docs: fix refine fn to encode works properly

imranbarbhuiya:
Just a small docs change, currently the first `encode` marked as passes also fails due to > 2000 check


colinhacks:
Good catch!

PR #5150, opened 2025-08-25
docs(codecs): update example to use payloadSchema

yusuke99:
You've probably meant to use `payloadSchema` instead of `stringToDate`?

Reference: https://zod.dev/codecs


colinhacks:
Thanks!

Issue #5147, opened 2025-08-24
Suggestion: improve `toJSONSchema` tests strategy


marcalexiei:
Proposal related to https://github.com/colinhacks/zod/issues/5143#issuecomment-3218366995

> Wow OpenAPI 3.0 is even worse than I thought 🫠 [...]. Definitely happy to have more tests here.

---

To avoid regression or oversights that might lead to invalid output (like #5146),
I would like to propose an addition to the `openapi-3.0` `toJSONSchema`  `target` tests:

How about building a custom matcher to validate this function’s output against the OpenAPI specification?  

I’ve used [@seriousme/openapi-schema-validator](https://github.com/seriousme/openapi-schema-validator) before to validate 3.x OpenAPI schemas,
and I think it could be really handy for creating a custom `vitest` matcher.

<details><summary>Proposal for a custom matcher</summary>
<p>

```ts
import { Validator } from '@seriousme/openapi-schema-validator';
import { expect } from 'vitest';

interface CustomMatchers<R = unknown> {
  toBeValidInsideOpenAPISchema30: () => R;
}

declare module 'vitest' {
  interface Assertion<T = any> extends CustomMatchers<T> {}
  interface AsymmetricMatchersContaining extends CustomMatchers {}
}

const validator = new Validator();

expect.extend({
  async toBeValidInsideOpenAPISchema30(
    this: { isNot: boolean },
    zodSchema: $ZodType,
  ): Promise<{ pass: boolean; message: () => string }> {
    const res = await validator.validate({
      openapi: '3.0.0',
      // ...
      components: { schemas: { test: zodSchema } }
    });

    const messages = [
      `OpenAPI schema is${this.isNot ? '' : ' not'} valid against ${validator.version}`,
    ];

    if (!res.valid) {
      messages.push(JSON.stringify(res.errors, null, 2));
    }

    return {
      pass: res.valid,
      message: () => messages.join('\n'),
    };
  },
});

```
</p>
</details> 

Using this custom matcher, we can ensure the schema is actually valid,  
rather than relying only on string snapshots,
which sometimes can be generated incorrectly (as happened to me earlier).

Personally, I like to use both snapshots and validation.  
Having an inline or file snapshot alongside a validator as a reference is really useful, IMHO.

---

What do you think? I understand this might feel like overkill, but I thought it was worth mentioning 🥲

colinhacks:
[...]

PR #5156, opened 2025-08-26
fix(v4): toJSONSchema - wrong tuple with `null` output when targeting `openapi-3.0`

### LINKED ISSUE #5155: [v4] toJSONSchema - `openapi-3.0` - when a tuple contains `null`  the output is not correct
With Zod 4.1.3, given the following code:

```ts
z.tuple([z.string(), z.number(), z.null()])
```

the output of

```ts
z.toJSONSchema(zodSchema, { target: "openapi-3.0" });
```

is

```ts
{
  "type": "array",
  "items": {
    "anyOf": [
      {
        "type": "string"
      },
      {
        "type": "number"
      },
      {
        "type": "null"
      }
    ]
  },
  "minItems": 3,
  "maxItems": 3
}
```

when it should be something like

```ts
{
  type: "array",
  items: {
    anyOf: [
      { type: "string" },
      { type: "number" },
    ],
    nullable: true
  },
  minItems: 2,
  maxItems: 2,
}
```

You can test the behaviour in this repository: 
https://github.com/marcalexiei/zod-playground/tree/main/examples/openapi-30.

marcalexiei:
- closes #5155

### DIFF
--- a/packages/zod/src/v4/core/to-json-schema.ts
+++ b/packages/zod/src/v4/core/to-json-schema.ts
@@ -528,9 +528,8 @@ export class JSONSchemaGenerator {
           case "nullable": {
             const inner = this.process(def.innerType, params);
             if (this.target === "openapi-3.0") {
-              Object.assign(_json, inner);
-              (_json as any).nullable = true;
               result.ref = def.innerType;
+              _json.nullable = true;
             } else {
               _json.anyOf = [inner, { type: "null" }];
             }


packages/zod/src/v4/classic/tests/to-json-schema.test.ts
```diff
@@ -1,7 +1,34 @@
...
+    // `console.error` should make `vitest` trow an unhandled error
+    // printing the validation messages in consoles
+    console.error(
+      `OpenAPI schema is not valid against ${openAPI30Validator.version}`,
+      JSON.stringify(res.errors, null, 2)
+    );
+  }
+
+  return true;
+};
```
marcalexiei, line 30: Since the changes required to solve the issue are minimal, 
I used this PR as a proof of concept for what I proposed in #5147.

Let me know if you’d like me to keep using this approach or if you’d prefer something else 😉.

colinhacks:
I slightly hijacked this PR to handle `z.null()` in a way I'm more comfortable with. It's less pretty but the new approach composes much better and avoids a lot of special casing logic. Still fixes the original issue (avoids `"type": "null"` entirely when targeting OAS 3.0) and keeps the new testing logic (very nice, thanks!).

PR #5164, opened 2025-08-27
feat(v4): add change and safeChange methods to modify existing schema properties

chribjel:
This adresses #2891

i understand that .extend has overlapping functionality, but it is nice to have a "safe" way to know that you ONLY change the schema, not extend it.

Very understandable if it is not merged, but there are many people wanting this functionality in the issue, and it still has activity after being closed.


colinhacks:
Though I see the utility here, I think it's not worth the complexity and bundle size increase. This could be implemented as a top-level `change(schema, changes)` function if anyone has a strong need for it in their codebase. Thanks for the good PR though!

PR #5166, opened 2025-08-27
docs: fix mentions of `v3.25.0` to `v3.25.1`

RiskyMH:
`v3.25.0` didn't upload the `dist` assets, so it shouldn't be mentioned as the baseline of supported versions.


colinhacks:
Perhaps. In practice it's very rare that this will be installed, and multiple early 3.25.x releases had various build issues for CJS before all the kinks were worked out. I don't want to confuse people around the lineages here. It's more important that people understand the 3.24.x -> 3.25.x -> 4.x progression.

RiskyMH:
Ah all good it's only really an issue for a few cases where lib authors didn't implement semver ^, so it only had one version available for v3.

PR #5168, opened 2025-08-27
Allow custom issue codes

CodingMeSwiftly:
Changes typing of custom issue class to allow arbitrary string values to be set for the code field.


colinhacks:
This breaks the discriminated union—every issue needs a a known string literal `code`. If you have several custom issues types you want to define, they should be sub-types of ZodCustomIssues (perhaps use `subcode` or something to discriminate between them).

Issue #4143, opened 2025-04-14
v4: what is the replacement for using z.function as a schema?


AndrewIngram:
The docs say this

> The result of z.function() is no longer a Zod schema. Instead, it acts as a standalone "function factory" for defining Zod-validated functions. The API has also changed; you define an input and output schema upfront, instead of using args() and .returns() methods.

We have Zod 3 code like this, which is used to give us type-safety (though presumably not runtime safety that a function matches a shape):

```
export const DynamicEnvVarSchema = z.object({
  from: z.literal('dynamic'),
  evaluate: z.function(z.tuple([]), z.promise(z.string())),
});
export type DynamicEnvVar = z.output<typeof DynamicEnvVarSchema>;
```

What would be the most appropriate way to do this in Zod 4?

sanadriu:
Any update on this?

shkreios:
@AndrewIngram @sanadriu What you can do is wrap the `z.function` inside `z.custom`, which calls `implement` or `implementAsync`.



## Implementation 
```ts
const functionSchema = <T extends z.core.$ZodFunction>(schema: T) =>
  z.custom<Parameters<T["implement"]>[0]>((fn) => schema.implement(fn))

const createAsyncFunctionSchema = <T extends z.core.$ZodFunction>(schema: T) =>
  z.custom<Parameters<T["implementAsync"]>[0]>((fn) => schema.implementAsync(fn))
```
## Example 
```ts
const DynamicEnvVarSchema = z.object({
  from: z.literal("dynamic"),
  evaluate: createAsyncFunctionSchema(z.function({ input: z.tuple([]), output: z.promise(z.string()) })),
})

type DynamicEnvVar = z.output<typeof DynamicEnvVarSchema>
```

```ts
// OutputType
type DynamicEnvVar = {
    from: "dynamic";
    evaluate: z.core.$InferInnerFunctionTypeAsync<z.ZodTuple<[], null>, z.ZodPromise<z.ZodString>>;
}
```

[...]

PR #5170, opened 2025-08-28
Update z.function() type to support array input

nicoespeon:
[It is documented](https://zod.dev/v4/changelog?id=zfunction) (and [in the main docs](https://zod.dev/api?id=functions) too) to work with array inputs like:

```ts
const myFunction = z.function({
  input: [z.object({
    name: z.string(),
    age: z.number().int(),
  })],
  output: z.string(),
});

myFunction.implement((input) => {
  return `Hello ${input.name}, you are ${input.age} years old.`;
});
```

But it doesn't work. Current type only supports `z.tuple()` (which is what's tested too).

Either the docs should be fixed to show a `z.tuple()` wrapper, or we could update the types to support both. In this PR, I went for the latter to have Zod v4 behave like the docs say. I've added a test to illustrate.

Feel free to decline and update the docs examples if you think it should not be like that.

Related to https://github.com/colinhacks/zod/issues/4143#issuecomment-3229560060

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -2059,11 +2059,11 @@ export const ZodFunction: core.$constructor<ZodFunction> = /*@__PURE__*/ core.$c
 );
 
 export function _function(): ZodFunction;
-export function _function<const In extends Array<core.$ZodType> = Array<core.$ZodType>>(params: {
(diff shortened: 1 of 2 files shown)


colinhacks:
Working for me on the latest versions. Make sure you're on the latest version.
Screenshot 2025-08-28 at 11 06 14

colinhacks:
I see, the problem is with readonly arrays specifically. Good catch.

PR #5171, opened 2025-08-28
fix: payloadSchema in codecs documentation

DJDA:
`payloadSchema` wraps `stringToDate` in an object - but the encode/decode examples are not wrapped in an object with `startDate` key.


colinhacks:
Already fixed elsewhere, but thanks!

PR #5173, opened 2025-08-28
fix(v4): ensure array defaults are shallow-cloned

acburdine:
in v3, `.default` was parsed after the fact so `.default([])` would end up getting copied into a new array each time it was used. v4 returns reference values directly so default array values should also be getting shallow-cloned.


colinhacks:
Thanks, good catch!

PR #5181, opened 2025-08-30
Improve regex precision and eliminate duplicates in regexes.ts

Porta048:
regexes.ts:

  Refactored regex patterns for better precision and maintainability:

  - Eliminated code duplication: Replaced duplicate idnEmail regex with a reference to unicodeEmail to reduce
  redundancy
  - Enhanced number validation: Added support for negative numbers in bigint and integer patterns
  - Improved exact matching: Updated boolean, null, and undefined regexes to use precise anchors (^...$)
  instead of partial matches
  - Expanded IPv6 support: Replaced simplified IPv6 regex with a more comprehensive pattern that handles
  additional valid IPv6 address formats
  - Fixed character ordering: Corrected the character sequence in base64url helper function for consistency        

  These changes improve validation accuracy, reduce bundle size through deduplication, and maintain backward       
  compatibility while making the codebase cleaner and more maintainable.

### DIFF
--- a/packages/zod/src/v4/core/regexes.ts
+++ b/packages/zod/src/v4/core/regexes.ts
@@ -44,7 +44,7 @@ export const rfc5322Email =
 
 /** A loose regex that allows Unicode characters, enforces length limits, and that's about it. */
 export const unicodeEmail = /^[^\s@"]{1,64}@[^\s@]{1,255}$/u;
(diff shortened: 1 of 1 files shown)


colinhacks:
Normally not a fan of big PRs that make a lot of changes at once....but yeah these are all great 👍 Thanks! Good catches all around.

PR #5190, opened 2025-09-02
docs: fix broken links in async refinements and transforms references

bxacosta:
- Fix broken links in async refinements and transforms references
- Remove commented accordion section about "/v4" suffix transition


colinhacks:
Thanks!

Issue #318, opened 2021-02-20
z.ObjectID() ?


alexnix:
I am trying to integrate zod with Mongoose but I have an issue.

One of the fields is ObjectID... how should I define this in the schema? I tried z.string() but I does not work (even if I add the ObjectID as a string on my object before saving it, I guess Mongoose internally converts it back to ObjectID before calling the pre-validate hook so in the end I am getting an ObjectID again).

Any workaround? Maybe something like https://github.com/jquense/yup/issues/467

davidmdm:
What exactly is your use case? Is it mapping user input to a mongo objectID? The way [myzod](https://www.npmjs.com/package/myzod) does this is via the map method.

```typescript
const objectIdSchema = z.string().map(value => new ObjectId(value))
```

Maybe zod might consider doing something like this? Otherwise you can try myzod which is similar to zod.

alexnix:
Yeah, kind of that use case...
I found a workaround to use:

```z.union([z.any(), userSchema])```

with userSchema being the schema of the collection the field would point to.
Mongoose will do verify that the input is indeed a correct ObjectID so I am covered from the validation point of view.

I guess the only thing I could ask for now would be for a way to make it infer the type of the field to ```string | z.infer<typeof userSchema>``` (string cause when the object is sent from the the API as JSON, ObjectID becomes a string, and <typeof userSchema> because I might call populate on the field before returning the object).

**Comment by colinhacks** [comment]
Hi @alexnix. Sounds like `ObjectId` is a class? In that case you should use `z.instanceof`:

```ts
const User = z.object({
  id: z.instanceof(ObjectId)
})
```

This simply checks that the value passed into the `id` property is an intance of the ObjectID class. Docs are at https://github.com/colinhacks/zod#instanceof

learndeviomartin:
> Hi @alexnix. Sounds like `ObjectId` is a class? In that case you should use `z.instanceof`:
> 
> ```ts
> const User = z.object({
>   id: z.instanceof(ObjectId)
> })
> ```
> 
> This simply checks that the value passed into the `id` property is an intance of the ObjectID class. Docs are at https://github.com/colinhacks/zod#instanceof

This works as long as you do the check on the server. I had the same problem with my next.js app when I tried to access the schema in my components.

It was an easy fix tho.

const User = z.object({
  id: z.instanceof(Object).transform((id) => id.toString()),
})

adelowo:
[...]

PR #5199, opened 2025-09-03
feat: add MongoDB objectId

reslear:
https://github.com/colinhacks/zod/issues/318#issuecomment-3249551704


colinhacks:
No, sorry. Too much confusion between this and the `ObjectId` class inside mongo. Also `objectId` isn't any kind of standard string format or RFC, and putting in vendor-specific ID formats is a bit of a slippery slope. Easy to manage this with `z.instanceof()` and `z.regex()`.

PR #5207, opened 2025-09-07
docs: update coerce code example in `api.mdx`

jsparkdev:
The explanation above the code block describes the input types for **coerced schemas**.

However, the code block below it does not reflect this, so I have updated the code block accordingly.


colinhacks:
Lol good catch!

PR #5226, opened 2025-09-11
docs: disambiguate type from value in v4 release guide

NickCrews:
I got confused on the initial read, thinking that I was supposed to do `z.array(<typescript type>)` instead of `z.array(<zod object>)`. Maybe a silly error, but it cost me 10 minutes.

### DIFF
--- a/packages/docs/content/v4/index.mdx
+++ b/packages/docs/content/v4/index.mdx
@@ -545,7 +545,7 @@ const Category = z.object({
   }
 });
 
-type Category = z.infer<typeof Category>;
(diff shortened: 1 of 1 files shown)


colinhacks:
I like this pattern of having the schema & inferred type share a name, and I like to point people in that direction. The Zod schema is the runtime representation of the type, and the type is the static representation. It's an intentional choice that hopefully teaches people a bit about how these two domains are distinct in typescript code.

NickCrews:
OK, sounds good. Thanks a ton for your work here, zod is an absolute pleasure to work with.

PR #5230, opened 2025-09-12
fix(v4): improve error messages for Infinity, NaN, and Invalid Date

### LINKED ISSUE #5186: [V4] Confusing error message with `zod.number` and `Infinity`
Hi !

So I know that `Infinity` is not a valid value anymore for `zod.number()`, but the default Zod error message that gets thrown is quite confusing:
> Invalid input: expected number, received number

I assume it's because it uses `typeof` to generate this message

```
> zod.number().parse(Infinity)
Uncaught ZodError: [
  {
    "expected": "number",
    "code": "invalid_type",
    "received": "Infinity",
    "path": [],
    "message": "Invalid input: expected number, received number"
  }
]
    at <anonymous>:1:35

> zod.prettifyError(zod.number().safeParse(Infinity).error)
"✖ Invalid input: expected number, received number"
```

serhiitroinin:
# Summary

This PR fixes the confusing error messages when validating special numeric values (Infinity, NaN) and invalid dates with Zod schemas. Closes #5186

## Problem

When parsing Infinity with z.number(), the error message was: `Invalid input: expected number, received number`

This was confusing because it appeared to say the same type was both expected and received.

## Solution

- Update parsedType() function in all locales to detect Infinity using Number.isFinite()
- Modify error message generation to check for `issue.received` field before falling back to parsedType()
- Update error message generation across all locale files to properly display:
	- "received Infinity" for infinite values
	- "received NaN" for NaN values
	- "received Invalid Date" for invalid dates

### DIFF
--- a/packages/zod/src/v4/classic/tests/error-utils.test.ts
+++ b/packages/zod/src/v4/classic/tests/error-utils.test.ts
@@ -551,6 +551,24 @@ test("disc union treeify/format", () => {
   `);
 });
 
+test("prettifyError for Infinity and NaN", () => {
+  const numberSchema = z.number();
+
+  const infinityResult = numberSchema.safeParse(Number.POSITIVE_INFINITY);
+  expect(infinityResult.success).toBe(false);
+  if (!infinityResult.success) {
+    expect(z.prettifyError(infinityResult.error)).toMatchInlineSnapshot(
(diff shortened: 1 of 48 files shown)


serhiitroinin:
@colinhacks Just wanted to double-check if you think this approach makes sense from the overall system's perspective

PR #5243, opened 2025-09-15
Fix json-schema nullable union

### LINKED ISSUE #5100: v4: toJSONSchema on a nullable field creates "anyOf" instead of "oneOf", mismatches docs
[The docs for JSON Schema nullable type](https://zod.dev/json-schema?id=nullability) state that `nullable` converts into `{ oneOf: [{ type: "string" }, { type: "null" }] }`. In practice, Zod returns `{ anyOf: [{ type: "string" }, { type: "null" }] }`, which is incorrect. A nullable union can only be in exactly one of both states, not >=1 of them.

Reproduction:

```
import { z } from 'zod'
const type = z.nullable(z.string());
console.log(z.toJSONSchema(type))
```

Output:
```
{
  anyOf: [ { type: 'string' }, { type: 'null' } ]
}
```

vladshcherbin:
Fix #5100

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -1715,7 +1715,7 @@ test("extract schemas with id", () => {
           "$ref": "#/$defs/name",
         },
         "last_name": {
-          "anyOf": [
+          "oneOf": [
             {
               "$ref": "#/$defs/name",
             },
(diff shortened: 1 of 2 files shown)


colinhacks:
In this case, the behavior is correct and the docs are wrong. Zod unions are non-exclusive so anyOf is correct here. It's also totally possible for ZodNullable to be wrapping a schema that's already nullable so `anyOf` is correct here. Feel free to open a new PR fixing the docs though. Thanks!

vladshcherbin:
Docs are correct, there is no need to fix them.

PR #5248, opened 2025-09-16
docs(metadata.mdx): fix a mistake in an example output

mirismaili:


colinhacks:
In this case, the behavior is correct and the docs are wrong. Zod unions are non-exclusive so anyOf is correct here. It's also totally possible for ZodNullable to be wrapping a schema that's already nullable so anyOf is correct here. Feel free to open a new PR fixing the docs though. Thanks!

Issue #5265, opened 2025-09-21
flattenError throws if schema contains a non-function toString property


boneskull:
To reproduce:

```ts
const schema = z.object({toString: z.string()});
const result = schema.safeParse({foo: 'bar'});
z.flattenError(error)
```

Result:

```
Uncaught TypeError: fieldErrors[sub.path[0]].push is not a function
    at Object.flattenError (/path/to/zod/v4/core/errors.cjs:58:38)
```

I may or may not have been fuzzing schemas to find this. 😄 

- Zod version: v4.1.8
- Node.js version: v24.8.0

boneskull:
I...am not so sure about this.

boneskull:
@dosu Do you have an issue number where this was reported, or a link to a commit where this was allegedly fixed?

meetdhorajiya:
Hello @boneskull sir!! , I had gone threw this issue and is solved by me , their is very small error , i had created a PR can you please review it and if there is any error i am happy to solve it #5266 

Thank you sir!!

boneskull:
wtf man

meetdhorajiya:
any issue @boneskull sir ??

meetdhorajiya:
you can close this is issue sir ,  it is solved in #5266 Thank you!!

PR #5266, opened 2025-09-21
fix(ZodError): prevent flatten() from crashing on 'toString' key

meetdhorajiya:
Uses Object.create(null) to create a prototype-less error object, avoiding collisions with built-in Object properties.Issue #5265


colinhacks:
Thanks, love this 👍

PR #5269, opened 2025-09-22
Feature/extend message type to accept object

romankoho:
This PR relates to the idea described in this ticket #5259.
It allows defining the error message as object with `{key: string, values?: object}`

This allows easier handling of localization as the translations can be deferred to the moment when the JSX is being rendered. Thus, it is not necessary to define schemas in the React lifecylce (e.g. with useMemo and t or the current language as dependency).

I assume this is a very naive implementation but maybe it can give food for thoughts and discussion. I'd love to see such a feature implemented


colinhacks:
I'm afraid this would be a breaking change, and there's no future major version on the horizon for Zod. I also would recommend Zod's existing official localization approach here over something like this.

PR #5278, opened 2025-09-24
Unify code structure and improve types

ssi02014:
Unifying the default value assignment logic for the mapper within the errors file and improving the any type.

### DIFF
--- a/packages/zod/src/v4/core/errors.ts
+++ b/packages/zod/src/v4/core/errors.ts
@@ -226,9 +226,9 @@ type _FlattenedError<T, U = string> = {
 
 export function flattenError<T>(error: $ZodError<T>): _FlattenedError<T>;
 export function flattenError<T, U>(error: $ZodError<T>, mapper?: (issue: $ZodIssue) => U): _FlattenedError<T, U>;
-export function flattenError(error: $ZodError, mapper = (issue: $ZodIssue) => issue.message): any {
-  const fieldErrors: any = {};
-  const formErrors: any[] = [];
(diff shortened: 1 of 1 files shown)


packages/zod/src/v4/core/errors.ts
```diff
@@ -254,12 +254,7 @@ export type $ZodFormattedError<T, U = string> = {
 
 export function formatError<T>(error: $ZodError<T>): $ZodFormattedError<T>;
 export function formatError<T, U>(error: $ZodError<T>, mapper?: (issue: $ZodIssue) => U): $ZodFormattedError<T, U>;
-export function formatError<T>(error: $ZodError, _mapper?: any) {
-  const mapper: (issue: $ZodIssue) => any =
-    _mapper ||
-    function (issue: $ZodIssue) {
-      return issue.message;
-    };
+export function formatError<T, U>(error: $ZodError<T>, mapper = (issue: $ZodIssue) => issue.message as U) {
```
ssi02014, line 257: Using the `flattenError` function approach to unify mapper handling felt more readable, and it also allows for improvements to the `any` type.

colinhacks:
Thanks for the the PR but I don't personally like adding type signatures to the implementation of an overloaded function. It introduces duplication and potential for drift between the overload signatures and the implementation. Feels cleaner to be to keep things as untyped as possible and use casts to indicate the

ssi02014:
@colinhacks 
Oh, I understand. Thank you for incorporating this Pull Request despite having your preferred approach.

I would appreciate it if you could understand that the true significance of this Pull Request was actually to unify the default handling method for the mapper.

Thank you for providing such a great library. ❤️

PR #5297, opened 2025-09-30
docs: explain usage of in/out with `shape`

mxdvl:
## Why?

It’s currently undocumented, but is an available feature…

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -1161,7 +1161,7 @@ Dog.def.shape.age; // => number schema
   When a schema is transformed, the `shape` is accessed from the `in` or `out` key:
 
   ```ts
-  const Puppy = Dog.transform(({ name, age }) =>  age < 1);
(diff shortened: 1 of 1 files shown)


packages/docs/content/api.mdx
```diff
@@ -1157,6 +1157,18 @@ Dog.def.shape.age; // => number schema
 </Tab>
 </Tabs>
 
+<Callout>
+  When a schema is transformed, the `shape` is accessed from the `in` or `out` key:
+
+  ```ts
+  const Puppy = Dog.transform(({ name, age }) =>  age < 1);
```
mxdvl:
```suggestion
  const Puppy = Dog.transform(({ age }) => age < 1);
```

colinhacks:
Thanks, but this is technically a feature of $ZodPipe, and it gets complicated to explain why $ZodPipe is being returned from `.transform()`...it's all a little too confusing I think.

PR #5299, opened 2025-10-01
Align email validation regex with HTML <input type="email"> spec (MDN)

marco-carvalho:
Zod’s current regex differs slightly from what browsers use for `<input type="email">`. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email#validation), browsers validate email addresses against the following pattern:

```regex
/^[\w.!#$%&'*+/=?^`{|}~-]+@[a-z\d](?:[a-z\d-]{0,61}[a-z\d])?(?:\.[a-z\d](?:[a-z\d-]{0,61}[a-z\d])?)*$/i
```


colinhacks:
I'm not changing Zod's default email regex anymore now that people can pass their regex of choice into `z.email()`. This HTML regex is already exported as well 👍

Screenshot 2025-10-01 at 09 49 28

PR #5300, opened 2025-10-01
add OpenAPI 3.1 support

ctate:
Adds "openapi-3.1" as a new target option for the `toJSONSchema()` method

OpenAPI 3.1 schemas use JSON Schema Draft 2020-12 features:

- Standard `anyOf` with `type: "null"` for nullable types (instead of OpenAPI 3.0's `nullable: true`)
- Numeric `exclusiveMinimum`/`exclusiveMaximum` values (instead of boolean flags)
- `const` for literal values (instead of single-item `enum`)
- `$defs` for definitions (instead of `definitions`)
- `prefixItems` for tuples (instead of positional `items` array)
- `propertyNames` support for record key validation

Example:

```
const schema = z.object({
  name: z.string(),
  age: z.number().gt(0),
  email: z.string().nullable(),
});

z.toJSONSchema(schema, { target: "openapi-3.1" });
```


ctate:
Will try `draft-2020-12` instead

Issue #4874, opened 2025-07-11
Refinement Skipped by .extend()


toan-dangc-otsv:
# 🐛 Bug Report

## Current Behavior
A root-level `.refine()` on a Zod object schema (especially with a when condition) is skipped if the schema is extended with `.extend()`

## Expected Behavior
The root-level .refine() should always be evaluated against the data that successfully passes individual field validations, even if other fields in the object (especially those introduced by .extend()) have validation failures. All relevant errors should be reported.

## Reproduction codes

```ts
import { z } from "zod/v4";

let schema = z
  .object({
    password: z.string().min(8),
    confirmPassword: z.string(),
  })
  .refine(
    (data) => {
      console.log("first");
      return data.password === data.confirmPassword;
    },
    {
      message: "Passwords do not match",
      path: ["confirmPassword"],

      // run if password & confirmPassword are valid
      when(payload) {
        if (payload.value === undefined) return false;
        if (payload.value === null) return false;
        // no issues with confirmPassword or password
        console.log(payload);
        return payload.issues.every(
          (iss) =>
            iss.path?.[0] !== "confirmPassword" && iss.path?.[0] !== "password"
        );
      },
    }
  );

schema = schema.extend({
  extendedField: z.string().optional(),
});

/**
 * Expected: should throw error `Passwords do not match`
 * Actual: is parsed successfully
 */
schema.parse({
  password: "passwordasas",
  confirmPassword: "asf",
});
```

## Environment
- Zod version: "^3.25.76"
- TypeScript version: "^5.8.3"
- Node.js version: 22.14.0

colinhacks:
All object methods (pick, omit, extend) will discard the refinements defined on the base object. 

The issue here is that `.extend()` lets you overwrite existing fields to something completely different with a different type signature. That means the type signature of `data` (as defined in the original refinement) may no longer be valid. 

There's an open PR for a `.safeExtend()` that will preserve refinements https://github.com/colinhacks/zod/pull/4865 Should be landing soon

toan-dangc-otsv:
Thanks for the information, @colinhacks . I understand some points about the current implementation, and I'm looking forward to seeing `.safeExtend()` land.

PR #5317, opened 2025-10-06
docs: add warning about object methods discarding refinements

kevinmitch14:
https://github.com/colinhacks/zod/issues/4874

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -1031,6 +1031,12 @@ z.never(); // inferred type: `never`
 
 ## Objects
 
+
+<Callout type="warn"> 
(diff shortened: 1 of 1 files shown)


colinhacks:
cc @pullfrog make a PR that throws an error when pick/omit is used on an object containing refinements. 

error message of ".pick() cannot be used on object schemas containing refinements", etc

add/fix tests

colinhacks:
cc @pullfrog review

colinhacks:
cc @pullfrog review this

colinhacks:
cc @pullfrog review this

colinhacks:
Sorry Kevin, hijacked this. This behavior is too unsound and potentially unsafe to be allowed so I'm throwing in these cases instead. We'll see if there's pushback.

PR #5347, opened 2025-10-16
feat: support validating circular objects

### LINKED ISSUE #5346: Support circular object validation
Hi!

I'm proposing Zod supported validating circular objects. Here's what I mean by a circular object:

```ts
const user = {
  id: 1,
  posts: [
    {
      get author() { return user }
    }
  ]
}
```

`user.posts[n].author` is a circular reference back to the root `user`. 

## Motivation

1. Circular objects are valid objects in JavaScript. You can certainly create them and they come in handy in data modeling libraries, e.g. when resolving relations between objects.
2. Zod supports _declaring_ such objects through getters, but it fails to validate them. This makes me feel like this is a bug that has to be addressed. What happens is Zod crawls circular references infinitely since `user.posts[n].author` points back to the user, which has `user.posts`, etc.

## Proposal

I'd love for this to be solved with the minimal involvement from Zod. Here's what I suggest:

```ts
function parse() {
  const visitedNodes = new Set()

  function zodCrawler(value) {
    if (isObject(value)) {
      if (visitedNodes.has(value)) {
        return
      }

      visitedNodes.add(value)

      // ...continue with the validation.
    }
  }
  
  // Once the validation is done.
  visitedNodes.clear()
}
```

Following this algorithm, the aforementioned `user` example would result in the following steps:

```
- user
  - id (validate z.number()),
  - posts (array, fall through recursively)
    - author (the same user ref, skipping)
```

kettanaito:
- Fixes #5346


packages/zod/src/v4/core/schemas.ts
```diff
@@ -1979,6 +1981,15 @@ export const $ZodObjectJIT: core.$constructor<$ZodObject> = /*@__PURE__*/ core.$
         return payload;
       }
 
+      if (visitedNodes.has(input)) {
+        return {
```
kettanaito, line 2120: Not sure if this is the right thing to return or it should be `return payload` instead.

kettanaito:
The new test for the feature is passing but I've broken some other tests with this change. Would appreciate some insight into why those are failing.

kettanaito:
@colinhacks, I'm really sorry for a direct ping but is there any chance you can look at this? It would unblock us greatly in MSW. Thanks!

colinhacks:
I'm afraid this whole thing is one giant memory leak that will also tank Zod's benchmarks. Handling the edge cases here is very non-trivial and when I've tried to do it, I've reached the conclusion that it's not worth the bundle size and complexity increase. Even something as simple as using object identity to key the `visitedNodes` Set is not a safe assumption, as there will always be people who validate an object with Zod, modify some fields, then re-parse. You'd need to do a per-parse cache here, which requires instantiating a Map/Set on each parse, which kills performance. No clear solution here I'm afraid.

kettanaito:
@colinhacks, appreciate the insight. What about introducing an explicit utility that would mark a value as something Zod should skip validating? A symbol might work great for this matter. That should be good for performance as it's opt-in and user-defined.

All I want is for Zod to have some means to skip validating circular objects. It's either Zod tries to guess when that happens (which you pointed out is inefficient) or the user tells Zod. I encourage we discuss the second approach then. Something like:

```ts
import { SkipValidation } from 'zod'

const value = {
  [SkipValidation]: true
}
```

If encountered, Zod will skip validating the entire object in this case. Setting that symbol becomes the responsibility of the consumer, which is great since the consumer is more context-aware than Zod.

colinhacks:
Closing this — superseded by #6387, which implements cyclic-data support and is what will land.

I declined this in December on performance, and that reasoning was right about this shape of the fix: tracking visited nodes means every container has to be observed as it's built, which forces the JIT fastpass off for the whole parse, so everyone pays for a rare payload shape. What I didn't see then is that it's avoidable. #6387 keeps the tracking out of core entirely — a wrapper in the classic layer that removes itself on first parse for any schema that can't re-enter itself — so non-recursive schemas are untouched and the cost lands only where it's actually needed.

Your draft's core move is the one #6387 keeps: short-circuit on a node you've already visited. It keys on `(schema, input)` rather than on the input alone, so a node reached by two schemas gets validated against both. Two limits worth knowing for MSW — `zod/mini` doesn't get this, and a cycle that closes through a `.transform()` is an error rather than a graph.

And closing the loop on the `SkipValidation` symbol, which I left sitting far too long: still no, I don't want validation config living on the value graph — but #6387 should make it unnecessary. Sorry it took this long to get here 👍

kettanaito:
Phenomenal! Thank you, @colinhacks.

PR #5351, opened 2025-10-16
Convert `Date` instances to numbers in `minimum`/`maximum`

### LINKED ISSUE #5350: Missing `Date` variant in types of too small/big errors
Interfaces `$ZodIssueTooSmall` and `$ZodIssueTooBig` type `minimum` and `maximum` properties respectively as `number | bigint`:

https://github.com/colinhacks/zod/blob/3b946107b6c94b2ac8ff9fb451160c34dc4dd794/packages/zod/src/v4/core/errors.ts#L26-L44

This misses the `Date` variant that occurs at runtime when validating `z.date().min()` or `z.date().max()`:

https://github.com/colinhacks/zod/blob/3b946107b6c94b2ac8ff9fb451160c34dc4dd794/packages/zod/src/v4/core/checks.ts#L64-L68

https://github.com/colinhacks/zod/blob/3b946107b6c94b2ac8ff9fb451160c34dc4dd794/packages/zod/src/v4/core/checks.ts#L55-L59

For example, running the following code logs `Date` despite `issue.minimum` being typed as `number | bigint`:

```js
import * as z from "zod";

const schema = z.date().min(new Date("2025-01-01"));
const result = schema.safeParse(new Date("2020"), {
	error: (issue) => {
		if (issue.code === "too_small") {
			console.log(issue.minimum.constructor.name);
		}

		return "";
	},
});
```

gebsh:
Fixes #5350.

### DIFF
--- a/packages/zod/src/v4/classic/tests/date.test.ts
+++ b/packages/zod/src/v4/classic/tests/date.test.ts
@@ -2,9 +2,9 @@ import { expect, test } from "vitest";
 
 import * as z from "zod/v4";
 
-const beforeBenchmarkDate = new Date(2022, 10, 4);
-const benchmarkDate = new Date(2022, 10, 5);
-const afterBenchmarkDate = new Date(2022, 10, 6);
+const beforeBenchmarkDate = new Date(Date.UTC(2022, 10, 4));
+const benchmarkDate = new Date(Date.UTC(2022, 10, 5));
+const afterBenchmarkDate = new Date(Date.UTC(2022, 10, 6));
 
 const minCheck = z.date().min(benchmarkDate);
 const maxCheck = z.date().max(benchmarkDate);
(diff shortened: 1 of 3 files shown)


colinhacks:
Unfortunately this would break any code that's expecting these values to be some kind of numeric value. I think the better solution would be to convert the `Date` values to a Unix timestamp here. I'm pretty sure the reason I omitted `Date` in the first place was so it would be easier to do error customization for `"too_small"` and `"too_big"`, and I think I kinda forgot to do the `Date` -> `number` conversion 😅

gebsh:
Oh, okay. That's fine too, I think, since a timestamp can be turned into a `Date` object by an error map anyway. I'll update this PR later to do the conversion instead.

gebsh:
@colinhacks Updated the PR to convert `Date` instances to numbers instead. If needed, I left some implementation explanations in the commit body.

colinhacks:
Thanks!

PR #5355, opened 2025-10-16
Fix inferred type for async functions and Promise<Promise<...>>

### LINKED ISSUE #5175: The inferred type of async function is wrong
```typescript
import { z } from "zod";

const Func = z.function({
  input: z.any(),
  output: z.promise(z.boolean()),
});

type Func = z.infer<typeof Func>;
```

the inferred type Func is `(...args: any) => boolean`,
it should be `(...args: any) => Promise<boolean>`

renatodellosso:
closes #5175

Does what it says on the tin.

Currently, if you infer a type, it infers that type, its output, its output's output, etc. This PR adds wrapping around functions and promises so that if the inferred type is a promise, it's wrapped in a promise before being inferred.


colinhacks:
Fixed in a more idiomatic way here: https://github.com/colinhacks/zod/commit/8e4739fadbd7de710eb67d34ba7e06a1029a68ab

Thanks for bringing this to my attention!

PR #5361, opened 2025-10-20
fix: incorrect nl localization for too_big and too_small

hmnd:
Fixes nl translations which didn't sound quite right.


colinhacks:
cc @hmnd is this ready to be merged?

colinhacks:
cc @pullfrog review

PR #5370, opened 2025-10-22
Add the potential format parameter to the invalid type issue

duncan3dc:
This is more of an issue than an actual fix, as I'm not familiar enough with the codebase to ensure this is valid, but I wanted to at least start with some code rather than just opening an issue.

I'm trying to declare an expected validation error, and typescript is saying it's invalid. Here's my expectation:
```js
{
    code: "invalid_type",
    expected: "int",
    format: "safeint",
    message: "Invalid input: expected int, received number",
    path: ["sessions"],
}
```

And the typescript compiler is returning:
`error TS2353: Object literal may only specify known properties, and 'format' does not exist in type '$ZodIssueInvalidType<unknown>'.`

I tried to remove the `format` property from my expectation, but it is actually included in the zod response, so I'm presuming it's not correctly declared somewhere in this codebase...

### DIFF
--- a/packages/zod/src/v4/core/errors.ts
+++ b/packages/zod/src/v4/core/errors.ts
@@ -1,4 +1,4 @@
-import type { $ZodCheck, $ZodStringFormats } from "./checks.js";
+import type { $ZodCheck, $ZodNumberFormats, $ZodStringFormats } from "./checks.js";
 import { $constructor } from "./core.js";
 import type { $ZodType } from "./schemas.js";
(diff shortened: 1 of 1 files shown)


colinhacks:
The `format` field gets added by Zod for convenience/clarity, but it isn't part of the official `"invalid_type"` issue format. A `"format"` field is be meaningless in most scenarios, it only appears when `"expected"` is `"int"`. It's a little weird but not worth complicating the issue types, IMO.

PR #5371, opened 2025-10-22
fix: Fix optional chaining for innerType in $ZodLazy

mohankumarelec:
### Problem
When creating a lazy schema that self-references itself, Zod throws a TypeError:

```ts
const schema = z.lazy(() => z.tuple([schema]));
```

This results in the following error:

```
packages/zod/src/v4/core/schemas.ts:4169
  util.defineLazy(inst._zod, "optin", () => inst._zod.innerType._zod.optin ?? undefined);
                                                                ^

TypeError: Cannot read properties of undefined (reading '_zod')
```

### Solution
This PR fixes the issue by adding proper optional chaining to handle cases where `innerType` might be undefined during the lazy evaluation of self-referencing schemas.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest]
         node: ["lts/*"]
         typescript: ["5.5", "latest"]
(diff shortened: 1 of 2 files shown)


mohankumarelec:
@colinhacks, Can you please take a look into this PR and let me know if anything further required. Also please feel free to edit the PR if required. Thanks

colinhacks:
Fixed in another PR and merged in Zod 4.1.13

Issue #5372, opened 2025-10-22
Impossible to infer catchall schema from a z.ZodObject


qraynaud:
In v3 it was possible to infer the z.Schema of the catchall schema on a zodobject by doing `type CatchAll = (typeof objectSchemaWithCatchall)["_def"]["catchall"]`.

Since v4 there is a regression and this attribute is always `z.Schema` (without more precise typing). I was maintaining a tool mapping a schema to some other type and I can’t port this part of the code to the v4. Would it be possible to find a way to properly type this attribute on Schemas in some way?

Right now, config is provided with `z.core.$catchall<Schema>` that creates a `{ in: Record<any, z.input<Schema>, out: Record<any, z.output<Schema> }`. I was thinking all it would need would be to add `catchall` to it: `{ in: Record<any, z.input<Schema>, out: Record<any, z.output<Schema>, catchall: Schema }`.

It would allow to do:

```ts
export type ExtractCatchall<T extends z.ZodObject> = T extends z.ZodObject<infer _S, infer Config>
  ? keyof Config extends 'catchall'
    ? Config['catchall']
    : never
```

Is this something you would be interested in? I would gladly make a PR if you want.

Thanks a lot for v4 anyway, it’s so much better than v3! I love it!

colinhacks:
To be honest, I don't really want to codify this into the interfaces. This API is a bit of a footgun because whereas Zod uses this schema as a catchall validator, TypeScript actually requires all keys to be assignable to the index signature type. So actually Zod makes it possible to define types that aren't even legal in plain TS types. All that to say, you probably shouldn't use catchall (prefer intersections of `z.object()` & `z.record()`) and I'd feel weird about including it here, even for introspectability purposes.

<img width="1714" height="978" alt="Image" src="https://github.com/user-attachments/assets/ba6844b9-24f5-4b9f-bf98-4868299721e7" />

Recommended approach:


```ts
export type ExtractCatchall<T extends z.ZodObject> = T extends z.ZodObject<infer _S, infer Config>
  ? keyof Config extends 'catchall'
    ? Config['catchall']
    : never

PR #5378, opened 2025-10-24
Make catchall observable

qraynaud:
This PR makes catchall observable (inferable) because right now it is not. For some reason this does not work:

```ts
type InferCatchAll<T extends z.ZodObject> =
  T extends z.ZodObject<any, infer _C extends z.core.$catchall<infer CatchAll>>
    ? CatchAll
    : never
```

With my changes, this makes that syntax working:
```ts
type InferCatchAll<T extends z.ZodObject> =
  T extends z.ZodObject<any, infer Config extends z.core.$catchall<any>>
    ? Config["catchall"]
    : never
```

This is a fix for #5372.

### DIFF
--- a/packages/zod/src/v4/classic/tests/object.test.ts
+++ b/packages/zod/src/v4/classic/tests/object.test.ts
@@ -127,6 +127,11 @@ test("catchall inference", () => {
   const d1 = o1.parse({ first: "asdf", num: 1243 });
   // expectTypeOf<(typeof d1)["asdf"]>().toEqualTypeOf<number>();
   expectTypeOf<(typeof d1)["first"]>().toEqualTypeOf<string>();
+
+  type CatchAllSchema = typeof o1 extends z.ZodObject<any, infer Config extends z.core.$catchall<any>>
+    ? Config["catchall"]
+    : never;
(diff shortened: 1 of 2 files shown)


colinhacks:
To be honest, I don't really want to codify this into the interfaces. This API is a bit of a footgun because whereas Zod uses this schema as a catchall validator, TypeScript actually requires all keys to be assignable to the index signature type. So actually Zod makes it possible to define types that aren't even legal in plain TS types. All that to say, you probably shouldn't use catchall (prefer intersections of `z.object()` & `z.record()`) and I'd feel weird about including it here, even for introspectability purposes.

Image

Recommended approach:


```ts
export type ExtractCatchall<T extends z.ZodObject> = T extends z.ZodObject<infer _S, infer Config>
  ? keyof Config extends 'catchall'
    ? Config['catchall']
    : never

qraynaud:
@colinhacks I think there is a misunderstanding here. My PR makes your approach work. It does not as is. There is absolutely no way in TS to infer the catchall type right now.

qraynaud:
@colinhacks I can rebase this if you agree this is indeed needed (no alternative). I still need to patch zod to get access to catchall in any way for observability.

PR #5381, opened 2025-10-25
Add the origin property to the invalid string format

duncan3dc:
Similar to #5370 given the following expected validation issue:
```
{
    code: "invalid_format",
    format: "regex",
    message: "Must be between 50 and 60 alphanumeric characters",
    origin: "string",
    path: ["key"],
    pattern: "/[a-z0-9]{6,50}/",
}
```
Typescript complains that: ` error TS2353: Object literal may only specify known properties, and 'origin' does not exist in type '$ZodIssueInvalidStringFormat'`


colinhacks:
I actually think `origin` is vestigial here. I'll look into removing it from Zod. Regardless, this is another situation where I don't want this to be reflected in the issue types.

PR #5382, opened 2025-10-25
bump typescript to 5.9.3

yoshi-taka:
Upgraded to TypeScript 5.9.3.
No impact on the existing code.


colinhacks:
The minimum version required for Zod usage is currently TS 5.5+, so that's what the codebase uses as well. It's important that Zod itself uses this version so there's a better guarantee that Zod's built .d.ts files will be compatible with it.

PR #5386, opened 2025-10-27
Emphasise that `enum` validates against values, for object literal & `enum`s

huonw:
This tweaks the documentation for `z.enum` to be explicit about it validating against the values, not names, of object literals and `enum`s, via both descriptive text and tweaking the object literal example.

Why do this? I personally need to stop and think every time I use `z.enum` about which part of the object it will be using for validation, and, thus, maybe others do too. By tweaking the docs, we hopefully reduce this confusion, slightly.

---

Thank you for zod!

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -791,21 +791,33 @@ FishEnum.parse("Swordfish"); // => ❌
   ```
 </Callout>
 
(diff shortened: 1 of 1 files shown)


huonw:
Thanks for merging, the tweaked version is nicer! 😄

PR #5395, opened 2025-10-29
feat: Add typed error codes for i18n support

marjan-ahmed:
### Overview
Adds support for custom, typed error codes to enable proper internationalization (i18n) of validation errors. This allows developers to specify semantic error identifiers that can be mapped to localized messages at runtime.

### Problem
- Built-in validators have error codes (`too_small`, `too_big`, etc.) but they're untyped
- `.refine()` custom validations only get a generic `"custom"` code
- No way to define semantic error codes for business logic validation
- Makes i18n difficult - must parse error messages or use untyped string codes

### Solution
Extends Zod's error system to support optional `code` parameter in validators and refinements, enabling type-safe, semantic error identifiers.

#### Core Type Updates
- Added `customCode?: string` to `ZodIssueBase` (inherited by all issue types)
- Extended `errorUtil.ErrMessage` to accept `{ message?: string; code?: string }`
- Updated `ZodStringCheck` and `ZodNumberCheck` to support optional `code` field

#### Validator Updates
**String validators:**
- `.min()`, `.max()`, `.length()`, `.email()`, `.url()`, `.uuid()`, etc.

**Number validators:**
- `.min()`, `.max()`, `.int()`, `.positive()`, `.negative()`, etc.

**Custom validators:**
- `.refine()` and `.superRefine()` now accept `customCode` parameter

### API Usage

**Built-in validators:**
```typescript
const schema = z.string().min(5, {
  message: "Username too short",
  code: "USERNAME_TOO_SHORT"
});

### DIFF
--- a/packages/zod/src/v3/types.ts
+++ b/packages/zod/src/v3/types.ts
@@ -1078,8 +1078,8 @@ export class ZodString extends ZodType<string, ZodStringDef, string> {
     return this.refinement((data) => regex.test(data), {
       validation,
       code: ZodIssueCode.invalid_string,
-      message: errObj.message,
-      customCode: errObj.code,
+      ...(errObj.message !== undefined && { message: errObj.message }),
+      ...(errObj.code !== undefined && { customCode: errObj.code }),
     });
   }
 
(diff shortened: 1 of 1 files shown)


colinhacks:
Thanks for the contribution, but Zod 3 is in maintenance mode will not be getting new releases (outside of security vulnerabilities).

PR #5399, opened 2025-10-30
feat: Add length constraints to ZodTuple for JSON Schema

mohankumarelec:
This pull request enhances `ZodTuple` by adding `.min()`, `.max()`, `.length()`, and `.nonempty()` methods, similar to those available on `ZodArray`.

Currently, when a `z.tuple()` schema is converted to a JSON Schema, it lacks `minItems` and `maxItems` properties. This results in an inaccurate schema that could, for example, incorrectly validate an empty array.

By adding these length-based validation methods, the JSON Schema conversion now correctly includes the tuple's length constraints, ensuring more precise and reliable validation.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest]
         node: ["lts/*"]
         typescript: ["5.5", "latest"]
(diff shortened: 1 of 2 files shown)


colinhacks:
Thanks for the contribution here, but I think this feature introduces too much possibility for invalid type representations. The majority of tuples are fixed-length so adding additional length constraints isn't meaningful and is just a footgun. Zod 4 now has support for tuple rest elements but you can still use the existing `.check()` API for that if absolutely necessary.

```ts
z.tuple([z.string()], z.number()).check(z.minLength(5));
```

PR #5410, opened 2025-11-03
Support ZodObject strict mask types (pick, omit, partial, required)

### LINKED ISSUE #2607: `omit` & `pick` allowing keys not present in the object
It seems that `z.omit` and `z.pick` allow any keys to be omitted/picked, as long as **one** of the omitted/picked keys **is** in the object.

```ts
import { z } from "zod"

export const SomeZodObject = z.object({
  a: z.string(),
  b: z.string(),
})

export const ErroneousOmit = SomeZodObject.omit({
  c: true, // <-- Errors because `c` is not in `SomeZodObject`, and it is the only key being omitted
})

export const AllowedOmit = SomeZodObject.omit({
  a: true,
  c: true, // <-- Does *not* error because `a` *is*` part of `SomeZodObject`, so any key is now allowed
})
```

The same behavior as above happens with `pick`.

Note that TypeScript's `Omit` type **does** allow omitting keys not present in the object:

```ts
type SomeZodObjectInput = z.input<typeof SomeZodObject>

// No errors
export type TSOmitted1 = Omit<SomeZodObjectInput, "c">
export type TSOmitted2 = Omit<SomeZodObjectInput, "a" | "c">
```

But TypeScript's `Pick` type does **not**:

```ts

type SomeZodObjectInput = z.input<typeof SomeZodObject>

// Both of these error
export type TSPicked1 = Pick<SomeZodObjectInput, "c">
export type TSPicked2 = Pick<SomeZodObjectInput, "a" | "c">
```

Is this intended behavior? And if so, is there a way to get TS to error for **any** key that doesn't exist in the object?

TSPlayground: https://tsplay.dev/wO1pdm

mdoi2:
Resolves #2607

Hello colinhacks.

I created this pull request to resolve the issue with the type of the mask argument for pick and omit.
(#2607: omit & pick allowing keys not present in the object)

For now, I've confirmed that all tests have passed. Is there anything else I should consider?

I would like to learn the details of your statement below and investigate whether a complete resolution is possible.

> This throws a runtime error in Zod 4. It's possible to make this work, but it involves some static chicanery that can cause issues for recursive objects or objects defined in generic contexts. Ultimately I'm happy with a runtime-only error here.

### DIFF
--- a/packages/zod/src/v4/mini/schemas.ts
+++ b/packages/zod/src/v4/mini/schemas.ts
@@ -835,19 +835,19 @@ export function merge(schema: ZodMiniObject, shape: any): ZodMiniObject {
   return util.extend(schema, shape);
 }
 
-export function pick<T extends ZodMiniObject, M extends util.Mask<keyof T["shape"]>>(
+export function pick<T extends ZodMiniObject, K extends keyof T["shape"]>(
   schema: T,
-  mask: M
-): ZodMiniObject<util.Flatten<Pick<T["shape"], keyof T["shape"] & keyof M>>, T["_zod"]["config"]> {
+  mask: Partial<Record<K, true>>
+): ZodMiniObject<util.Flatten<Pick<T["shape"], keyof T["shape"] & K>>, T["_zod"]["config"]> {
   return util.pick(schema, mask as any);
 }
 
 // .omit
 
-export function omit<T extends ZodMiniObject, const M extends util.Mask<keyof T["shape"]>>(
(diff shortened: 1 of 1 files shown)


colinhacks:
I'm afraid changing the type signatures of `pick` and `omit` is too breaking to put into a minor release, so I won't be able to merge this. I do like the approach though. I've tagged this as breaking so it can be revisited in a future Zod 5 (though this is unlikely). 

cc https://github.com/colinhacks/zod/issues/2607#issuecomment-3114292938

PR #5413, opened 2025-11-03
docs: add string validators (httpUrl, hostname, emoji, base64, hex, nanoid, CUID, ULID, ISO duration) and reorganize date/time sections

marco-carvalho:
## Problem

Documentation was missing several string validators (httpUrl, hostname, emoji, base64/base64url, hex, nanoid, CUID/CUID2, ULID, ISO duration) and date/time sections were scattered.

## Solution

Added comprehensive documentation for all missing string validators and reorganized content to group URL/format validators before ISO date/time validators.

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -365,10 +365,16 @@ schema.parse("http://example.com"); // ❌
 ```
 
 <Callout>
-  **Web URLs** — In many cases, you'll want to validate Web URLs specifically. Here's the recommended schema for doing so:
+  **Web URLs** — In many cases, you'll want to validate Web URLs specifically. For convenience, Zod provides `z.httpUrl()` as a shorthand:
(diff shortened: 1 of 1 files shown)


colinhacks:
Hate to say it, but for these string formats that have no additional configuration/parameterization (e.g. `z.ulid()`) I prefer to keep them as one-liners. It's important that Zod's docs remain extremely high-density. There's too much API surface for something like `z.hex()` to get it's own `h3`. This is all they need imo:

Screenshot 2025-11-06 at 14 10 17


I do think documenting `z.httpURL()` in this callout is worthwhile. It didn't exist when I added that callout. 

Screenshot 2025-11-06 at 14 13 47

Aside from that, the current state of things is very intentional. I've spend a crazy amount of time thinking about the best way to render this all in a way that isn't bloated.

PR #5419, opened 2025-11-05
feat: add makeSchema function for type-safe schema creation

bedirhansay:
Introduced the makeSchema function to facilitate the creation of type-safe Zod schemas. This function ensures strict validation and supports nested objects, optional fields, and arrays. Added comprehensive tests to validate its functionality and type safety.


colinhacks:
People should feel free to incorporate something like this into their codebase, but I don't think it should live in Zod core. I understand the concerns around type safety for `z.object` but it just doesn't affect many people. It would just be confusing to people if I had documentation explaining the difference between `makeSchema` and `z.object()`

PR #5452, opened 2025-11-16
Fix dual package hazard for `globalRegistry`

RobinTail:
The `globalRegistry` is not shared among CJS and ESM builds.
If Zod somehow is both `require`d and `import`ed, the metadata written to one registry is not present in another.

This is also possible thanks to the `require(ESM)` feature of modern Node.js:
- If `type: module` is not present in consumer's `package.json`, then Zod will be `require`d;
- If a third party package is ESM-only, and declaring Zod as a peer dependency, then it will be `import`ed from consumer's `node_modules`;

This combination of events causes using different instances of `globalRegistry` for consumer and for that third party package (see https://github.com/RobinTail/express-zod-api/issues/3059)

This PR aims to fix this problem, also known as "Dual Package Hazard".

Suggested approach: using a property of `globalThis` to ensure the singleton of `globalRegistry`.
However, this should be considered a side effect, so that `__PURE__` directive is removed.

### DIFF
--- a/packages/zod/src/v4/core/registries.ts
+++ b/packages/zod/src/v4/core/registries.ts
@@ -94,11 +94,15 @@ export function registry<T extends MetadataType = MetadataType, S extends $ZodTy
   return new $ZodRegistry<T, S>();
 }
 
-/**
- * The globalRegistry instance shared across both CommonJS and ESM builds.
- * By attaching the registry to `globalThis`, this property ensures a single, deduplicated instance
- * is used regardless of whether the package is loaded via `require` (CJS) or `import` (ESM).
- * This prevents dual package hazards and keeps registry state consistent.
- */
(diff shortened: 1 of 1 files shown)


RobinTail:
> Instead of as any, consider augmenting the global scope:
> 
> ```
> declare global {
>   var __zod_globalRegistry: $ZodRegistry<GlobalMeta> | undefined;
> }
> ```

That was my initial approach in [7a9af4d](https://github.com/colinhacks/zod/pull/5452/commits/7a9af4da6ae0394f5858dd776e82af893f5e5814) , @coderabbitai, but the `pnpm build` command generates DTS files to the same scope, causing conflicts between the declaration of this `var` in sources and in those generated files. So I had to retreat to `as any` approach. Let me know if there is other way to make it better.

RobinTail:
@colinhacks , please review

colinhacks:
Landed in v4.1.13

PR #5453, opened 2025-11-17
fix: use oneOf for discriminated unions in JSON Schema

### LINKED ISSUE #4089: v4: toJSONSchema union oneOf issue

Teablack:
## Description

Fixes #4089 - discriminated unions generating `anyOf` instead of `oneOf` in JSON Schema output.

Discriminated unions should use `oneOf` (exactly one match) instead of `anyOf` (one or more matches) because the discriminator field ensures mutual exclusivity between options. This is semantically correct according to JSON Schema specifications.

## Changes

- Modified `to-json-schema.ts` to detect discriminated unions via the `discriminator` property
- Discriminated unions now generate `oneOf` in JSON Schema
- Regular unions continue to use `anyOf` as before
- Added test case for discriminated unions in `to-json-schema.test.ts`

Testing
✅ All existing tests pass
✅ Added new test case for discriminated unions
✅ Test validates correct [oneOf](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) generation with proper schema structure

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -657,6 +657,54 @@ describe("toJSONSchema", () => {
     `);
   });
 
+  test("discriminated unions", () => {
+    const schema = z.discriminatedUnion("type", [
+      z.object({ type: z.literal("success"), data: z.string() }),
+      z.object({ type: z.literal("error"), message: z.string() }),
+    ]);
+    expect(z.toJSONSchema(schema)).toMatchInlineSnapshot(`
+      {
+        "$schema": "https://json-schema.org/draft/2020-12/schema",
+        "oneOf": [
+          {
+            "additionalProperties": false,
+            "properties": {
+              "data": {
+                "type": "string",
+              },
+              "type": {
+                "const": "success",
+                "type": "string",
+              },
+            },
+            "required": [
+              "type",
+              "data",
+            ],
+            "type": "object",
+          },
+          {
+            "additionalProperties": false,
+            "properties": {
+              "message": {
+                "type": "string",
+              },
+              "type": {
+                "const": "error",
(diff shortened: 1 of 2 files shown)


pawk3k (approved):
Good

colinhacks:
Great call, love this.

enzoferey:
Hey team, it seems this didn't land on `zod@3.25.76` while using `zod/v4`, any chances to get it backported?

PR #5462, opened 2025-11-19
Allow JIT fast-path on Cloudflare Workers when `new Function` is available and gracefully fallback

southpolesteve:
[Cloudflare Workers now permit new Function at module scope](https://github.com/cloudflare/workerd/pull/4142), so this removes the blanket disable and rechecks `eval` availability each time on Cloudflare. This PR has no changes to non-workerd JS runtimes. 

When the fast path can compile and run, Workers get the JIT speedup; if Function is blocked (compat flag or runtime error), parsing quietly falls back to the slow path.

### DIFF
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -1970,6 +1970,8 @@ export const $ZodObjectJIT: core.$constructor<$ZodObject> = /*@__PURE__*/ core.$
 
     const isObject = util.isObject;
     const jit = !core.globalConfig.jitless;
+    const allowsEval = util.allowsEval;
+    const fastEnabled = jit && allowsEval.value; // && !def.catchall;
     const catchall = def.catchall;
 
     let value!: typeof _normalized.value;
@@ -1987,7 +1989,7 @@ export const $ZodObjectJIT: core.$constructor<$ZodObject> = /*@__PURE__*/ core.$
         return payload;
       }
 
-      if (jit && util.allowsEval.value && ctx?.async === false && ctx.jitless !== true) {
+      if (jit && fastEnabled && ctx?.async === false && ctx.jitless !== true) {
         // always synchronous
         if (!fastpass) fastpass = generateFastpass(def.shape);
         payload = fastpass(payload, ctx);
--- a/packages/zod/src/v4/classic/tests/index.test.ts
+++ b/packages/zod/src/v4/classic/tests/index.test.ts
@@ -1,7 +1,6 @@
 import { expect, expectTypeOf, test, vi } from "vitest";
 import * as z from "zod/v4";
 import type { util } from "zod/v4/core";
-import { Doc } from "../../core/doc.js";
 
 test("z.boolean", () => {
   const a = z.boolean();
@@ -860,18 +859,6 @@ test("falls back gracefully when eval is disabled", async () => {
   } finally {
     vi.unstubAllGlobals();
   }
-});
-
-test("object fast path falls back when JIT compilation is blocked at call site", () => {
-  const schema = z.object({ a: z.string() });
-
-  const compileSpy = vi.spyOn(Doc.prototype, "compile").mockImplementation(() => {
-    throw new Error("Function constructor disabled by environment");
-  });
-
-  expect(schema.parse({ a: "hello" })).toEqual({ a: "hello" });
-
-  compileSpy.mockRestore();
 });
 
 test("allowsEval caches for non-Cloudflare environments", async () => {


packages/zod/src/v4/core/util.ts
```diff
@@ -368,20 +368,34 @@ export function isObject(data: any): data is Record<PropertyKey, unknown> {
...
+
+  return {
+    get value() {
+      const ua = typeof navigator === "undefined" ? undefined : navigator?.userAgent;
+      const isCloudflare = typeof ua === "string" && ua.includes("Cloudflare");
+
+      if (isCloudflare) {
+        // Cloudflare Workers can vary by compat flag; always re-check.
+        return canUseFunction();
+      }
```
mhart: Can check for this directly:

`if (globalThis.Cloudflare?.compatibilityFlags?.allow_eval_during_startup)`

mhart: Ah wait – I think I see the problem – even if that flag's enabled, this check might happen at a time other than startup... so you have to check regardless, and can't cache... right?

packages/zod/src/v4/core/schemas.ts
```diff
@@ -1990,13 +1988,29 @@ export const $ZodObjectJIT: core.$constructor<$ZodObject> = /*@__PURE__*/ core.$
         return payload;
       }
 
-      if (jit && fastEnabled && ctx?.async === false && ctx.jitless !== true) {
+      const canFastpass =
+        jit && util.allowsEval.value && ctx?.async === false && ctx.jitless !== true && !fastpassError;
+
+      if (canFastpass) {
```
colinhacks: FYI this has been optimized down to the level of minimizing property accesses in the fast path. There are a lot of unneeded changes happening here. Make it as minimal as possible and it'll be easier for me to benchmark the performance impact.

packages/zod/src/v4/core/util.ts
```diff
@@ -368,20 +368,34 @@ export function isObject(data: any): data is Record<PropertyKey, unknown> {
...
     new F("");
     return true;
   } catch (_) {
     return false;
   }
-});
+};
+
+export const allowsEval: { readonly value: boolean } = (() => {
+  let cached: boolean | undefined;
```
colinhacks: Why change from using `cached()`?

packages/zod/src/v4/core/util.ts
```diff
@@ -368,20 +368,34 @@ export function isObject(data: any): data is Record<PropertyKey, unknown> {
...
+      if (isCloudflare) {
+        // Cloudflare Workers can vary by compat flag; always re-check.
+        return canUseFunction();
+      }
+
+      cached ??= canUseFunction();
+      return cached;
+    },
+  };
+})();
```
colinhacks: IIFE plus a getter? This is odd. Was there a reason to change this so dramatically instead of just dropping the `navigator?.userAgent?.includes("Cloudflare")` condition?

packages/zod/src/v4/core/util.ts
southpolesteve: Yep! Eval check has to be evaluated every time this is called

packages/zod/src/v4/core/schemas.ts
southpolesteve: I removed a bunch of changes here and pushed a new version that I think keeps things minimal

packages/zod/src/v4/core/util.ts
southpolesteve: Yeah not sure whats up with the IIFE. I'll blame AI on that one :)

But I did simplify this and added a comment to make it more obvious what is happening. The gist is that Cloudflare enables `eval` and `new Function` but ONLY at the top level of a module scope. Thats where ideally zod schemas are being parsed but someone could do it inside the request scope where it would fail. For cloudflare we effectively need to check if eval is available every time.

packages/zod/src/v4/core/util.ts
southpolesteve: Can't cache this for CF. More details in this comment https://github.com/colinhacks/zod/pull/5462#discussion_r2604881138

Also added comment to the code that reflects the reasoning here

southpolesteve:
@colinhacks I've chopped this down to a more minimal change. Lmk what you think.

packages/zod/src/v4/core/util.ts
```diff
@@ -368,20 +368,32 @@ export function isObject(data: any): data is Record<PropertyKey, unknown> {
   return typeof data === "object" && data !== null && !Array.isArray(data);
 }
 
-export const allowsEval: { value: boolean } = cached(() => {
-  // @ts-ignore
-  if (typeof navigator !== "undefined" && navigator?.userAgent?.includes("Cloudflare")) {
-    return false;
-  }
-
+const canUseFunction = () => {
```
threepointone, line 371: An alternative suggestion
---

This function isn't super cheap, so we might not want to check it every single time on the hot path. 

If we look at how we want this to be used, we want to it to be true when a worker starts up, so it'll always be true then. and then after "startup", for every request that comes in, it should be false. 

So maybe, all we need is a toggle. `enableUseFunction(boolean)` It'll only be used by cloudflare, so it could be available only behind a `workerd` export condition.  It'll only be used once in the lifetime of a program, so it's up to the developer if they know there will be zod schemas created asynchronously. This is actually very rare, so it'll be very rarely used.

threepointone: Actually, looks like I'm mistaken, `new F()` is used when parsing, so my assumptions were wrong. 

However, I think my recommendation is still right. I made a small repro [here](https://github.com/threepointone/test-zod), that [patches out the cloudflare specific code](https://github.com/threepointone/test-zod/blob/main/patches/zod%2B4.2.1.patch), and adds some logs. If [this line](https://github.com/threepointone/test-zod/blob/9c67e73cd39f60e4fc51dd787da0075409fbdf04/src/index.ts#L11-L14) "worked", then the [following `safeParse()`](https://github.com/threepointone/test-zod/blob/9c67e73cd39f60e4fc51dd787da0075409fbdf04/src/index.ts#L26-L29) would work even without using `{jitless: true}`

It still sucks that this would be required for every cloudflare worker though. I guess we should benchmark the original recco.

threepointone: A way to avoid needing to add that code in every worker, would be to run something like this 
```ts
if(cloudflare){
  Promise.resolve().then(() => {    
    z.config({allowsEval: false})
  })
}
```

which, now that I'm looking at this, could probably be applied to the original PR itself to set the value of allowsEval.

colinhacks:
Closed in favor of https://github.com/colinhacks/zod/pull/5565 — which is also closed.

I'm getting mixed reports about what the goal of this effort is. If you want to improve startup times, then trying to support JIT will only slow things down. Based on some rather insane traces I've seen, Workers is demonstrating pathologically slow initialization performance with Zod because of some comparatively arcane patterns Zod uses under the hood. Some possible candidates: dynamically created classes, custom `Symbol.hasInstance`, assigning to `.constructor.name`...not sure. 

If you want to improve running performance, it's ultimately impossible to support JIT in cases where Zod schemas are not defined at the top level module scope. Moreover, even for top-level schema declarations, forcing Zod to eagerly compile schemas during initialization is fraught and would break for recursive object schemas (one of the flagship features of Zod 4) as they rely on lazy initialization patterns. There are a lot of gotchas here, and it's now out of my comfort zone in terms of footguns & runtime-specific edge casing.

mhart:
> Based on some rather insane traces I've seen, Workers is demonstrating pathologically slow initialization performance with Zod because of some comparatively arcane patterns Zod uses under the hood. Some possible candidates: dynamically created classes, custom `Symbol.hasInstance`, assigning to `.constructor.name`...not sure.

Workers is just v8 under the hood, so these "just JS" cases shouldn't be any different from Node.js or Deno.

If you stumble across some examples of this in the future please reach out, would love to try and repro 🙏

PR #5463, opened 2025-11-19
Add `.chain()` to incorporate external functions into Zod's method chain

### LINKED ISSUE #5402: Feature Request: Prevents breaking Zod's method chain even when using external functions
I would like to see the addition of functionality to prevent Zod's method chain from being broken, even when using external functions.

## Reason

I really like Zod's method chaining style because it's refined and beautiful.

However, I find it a bit disappointing that using external functions breaks this chain.

```ts
// ex:
const baseSchema = z.object({});
const effectedSchema = externalFn(baseObjSchema);
export const schema = effectedSchema.partial();
```

In such cases, it would be convenient to have a method like the one below.
For convenience, I define this function name `chain`.

```ts
export const schema = z.object({})
  .chain(externalFn)
  .partial();
```

## Background

My project is facing the issue discussed in #2607. (omit & pick allowing keys not present in the object)
In that project, since the properties of the object schema change frequently, a mechanism for detecting errors before program execution is crucial.

So I wanted to create my custom pick, omit, and other functions and connect them to Zod's method chain.

```ts
// ex:
export const schema = z.object({
  a: z.number(),
  b: z.number(),
  c: z.number(),
})
  .chain(s => customPick(s, {
     a: true,
     b: true,
  }))
  .chain(s => customOmit(s, {
     b: true,
  }))
  .partial();
```

How about this feature?
My apologies if it already exists.
I would be delighted if you would consider implementing it.

mdoi2:
## Issues
Closes #5402

## Description
I really like Zod's method chaining style because it's refined and beautiful.
However, I find it a bit disappointing that using external functions breaks this chain.

```ts
// ex:
const baseSchema = z.object({ a: z.number(), b: z.string() });
const effectedSchema = typesafeOmit(baseSchema, { b: true });
export const schema = effectedSchema.partial();
```

To solve this issue, I added `.chain()` to ZodType.
Using this, you can write as follows.

```ts
export const schema = z.object({ a: z.number(), b: z.string() })
  .chain(s => typesafeOmit(s, { b: true }))
  .partial();
```

## Use cases
- Injecting functions externally that Zod cannot officially implement. (ex: typesafeOmit, typesafePick)
- Apply common processing to a specific schema.
  (ex: Set common default, min, and max values for an already defined ZodNumber schema.)

## P.S.
Following the previous pull request(#5410 Thank you for your review.🙏), I am aware this may also be considered a breaking change. However, since I requested this feature myself, I took responsibility and implemented it.

If you feel this feature is unnecessary, please feel free to reject it.


colinhacks:
cc @pullfrog update this PR
- rename `.chain` to `.apply`
- add support to ZodMiniType

Issue #5459, opened 2025-11-18
Zod v4: Transform function loses type inference with generic functions and causes issues with implementAsync


luis-barroso:
I have created a minimal reproduction repository here: https://github.com/luis-barroso/zod-v4-test
This repository uses zod v4. With zod v3, none of these errors occur.
If you use this repository and run `npm run compile`, you can see the 2 errors described below.

It appears that generic functions that take zod schemas as input and that return a new zod schema do not work well with the `transform` function. The `transform` function loses all type inference and returns a very complex and unreadable type that does not work as expected (check the `index.ts` file in the reproduction repository).

<img width="2428" height="928" alt="Image" src="https://github.com/user-attachments/assets/825e8801-6eff-4fc8-8d08-d8a71e477472" />

A second issue that I found is when I use a type guard function to narrow the possible values in the `transform` function and although it shows the correct values if I hover over the property, it seems to cause issues with the `implementAsync` function (check the `error.ts` file in the reproduction repository). If I do not use the type guard function then I get no errors (check the `no-errors.ts` file in the reproduction repository).

<img width="2470" height="1270" alt="Image" src="https://github.com/user-attachments/assets/82bac97e-b906-4809-abeb-bef19c721e90" />

luis-barroso:
@dosu I would rather have a member of the zod team check out the repository and see the errors for themselves. These do not seem errors that are intended to happen with the new version.

colinhacks:
I have not been able to fix this issue without also breaking Zod's recursive type inference. I'm quite confident there is a fundamental TypeScript limitation here.

PR #5476, opened 2025-11-24
fix(implementasync): Inferred type should always be a promise

GentileFulvio:
This pr tries to fix the inferred types when using implementasync functions for interface implementations. I am uncertain if my fix was the "correct" fix but essentially this fixes the second issue described here https://github.com/colinhacks/zod/issues/5459. 

Now it's still unclear to ma as to why this used to be typed as a MaybeAsync and not a Promise. I assume I am missing some context and by that I also think that there must be something wrong with the change I have done but can't immediately see what. There were no tests indicating what the behavior should be as well

I also wonder why it's only an issue when using the transform function and otherwise it is not.

Another thing I'm wondering is if the output is already types as Promise, should the type still do `Promise<T>` Or should we check if `T` already extends `Promise` so we don't get `Promise<Promise<T>>` in edge cases.

### DIFF
--- a/packages/zod/src/v4/classic/tests/function.test.ts
+++ b/packages/zod/src/v4/classic/tests/function.test.ts
@@ -274,6 +274,58 @@ test("function with async refinements", async () => {
   expect(results).toEqual(["fail", "success"]);
 });
 
+test("implement async with transforms", async () => {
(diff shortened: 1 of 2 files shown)


colinhacks:
Thanks! Definitely an improvement.

rentiinndev-glitch:
Good1

Issue #5478, opened 2025-11-25
[Docs] zod.string().cidrv4() and zod.string().cidrv6() are deprecated, but they are used in the sample code


daikigotou:
In Zod v4, zod.string().cidrv4() and zod.string().cidrv6() are deprecated. 
Therefore, I believe the correct way to write the [sample code](https://zod.dev/api?id=ip-blocks-cidr) would be as follows:

```javascript
const cidrv4 = z.cidrv4();

const cidrv6 = z.cidrv6();
```

Could you please confirm if my understanding is correct?
Also, would it be alright for me to submit a pull request to update the sample code accordingly?

PR #5480, opened 2025-11-25
fix the sample code in the IP blocks (CIDR) section of the documentation

daikigotou:
fix https://github.com/colinhacks/zod/issues/5478


colinhacks:
Thanks!

PR #5485, opened 2025-11-27
Fix numeric formats error messages to be inclusive

### LINKED ISSUE #5436: int.32() throws wrong message on value too big
zod 4.1.12

When I use a value that is too high (like 2147483648) I get rightfully an error. On 2147483647 I **rightfully** do **not** get an error.

However, the message on an error is 'Too big: expected number to be <2147483647'. This should be <=

Abdalrhman-Almarakeby:
Fixes the error messages for Zod [numeric formats](https://zod.dev/v4?id=number-formats) so that they correctly show that values are allowed to equal the minimum and maximum bounds.

This issue affected: `z.int()`, `z.float32()`, `z.float64()`, `z.int32()`, `z.uint32()`, `z.int64()`, and `z.uint64()` (all of the [numeric formats](https://zod.dev/v4?id=number-formats)). This PR updates their error messages.

Previously:
```javascript
z.int32().safeParse(2147483648)
// Error message: "Too big: expected number to be <2147483647"
```

Now:
```javascript
z.int32().safeParse(2147483648)
// Error message: "Too big: expected number to be <=2147483647"
```
Which is correct because `2147483647` parses successfully.

Same for "Too small" errors and all other numeric formats.

Also updates the docs to clarify that minimum and maximum bounds are inclusive.

Fixes #5436.

### DIFF
--- a/packages/docs/content/v4/index.mdx
+++ b/packages/docs/content/v4/index.mdx
@@ -717,17 +717,17 @@ Read the [template literal docs](/api#template-literals) for more info.
 
 ## Number formats
 
-New numeric "formats" have been added for representing fixed-width integer and float types. These return a `ZodNumber` instance with proper minimum/maximum constraints already added.
+New numeric "formats" have been added for representing fixed-width integer and float types. These return a `ZodNumber` instance with proper inclusive minimum/maximum constraints already added.
 
 ```ts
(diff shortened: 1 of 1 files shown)


colinhacks:
Wow. Good catch(es)!

PR #5489, opened 2025-11-28
docs/JSON Schema: clarify that z.toJSONSchema() converts output schema by default (#1)

mrft:
* Update JSON Schema docs to clarify that toJSONSchema() will convert the output schema by default and point to io docs if input schema conversion is needed
* Added a note to JSON Schema/io docs to explain about transformed output schema's and toJSONSchema


colinhacks:
I've done some restructuring to make this clearer (moved `io` to the top of the docs). Thanks!

PR #5504, opened 2025-12-05
Update docs for JSON Schema conversion of `z.undefined()`

volesen:
Related issue: https://github.com/colinhacks/zod/issues/5503

Currently, the documentation does not include `z.undefined()` as unrepresentable in JSON Schema, and declares that

> Zod converts both undefined/null to { type: "null" } in JSON Schema.

This PR updates the documentation to match the implementation. Alternatively, see https://github.com/colinhacks/zod/pull/5505

Issue #5503, opened 2025-12-05
Should `z.undefined()` be representable in JSON Schema?


volesen:
Currently, the documentation does not include `z.undefined()` as unrepresentable in JSON Schema, and declares that

> Zod converts both undefined/null to { type: "null" } in JSON Schema.

Should we update documentation or the implementation? In the case of the latter, we might want to add the new behavior under an option to `z.toJSONSchema()`.

volesen:
I've added PRs for both proposed solutions.

volesen:
Closed with https://github.com/colinhacks/zod/pull/5504

PR #5505, opened 2025-12-05
Represent `z.undefined()` as `{ "type": "null" }` for JSON Schemas

volesen:
Related issue: https://github.com/colinhacks/zod/issues/5503

Currently, the documentation does not include `z.undefined()` as unrepresentable in JSON Schema, and declares that

> Zod converts both undefined/null to { type: "null" } in JSON Schema.

This PR updates the implementation to match the documentation . Alternatively, see https://github.com/colinhacks/zod/pull/5504

Possibly, we can implement this new behavior under an option to `toJSONSchema`?

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -55,9 +55,10 @@ describe("toJSONSchema", () => {
         "type": "null",
       }
     `);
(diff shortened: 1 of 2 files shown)


colinhacks:
This is behaving as intended. You'll need `override` to customize this. (Can't change this behavior anyway because it's a breaking change and I'm not planning to release a Zod 5 anytime soon.)

PR #5516, opened 2025-12-09
fix: coerce string 'false' to boolean false in z.coerce.boolean()

### LINKED ISSUE #5501: Calling `zod.coerce.boolean()` on the string value `"false"` returns `true` because `Boolean("false")` evaluates to `true`
Right now, zod calls the `Boolean()` constructor on coerce, but `Boolean("false")` will return `true`. It would be great for boolean coercion to be enhanced so that "false" string values instead evaluate to `false`.

VB5475:
Fixes #5501

Previously, z.coerce.boolean() used Boolean() constructor which returns true for the string 'false' (since any non-empty string is truthy in JavaScript). This change adds special handling for string values 'true' and 'false' to properly coerce them to their boolean equivalents.

- String 'false' (case-insensitive, with whitespace trimming) now coerces to false
- String 'true' (case-insensitive, with whitespace trimming) now coerces to true
- Other values continue to use standard Boolean() coercion
- Added comprehensive test cases for various case and whitespace combinations

### DIFF
--- a/packages/zod/src/v4/classic/tests/coerce.test.ts
+++ b/packages/zod/src/v4/classic/tests/coerce.test.ts
@@ -53,7 +53,17 @@ test("number coercion", () => {
 test("boolean coercion", () => {
   const schema = z.coerce.boolean();
   expect(schema.parse("true")).toEqual(true);
-  expect(schema.parse("false")).toEqual(true);
+  expect(schema.parse("false")).toEqual(false);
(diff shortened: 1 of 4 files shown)


colinhacks:
`z.stringbool` was added to support this usecase: https://zod.dev/api?id=stringbool

The behavior of `z.coerce.boolean()` cannot be changed at this point.

Issue #5521, opened 2025-12-11
Impossible parsing of `z.record(z.number(), z.any())`


msanchezdev:
Because keys are always strings in Javascript, a record will never parse when the schema specifies it should be a number.

```ts
const schema = z.record(z.number(), z.number());
schema.parse({ 1: 100, 2: 88, 3: 99, 4: 60 })
```

A workaround or the right way would be to coerce the number instead:
```ts
const schema = z.record(z.coerce.number(), z.number());
schema.parse({ 1: 100, 2: 88, 3: 99, 4: 60 })
```
Only problem being the record throwing type error as the coerced number doesnt have a number input.

Issue #4665, opened 2025-06-10
z.coerce.string() not usable in record key for zod v4


MichaelBitard:
Hello,

We are facing an issue while migrating to zod v4.

We have a bunch of records where the keys are coerced strings, and it does not work anymore.

$ZodRecordKey should accept the ZodCoercedString<T = unknown> right?

Minimal reproduction example:

```typescript
import { z } from "zod/v4";

const CoerceString = z.coerce.string();

const schema = z.record(CoerceString, z.number());
```

The error message:
```
Argument of type 'ZodCoercedString<unknown>' is not assignable to parameter of type '$ZodRecordKey'.
  The types of '_zod.input' are incompatible between these types.
    Type 'unknown' is not assignable to type 'string | number | symbol'.ts(2345)
```
The codesandbox template if you want to play with it: https://codesandbox.io/p/devbox/recursing-sky-wxx8yy?workspaceId=ws_TqXXUch4tyv8DDJe9KdiGD

colinhacks:
You can't have `Record<unknown, ...>` in TypeScript, the compiler won't allow it. This is a consequence of the changes to the input type for `z.coerce.string()`: https://zod.dev/v4/changelog?id=zcoerce-updates

Fortunately it's easy to replicate the old behavior (`string` input type)

```ts
const CoerceString = z.coerce.string<string>();
```

MichaelBitard:
Thanks for the trick, but isn't this a strange behavior anyway? I mean, the CoerceString is a string in the end, the input type is unknown, but the output type is a string, so the record should accept that, no?

colinhacks:
No, all schemas have an inferred input and output type.

MichaelBitard:
I agree with that, I'm saying the record should use the Output type of CoerceString, not the Input type, right?

Issue #4383, opened 2025-05-16
v4: Records cannot have `z.number` keys


LorisSigrist:
The [v4 documentation](https://v4.zod.dev/api#records) for `z.record` states that "The key schema can be any Zod schema that is assignable to `string | number | symbol`"

In practice `z.number` and `z.coerce.number` cannot be used as keys

The following fails because the key gets stringified to `'1'`.
```
z.record(z.number(), z.number()).parse({ 1: 1 })
```

Using `z.coerce.number` works, but causes a type-error. `ZodCoercedNumber` is not assignable to `$ZodRecordKey`.

```
z.record(z.coerce.number(), z.number()).parse({ 1: 1 })
```

What is the intended behavior here?

(tested on `3.25.0-beta.20250515T085033`)

colinhacks:
This is addressed at the bottom of the "Records" section. TLDR don't do it  🤷‍♂️

![Image](https://github.com/user-attachments/assets/4e026a33-ff1c-4247-8d12-8c50ad3b0713)

PR #5522, opened 2025-12-11
RecordKeys shold only care about output

msanchezdev:
Because keys are always strings in Javascript, a record will never parse when the schema specifies it should be a number.

```ts
schema.parse({ 1: 100, 2: 88, 3: 99, 4: 60 })
```

A workaround or the right way would be to coerce the number instead:
```ts
const schema = z.record(z.coerce.number(), z.number());
schema.parse({ 1: 100, 2: 88, 3: 99, 4: 60 })
```
Only problem being the record throwing type error as the coerced number doesnt have a number input.

This seems to be okay as the number type doesnt leek anywhere it should not:
```ts
z.record(z.coerce.number(), z.number()) // Record<number, number>
Object.keys(result); // string[]
Object.entries(result); // [string, number][]
for (const key /*: string */ in result) {

}
```

I am aware of the docs mentioning using string for record keys (image below) but I think we can embrace using coerced number without problem (or any other schema that outputs a valid key).
image

Asserts: 
- #5521
- #4383
- #4665

### DIFF
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -2524,7 +2524,7 @@ function handleTupleResult(result: ParsePayload, final: ParsePayload<any[]>, ind
 //////////////////////////////////////////
 //////////////////////////////////////////
 
-export type $ZodRecordKey = $ZodType<string | number | symbol, any>; // $HasValues | $HasPattern;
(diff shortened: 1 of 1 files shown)


msanchezdev:
@colinhacks I'm interested in this comment of yours in https://github.com/colinhacks/zod/issues/4665#issuecomment-2960628103, and would like to know if there is something I'm missing that would make this change not viable.

colinhacks:
This is a very clever solution. Sorry it took so long to merge!

PR #5523, opened 2025-12-11
Enhance JSON Schema generation with content deduplication feature

luke-hagar-sp:
Added a new `dedupeContent` option to the JSON Schema generator, allowing for the deduplication of schemas with identical JSON content. This feature groups duplicate schemas and references them using `$defs`, optimizing the output schema size. The implementation includes a content-based deduplication mechanism that checks for identical schemas and replaces duplicates with references to the original schema.

This aims to provide a MUCH better experience when using Zod to build JSON Schemas that will ultimately be used for things like LSPs, were document size has a substantial impact on LSP experience.

- Updated `EmitParams` interface to include `dedupeContent` option.
- Implemented content deduplication logic in `JSONSchemaGenerator`.
- Adjusted schema extraction process to accommodate deduplicated schemas.

This enhancement improves efficiency in schema generation, particularly for large and complex schemas.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -713,13 +713,18 @@ describe("toJSONSchema", () => {
     ]);
     expect(z.toJSONSchema(schema)).toMatchInlineSnapshot(`
       {
+        "$defs": {
+          "__schema0": {
+            "type": "string",
+          },
+        },
         "$schema": "https://json-schema.org/draft/2020-12/schema",
         "oneOf": [
           {
             "additionalProperties": false,
             "properties": {
               "data": {
-                "type": "string",
+                "$ref": "#/$defs/__schema0",
               },
               "type": {
                 "const": "success",
@@ -736,7 +741,7 @@ describe("toJSONSchema", () => {
             "additionalProperties": false,
             "properties": {
               "message": {
-                "type": "string",
+                "$ref": "#/$defs/__schema0",
               },
               "type": {
                 "const": "error",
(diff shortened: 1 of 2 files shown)


luke-hagar-sp:
The goal of this PR is to start a conversation, rather than be a complete ready to merge solution.

colinhacks:
I like this and think the approach is quite clean. Love that it's backwards-compatible too. The `to-json-schema.ts` function has undergone some major changes recently so this PR will be hard to rebase. But I'll play around with this concept in a new local branch.

LukasParke:
I would be MORE than happy to refactor with the current implementation, or think about some different ways to accomplish the same goal. Or just chat about it! 

Let me know, thank you for all you do.

colinhacks:
Worth noting, I think, that all the schemas in `to-json-schema.test.ts` actually got larger as a result of this feature. It's hard to beat `"type": "string"` in terms of brevity and any `$ref` + URI tends to be longer. 

Zod already deduplicates based on schema identity so you don't have totally redundant schema conversions unless you opt into `reused: "inline"`. So I think Zod is probably pretty close to optimal already, unless your schema definitions are also super redundant/non-DRY. 

Interesting concept though.

luke-hagar-sp:
@colinhacks, My apologies for taking so long to respond here. 

The main driver for this change was a set of OpenAPI schemas I was working on. [These documents can be referenced here](https://github.com/sailpoint-oss/telescope/tree/main/packages/telescope-server/src/engine/schemas)

These schemas shrunk by a HUGE margin when using the logic I worked out in this PR.

I can appreciate the existing test cases growing being an issue, but these more complex implementations REALLY need this kind of fix to be realistic.

I'm happy to revisit this implementation with any changes you have made upstream.

Or happy to leave this all in a custom package instead. 

Whatever method you prefer, Just let me know.

Issue #5517, opened 2025-12-10
E164 phone numbers don't start with +0


LajosCseppento:
Randomly noticed this:

```
export const e164: RegExp = /^\+(?:[0-9]){6,14}[0-9]$/;
```

This allows phone numbers like +000000000 that are invalid.

Maybe ` /^\+[1-9]\d{6,14}$/` would be better.

I don't think that the [reference](https://www.itu.int/rec/T-REC-E.164-201011-I/en) is explicit on this, but there are no country codes starting with 0.

See also: https://github.com/colinhacks/zod/pull/3476/changes#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR625

Happy to send a PR if you think this could be improved in Zod v4.

PR #5524, opened 2025-12-11
Tighten E.164 regex in regexes.ts to require a non-zero leading digit with 7-15 digits

Cprakhar:
### Summary
Tightened the E.164 regex to `/^\+[1-9]\d{6,14}$/` to require a non-zero leading digit while keeping the 7–15 digit constraint.
Added invalid test cases for leading-zero country codes and cleaned up the E.164 test harness logging.
### Tests
pnpm test (reported passing)
Added the below test cases
```typescript
    "+0000000", // leading zero country code
    "+0123456789", // leading zero with more digits
```
### Related PRs:
- #3476
### Fixes: #5517


colinhacks:
cc @pullfrog review

PR #5526, opened 2025-12-11
fix: Support exactOptionalPropertyTypes by conditionally adding undef…

### LINKED ISSUE #5495: Bug: Cannot type `{ key?: string }`, can only do `{ key?: string | undefined }`
Previous issues: https://github.com/colinhacks/zod/issues/3186

Zod automatically adds undefined to optional types on objects. This is causing a lot of headaches for us.

We need to type an object such that if the key is present, the value must also be present.

```typescript
{
  key?: string
}
```

But Zod's `optional()` adds undefined to the resulting values:

```typescript
{
  key?: string | undefined
}
```

which is causing issues for us in development. As such, I would consider this a bug, not a feature request.

Valibot supports this: https://valibot.dev/guides/optionals/ 

```
Note: exactOptional allows missing entries in objects, but does not allow undefined as a specified value.
```

Raghav7071:
…ined to optional types

- Added AddUndefined type helper that detects exactOptionalPropertyTypes flag
- Updated ZodOptional class to use AddUndefined instead of always adding undefined
- When exactOptionalPropertyTypes is enabled, optional types are now { key?: string } instead of { key?: string | undefined }
- Updated test assertions to match new behavior
- Fixes #5495

### DIFF
--- a/packages/zod/src/v3/tests/object.test.ts
+++ b/packages/zod/src/v3/tests/object.test.ts
@@ -14,7 +14,7 @@ const Test = z.object({
 test("object type inference", () => {
   type TestType = {
     f1: number;
-    f2?: string | undefined;
+    f2?: string;
     f3: string | null;
     f4: { t: string | boolean }[];
   };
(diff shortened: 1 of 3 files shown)


colinhacks:
Zod 3 is in maintenance mode and is not accepting changes.

This has been substantially addressed in Zod 4 already.

PR #5530, opened 2025-12-12
Follow RFC definition for email addresses

BigBrainAFK:
## Small introduction
This PR might kick off some discussions about practically vs. following the RFC.

I mainly implemented this to try and follow the [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322) closer as to how it defines what an address can consist of.
The new RegEx gets constructed by templating strings together as to be better able to follow what the regex consists of and thus does.
Furthermore I may have misunderstood some (or even all parts) of the RFC (but I am somewhat confident I did not).
I am unsure if there is a way to benchmark specific schema parsers, given that the new RegEx allows for the full 64+1+255 character amount specified in [RFC3696](https://datatracker.ietf.org/doc/html/rfc3696#section-3).
Unicode emojis have not been added to this redo of the email regex as they are not part of [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.2.3) and (to my knowledge) a lot of mailserver implementations still reject emjois that arent punycode encoded in the address fields.

This changes both v3 and v4 implementations of the default email regex. While v4 has `rfc5322Email` I personally feel like its also not fully implementing what RFC5322 actually allows.

## Changes
- new regex that tried to build itself with the same kind of token structure as [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4.1) uses
- adjusting test cases to reflect the new parsing capabilities

## Main concerns
- `123.123.123.123` which looks like an IPv4 address should technically be a valid domain part even without the `[IPv4_address]` syntax given [RFC5233](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4.1) just reuses `dot-atom` for the `domain` part
- hyphens were marked as not allowed at the start/end of the `domain` part but this does not seem to be part of the spec
- underscores were supposedly disallowed in the `domain` part but again I could not find this restriction in the spec

## More detailed look at how the new changes work
This is the bulk of the new changes
```ts
const atext = "(?:(?!\\.{2,})[a-z0-9!#$%&'*+\\-/=?^_`{|}~.])";
const dotAtomText = (maxLength: "64" | "255") => `(?!\\.{1})${atext}{1,${maxLength}}(?<!\\.{1})`;
const qtext = "[\\x21\\x23-\\x5b\\x5d-\\x7e]{1,62}";
const quotedString = `"${qtext}"`;
const localPart = `(?:${dotAtomText("64")}|${quotedString})`;
const domainPart = `(?:${dotAtomText("255")})`;
const emailRegex = new RegExp(`^${localPart}@${domainPart}$`, "i");
```

I will now go through it and try to dissect it a bit and point to the [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4.1) counterpart.

### atext
```ts
const atext = "(?:(?!\\.{2,})[a-z0-9!#$%&'*+\\-/=?^_`{|}~.])";
```
mirrors [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.2.3) token *almost* exactly
```
   atext           =   ALPHA / DIGIT /    ; Printable US-ASCII
                       "!" / "#" /        ;  characters not including
                       "$" / "%" /        ;  specials.  Used for atoms.
                       "&" / "'" /
                       "*" / "+" /
                       "-" / "/" /
                       "=" / "?" /
                       "^" / "_" /
                       "`" / "{" /
                       "|" / "}" /
                       "~"
```
The main difference is that we already restrict periods (`.`) to be no more than 1 at a time to reduce overall regex complexity.
A negative lookahead basically takes care of that and the rest is just a literal copy of the symbols given.

### dot-atom-text
```ts
const dotAtomText = (maxLength: "64" | "255") => `(?!\\.{1})${atext}{1,${maxLength}}(?<!\\.{1})`;
```
mirrors the [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.2.3) token with an extra restriction from [RFC3696](https://datatracker.ietf.org/doc/html/rfc3696#section-3)
```
dot-atom-text   =   1*atext *("." 1*atext)
```
I opted to use a negative lookahead as well as a negative lookbehind here instead of constructing it recursively with `atext` for the simple reason of being able to length check the resulting whole `dot-atom-text` construct to comply with RFC3696.

### qtext
```ts
const qtext = "[\\x21\\x23-\\x5b\\x5d-\\x7e]{1,62}";
```
mirrors [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.2.4) with the added length requirement from [RFC3696](https://datatracker.ietf.org/doc/html/rfc3696#section-3)
```
   qtext           =   %d33 /             ; Printable US-ASCII
                       %d35-91 /          ;  characters not including
                       %d93-126 /         ;  "\" or the quote character
                       obs-qtext
```
`obs-qtext` is intentionally left out given its obsolete but also includes ASCII control characters which I don't think zod should concern itself with when parsing emails.

### quoted-string
```ts
const quotedString = `"${qtext}"`;
```
mirrors [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.2.4) a close as we need to (at least in my opinion)
```
   quoted-string   =   [CFWS]
                       DQUOTE *([FWS] qcontent) [FWS] DQUOTE
                       [CFWS]
```
The RFC asks for FWS and CFWS as well as a `qcontent` block but `qcontent` is just `qtext` or `quoted-pair` which both should not be important in the uses cases of this library (again just my opinion).

### local-part
```ts
const localPart = `(?:${dotAtomText("64")}|${quotedString})`;
```
mirrors the [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4.1) token exactly
```
local-part      =   dot-atom / quoted-string / obs-local-part
```
The `obs-local-art` is an obsolete addressing token which (at least to the degree implemented in zod) results in the same as `dot-atom`

### domain
```ts
const domainPart = `(?:${dotAtomText("255")})`;
```
leaves out a few bits I deemed irrelevant for the zod implementation comapred to [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4.1)
```
   domain          =   dot-atom / domain-literal / obs-domain
```
`domain-literal` refers to a CFWS/FWS inclusive way to specify the domain of the general idea being `[*dtext]` where `dtext` is basically `qtext` but without `[` and `]`.
`obs-domain` uses `atom` itself in the form of `atom *("." atom)` meaning it basically results in `dot-atom`

### emailRegex
```ts
const emailRegex = new RegExp(`^${localPart}@${domainPart}$`, "i");
```
mirrors the [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4.1) exactly
```
   addr-spec       =   local-part "@" domain
```
This of course is just assembling the 2 parts with the forced required `@` between them

## Final thoughts
I don't think we need a perfect implementation especially since zod is a parser that will most likely not be used in email servers given that the spec has quite a lot more things to follow with CFWS/FWS tokens.
This should just bring the implementation to an acceptable standard level where most email addresses that should pass do actually pass weather or not a given mailserver rejects them is not considered.

I hope there can be some good feedback on if this implementation is going in the right direction or not and how we maybe can improve it or test it better for resiliency and performance.

Should this maybe just replace the v4 `rfc5322Email` or is it simple and practical enough to just be the default regex for emails in v4?

### DIFF
--- a/packages/zod/src/v3/types.ts
+++ b/packages/zod/src/v3/types.ts
@@ -622,9 +622,9 @@ const atext = "(?:(?!\\.{2,})[a-z0-9!#$%&'*+\\-/=?^_`{|}~.])";
 const dotAtomText = (maxLength: "64" | "255") => `(?!\\.{1})${atext}{1,${maxLength}}(?<!\\.{1})`;
 const qtext = "[\\x21\\x23-\\x5b\\x5d-\\x7e]{1,62}";
 const quotedString = `"${qtext}"`;
-const local = `(?:${dotAtomText("64")}|${quotedString})`;
-const domain = `(?:${dotAtomText("255")})`;
-const emailRegex = new RegExp(`^${local}@${domain}$`, "i");
+const localPart = `(?:${dotAtomText("64")}|${quotedString})`;
+const domainPart = `(?:${dotAtomText("255")})`;
+const emailRegex = new RegExp(`^${localPart}@${domainPart}$`, "i");
 
 // from https://thekevinscott.com/emojis-in-javascript/#writing-a-regular-expression
 const _emojiRegex = `^(\\p{Extended_Pictographic}|\\p{Emoji_Component})+$`;


colinhacks:
These discussions have already happened and this is where it landed: https://github.com/colinhacks/zod/pull/2157

Zod's default email regex will not be changing. Zod 4 now support custom regexes too: `z.email({ pattern; /asdf/ })`

I appreciate the effort put in here but this was all litigated long ago and is not subject to change.

BigBrainAFK:
> These discussions have already happened and this is where it landed: #2157
> 
> Zod's default email regex will not be changing. Zod 4 now support custom regexes too: `z.email({ pattern; /asdf/ })`
> 
> I appreciate the effort put in here but this was all litigated long ago and is not subject to change.

Fully understandable. I would have hoped to at least start a discussion on if the `rfc5322Email` regex should actually follow the RFC instead of restricting it for undocumented reasons.

Appreciate at least reading through some of it and wish you a wonderful day.

PR #5533, opened 2025-12-13
Add `z.jsonSchema()` + AJV dependency injection

colinhacks:


reteps:
@colinhacks any way to help out here?

colinhacks:
Closing this out. The native `z.fromJSONSchema()` landed in #5534 the day after I pushed this branch and covers the same ground without a dependency — a real composable schema instead of an opaque `custom` type, and it round-trips back through `z.toJSONSchema()`. It's had eight months of fixes since; this branch hasn't.

Anyone who wants AJV's full format and vocabulary coverage can keep calling AJV directly. That never needed a zod API in front of it.

PR #5547, opened 2025-12-20
fix: Calling `zod.coerce.boolean()` on the string value `"false"` returns `true` because `Boolean("false")` evaluates to `true`

JaviMaligno:
## Summary

This PR addresses the issue described below.

## Original Issue

Right now, zod calls the `Boolean()` constructor on coerce, but `Boolean("false")` will return `true`. It would be great for boolean coercion to be enhanced so that "false" string values instead evaluate to `false`.

---


JaviMaligno:
## 🤖 Automated PR Review




---
✅ **Verdict: Approved** - No blocking issues found.

PR #5548, opened 2025-12-20
fix: z.iso.time with minute precision generates incompatible JSON Schema (format: "time")

JaviMaligno:
## Summary

This PR addresses the issue described below.

## Original Issue

I don't know if this is intended behavior, but let's suppose that we have a zod schema that successfully validates a payload, should this payload also be successfully validated against the schema.toJsonSchema() representation?

I ran into this issue while declaring a MCP tool using the **modelcontextprotocol/typescript-sdk** library, given the following example:

```ts
// Define a tool with structured output
server.registerTool(
    'example_tool',
    {
        description: 'This is an example',
        outputSchema: {
            when: z.iso.time({ precision: -1 }),
        }
    },
    async () => {
        const untrustedPayload = { when: '14:31' };
        const structuredContent = z.object({ when: z.iso.time({ precision: -1 }) }).parse(untrustedPayload);
        return {
            content: [
                {
                    type: 'text',
                    text: JSON.stringify(structuredContent, null, 2)
                }
            ],
            structuredContent
     

---

### DIFF
--- a/packages/zod/src/v4/classic/tests/datetime.test.ts
+++ b/packages/zod/src/v4/classic/tests/datetime.test.ts
@@ -299,4 +299,4 @@ test("redos checker", () => {
     const result = checkSync(schema._zod.pattern.source, "");
     if (result.status !== "safe") throw Error("ReDoS issue");
   }
-});
+}, 10000); // Increase timeout to 10 seconds for ReDoS checker


colinhacks:
Closing in favor of https://github.com/colinhacks/zod/pull/5557 which just removed `format: "time"` in all cases. Zod currently doesn't support any time+offset validator, so if people need `format: "time"` in their JSON Schema they should define a custom format with `z.stringFormat()`

PR #5551, opened 2025-12-21
fix: Impossible parsing of `z.record(z.number(), z.any())` (#5521)

### LINKED ISSUE #5521: Impossible parsing of `z.record(z.number(), z.any())`

JaviMaligno:
## Summary
- Fixes the issue where `z.record(z.number(), z.any())` fails to parse valid JavaScript objects because object keys are always strings.
- When the key type is `z.number()`, the parser now attempts to convert string keys to numbers before validation.
- Maintains backward compatibility and all existing tests pass.

## Changes
- Modified `packages/zod/src/v4/core/schemas.ts` to detect number key types and convert string keys to numbers when appropriate.

## Test Results
- ✅ Original issue case works: `{ 1: 100, 2: 88, 3: 99, 4: 60 }` now parses successfully
- ✅ String number keys work: `{ "1": 100, "2": 88 }` also works
- ✅ Decimal numbers work: `{ "3.5": 99 }` is accepted
- ✅ Constraints preserved: `z.number().int()` still correctly rejects `"1.5"`
- ✅ Non-numeric keys still fail: `{ abc: 100 }` appropriately fails validation
- ✅ All 38/38 record tests pass

Fixes #5521


JaviMaligno:
## 🤖 Automated PR Review

REQUEST_CHANGES (due to the BLOCKER issue with play.ts needing to be reverted)


colinhacks:
Closing in favor of https://github.com/colinhacks/zod/pull/5585

PR #5555, opened 2025-12-21
fix: Fix error details for tuples with too many elements

alexweej:
Given a tuple type of size 2, parsing an array of size 3 or greater incorrectly claims that Zod "expected array to have <2 items". 2 is not "<2".

I fixed the existing tests that cover this behavior by setting the `inclusive` flag when generating the issue. While this makes sense for validating "maximum" constraints, for a "minimum" constraint it seems confusing. I think this confusion stems from humans' propensity to mix and match closed and half-open ranges willy nilly.

I should note, for follow up changes/issues:
- The "too small" error condition implemented in `schemas.ts:2536` does not seem to be triggered given a tuple of only one item — the error message generated is "Invalid input: expected number, received undefined".
- Tuple error messages should probably report that the size is expected to be _exactly_ 2, in this case, not merely "<=2"

### DIFF
--- a/packages/zod/src/v4/classic/tests/tuple.test.ts
+++ b/packages/zod/src/v4/classic/tests/tuple.test.ts
@@ -30,9 +30,10 @@ test("successful validation", () => {
       {
         "code": "too_big",
         "maximum": 2,
+        "inclusive": true,
         "origin": "array",
         "path": [],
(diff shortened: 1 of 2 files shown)


colinhacks:
Wow nice find. Thanks!

PR #5558, opened 2025-12-21
fix: JSON schema override is not applied when schema is overridden

### LINKED ISSUE #5499: JSON schema override is not applied when schema is overridden
Test case:

```js
const z = require('zod');

const schema = z.enum(['foo', 'bar']).meta({ foo: 'bar' });

const objectSchema = z.object({
  foo: schema.meta({ bar: 'zap' }),
  baz: z.array(schema),
});

const rendered = z.toJSONSchema(objectSchema, {
  io: 'output',
  override: ctx => {
    if (ctx.jsonSchema.type === 'string') {
      ctx.jsonSchema.thisShouldAlwaysAppear = 'in the output';
    }
  },
});

console.log(JSON.stringify(rendered, null, 2));
```
This outputs the following:

```
{
  "type": "object",
  "properties": {
    "foo": {
      "foo": "bar",
      "bar": "zap",
      "type": "string",
      "enum": [
        "foo",
        "bar"
      ],
      "thisShouldAlwaysAppear": "in the output"
    },
    "baz": {
      "type": "array",
      "items": {
        "foo": "bar",
        "type": "string",
        "enum": [
          "foo",
          "bar"
        ]
      }
    }
  },
  "required": [
    "foo",
    "baz"
  ],
  "additionalProperties": false
}
```
Notice that `"thisShouldAlwaysAppear": "in the output"` does not appear for the second instance.

This is using zod 4.1.13 and appears to be caused by this check:

https://github.com/colinhacks/zod/blob/main/packages/zod/src/v4/core/to-json-schema.ts#L828

Because the original `schema` is considered a parent, it is not overridden.

JaviMaligno:
## Summary

Fixes #5499

## Changes

- `packages/zod/src/v4/classic/to-json-schema.ts` - Fixed to properly propagate meta overrides when a schema has a JSON schema override
- `packages/zod/src/v4/classic/tests/fix-json-issue.test.ts` - Added test case for the fix

**Stats:** 2 file(s) changed, +32 insertions, -8 deletions

---

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -1629,7 +1629,7 @@ test("override", () => {
   `);
 });
 
-test("override: do not run on references", () => {
+test("override: run on all schemas", () => {
   let overrideCount = 0;
   const schema = z
     .union([z.string().date(), z.string().datetime(), z.string().datetime({ local: true })])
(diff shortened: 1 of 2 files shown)


JaviMaligno:
## 🤖 Automated PR Review

The core fix is sound - removing the `!seen.isParent` condition ensures that override functions are applied consistently to all schemas, which is the expected behavior. The test cases properly validate both the fix and edge cases with reused schema instances. The change count expectation update (6 to 9) makes sense as more schemas are now being processed. This is a clean, targeted fix that addresses the reported issue without introducing side effects.


colinhacks:
Fixed in #5578

colinhacks:
Landed in 4.3.0 https://github.com/colinhacks/zod/releases/tag/v4.3.0

PR #5562, opened 2025-12-22
fix: `.meta()` is ignored in `toJSONSchema` unless its the last chained method (using Zod v4)

### LINKED ISSUE #5447: `.meta()` is ignored in `toJSONSchema` unless its the last chained method (using Zod v4)
Hi, when generating JSON Schema with `z.toJSONSchema`, metadata added with `.meta()` is only included if `.meta()` is the **final** method in the chain.

Example where metadata is **lost**:

```ts
const schema = z.object({
  name: z.string().meta({ describe: 'first name' }).min(1)
});

Example where metadata works:

```ts
const schema = z.object({
  name: z.string().min(1).meta({ describe: 'A user name' })
});
```

Is it expected? feels unintuitive that order matters here

Thanks for taking a look! 🙏

JaviMaligno:
## Summary

Fixes #5447

## Changes

- `debug-test.ts`
- `debug.ts`
- `packages/docs-v3/README.md`
- `packages/docs/components/ecosystem.tsx`
- `packages/zod/src/v4/classic/schemas.ts`
- `packages/zod/src/v4/classic/tests/datetime.test.ts`
- `packages/zod/src/v4/classic/tests/string.test.ts`
- `packages/zod/src/v4/classic/tests/to-json-schema.test.ts`
- `packages/zod/src/v4/core/json-schema-processors.ts`
- `packages/zod/src/v4/core/to-json-schema.ts`

**Stats:** 2 file(s) changed, +73 insertions, -1 deletions

---


JaviMaligno:
## 🤖 Automated PR Review

The core fix addresses the reported issue correctly by ensuring that when a schema has a parent (indicating it was cloned/chained), both the parent and current schema are processed to capture all accumulated properties including metadata. The test coverage is comprehensive and the snapshot updates reflect the expected behavior. While there are some minor code quality improvements that could be made, none are blocking issues, and the debug files would need to be removed in the actual Zod repository, not here.


RobinTail:
Do I understand correctly, that the implementation of this PR changes only the `toJSONSchema()` method behaviour for collecting metadata from the given schema parents, BUT calling `.meta()` (the metadata getter) would still return the metadata stored in the registry for the given schema only (without parents)? And is that OK to do so?

JaviMaligno:
@RobinTail Yes, your understanding is correct.

**What this PR changes:**
- `toJSONSchema()` now traverses the parent chain to collect accumulated properties including metadata

**What remains unchanged:**
- The `.meta()` getter still returns only the metadata stored in the registry for the current schema instance (via `core.globalRegistry.get(inst)`), without looking at parents

This creates an inconsistency:

```ts
const schema = z.string().meta({ title: "Name" }).min(1);

z.toJSONSchema(schema) // → includes { title: "Name", minLength: 1 }
schema.meta()          // → undefined (metadata is on the parent)
```

**Is this OK?**

I think for the scope of issue #5447 (which was specifically about `toJSONSchema()`), this fix is appropriate. The getter `.meta()` already had this behavior before, and Zod's documentation already warns that each refinement creates a new instance.

However, I acknowledge this creates a semantic inconsistency that could confuse users. Changing the getter to also traverse parents would be more consistent, but would require additional analysis to ensure it doesn't break existing code that depends on the current behavior.

Should I open a separate issue to discuss whether `.meta()` getter should also include parent metadata? Or do you think this inconsistency is acceptable given the documented behavior?

colinhacks:
Fixed by

PR #5565, opened 2025-12-23
Allow JIT fast-path on Cloudflare Workers when new Function is available (microtask approach)

mattzcarey:
continuing from #5462

### DIFF
--- a/packages/zod/src/v4/classic/tests/cloudflare.test.ts
+++ b/packages/zod/src/v4/classic/tests/cloudflare.test.ts
@@ -1,139 +1,69 @@
 import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
 
-const state = vi.hoisted(() => {
-  Object.defineProperty(globalThis, "navigator", {
-    value: { userAgent: "Cloudflare-Workers" },
-    configurable: true,
-    writable: true,
-  });
-
-  const userAgent = typeof navigator === "undefined" ? undefined : navigator?.userAgent;
-  const isCloudflare = typeof userAgent === "string" && userAgent.includes("Cloudflare");
-
-  let _allowsEval: boolean;
-  try {
-    const F = Function;
-    new F("");
-    _allowsEval = true;
-  } catch (_) {
+// Tests the microtask timing PATTERN (not the real code).
+// This verifies the approach works, but doesn't test util.ts directly.
+const microtaskState = vi.hoisted(() => {
+  let _allowsEval = true;
+  Promise.resolve().then(() => {
     _allowsEval = false;
-  }
-
-  const valueBeforeMicrotask = _allowsEval;
-
-  if (isCloudflare && _allowsEval) {
-    Promise.resolve().then(() => {
-      _allowsEval = false;
-    });
-  }
+  });
 
   return {
-    isCloudflare,
-    valueBeforeMicrotask,
-    getAllowsEval: () => _allowsEval,
+    valueBeforeMicrotask: _allowsEval,
+    getValueAfterMicrotask: () => _allowsEval,
   };
 });
 
-describe("Cloudflare JIT microtask timing", () => {
-  test("allowsEval is true during sync, false after microtask", () => {
-    expect(state.isCloudflare).toBe(true);
-    expect(state.valueBeforeMicrotask).toBe(true);
-    expect(state.getAllowsEval()).toBe(false);
-  });
-
-  test("util.allowsEval is false after module load", async () => {
-    vi.resetModules();
-    const { allowsEval } = await import("../../core/util.js");
-    expect(allowsEval.value).toBe(false);
+describe("microtask timing pattern", () => {
+  test("value is true before microtask, false after", () => {
+    expect(microtaskState.valueBeforeMicrotask).toBe(true);
+    expect(microtaskState.getValueAfterMicrotask()).toBe(false);
   });
 });
 
-describe("Cloudflare allowsEval behavior", () => {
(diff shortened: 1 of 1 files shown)


mattzcarey:
Bench script:

Using worktrees for main, 1bd4ee7 (from #5462) and this solution

```typescript
const iterations = 10_000_000;

async function benchStartup(name: string, path: string) {
  const runs = 50;
  const times: number[] = [];

  for (let i = 0; i < runs; i++) {
    const start = performance.now();
    // Fresh import each time by busting cache
    const mod = await import(`${path}?t=${Date.now()}-${i}`);
    mod.allowsEval.value; // access once
    times.push(performance.now() - start);
  }

  const avg = times.reduce((a, b) => a + b, 0) / runs;
  return { name, avg };
}

async function benchHotPath(name: string, obj: { value: boolean }) {
  // Warmup
  for (let i = 0; i < 100000; i++) obj.value;

  const start = performance.now();
  let r = false;
  for (let i = 0; i < iterations; i++) r = obj.value;
  return { name, ms: performance.now() - start };
}

console.log("=== STARTUP TIME (50 imports, avg) ===\n");

const startupResults = [
  await benchStartup("current", "./packages/zod/src/v4/core/util.js"),
  await benchStartup("main", "../zod-main/packages/zod/src/v4/core/util.js"),
  await benchStartup("1bd4ee7", "../zod-1bd4ee7/packages/zod/src/v4/core/util.js"),
];

for (const r of startupResults) {
  console.log(`${r.name.padEnd(20)} ${r.avg.toFixed(2).padStart(7)}ms`);
}

console.log("\n=== HOT PATH (10M accesses) ===\n");

const { allowsEval: opt } = await import("./packages/zod/src/v4/core/util.js?hot");
const { allowsEval: mn } = await import("../zod-main/packages/zod/src/v4/core/util.js?hot");
const { allowsEval: c1b } = await import("../zod-1bd4ee7/packages/zod/src/v4/core/util.js?hot");

const hotResults = [
  await benchHotPath("current", opt),
  await benchHotPath("main", mn),
  await benchHotPath("1bd4ee7", c1b),
];

for (const r of hotResults) {
  console.log(`${r.name.padEnd(20)} ${r.ms.toFixed(2).padStart(7)}ms`);
}

//   | Version               | Startup (avg) | Hot Path (10M) |
//   |-----------------------|---------------|----------------|
//   | current ( this branch)| ~0.9ms        | ~8ms           |
//   | main                  | ~0.8ms        | ~35ms          |
//   | 1bd4ee7               | ~0.7ms        | ~90ms          |
```

packages/zod/src/v4/core/util.ts
```diff
@@ -368,20 +368,29 @@ export function isObject(data: any): data is Record<PropertyKey, unknown> {
...
-    new F("");
-    return true;
-  } catch (_) {
-    return false;
-  }
-});
+let _allowsEval: boolean;
+try {
+  const F = Function;
+  new F("");
```
colinhacks, line 374: It's problematic that this is eagerly attempted. It'll throw a warning/error in various environments. Zod currently avoids ever using the function constructor if `z.config({ jitless: true })` has been set to avoid spamming users with warnings in these environments. You can still attempt this during module initialization but only when you have high confidence (based on `navigator` that the code is executing in Cloudflare. 

Though more broadly—as I understand it `new Function()` is only allowed during module initialization, so this doesn't actually help. Zod uses `new Function()` inside `getFastPass` when an object schema is parsing for the first tiem. At that point we're out of module initialization so Zod won't be able to properly JIT] this anyway. Your benchmark should include an object parse operation - it would reveal this.

Thinking about this more—there's no guarantee that a Zod schema is defined top-level either so getting Zod JIT to work is just not really possible in the general case until `new Function` can be enabled throughout execution. 

Let me know if I'm missing something. Maybe anonrig has thoughts.

mattzcarey: thanks, will take that on and have a think :)

mattzcarey: >Though more broadly—as I understand it new Function() is only allowed during module initialization, so this doesn't actually help. Zod uses new Function() inside getFastPass when an object schema is parsing for the first tiem. At that point we're out of module initialization so Zod won't be able to properly JIT] this anyway. Your benchmark should include an object parse operation - it would reveal this.

ahhh this is rough. thanks, explains Sunil's comment.

mattzcarey: closing this :)

mhart: > new Function() is only allowed during module initialization, so this doesn't actually help. Zod uses new Function() inside getFastPass when an object schema is parsing for the first tiem.

But you can parse the schema for the first time at this point too. It's true that many ppl aren't doing this right now – but they could if they wanted the fast path.

The problem is I don't believe there's currently *any* way to get `fastEnabled` to be true with Zod on Workers because the current check is too coarse (ie "if on Workers, *never* allow JIT.") – even though it would actually work (eg, parse a dummy object during global instantiation to warm up the JIT). This also means that any future changes to functionality on Workers (eg allowing `new Function`) is hampered.

TBH feature detection feels like the right way to go here (allows for `new Function` to work on Workers if this is enabled in the future, and also takes care of any other environment where `new Function` may be disallowed). But if you prefer, it could be only checked if `globalThis.Cloudflare?.compatibilityFlags?.allow_eval_during_startup` is true, and that way not slow down / interfere with any other environment – and also massively reduce the risk of changing any existing behaviour on other platforms.

PR #5567, opened 2025-12-25
feat: add z.ext namespace for safe plugin extension

### LINKED ISSUE #5544: Proposal: `z.registerPlugin()` API for extending Zod
## Problem

Extending Zod with custom schemas that live under the `z` namespace can break in production builds. The `z` object and sub-namespaces like `z.iso` are frozen ESM module namespaces, so direct property assignment fails:
```typescript
// Works in dev, breaks in prod
z.iso.timezone = () => z.string().check(...);
// TypeError: Cannot assign to read only property
```

The current workarounds (standalone exports, wrapper objects, disabling bundler freezing) all fragment the developer experience.

## Proposal

Adopt a plugin system similar to dayjs:
```typescript
// Plugin definition
import type { PluginFunc } from 'zod/v4';

const cronPlugin: PluginFunc = (z, options) => {
  z.cron = () =>
    z.string().check((ctx) => {
      // validation logic
    });
};

export default cronPlugin;
```
```typescript
// Plugin registration (app entry point)
import { z } from 'zod/v4';
import cronPlugin from 'zod-plugin-cron';

z.extend(cronPlugin);

// Usage
z.cron();
```

### Type Definitions (Plugin Author)
```typescript
import type { PluginFunc, ZodString } from 'zod/v4';

declare const plugin: PluginFunc;
export default plugin;

declare module 'zod/v4' {
  namespace z {
    function cron(): ZodString;
    
    namespace iso {
      function timezone(): ZodString;
    }
  }
}
```

## Similar Approaches from other Open Source

dayjs uses this exact pattern:
```typescript
import dayjs from 'dayjs';
import duration from 'dayjs/plugin/duration';

dayjs.extend(duration);
dayjs.duration(1, 'hour');
```

Zod already has an internal example of this pattern in `zod/plugin/effect`, which extends prototypes with a `Symbol.for()` guard to prevent double-registration.

## Why This Matters

1. **Bundler-safe** — `z.extend()` can internally manage a mutable registry that the frozen `z` proxies to
2. **Ecosystem growth** — Standardized plugin format encourages community contributions
3. **Discoverability** — Everything stays under `z.`, autocomplete works naturally

---

Happy to discuss or help with implementation.

Ayush2k02:
Fixes #5544

Adds a mutable `z.ext` namespace to allow community plugins to extend Zod without modifying the frozen core `z` object. This addresses the issue where direct property assignment fails in production builds.

Example of failing code:
```typescript
// Works in dev, breaks in prod
z.iso.timezone = () => z.string().check(...);
// TypeError: Cannot assign to read only property
```

## Solution
Implemented `z.ext` as a dedicated, mutable namespace specifically for plugins. This approach:

- ✅ Maintains discoverability through `z.ext.*` with IDE autocomplete
- ✅ Avoids any breaking changes to the core `z` object

## Usage Example

**Plugin Author:**
```typescript
import { z } from 'zod';

// Register plugin on z.ext
z.ext.cron = () => {
  return z.string().refine(
    (val) => /^(\*|([0-5]?\d)) (\*|([01]?\d|2[0-3]))/.test(val),
    { message: "Invalid cron expression" }
  );
};

// TypeScript support via declaration merging
declare module 'zod' {
  namespace z {
    namespace ext {
      function cron(): ZodString;
    }
  }
}
```

**Consumer:**
```typescript
import 'my-zod-cron-plugin'; // Side-effect import
import { z } from 'zod';

const cronSchema = z.ext.cron();
cronSchema.parse("0 0"); // ✅ Valid
cronSchema.parse("invalid"); // ❌ Throws error
```

## Test Results

- ✅ Basic plugin extension
- ✅ Multiple plugins coexisting
- ✅ Core `z` object remains unaffected


Ayush2k02:
Hey, any updates on this !

PR #5584, opened 2025-12-30
Improve locale translation approach

colinhacks:
- **Add type dictionaries, use invalid_type for instanceof**


colinhacks:
cc @pullfrog review and pay special attention to any behavioral changes in each locale. the old translations in `parsedType` or elswhere need to be accurately represented in the new translation dictionaries.

PR #5596, opened 2026-01-02
Fix/z function unknown type

### LINKED ISSUE #5080: z.function returned 'unknown' type in zod v4
I don't know why this happens, it should return type 'function' but zod returns 'unknown'

Image

this is my zod schema
Image

Victor1890:
# Support ReadonlyArray in .function() schemas

Fixes #5080

### What’s changing?
This PR makes Zod's function schemas a lot more flexible by allowing `ReadonlyArray` as an input for `.function()`. Previously, it was too strict with mutable arrays, which caused issues when passing `as const` tuples or readonly argument lists.

### Key Updates
* **Flexibility:** You can now use both `Array` and `ReadonlyArray` for function arguments.
* **Smart Inference:** Updated the `_function` implementation to handle these inputs correctly, ensuring type inference and runtime validation stay in sync.
* **Better Testing:** Swapped out the placeholder tests for a robust suite in `functions.test.ts`. This covers:
    * Sync and async function parsing.
    * Method chaining and refinements.
    * Proper error handling for invalid inputs/outputs.
    * Accurate type inference checks using `expectTypeOf`.

### Why?
Many TypeScript projects use readonly arrays for safety or performance (especially with constants). This change ensures Zod doesn't complain when your function arguments are marked as readonly.

### Demostration

Before to implement this correction
```ts
const myFunction = z.function({
  input: [z.string()],
  output: z.object({ foo: z.any() }),
});
type Foo = z.infer<typeof myFunction>; // unknown
```

After to implement this correction
image

### DIFF
--- a/packages/zod/src/v4/mini/tests/functions.test.ts
+++ b/packages/zod/src/v4/mini/tests/functions.test.ts
@@ -1,5 +1,222 @@
-import { expect, test } from "vitest";
+import { expect, expectTypeOf, test } from "vitest";
+import * as z from "zod/mini";
 
-test("z.function", () => {
-  expect(true).toEqual(true);
+const args1 = z.tuple([z.string()]);
+const returns1 = z.number();
+const func1 = z.function({
+  input: args1,
+  output: returns1,
+});
+
+test("function parsing", () => {
+  const parsed = func1.implement((arg: any) => arg.length);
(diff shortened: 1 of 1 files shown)


colinhacks:
This has been fixed for a long time. https://github.com/colinhacks/zod/pull/5170 Thanks though.

PR #5601, opened 2026-01-03
test: add edge case tests for boundary values

uchkunr:
- Add tests for string boundary cases: length(0), min(0), max(0)
- Add tests for number negative zero (-0) edge case
- Improves test coverage for edge cases that were previously untested


colinhacks:
Nice 👍

colinhacks:
Landed in Zod 4.4

PR #5604, opened 2026-01-03
[Docs] Fix order of brand method examples in api.mdx

onurtemiz:
Fixes the function parameter names for brand function.


colinhacks:
Nice catch ✨

colinhacks:
Landed in Zod 4.4

PR #5606, opened 2026-01-04
docs: fix z.custom example and reference z.templateLiteral

### LINKED ISSUE #5605: [Docs] Maybe new example for z.custom
Since z.templateLiteral is added in v4, the custom example description is not correct:

https://zod.dev/api?id=custom

> This is useful for creating schemas for types that are not supported by Zod out of the box, such as template string literals.

It would be better to either create a new example for z.custom or remove the "such as template string literals." statement to not confuse the reader for v4.

GabrielCardoso-0:
### What changed

This PR improves the `z.custom()` documentation by adding a clearer, more representative
example that demonstrates its intended use with unsupported or external types, such as
class instances.

A note was also added to clarify that `z.templateLiteral()` should be preferred when
working with template string literal schemas.

### Why

The previous documentation could be confusing, as it referenced template string literals
without clearly distinguishing when `z.custom()` is appropriate versus when a dedicated
schema like `z.templateLiteral()` should be used.

This change improves clarity without removing or altering any existing API functionality.

### Scope

- Documentation only

### Related issue

Closes #5605

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -3080,17 +3080,21 @@ computeTrimmedLengthAsync("sandwich"); // => Promise<8>
 
 ## Custom
 
-You can create a Zod schema for any TypeScript type by using `z.custom()`. This is useful for creating schemas for types that are not supported by Zod out of the box, such as template string literals.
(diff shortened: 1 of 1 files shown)


BrendanC23:
I don't think that this is the best example for `z.custom()` because `z.instanceof()` already exists. See [the docs](https://zod.dev/api#instanceof).

colinhacks:
Merged in a different change to this bit of docs that addresses these concerns. Thanks for the contribution, though!

PR #5621, opened 2026-01-10
docs: Add `enumToString` function to codec docs

jtannas:
Adds a helper function `enumToString` that demonstrates a codec for converting between an integer enum and a more readable string equivalent.

There is an extra tweak that adds type safety to the enum integers by deriving the valid indexes from the string array.

P.S. Sorry about all the diff noise from the removed trailing whitespace. You can view the changes without it at https://github.com/colinhacks/zod/pull/5621/changes?w=1


colinhacks:
Thanks, but this isn't a pattern that I'd really recommend, and it's too complex to put here imo.

PR #5623, opened 2026-01-10
fix(v4): z.coerce.date() error message not showing original input type

### LINKED ISSUE #5408: Weird error message when z.coerce.date() fails to coerce
## Summary

When using `z.coerce.date()` with a value that cannot be coerced into a valid `Date`, the error message is misleading:

```ts
z.coerce.date().parse('www');
// => ZodError: Invalid input: expected date, received Date
```

This reads as if a `Date` was provided, when in fact a **string** was provided and the coercion produced an **invalid Date**. It would be helpful if the error communicated that coercion failed and/or showed the original input type/value.

---

## Repro

```ts
import { z } from 'zod';

const schema = z.coerce.date();

schema.parse('www');          // Invalid input: expected date, received Date
schema.parse('not-a-date');   // same
schema.parse('2025-13-99');   // same (invalid calendar date)
```

(Using Zod v4 — reproducible in a fresh project.)

---

## Actual behavior

* Error message: `Invalid input: expected date, received Date`
* This suggests the input was a `Date`, which is incorrect. The input was a string and the coercion produced `new Date("www")` → `Invalid Date`.

---

## Expected behavior

A clearer message, for example (any of these would be great):

* `Invalid date: could not coerce value "www" into a valid Date`
* `Invalid input: expected a date, received string "www" (coercion failed)`
* `Invalid input: expected valid Date after coercion, got Invalid Date (from "www")`

Key points:

* Indicate original input type/value (string `"www"`).
* Indicate that **coercion** happened and failed because the result is an **Invalid Date**.

---

## Why this matters

* The current message can send users on a wild goose chase, thinking they’re passing a `Date` incorrectly.
* Clearer messaging speeds up debugging, especially in schemas with nested objects where the path matters.

---

## Possible directions (implementation ideas)

* In the `coerce.date()` pipeline, after coercion (`new Date(value)`), detect `Number.isNaN(date.getTime())`:

  * If `Invalid Date`, raise a custom issue code (e.g., `"invalid_date"`) that includes:

    * the original input type and a safe preview of the value,
    * a hint that coercion failed.
* Alternatively, adjust the error constructor for `ZodDate` to include a special branch when the input is already a `Date` but invalid vs. when the input was coerced from a non-Date type and resulted in `Invalid Date`.
* (Nice-to-have) Add a short value preview with truncation and JSON-stringify to avoid giant messages.

---

## Environment

* Zod: v4.x
* Node: v18+/v20 (reproducible)
* TS: 5.x

---

## Minimal code sandbox

```ts
import { z } from 'zod';
z.coerce.date().parse('www');
```

Thanks for the awesome library! 🙌

jake158:
## Summary
Fixes misleading error message from `z.coerce.date()` when invalid inputs are provided. Closes #5408

```ts
z.coerce.date().parse('www');
// BEFORE:
// => ZodError: Invalid input: expected date, received Date
// NOW:
// => ZodError: Invalid input: expected date, received string
```

## Problem
`$ZodDate` parse function (`packages/zod/src/v4/core/schemas.ts:1537`) did not preserve the original input during coercion. Coercion overwrote `payload.value`, which was later pushed to the `input` field of `payload.issues`.

## Solution
- Modified `$ZodDate` parse function to preserve original input before coercion
- Added tests for error messages with various invalid inputs (`packages/zod/src/v4/classic/tests/date.test.ts`)

### DIFF
--- a/packages/zod/src/v4/classic/tests/date.test.ts
+++ b/packages/zod/src/v4/classic/tests/date.test.ts
@@ -60,3 +60,57 @@ test("min max getters", () => {
   expect(maxCheck.maxDate).toEqual(benchmarkDate);
   expect(maxCheck.max(beforeBenchmarkDate).maxDate).toEqual(beforeBenchmarkDate);
 });
+
+test("invalid string error message", () => {
+  const schema = z.coerce.date();
+  const result = schema.safeParse("www");
+
+  expect(result.success).toEqual(false);
+  expect(result.error!.issues).toMatchInlineSnapshot(`
+    [
+      {
+        "code": "invalid_type",
+        "expected": "date",
+        "message": "Invalid input: expected date, received string",
+        "path": [],
+        "received": "Invalid Date",
+      },
+    ]
+  `);
+});
+
+test("invalid number error message", () => {
+  const schema = z.coerce.date();
+  const result = schema.safeParse(Number.NaN);
+
+  expect(result.success).toEqual(false);
+  expect(result.error!.issues).toMatchInlineSnapshot(`
+    [
+      {
+        "code": "invalid_type",
+        "expected": "date",
+        "message": "Invalid input: expected date, received NaN",
+        "path": [],
+        "received": "Invalid Date",
+      },
+    ]
+  `);
+});
+
+test("invalid date object error message", () => {
+  const schema = z.coerce.date();
+  const result = schema.safeParse(new Date("invalid"));
+
+  expect(result.success).toEqual(false);
+  expect(result.error!.issues).toMatchInlineSnapshot(`
+    [
+      {
+        "code": "invalid_type",
+        "expected": "date",
+        "message": "Invalid input: expected date, received Date",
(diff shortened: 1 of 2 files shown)


jake158:
It seems that this fix goes against the current codebase pattern. Coerce functions pass `payload.value` as `input` after it's been overwritten by coercion.

To me, the input field should represent - "what did the user actually provide?" rather than - "what intermediate value did we transform it to?"

Example:
```ts
z.coerce.number().parse("abc")
```
```
ZodError: [
  {
    "expected": "number",
    "code": "invalid_type",
    "received": "NaN",
    "path": [],
    "message": "Invalid input: expected number, received NaN"
  }
]
```

The fact non-coerce functions return input as the original value, while coerce functions return the intermediate form, in my eyes is a little strange.

I'd be happy to implement the fix differently or to change the pattern across the codebase, whatever is preferred (or whatever you suggest). @colinhacks

PR #5624, opened 2026-01-12
feat: Loosen `z.strictObject()` inside union

chiichen:
similar to #5587


colinhacks:
That other PR was specifically to align Zod's behavior better with TypeScript's intersection behavior. Same logic doesn't apply for unions I'm afraid.

PR #5626, opened 2026-01-12
Fix $ZodErrorTree type: Allow undefined in error items array

cmfcmf:
As per the docs, the `items` property of an error tree may include `undefined`.


cmfcmf:
It looks like a test is failing - I'll take a look.

cmfcmf:
Test failure is fixed :)

PR #5628, opened 2026-01-13
fix: typo in codec.test.ts file

bgk614:
There is a small typo in one of the test names.


colinhacks:
cc @pullfrog check for similar typos and fix

PR #5631, opened 2026-01-13
Fix JIT checks with object containing promise results

### LINKED ISSUE #5630: Object JIT erros when an async schema is found.
In this example the .$default() function makes the field accept `undefined` input and if it's `undefined` and a database connection was provided the check returns a promise result instead.

```ts
const User = z.table("user").fields({
  id: z.tuple([z.string(), z.string().$default(surql`rand::ulid()`)]),
});

const data = User.dto().decode(
  {
    id: User.record().fromId(["client"]),
  },
  { db: surreal },
);
```

Image

msanchezdev:
Fixes #5630

### DIFF
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -1935,6 +1935,7 @@ export const $ZodObjectJIT: core.$constructor<$ZodObject> = /*@__PURE__*/ core.$
         return `shape[${k}]._zod.run({ value: input[${k}], issues: [] }, ctx)`;
       };
 
+      doc.write(`const promises = [];`);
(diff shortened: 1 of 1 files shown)


msanchezdev:
No review yet?

colinhacks:
Closing this. The JIT fastpass deliberately doesn't handle async — `parseAsync` already bypasses it and goes through `$ZodObject.parse`, which handles per-field promises correctly. All this PR actually improves is the error you get when you forget to call `parseAsync` on a schema with an async field, and that's not worth the bundle-size cost on a hot path we've worked to keep lean. The current `TypeError` plus a stack trace into the compiled body is enough signal for that case. There were also some minor perf regressions in the benchmarks.

PR #5632, opened 2026-01-14
fix: preserve context immutability in parse functions

### LINKED ISSUE #5466: Mutation of ctx param is unintuitive
Lines such as this in the zod codebase result in mutation of the passed context object.
```const ctx = _ctx ? Object.assign(_ctx, { direction: "backward" as const }) : { direction: "backward" as const };```

This causes weird issues if the same context object is used in multiple calls.

E.g.

```
const myOptions = { reportInput: true } as const;
z.decode(stringToDateCodec, someDateString, myOptions);
z.decode(stringToDateCodec, someOtherDateString, myOptions);
```

In my particular case it was causing a later decode call to run in reverse which passed a string to the encode function instead of a date.

The reason I am sharing an options object is because i'm trying to mimic global reportInput: true by simply passing it explicity in each call.

### PR DESCRIPTION
**Pull request by bgk614**
## Summary
Fixes mutation of user-provided `ctx` parameter in parse functions by replacing `Object.assign` with object spread syntax.

## Problem
Using `Object.assign(_ctx, {…})` mutates the original context object passed by the user.

**Before Example:**
```ts
const ctx: schemas.ParseContextInternal = _ctx ? Object.assign(_ctx, { async: false }) : { async: false };
```
**After Example:**
```ts
const ctx: schemas.ParseContextInternal = _ctx ? { ..._ctx, async: false } : { async: false };
```

## Affected Functions
- _parse
- _parseAsync
- _safeParseAsync
- _encode
- _encodeAsync
- _safeEncode
- _safeEncodeAsync

Note: `_safeParse` already used spread operator.

## Changes
- `packages/zod/src/v4/core/parse.ts` - Replace Object.assign with spread operator
- `packages/zod/src/v4/classic/tests/codec.test.ts` - Add context immutability tests

Fixes #5466

### DIFF
--- a/packages/zod/src/v4/classic/tests/codec.test.ts
+++ b/packages/zod/src/v4/classic/tests/codec.test.ts
@@ -1,4 +1,4 @@
-import { expect, expectTypeOf, test } from "vitest";
+import { describe, expect, expectTypeOf, test } from "vitest";
 import * as z from "zod/v4";
 
 const isoDateCodec = z.codec(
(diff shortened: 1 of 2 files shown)


colinhacks:
Good fix. The object.assign approach benchmarked faster at some point, but I'm not able to reproduce that. And using spread here is a lot faster (30ish%) when an options object is actually specified. Since there seems to be no more performance advantage I'll happily take this correctness win.

colinhacks:
Landed in Zod 4.4

PR #5636, opened 2026-01-14
fix: check `refSeen.def` when flattening ref

ahmedriad1:
Fixes #5635

PR #5638, opened 2026-01-15
Avoid non null assertion in toJSONSchema

samchungy:
Howdy doody, this little line is throwing and reported via one of my repos.

Minimal Repro:

```ts
import * as z from 'zod';

console.log(
  z.toJSONSchema(
    z
      .string()
      .meta({ id: 'foo', description: 'foo' })
      .describe('bar')
      .optional(),
  ),
);
```

To be honest, I haven't taken too much time to understand what this bit of code is doing but adding a conditional before seems to work.

Resolves https://github.com/samchungy/fastify-zod-openapi/issues/352

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2491,6 +2491,21 @@ test("_ref", () => {
       "type": "string",
     }
   `);
+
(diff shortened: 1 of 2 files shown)


dan-lee:
We are also seeing this error, would be great if fixed.

Found similar PRs: https://github.com/colinhacks/zod/pull/5639 https://github.com/colinhacks/zod/pull/5644 https://github.com/colinhacks/zod/pull/5636

PR #5639, opened 2026-01-15
Add undefined check to refSeen.def

Matt-Ord:
Fix for a bug we found when updating zod
"TypeError: Cannot use 'in' operator to search for 'default' in undefined


colinhacks:
Fixed in https://github.com/colinhacks/zod/pull/5638, merged in v4.3.6

colinhacks:
Fixed in https://github.com/colinhacks/zod/pull/5638, merged in v4.3.6

PR #5644, opened 2026-01-18
Add safety check to handle undefined refSeen.def

brunolemos:
Fix the following issue after upgrading from `4.2.1` to `4.3.5`:

```
node_modules/.pnpm/zod@4.3.5/node_modules/zod/v4/core/to-json-schema.cjs:267
                    if (key in refSeen.def && JSON.stringify(schema[key]) === JSON.stringify(refSeen.def[key])) {
                            ^
TypeError: Cannot use 'in' operator to search for 'id' in undefined
    at flattenRef (node_modules/.pnpm/zod@4.3.5/node_modules/zod/v4/core/to-json-schema.cjs:267:29)
```

As per flagged by copilot bot here: https://github.com/colinhacks/zod/pull/5578#discussion_r2652580788

Confirmed this change fixes the issue for me by using `pnpm patch`.

cc @colinhacks

### DIFF
--- a/packages/zod/src/v4/core/to-json-schema.ts
+++ b/packages/zod/src/v4/core/to-json-schema.ts
@@ -409,7 +409,7 @@ export function finalize<T extends schemas.$ZodType>(
       if (refSchema.$ref) {
         for (const key in schema) {
           if (key === "$ref" || key === "allOf") continue;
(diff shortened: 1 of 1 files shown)


colinhacks:
Merged equivalent fix in https://github.com/colinhacks/zod/pull/5638 

Thanks for the PR, apologies for the inconvenience here.

PR #5646, opened 2026-01-19
Fix invalid `robots.txt`

Abdalrhman-Almarakeby:
Fix invalid `robots.txt` by adding a `User-agent` directive and allowing all paths, same as the v3 docs [`robots.txt`](https://github.com/colinhacks/zod/blob/main/packages/docs-v3/robots.txt).


colinhacks:
cc @pullfrog review

PR #5649, opened 2026-01-21
docs: fix refinements examples in api.mdx

playoffthecuff:
Replace a deprecated "message" field with an "error" in params arg of schema refine method in code examples.


colinhacks:
Landed in Zod 4.4

Issue #5593, opened 2026-01-02
discriminatedUnion fails on encode() when discriminator is a ZodCodec


MadL0rd:
## Description

When using `z.discriminatedUnion` with a discriminator field implemented as a `ZodCodec`,
`decode()` works correctly, but `encode()` fails at runtime with
`No matching discriminator`.

This happens even though:

- the types of `decode()` / `encode()` inputs are correct
- the codec itself is fully bidirectional
- the decoded domain object is valid

The issue appears to be that `discriminatedUnion` attempts to select a branch
**before encoding the discriminator field**, using the domain value instead of
the encoded (raw) value.

---

## Reproducible example

```ts
const LiteralTestCodec1 = z.codec(z.literal(1), z.literal('one'), {
    decode: (raw) => 'one' as const,
    encode: (domain) => 1 as const,
})

const LiteralTestCodec2 = z.codec(z.literal(2), z.literal('two'), {
    decode: (raw) => 'two' as const,
    encode: (domain) => 2 as const,
})

const TestUnionSchema = z.discriminatedUnion(
    'type',
    [
        z.object({
            type: LiteralTestCodec1,
            valueString: z.string(),
        }),
        z.object({
            type: LiteralTestCodec2,
            valueNumber: z.number(),
        }),
    ]
)

const domain = TestUnionSchema.decode({
    type: 1,
    valueString: 'test value',
})

const raw = TestUnionSchema.encode({
    type: 'one',
    valueString: 'test value',
})
```

---

## Actual result

`decode()` works correctly.

`encode()` throws the following error:

```json
ZodError: [
  {
    "code": "invalid_union",
    "errors": [],
    "note": "No matching discriminator",
    "discriminator": "type",
    "path": ["type"],
    "message": "Invalid input"
  }
]
```

---

## Expected behavior

Given that:

- the discriminator codec is bidirectional
- `encode()` should produce a valid raw representation

`encode()` is expected to succeed and return:

```ts
{
    type: 1,
    valueString: "test value"
}
```

---

## Notes / Workaround

At the application level, this issue can be worked around by enabling
`{ unionFallback: true }` when creating the `discriminatedUnion`.

This forces Zod to fall back to full union validation instead of the fast
discriminator-based path.

However, this behavior is surprising when combining:

- `z.discriminatedUnion`
- codecs used as discriminator fields
- `encode()` direction

---

## Possible directions

[...]

PR #5650, opened 2026-01-21
docs: warns about codec usage in discriminatedUnion

MadL0rd:
This PR adds a warning to the Codecs documentation about using `z.codec()` as a
discriminator in `z.discriminatedUnion()`.

When a discriminator is implemented as a codec, the fast discriminator path
relies on the input (`in`) schema only, which can cause `.encode()` to fail with
"No matching discriminator".

The added note highlights this limitation and documents available workarounds
(`{ unionFallback: true }` or avoiding codecs in discriminator fields).

Related issue: https://github.com/colinhacks/zod/issues/5593

### DIFF
--- a/packages/docs/content/codecs.mdx
+++ b/packages/docs/content/codecs.mdx
@@ -108,6 +108,12 @@ payloadSchema.decode({
 }); // => { startDate: Date }
 ```
 
+<Callout type="warn">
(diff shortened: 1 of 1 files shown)


khanbilal732 (approved):

colinhacks:
This has been fixed.

khanbilal732 (approved):

PR #5654, opened 2026-01-22
fix: superRefine() errors surfacing on optional() fields

### LINKED ISSUE #5653: superRefine() errors no longer surfaced if used in combination with optional() on an object() field that is not defined
When upgrading from zod 4.2.1 to 4.3.0 I ran into a behavioral change that broke a test. It looks like the behavioral change is introduced in the `Add .exactOptional()` PR (#5589). The `handlePropertyResult` function in packages/zod/src/v4/core/schemas.ts no longer returns errors if `isOptionalOut && !(key in input)`. I'm just wondering if this was meant to change how `optional()` worked, or meant solely for `exactOptional()`, as I did not see any mention of this in the release notes.

I have a test suite that was used to ensure a schema validation wrapper class functioned the same when we switched from `convict` to `zod`. One of the tests calls `superRefine()` on an optional field of an object, expecting the function to throw and add an issue to `ctx`, even if the field is not defined on the object.

Here is a simplified example. In 4.2.1 the following schema would throw if we parse an object where the field is not set. In 4.3.0 schema.parse() no longer throws.
```js
const schema = z.object({
    field: z.any().optional().superRefine((val, ctx) => {
        ctx.addIssue({
            code: 'custom',
            message: 'This validation should fail',
        });
    })
});

const result = schema.parse({}); // throws in 4.2.1 only, not in 4.3.0
```

If I set field to undefined the schema will throw:
```js
const result = schema.parse({ field: undefined }); //throws
```

If I parse the field's schema directly (not wrapped in an object) the error is thrown:
```js
const worksOutsideObject = z.any().optional().superRefine((val, ctx) => {
    ctx.addIssue({
        code: 'custom',
        message: 'This validation should fail',
    });
});

const result = worksOutsideObject.parse(undefined); // throws
```

This works the same for other functions that throw errors, like `refine()` and `check()`.

I think this new behavior makes sense, I just want to confirm that it's intentional and will not change. If so, I can send over a PR to update the documentation if you are interested.

Thanks

tushargr0ver:
Fixes #5653.

This PR updates ZodOptional to ensure validation chains (like superRefine) continue execution even if the value is undefined, correcting a regression where errors were swallowed.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1857,9 +1857,9 @@ export interface ZodExactOptional<T extends core.SomeType = core.$ZodType>
 export const ZodExactOptional: core.$constructor<ZodExactOptional> = /*@__PURE__*/ core.$constructor(
   "ZodExactOptional",
   (inst, def) => {
-    core.$ZodExactOptional.init(inst, def);
+    core.$ZodExactOptional.init(inst as any, def);
     ZodType.init(inst, def);
-    inst._zod.processJSONSchema = (ctx, json, params) => processors.optionalProcessor(inst, ctx, json, params);
+    inst._zod.processJSONSchema = (ctx, json, params) => processors.optionalProcessor(inst as any, ctx, json, params);
 
     inst.unwrap = () => inst._zod.def.innerType;
   }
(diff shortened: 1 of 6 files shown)


colinhacks:
Ultimately, this is behaving as intended. When you call `.optional()` on a schema, it means that omitting that field from any surrounding object is a valid state. That property bubbles up through the `.superRefine()` call. You can't add additional refinements on top of a absent value. There are other APIs for different types of optionality:

```ts
.or(z.undefined)  // | undefined
.exactOptional()  // adds question mark, does not allow literal `undefined`
```

These various kinds of optionality can be composed however you like to craft this logic. But if you want to add special refinements that only apply when a given key is present or not present on an object, that refinement needs to live at the object level.

PR #5684, opened 2026-02-03
Remove deprecated downlevelIteration option

RyanCavanaugh:
This option is deprecated in TypeScript 6.0. It doesn't do anything when `target` is `es2020`, so can be safely removed from this config file.


colinhacks:
Landed in Zod 4.4

PR #5687, opened 2026-02-05
fix(v4): handle multi-digit exponents in floatSafeRemainder

shakecodeslikecray:
## Summary

The regex in `floatSafeRemainder` only matches single-digit exponents, causing `multipleOf` validation to incorrectly fail for steps with exponents >= 10 (e.g., `1e-10`, `1e-15`).

## The Bug

```typescript
// Before: \d? matches 0 or 1 digit
if (stepDecCount === 0 && /\d?e-\d?/.test(stepString)) {
  const match = stepString.match(/\d?e-(\d?)/);
```

For `step = 1e-10`:
- Regex matched `"1e-1"` instead of `"1e-10"`
- Captured exponent: `"1"` instead of `"10"`
- `stepDecCount` set to `1` instead of `10`
- `toFixed(1)` rounds both values to `"0.0"`
- `stepInt` becomes `0`, remainder becomes `NaN`
- Valid values incorrectly rejected

## The Fix

Changed `\d?` to `\d+` to capture multi-digit exponents:

```typescript
// After: \d+ matches 1 or more digits
if (stepDecCount === 0 && /\d?e-\d+/.test(stepString)) {
  const match = stepString.match(/\d?e-(\d+)/);
```

## Reproduction

```typescript
const step = 1e-10;
const regex = /\d?e-(\d?)/;
const match = step.toString().match(regex);
console.log(match); // ["1e-1", "1"] - captures only first digit
```

## Discovery

This bug was discovered by [whiterose](https://github.com/shakecodeslikecray/whiterose), an AI-powered bug hunter that uses LLM-based static analysis.

---

Scanned with [whiterose](https://github.com/shakecodeslikecray/whiterose)

### DIFF
--- a/packages/zod/src/v4/classic/tests/number.test.ts
+++ b/packages/zod/src/v4/classic/tests/number.test.ts
@@ -153,6 +153,22 @@ test(".multipleOf() with negative divisor", () => {
   expect(() => schema.parse(7.5)).toThrow();
 });
 
+test(".multipleOf() with scientific notation (multi-digit exponents)", () => {
+  // Regression test for https://github.com/colinhacks/zod/pull/5687
+  // The regex was using \d? which only matches single-digit exponents
+  const schema = z.number().multipleOf(1e-10);
+
+  // These should all pass - they are valid multiples of 1e-10
+  expect(schema.parse(1e-10)).toEqual(1e-10);
+  expect(schema.parse(5e-10)).toEqual(5e-10);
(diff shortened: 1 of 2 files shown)


colinhacks:
Thanks!

shakecodeslikecray:
> Thanks!

Did it using [Whiterose](https://github.com/shakecodeslikecray/whiterose)

Thank you for your response!

colinhacks:
Landed in Zod 4.4

Issue #5641, opened 2026-01-16
zod/mini is still not tree-shaken correctly inside NextJS


DoroGi:
I saw there are a number of cases when this issue has been figured out, but I was still seeing the issue and I managed to create a clean NextJS example to reproduce it.
I think that NextJS interrupts the tree-shaking after a certain number of nested imports. Maybe Zod internal folder structure is not helping Next to tree shake correctly.

### Link to the code that reproduces this issue

https://github.com/DoroGi/nextjs-zod-tree-shaking-issues/tree/main

I created this repo running: `npx create-next-app@latest my-app --yes` as suggested by NextJS "getting started".

Then, in a second commit, I installed Zod and created a simple client component called Zodder. This component simply imports Zod using `import * as z from zod/mini` and parses a string.

### To Reproduce

- npm i
- npm run build
- npx next experimental-analyze
- navigate to http://localhost:4000/

### Current vs. Expected behavior

You will see that the bundle contains all Zod locales, which are unused and very heavy;
I expect they should be tree-shaken away.

### Additional context

A few workarounds to make it work, that may be useful to debug:

To try this methods, remember to run each time the line below, so you update the .next and you can check whether the locales are present in the bundle or not:

npm run build && npx next experimental-analyze

1. Importing zod instead of zod/mini in the zodder component seems to work. Locales are not present anymore.
2. Going into node_modules/zod/package.json and editing the ./mini exports configuration to this seems to work:

```
"./mini": {
  "@zod/source": "./src/mini/index.ts",
  "types": "./mini/index.d.cts",
  "import": "./v4/mini/external.js",   <-- only line changed
  "require": "./mini/index.cjs"
},
```

I did this because ./mini in the package.json links to `zod/mini/index.js`, which just imports `zod/v4/mini/index.js`, which just imports `zod/v4/mini/externals.js`.

3. Removing the line `export { z };` in `zod/v4/mini/index.js` seems to work. Notice that the import line when using the standard `import * as z from 'zod'` (which is just `zod/index.js` -> `./v4/classic/external.js`) has a similar line in `zod/index.js`, but it's not causing any issue apparently.

### Related issues

This seems to have been talked a lot in the zod issues. Many seem to have found their solution, the only remaining ones are the issues related to NextJS

https://github.com/colinhacks/zod/issues/4433
https://github.com/colinhacks/zod/issues/4572
https://github.com/colinhacks/zod/issues/4637
https://github.com/colinhacks/zod/issues/4798
[...]

Issue #4433, opened 2025-05-20
v4: bundle size increased significantly 


Codexphere92:
I am not sure if I am doing something wrong but using zod/v4 shows this with the import cost extension in vscode and my "first load of js" went up by 30kb in my nextjs app.

![Image](https://github.com/user-attachments/assets/3bfff44f-7a79-433b-991b-b81fb5801977)

Codexphere92:
@coderrshyam didn't you run into a similar issue?

Codexphere92:
Hi @colinhacks, let me know if you want me to provide a reproducible example for my next js app or if you are aware of my problem, thanks!!

coderrshyam:
> [@coderrshyam](https://github.com/coderrshyam) didn't you run into a similar issue?

Yes! @Codexphere92

coderrshyam:
> I am not sure if I am doing something wrong but using zod/v4 shows this with the import cost extension in vscode and my "first load of js" went up by 30kb in my nextjs app.
> 
> ![Image](https://github.com/user-attachments/assets/3bfff44f-7a79-433b-991b-b81fb5801977)

@Codexphere92 I believe the issue is caused by multiple versions of the same package being bundled together—for example, `zod`, `zod/v3`, `zod/v4`, and `zod/v4/mini`.

## To avoid this, I recommend the following:

- Publish version 4 separately under `zod@v4`.

- Keep version 3 under a version constraint like `<3.25`.

- For `zod/v4/mini`, consider publishing it as a separate package, such as `@zod/mini`.

Codexphere92:
I am sure @colinhacks had reasons for doing this but yes the bundle size is messed up right now

WtfJoke:
He explained his reasons for publishing v4/v3 in the zod v3.2.5 here: https://zod.dev/v4?id=versioning

>I believe the issue is caused by multiple versions of the same package being bundled together—for example, zod, zod/v3, zod/v4, and zod/v4/mini.

I think the idea here is to stick to one of those: 
- zod/v3
- zod/v4
- zod/v4/mini

piotr-czerwik:
I'm actually using "module": "ESNext" + "moduleResolution": "Bundler"

EDM115:
@Codexphere92 well I guess this might be (in part) the reason why : https://github.com/colinhacks/zod/issues/4477#issuecomment-2902605756

Codexphere92:
[...]

PR #5689, opened 2026-02-05
fix: add sideEffects: false to stub package.json files

jesse-holden:
## Summary                                                                                                                                                                    
  - Adds `sideEffects: false` to stub package.json files to enable proper tree-shaking

## Context
The stub package.json files introduced in #5222 were missing the `sideEffects` field, preventing bundlers from tree-shaking unused submodules.

Related to:

- https://github.com/colinhacks/zod/issues/4433
- https://github.com/colinhacks/zod/issues/5641

## Reproduction

https://stackblitz.com/edit/vitejs-vite-dqpihhpx?file=src%2Findex.ts

When bundling for the web with webpack 5.105.0 using the `import * as z from "zod";`  syntax:

### DIFF
--- a/scripts/write-stub-package-jsons.ts
+++ b/scripts/write-stub-package-jsons.ts
@@ -7,7 +7,8 @@ const STUB_PACKAGE_JSON_CONTENT = `{
   "type": "module",
   "main": "./index.cjs",
   "module": "./index.js",
-  "types": "./index.d.cts" 
+  "types": "./index.d.cts",
+  "sideEffects": false 
 }
 `;
 


jesse-holden:
Added note here, the bundle size noticeably increased between v4.2.0 (57.4kb) and v4.2.1 (95.2kb) so also relates to https://github.com/colinhacks/zod/commit/5b5b129315fbc94a3b0d6244185eaeefcbe438d1

Reverting the change to `from-json-schema.ts` alone drops the bundle size to 60.7kb without needing to add the `sideEffects: false` change to `write-stub-package-jsons.ts`, so avoiding a hard dependency and reverting to circular import does work better for bundlers that I've tested.

dorbn4-team:
@colinhacks thank you for merging the PR and also for this amazing package. Do you think it would be possible to release this fix soonish so that we can reduce our bundle sizes again? 😅🫶

osdiab:
fyi tried doing this with a pnpm patch since it's not released yet.

With NextJS, if you use `import { z } from "zod"` it still does not tree shake, but if you do `import * as z from "zod"` then it does appear to treeshake properly (i don't see all the locales being included anymore).

Alternatively, even without this change, if you add zod to `experimental.optimizePackageImports` and do the above import strategy, then it achieves the same end result.

colinhacks:
Landed in Zod 4.4

PR #5699, opened 2026-02-08
fix: add missing addIssue to transform context

### LINKED ISSUE #5678: `transform` callback `ctx.addIssue` not in Typescript
According to the source (and various articles) I can see that a transform callback receives a context object containing an `addIssue` method:

```ts
const schema = z.transform((value, ctx) => {
  if(isNotValid(value)) {
    ctx.addIssue({
      code: "invalid_element",
      message: "Not valid",
    });
  }
  return value;
});

schema.parse(someValue);
```

This work at runtime. However, in Typescript I get an error telling me `addIssue` doesn't exist.

Is this method deprecated? It certainly wasn't removed, and TS doesn't mark it as deprecated, but rather absent completely.

Confirmed in Zod v4.2.1 and v4.3.6 (presumably between as well).

-----

Resolution:

A. If the method is deprecated, the Typescript needs to be updated to indicate that it's deprecated, but not absent (until it is in fact removed).

B. If the method is *not* deprecated, it needs to be added to the Typescript definition.

F-A-N-D-E:
This PR adds the missing addIssue method to the transform callback context, resolving the issue where manual issue reporting was unavailable in transformations.

Closes #5678

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1814,7 +1814,7 @@ export const ZodTransform: core.$constructor<ZodTransform> = /*@__PURE__*/ core.
 );
 
 export function transform<I = unknown, O = I>(
-  fn: (input: I, ctx: core.ParsePayload) => O
+  fn: (input: I, ctx: core.$RefinementCtx) => O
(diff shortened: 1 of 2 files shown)


colinhacks:
Landed in Zod 4.4

PR #5700, opened 2026-02-09
fix(v4): JSON schema min/max intersection for draft-04 and openapi-3.0

ebroder:
When both minimum and exclusiveMinimum exist (e.g., from .int().positive()), the conversion now picks the more restrictive constraint. Previously, the inclusive minimum would overwrite the exclusive one, producing incorrect output like `minimum: -9007199254740991` instead of `minimum: 0`.

Same fix applied for maximum/exclusiveMaximum.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -647,6 +647,47 @@ describe("toJSONSchema", () => {
     `);
   });
 
+  test("number constraints intersection draft-04", () => {
+    // When both minimum (from .int()) and exclusiveMinimum (from .positive()) exist,
+    // the more restrictive constraint should be used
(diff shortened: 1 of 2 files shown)


colinhacks:
@pullfrog refactor to be a little tighter. fewer intermediate vars.  ensure tests still pass.

colinhacks:
Solid catch. With both `minimum` and `exclusiveMinimum` (and the same for `maximum` / `exclusiveMaximum`) coming out of things like `.int().positive()`, the inclusive bound was clobbering the exclusive one on `draft-04` / `openapi-3.0`. Bot tightened up the `numberProcessor` intersection logic on top — collapsed the per-target branches into a single `exMin`/`exMax`/`legacy` form and dropped the intermediate booleans. Thanks for digging into this!

colinhacks:
Landed in Zod 4.4

PR #5701, opened 2026-02-09
fix(v4): avoid array item parsing for any/unknown validator

kamikazePT:
This pull request introduces an optimisation to array parsing, specifically for arrays of `unknown` and `any` types. The improvement is that per-item parsing is skipped for these "noop" validators.

### Context
I had a use case, where I checked if a field in a response was an array, the item parsing was postponed to another step, to manually check each item to support partially valid arrays. I was using `z.array(z.unknown())` with the expectation of being a zod equivalent to `Array.isArray` check.

### Changes
Added logic in `$ZodArray` (in `schemas.ts`) to skip per-item parsing when the element schema is `unknown` or `any`, directly returning the input array as the parsed value.

### Testing
Introduced a new test in `array.test.ts` to verify that per-item parsing is skipped for arrays of `unknown` and `any`, using spies to ensure the schema's `run` method is not called.

### DIFF
--- a/packages/zod/src/v4/classic/tests/array.test.ts
+++ b/packages/zod/src/v4/classic/tests/array.test.ts
@@ -1,4 +1,4 @@
-import { expect, expectTypeOf, test } from "vitest";
+import { expect, expectTypeOf, test, vi } from "vitest";
 import * as z from "zod/v4";
 
 test("type inference", () => {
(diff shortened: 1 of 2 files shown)


kamikazePT:
@colinhacks a comment on the close would be nice 👀

Issue #5390, opened 2025-10-28
`z.intersection()` compiles to unsatisfiable schema


za-creature:
The intersection of two objects

```js
z.toJSONSchema(z.intersection(
    z.object({foo: z.boolean()}),
    z.object({bar: z.boolean()}),
))
```

compiles to:

```json
{ "allOf": [{
    "type": "object",
    "properties": { "foo": { "type": "boolean" } },
    "required": ["foo"],
    "additionalProperties": false
}, {
    "type": "object",
    "properties": { "bar": { "type": "boolean" } },
    "required": [ "bar" ],
    "additionalProperties": false
}]}
```

where `additionalProperties: false` prevents `{ foo: true, bar: false }` from matching at least one of the rules.

`.looseObject` works around this but isn't strict enough for my taste so for now I'm stuck with extending union cases

Maybe [this](https://json-schema.org/draft/2020-12/json-schema-core#name-unevaluatedproperties) helps?

za-creature:
> This matches Zod's output type behavior

```ts
let Intersection = z.intersection(
    z.object({foo: z.boolean()}),
    z.object({bar: z.boolean()}),
)
let obj : z.infer<typeof Intersection> = { foo: true, bar: false }
```

Remember when GPUs were used to make kids happy by rendering triangles?

PR #5702, opened 2026-02-10
fix(v4): allOf + additionalProperties: false rejects valid documents (#5390)

ebroder:
In JSON Schema, additionalProperties only considers properties in the same schema object, not sibling allOf entries. This made intersections of strict objects unsatisfiable.

For draft-2020-12 (when all entries are strict): replace per-entry additionalProperties: false with unevaluatedProperties: false on the wrapper. For older drafts (or mixed strict/loose): add {} placeholders for sibling property names.


colinhacks:
@pullfrog fresh review

percymcn:
The diagnosis in the description holds up, and this still reproduces on **4.4.3**: `z.toJSONSchema(z.intersection(A, B))` (and `A.and(B)`) emits two closed branches that each forbid the other's required property, so no object satisfies the schema — while `z.intersection(A, B).safeParse({ ... })` accepts it. The emitted schema contradicts zod's own parser.

One thing I could add that isn't in the thread: I measured both encodings this PR can produce against the three **constrained-decoding** libraries, since that is where a large share of `toJSONSchema()` output ends up (vLLM / SGLang structured output call these directly). Versions: `xgrammar` 0.2.5, `outlines-core` 0.2.14, `lm-format-enforcer` 0.11.3. Schema = `A = {reasoning: string}`, `B = {answer: string}`, both strict.

Legal instance `{reasoning, answer}` must be accepted; `{reasoning}` and `{reasoning, answer, zz}` must not — i.e. `Y . .`

| emitted form | xgrammar | outlines-core | lm-format-enforcer |
|---|---|---|---|
| 4.4.3 today (two closed branches) | `Y Y Y` | `. . .` | `Y . .` |
| this PR, `unevaluatedProperties: false` | `Y Y Y` | `. . .` | `Y . .` |
| this PR, `{}` placeholders (draft-07 / mixed) | `Y Y Y` | `. . .` | **error** |
| single merged object | `Y . .` | `Y . .` | `Y . .` |

Two things fall out of that:

**1. Neither branch changes the outcome on xgrammar or outlines-core.** Both ignore or refuse `allOf` regardless of how the closedness is spelled — xgrammar compiles a grammar with the intersection dropped (it warns `Support for allOf with multiple options is still ongoing`), and outlines-core's `build_regex_from_schema` produces a guide that matches nothing. So this PR makes the schema satisfiable for a *validator*, and the constrained-decoding path is unchanged: the field is still unconstrained on one engine and ungenerable on the other. That's not an argument against merging it — it's just worth knowing the fix doesn't reach that consumer.

**2. The `{}`-placeholder branch is a regression for lm-format-enforcer**, and that branch is the one used for `target: "draft-07"` and for mixed strict/loose on 2020-12. It emits `"name": {}` as a property schema, and lm-format-enforcer raises on a bare `{}` under `properties`:

```python
from lmformatenforcer import JsonSchemaParser
JsonSchemaParser({"type":"object","properties":{"name":{"type":"string"}},
                 "required":["name"],"additionalProperties":False})   # builds
JsonSchemaParser({"type":"object","properties":{"name":{}},
                 "required":[],"additionalProperties":False})          # Exception: Unsupported type None
```

(The parser is lazy, so it has to be driven character by character to surface this — constructing it is not enough.) Today those users get a working request; after this branch they get an exception at parser construction.

If you want a form all three honour, it is the one the intersection actually *means* when every entry is a plain object: a single object with the union of `properties`, the union of `required`, and `additionalProperties: false`. That is only sound when the entries are objects and no property name is declared on two entries with different schemas — a `oneOf` entry like the one in the new discriminated-union test can't be flattened that way — so it would have to be a narrow case rather than a replacement for either branch here.

**Test-design note:** the added tests are inline snapshots of the emitted JSON, and an assertion that a JSON Schema validator accepts `{name, age}` passes for *both* branches, so nothing in the suite distinguishes them. If it's worth guarding the placeholder case specifically, the discriminating assertion is that no property schema in the output is an empty object.

Scope: I have no vLLM/SGLang instance here, so this is the library layer those servers call, not an observation of a running server.

ebroder:
Thanks for running these. I reproduced the whole table locally and it matches (xgrammar 0.2.5 needs a cp310 venv on macOS, but all twelve cells agree).

Two things I'd push back on.

The lm-format-enforcer breakage isn't specific to this PR. It raises `Unsupported type None` on any property schema that doesn't declare a `type`, and zod already emits a bare `{}` for `z.any()` and `z.unknown()`. So `z.object({ x: z.any() })` fails there today with no `allOf` anywhere. This widens an existing incompatibility rather than introducing one, and what it replaces for those users is a schema no validator can satisfy at all.

On tests: the five added cases are all inline snapshots, which pin the emitted JSON exactly, `{}` placeholders and all, so changing either encoding fails them. There's also no validator assertion in the suite to be encoding-agnostic, because there's no JSON Schema validator here to draw on: `@seriousme/openapi-schema-validator` checks OpenAPI documents rather than instances, and ajv is present only transitively through it (not resolvable under pnpm). Asserting on instances would mean promoting ajv to a direct devDependency, which seemed like a wider call than this PR.

On the merged-object suggestion, agreed that it's the most portable encoding, but it only applies when every entry is a plain object with no composition keywords, no `$ref`, and no overlapping property names, so it'd be a third strategy rather than a replacement for either branch. Happy to open an issue for it separately.

None of this seems like it changes the case for the PR as it stands, but say the word if you'd rather I take it in a different direction.

colinhacks:
Closing in favor of #6461, which fixes the same bug by folding the intersection into the single object it describes instead of rewriting the emitted `allOf`.

The diagnosis here is right, and the encoding works for the plain case. What decided it: rewriting each member means cloning it, and def extraction swaps a schema for its `$ref` by mutating that schema in place — so a member carrying an id loses its reference and ships its internal `id` as a real keyword, which ajv then refuses to compile at all. The mixed strict/loose case also still rejects a value the parser accepts.

Thanks for the work on this one 👍

PR #5708, opened 2026-02-11
fix(v4): treeify error nested union bug

dstashevskyi:
### Problem

In `treeifyError`, when recursing into nested `invalid_union`, `invalid_key`, or `invalid_element` issues, the function passed only `issue.path` to the recursive call — discarding the `path` accumulated from parent recursions. This caused nested union errors to lose their parent path context, resulting in incorrect (shallow) paths in the output tree.

For example, given a schema like `z.object({ a: z.union([z.string(), z.number()]) })`, if the union inside property `a` failed, the inner issues' paths would not include the parent `"a"` segment, so errors would be placed at the root of the tree instead of under `properties.a`.

### Fix

In the `processError` inner function of `treeifyError`, changed the recursive calls to prepend the accumulated `path`:

```diff
-  issue.errors.map((issues) => processError({ issues }, issue.path));
+  issue.errors.map((issues) => processError({ issues }, [...path, ...issue.path]));

-  processError({ issues: issue.issues }, issue.path);
+  processError({ issues: issue.issues }, [...path, ...issue.path]);
```

This matches the pattern already used when computing `fullpath` for leaf issues (`const fullpath = [...path, ...issue.path]`) and ensures the full path context is preserved through arbitrarily nested union/key/element structures.

### Affected file

- errors.ts

### DIFF
--- a/packages/zod/src/v4/classic/tests/error-utils.test.ts
+++ b/packages/zod/src/v4/classic/tests/error-utils.test.ts
@@ -80,7 +80,10 @@ test(".flatten()", () => {
 
 test("custom .flatten()", () => {
   type ErrorType = { message: string; code: number };
-  const flattened = parsed.error!.flatten((iss) => ({ message: iss.message, code: 1234 }));
+  const flattened = parsed.error!.flatten((iss) => ({
+    message: iss.message,
+    code: 1234,
+  }));
   expectTypeOf(flattened).toMatchTypeOf<{
     formErrors: ErrorType[];
     fieldErrors: {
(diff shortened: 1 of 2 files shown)


colinhacks:
@pullfrog do a fresh review

colinhacks:
@pullfrog fix the formatting changes. Our linting tests should either have caught this or auto-formatted it properly during git lifecycle.

colinhacks:
@pullfrog fix formatError too.

colinhacks:
@pullfrog resolve merge conflicts

colinhacks:
Solid catch. The `processError` inner in `treeifyError` was dropping the accumulated `path` when recursing into `invalid_union`/`invalid_key`/`invalid_element`, so nested union issues collapsed to the root instead of nesting under their parent. pullfrog extended the same fix to `formatError` in commit `a429a568` since it had the identical shape. Thanks for tracking this down 👍

colinhacks:
Landed in Zod 4.4

PR #5709, opened 2026-02-11
Add JSONType export (fixes #5200)

### LINKED ISSUE #5200: [v4] Type portability issue when an `object` containing a `json` field is run through `transform`
I'm using v4 Mini to build some validators, and it took quite a while to figure out where this was coming from. Here's some minimal code I whipped together which triggers a TypeScript error:
```ts
import { json, nullable, object, pipe, transform, type infer as _infer } from 'zod/mini'

const DataType = object({
  data: json(),
})
type DataType = _infer<typeof DataType>

export const Container = object({
  contained: pipe(
    nullable(DataType),
    transform<DataType | null>(
      (v) =>
        v ?? {
          data: '',
        },
    ),
  ),
})
```

The following error is raised for `Container`:
```
ts: The inferred type of 'Container' cannot be named without a reference to '../../../node_modules/zod/v4/core/util.cjs'. This is likely not portable. A type annotation is necessary. [2742]
```

I had seen an odd work-around in a similar but much older issue that suggested to add an import for `z` that does not need to actually be used. I added `z as _z` to the list of imports from `zod/mini` and the error disappeared.

RobinVdBroeck:
I ran into the issue described in #5200, where I got errors like the following when using inferred z.json types 

```
error TS2742: The inferred type of 'getData' cannot be named without a reference to '../node_modules/zod/v4/core/util.cjs'. This is likely not portable. A type annotation is necessary.
```

I tested that this was resolved this in my minimal reproduction repository (https://github.com/RobinVdBroeck/zod-ts2742-repro) by running `pnpm link` here and `pnpm link zod` in the reproduction. I also tested this in proprietary code which faced this issue, and it's also resolved there. (I cannot share this code)

I'm not sure how to write unit tests for this. I tried to add some tests based on https://github.com/colinhacks/zod/pull/5511, but those succeeded without any issues. This only shows up when running a typescript build, so I'm not even sure that you can test it?

### DIFF
--- a/packages/zod/src/v4/classic/external.ts
+++ b/packages/zod/src/v4/classic/external.ts
@@ -11,6 +11,7 @@ import en from "../locales/en.js";
 config(en());
 
 export type { infer, output, input } from "../core/index.js";
+export type { JSONType } from "../core/util.js";
 export {
   globalRegistry,
   type GlobalMeta,
--- a/packages/zod/src/v4/mini/external.ts
+++ b/packages/zod/src/v4/mini/external.ts
@@ -4,6 +4,7 @@ export * from "./schemas.js";
 export * from "./checks.js";
 
 export type { infer, output, input } from "../core/index.js";
+export type { JSONType } from "../core/util.js";
 export {
   globalRegistry,
   registry,


colinhacks:
@pullfrog do a fresh review

colinhacks:
Merged. Good investigation — TS2742 around inferred `z.json` types is a pain to chase, and re-exporting `JSONType` at the top level is the right minimal fix. Thanks for sorting it 👍

colinhacks:
Landed in Zod 4.4

PR #5712, opened 2026-02-13
Fix ESM type resolution in package.json exports

### LINKED ISSUE #5686: Zod incorrectly points to CommonJS declaration when used with ESM

More details:

```ts
import z from 'zod';

  name: z.string(),
});

```

We get this output:

```ts
import z from 'zod';
    name: z.ZodString;
}, z.z.core.$strip>;
```

```
    ".": {
      "import": {
        "default": "./index.js",
        "types": "./index.d.ts"
      },
      "require": {
        "default": "./index.cjs",
        "types": "./index.d.cts"
      }
    },
```

We get a correct export:

```ts
import z from 'zod';
    name: z.ZodString;
}, z.core.$strip>;
```

veeceey:
Fixes #5686

## Problem
When using Zod in an ESM project with TypeScript's `module: "nodenext"`, the type declarations were resolving incorrectly. The `package.json` exports field was pointing to CommonJS declaration files (`index.d.cts`) for both import and require conditions, causing TypeScript to treat Zod as a CommonJS module. This resulted in broken type paths like `z.z.core.$strip` instead of the correct `z.core.$strip`.

## Solution
Updated the exports field to use the correct declaration files for each condition:
- **ESM (import)**: `index.d.ts` 
- **CommonJS (require)**: `index.d.cts`

This change was applied to all export entries in the package.json.

## Testing
With the fix applied, TypeScript now correctly resolves module types:
- Before: `Module name 'zod' was successfully resolved to 'node_modules/zod/index.d.cts'`
- After: `Module name 'zod' was successfully resolved to 'node_modules/zod/index.d.ts'`

And generates correct type declarations without the erroneous `z.z` path.

### DIFF
--- a/packages/resolution/attw.test.ts
+++ b/packages/resolution/attw.test.ts
@@ -38,7 +38,7 @@ describe("Are The Types Wrong (attw) tests", () => {
     // remove first line
     const outputWithoutFirstLine = output.split("\n").slice(2).join("\n").trim();
     expect(outputWithoutFirstLine).toMatchInlineSnapshot(`
-      "🎭 Import resolved to a CommonJS type declaration file, but an ESM JavaScript file. https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseCJS.md
+      "No problems found 🌟
 
 
       "zod/package.json"
@@ -54,7 +54,7 @@ describe("Are The Types Wrong (attw) tests", () => {
 
       node10: 🟢 
       node16 (from CJS): 🟢 (CJS)
-      node16 (from ESM): 🎭 Masquerading as CJS
+      node16 (from ESM): 🟢 (ESM)
       bundler: 🟢 
 
       ***********************************
@@ -63,7 +63,7 @@ describe("Are The Types Wrong (attw) tests", () => {
 
       node10: 🟢 
       node16 (from CJS): 🟢 (CJS)
-      node16 (from ESM): 🎭 Masquerading as CJS
+      node16 (from ESM): 🟢 (ESM)
       bundler: 🟢 
 
       ***********************************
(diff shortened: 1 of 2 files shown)


veeceey:
Good point about the condition ordering — just pushed a fix to put `types` before `default` in all export entries. That should ensure TypeScript resolves the correct declaration file before falling through to the runtime module.

Regarding the snapshot update for `attw.test.ts` — I'll look into regenerating that once CI confirms the ordering fix works as expected.

veeceey:
Following up on the snapshot update — CI is green now with the condition ordering fix. I looked into `packages/resolution/attw.test.ts` and it seems the current CI suite doesn't run that specific test, which is why it's passing. I'll regenerate the snapshot locally with `pnpm test packages/resolution/attw.test.ts -u` and push the update to make sure everything is consistent. Thanks for flagging that @pullfrog\!

veeceey:
Updated the `attw.test.ts` snapshot as suggested by @pullfrog. The snapshot now correctly reflects the fix -- all export paths show green ESM resolution instead of "Masquerading as CJS" warnings, and the top-level summary is "No problems found". All 323 test files pass.

veeceey:
checking in on this one - the ESM type resolution fix is needed for projects using moduleResolution: bundler. anything I should update?

veeceey:
hey, just checking in - the ESM type resolution fix and snapshot update are both done. anything else needed?

colinhacks:
@pullfrog do a fresh review

colinhacks:
This problem is a lot more complicated than this. The current approach used here is intentional and prevents out of memory errors in some more arcane dual package environments. Ultimately the `z.z` stuff is cosmetic but in scenarios where Zod is being loaded both from ESM and CJS environments in the same project (this happens more than you think) The fact that Zod resolves to different declaration files causes some catastrophic type comparison performance issues. 

Using `import * as z` should fix this on your end. 

https://github.com/colinhacks/zshy/issues/65#issuecomment-4331089533

PR #5713, opened 2026-02-13
fix: use #/components/schemas/ refs for openapi-3.0 recursive schemas

### LINKED ISSUE #5693: toJSONSchema with target "openapi-3.0" generates #/definitions/ refs for recursive schemas
## Description

When using `toJSONSchema()` with `target: "openapi-3.0"` on recursive schemas (`z.lazy()`), the output contains `#/definitions/` references and a `definitions` block, which are JSON Schema format—not OpenAPI 3.0.

OpenAPI 3.0 expects `#/components/schemas/` for `$ref` values and does not use `definitions`.

This breaks tools like Redoc that strictly follow the OpenAPI spec.

## Reproduction

```javascript
import { z } from "zod";

const Recursive = z.lazy(() => z.object({ child: Recursive.optional() }));
const Schema = z.object({ data: Recursive });

console.log(JSON.stringify(z.toJSONSchema(Schema, { target: "openapi-3.0" }), null, 2));
```

**Output:**
```json
{
  "type": "object",
  "properties": {
    "data": {
      "$ref": "#/definitions/__schema0"
    }
  },
  "required": [
    "data"
  ],
  "additionalProperties": false,
  "definitions": {
    "__schema0": {
      "type": "object",
      "properties": {
        "child": {
          "allOf": [
            {
              "$ref": "#/definitions/__schema0"
            }
          ]
        }
      },
      "additionalProperties": false
    }
  }
}
```

## Expected behavior

With `target: "openapi-3.0"`, recursive refs should either:
- Be inlined (if possible)
- Or not use `#/definitions/` format

## References

According to the [OpenAPI 3.0.3 specification](https://spec.openapis.org/oas/v3.0.3#reference-object), `$ref` values must point to the Components Object, using paths like `#/components/schemas/...`.

The `#/definitions/` path is a JSON Schema / Swagger 2.0 convention and is **not valid** in OpenAPI 3.x.

See also: [OpenAPI 3.0 Components Object](https://spec.openapis.org/oas/v3.0.3#components-object)

## Environment

- Zod version: 4.3.5
- Node version: 20.x

veeceey:
Fixes #5693

When `toJSONSchema()` is called with `target: "openapi-3.0"` on recursive schemas (`z.lazy()`), the output currently contains `#/definitions/` references and a `definitions` block, which are JSON Schema conventions — not valid OpenAPI 3.0.

According to the [OpenAPI 3.0.3 spec](https://spec.openapis.org/oas/v3.0.3#reference-object), `$ref` values must use paths like `#/components/schemas/...`, and the `definitions` keyword is not part of the OpenAPI 3.0 Schema Object.

### Changes

Two small changes in `to-json-schema.ts`:

1. **`makeURI`**: When target is `openapi-3.0`, use `components/schemas` as the defs segment instead of `definitions`
2. **`finalize`**: When target is `openapi-3.0`, place extracted defs under `result.components = { schemas: defs }` instead of `result.definitions`

### Before

```json
{
  "type": "object",
  "properties": {
    "data": { "$ref": "#/definitions/__schema0" }
  },
  "definitions": {
    "__schema0": { ... }
  }
}
```

### After

```json
{
  "type": "object",
  "properties": {
    "data": { "$ref": "#/components/schemas/__schema0" }
  },
  "components": {
    "schemas": {
      "__schema0": { ... }
    }
  }
}
```

### Tests

Added two tests:
- Recursive schema with `openapi-3.0` verifying `#/components/schemas/` refs and `components.schemas` structure
- Self-referencing root schema with `openapi-3.0` verifying `#` self-refs still work correctly

All existing tests pass.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -1794,6 +1794,96 @@ describe("toJSONSchema", () => {
     `
     );
   });
+
+  test("recursive schema with openapi-3.0 uses components/schemas refs", () => {
+    const Recursive: z.ZodType = z.lazy(() => z.object({ child: Recursive.optional() }));
+    const Schema = z.object({ data: Recursive });
+
+    const result = z.toJSONSchema(Schema, { target: "openapi-3.0" });
+
+    // OpenAPI 3.0 should use #/components/schemas/ refs, not #/definitions/
+    expect(result).toMatchInlineSnapshot(`
+      {
+        "additionalProperties": false,
+        "components": {
+          "schemas": {
+            "__schema0": {
+              "additionalProperties": false,
+              "properties": {
+                "child": {
+                  "allOf": [
+                    {
+                      "$ref": "#/components/schemas/__schema0",
+                    },
+                  ],
+                },
+              },
+              "type": "object",
+            },
+          },
+        },
+        "properties": {
+          "data": {
+            "$ref": "#/components/schemas/__schema0",
+          },
+        },
+        "required": [
+          "data",
+        ],
+        "type": "object",
+      }
+    `);
+
+    // Verify no #/definitions/ refs exist
+    const json = JSON.stringify(result);
+    expect(json).not.toContain("#/definitions/");
+    expect(json).not.toContain('"definitions"');
+  });
+
+  test("recursive schema with openapi-3.0 self-referencing root", () => {
+    interface Category {
+      name: string;
+      subcategories: Category[];
+    }
+
+    const categorySchema: z.ZodType<Category> = z.object({
+      name: z.string(),
+      subcategories: z.array(z.lazy(() => categorySchema)),
+    });
+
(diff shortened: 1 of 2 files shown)


veeceey:
Good catch on the `components` placement. You're right that `components` is only valid at the document root level per the OpenAPI 3.0 spec, so embedding it inside the schema object itself is technically invalid.

I think the best approach here is option 3 — keep the `#/components/schemas/` refs (which are the correct ref format for OpenAPI 3.0), and have the definitions returned separately so consumers can merge them into their document-level `components.schemas`. This way the refs actually resolve correctly when the schema gets embedded in a proper OpenAPI document.

I'll update the PR to return the definitions separately rather than embedding them in the schema output. That way the output schema itself stays valid, and the consumer is responsible for placing the defs where they belong in their OpenAPI doc.

veeceey:
Closing this — the pullfrog review was right, embedding `components` inside Schema Objects isn't valid OpenAPI 3.0. A proper fix for recursive schemas with the openapi-3.0 target would need API-level changes to return definitions separately (so consumers can merge them into their document root). Going to file a separate issue to track that instead of pushing an invalid fix.

PR #5720, opened 2026-02-15
refactor(v4): remove unnecessary type assertions

chisaki66:
Removes as any type assertions to improve type safety and leverage TypeScript's type inference.
## Changes
`from-json-schema.ts:` Removed as any from array/tuple constraint methods and default() calls
`parse.ts:` Removed as any from parse, parseAsync, encode, decode, encodeAsync, decodeAsync
`schemas.ts:` Removed as any from string(), number(), boolean(), bigint(), array.element, and nativeEnum()


colinhacks:
Nice cleanup. Most of these `as any` casts were leftover from earlier refactors and the inference handles them fine now — thanks for going through these 👍

colinhacks:
Landed in Zod 4.4

PR #5721, opened 2026-02-16
Preprocess is not deprecated

mxdvl:
## Why are we doing this?

To make it clear that it isn’t.

## What does this change?

Remove the commented out JSDoc suggesting it might be. Note it could be the other way around and this should in fact be marked as deprecated 😆


colinhacks:
@pullfrog fix merge conflicts.

colinhacks:
Merged — `z.preprocess` is staying, the `JSDoc` was a holdover from when I was on the fence and never followed through. Heads up: the conflict resolution dropped a few workflow files, which I restored in commit d518dc13 before landing — nothing on your end. Thanks for cleaning this up!

colinhacks:
Landed in Zod 4.4

PR #5726, opened 2026-02-20
feat: loosen z.strictObject() inside intersection

vatsal9117:
### What does this PR do?
This PR resolves an issue where `z.strictObject()` is functionally incompatible with `z.intersection()`. Previously, intersecting a strict object with another object would cause the strict object to throw an `unrecognized_keys` error for any keys defined by the other half of the intersection.

This update loosens the strictness exclusively within the context of intersections.

### How it works:
In `src/v4/core/schemas.ts`, the `handleIntersectionResults` function now tracks unrecognized keys from both the `left` and `right` schemas using a tracking Map. 
* Any key recognized by at least *one* side of the intersection will now pass validation.
* `ZodError` is only thrown for keys that are completely unrecognized by *both* sides.
* Extra keys are stripped according to the standard object's default behavior if mixed.

### Testing
Added test coverage in `src/v4/classic/tests/intersection.test.ts` to verify:
- `strict + strip` behavior (allows shared keys, strips unknown)
- `strict + strict` behavior (allows shared keys, throws on completely unknown keys)

### DIFF
--- a/packages/zod/src/v4/classic/tests/intersection.test.ts
+++ b/packages/zod/src/v4/classic/tests/intersection.test.ts
@@ -196,3 +196,23 @@ test("invalid deep merge of object and array combination", async () => {
     `[Error: Unmergable intersection. Error path: ["students",0,"name"]]`
   );
 });
+test("object intersection: strict + strip", () => {
+  const A = z.strictObject({ a: z.string() });
+  const B = z.object({ b: z.string() });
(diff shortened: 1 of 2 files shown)


vatsal9117:
feat: loosen z.strictObject() inside intersection #5726

colinhacks:
@pullfrog review this. I believe this has already been fixed elsewhere.

PR #5730, opened 2026-02-23
feat(v4): cron expression validation

ysknsid25:
# Overview

This PR adds a new built-in string format validator z.cron() for validating standard 5-field cron expressions (minute, hour, day-of-month, month, day-of-week).

# Why this is useful

Cron expressions are widely used in task scheduling, CI/CD pipelines, cloud functions (AWS EventBridge, GCP Cloud Scheduler, etc.), and configuration files. Validating them at the schema layer catches malformed expressions before they reach the scheduler, preventing silent failures or unexpected behavior at runtime.

Before this PR, users had to reach for a custom .refine() with a hand-written regex or a third-party package:

```typescript
// Before: verbose and error-prone
const schema = z.string().refine(val => /^(\*|[0-5]?\d).../.test(val), {
  message: "Invalid cron expression",
});
```

With `z.cron()`, this is a single, well-tested, semantically clear call.

# Usage

## Standalone schema

```typescript
import { z } from "zod";

const CronJob = z.object({
  name: z.string(),
  schedule: z.cron(),
});

CronJob.parse({ name: "cleanup", schedule: "0 2 * * *" }); // OK
CronJob.parse({ name: "cleanup", schedule: "60 * * * *" }); // NG. minute out of range
```

## As a string check

```typescript
z.string().cron();
```

## Supported syntax

```typescript
z.cron().parse("* * * * *");        // wildcard
z.cron().parse("0 9-17 * * 1-5");   // ranges (weekdays 9am–5pm)
z.cron().parse("*/15 * * * *");     // steps (every 15 min)
z.cron().parse("0 0 1,15 * *");     // lists (1st and 15th)
z.cron().parse("0-30/5 * * * *");   // combined (every 5 min for first 30 min)
```

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -370,6 +370,8 @@ export interface ZodString extends _ZodString<core.$ZodStringInternals<string>> 
   cidrv6(params?: string | core.$ZodCheckCIDRv6Params): this;
   /** @deprecated Use `z.e164()` instead. */
   e164(params?: string | core.$ZodCheckE164Params): this;
+  /** @deprecated Use `z.cron()` instead. */
+  cron(params?: string | core.$ZodCheckCronParams): this;
 
   // ISO 8601 checks
   /** @deprecated Use `z.iso.datetime()` instead. */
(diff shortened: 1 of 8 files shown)


colinhacks:
After doing some research, I think cron is not a good area for Zod to come in with an opinionated API. There are subtly different formats and various cron extensions that will inevitably be asked about. Ultimately, I don't think the syntax is quite standard enough, and I'm increasingly hesitant to add support for validators like this with different flavors or modes. Thanks for your work on this, though.

PR #5737, opened 2026-03-01
fix(v4): parsedType() returns "infinity" for Infinity/-Infinity, fixing misleading error messages

### LINKED ISSUE #5697: Weird error message after update | 4.3.6

```
 const schema = z.object({
          id: z.number().finite()
        }).required({ id: true })

const value = Infinity

// validate schema with value

// error code: invalid_type

```

berketok-git:
## Problem

When `Infinity` or `-Infinity` is passed to `z.number()`, the error message is confusingly self-contradictory:

```ts
z.number().parse(Infinity)
// ZodError: Invalid input: expected number, received number  ← ???
```

The same message appears with `.finite()`:

```ts
z.number().finite().parse(Infinity)
// ZodError: Invalid input: expected number, received number
```

This was reported in #5697.

## Root Cause

`parsedType()` in `util.ts` only checks `Number.isNaN` but not `Number.isFinite`:

```ts
case "number": {
  return Number.isNaN(data) ? "nan" : "number";  // ← Infinity falls through as "number"
}
```

The locale formatter then calls `parsedType(issue.input)` to generate the `received` string in the message, so it produces `"number"` for both finite and infinite numbers.

Note: `schemas.ts` was already correctly setting `received: "Infinity"` on the raw issue object — the bug was only in the message rendering path through `parsedType`.

## Fix

1. **`util.ts`** — extend the `number` case to return `"infinity"` for non-finite values:

```ts
case "number": {
  if (Number.isNaN(data)) return "nan";
  if (!Number.isFinite(data)) return "infinity";
  return "number";
}
```

2. **All locale files** — add `infinity: "Infinity"` to each locale's `TypeDictionary`, following the same pattern as the existing `nan: "NaN"` entry.

## Result

```ts
z.number().parse(Infinity)
// ZodError: Invalid input: expected number, received Infinity  ← clear
```

All 3575 tests pass.

Fixes #5697

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ packages/tsc-perf
 tsconfig.vitest-temp.json
 *.tgz
 
+bun.lock
--- a/packages/zod/src/v4/classic/tests/number.test.ts
+++ b/packages/zod/src/v4/classic/tests/number.test.ts
@@ -22,7 +22,7 @@ test("Infinity validation", () => {
         "code": "invalid_type",
         "received": "Infinity",
         "path": [],
-        "message": "Invalid input: expected number, received number"
+        "message": "Invalid input: expected number, received Infinity"
       }
     ]],
       "success": false,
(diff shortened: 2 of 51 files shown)


colinhacks:
This is a decent idea, but changing error messages is considered a breaking change, and there are no major versions of Zod planned in the horizon.

Issue #5739, opened 2026-03-02
`.superRefine` is missing "when" parameter


vilvai:
`.refine()` has a `when` parameter that allows running a refinement even when the base schema validation fails: https://github.com/colinhacks/zod/pull/4838

However, `.superRefine` is missing this parameter.

Example
```
    const baseSchema = z.object({
      foo: z.number(),
      bar: z.number(),
    });

    const schema = baseSchema.superRefine(
      (data, ctx) => {
        if (data.foo > 10) {
          ctx.addIssue({
            code: "custom",
            message: "foo must be less than 10",
          });
        }
      },
      {
        when: ({ value }) =>
          baseSchema.pick({ foo: true }).safeParse(value).success,
      },
    );

    const result = schema.safeParse({
      foo: 11,
    });

// 
```

In the above example, `result.error.issues` will only contain one error: "Invalid input: expected number, received undefined" and not the error from `.superRefine`

The above example could be replaced with `.refine` since it only adds one issue, but for more complex use cases where `.superRefine` is needed, having `.when` would be useful.

This is a widely requested feature, see for example:
https://github.com/colinhacks/zod/issues/479#issuecomment-3365086928

vilvai:
I already made an example PR about this: https://github.com/colinhacks/zod/pull/5741

colinhacks:
Landed in Zod 4.4

PR #5741, opened 2026-03-02
feat: allow 'when' parameter in .superRefine

vilvai:
This PR adds the `params` object to `.superRefine` which allows using `when` parameter to run the superrefine validation even when the base schema fails, similar to how `.refine` works currently

See my issue about this: https://github.com/colinhacks/zod/issues/5739


colinhacks:
Narrowed the `params` type on `.superRefine` to just `{ when }` for now, rather than the full `.refine` params shape — `when` is really the only one that makes sense here, since the refinement pushes its own issues via `ctx.addIssue` (so `error`/`message`/`path` aren't read) and `abort` is redundant with the per-issue `fatal` flag. Thanks for picking this up 👍

colinhacks:
Landed in Zod 4.4

PR #5742, opened 2026-03-02
perf: avoid delete in finalizeIssue to keep issue objects in V8 fast …

### LINKED ISSUE #5717: perf: avoid `delete` in `finalizeIssue` to keep issue objects in V8 fast mode
`finalizeIssue()` uses `delete` to strip internal properties (`inst`, `continue`, `input`) from issue objects before returning them. This causes V8 to transition the objects from hidden class mode to [dictionary mode](https://v8.dev/blog/fast-properties), which makes all subsequent property access on those objects slower.

Since every finalized issue goes through this path, it affects `flattenError`, `formatError`, `treeifyError`, etc.

The fix is straightforward — destructure out the internal properties instead of deleting them, so the returned object is clean from the start. 1 function, net -1 line, all 3574 tests pass unchanged.

I benchmarked three realistic scenarios (form validation + flatten, API response + format, config + treeify) with static fixture data and saw **6-8% improvement** on the error consumption path.

We know this is small stuff, but given how widely Zod is used and that this is the hot path for error handling, these minor improvements add up.

Happy to submit a PR with the fix and benchmark if this sounds reasonable.

andrewdamelio:
Replace property deletion with destructuring to prevent V8 from transitioning issue objects to dictionary mode. This improves performance of error consumption paths (flattenError, formatError, treeifyError) by 6-8%.

Fixes #5717

### DIFF
--- a/packages/zod/src/v4/core/util.ts
+++ b/packages/zod/src/v4/core/util.ts
@@ -840,12 +840,14 @@ export function finalizeIssue(
   ctx: schemas.ParseContextInternal | undefined,
   config: $ZodConfig
 ): errors.$ZodIssue {
-  const full = { ...iss, path: iss.path ?? [] } as errors.$ZodIssue;
+  // Destructure out internal properties to avoid V8 deoptimization from delete
+  const { inst, continue: continueFn, input, ...rest } = iss as any;
+  const full = { ...rest, path: iss.path ?? [] } as errors.$ZodIssue;
 
   // for backwards compatibility
   if (!iss.message) {
     const message =
-      unwrapMessage(iss.inst?._zod.def?.error?.(iss as never)) ??
+      unwrapMessage(inst?._zod.def?.error?.(iss as never)) ??
       unwrapMessage(ctx?.error?.(iss as never)) ??
       unwrapMessage(config.customError?.(iss)) ??
       unwrapMessage(config.localeError?.(iss)) ??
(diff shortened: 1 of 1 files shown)


idirdev (commented):
Clean and well-targeted optimization. The V8 deoptimization from `delete` is well-documented — once an object transitions to dictionary mode, all subsequent property lookups on that object go through a hash table instead of inline caches, which matters on a hot path like `finalizeIssue`.

The destructuring approach is the right solution here: `full` is constructed without the internal properties from the start, so V8 never needs to transition its hidden class. A few notes:

1. **`continue` → `continueFn` alias** — nice use of destructuring rename to avoid the reserved word conflict.

2. **`input` logic inversion** — the old code was "include by default, delete if not `reportInput`". The new code is "exclude by default, add if `reportInput`". Semantically equivalent, but slightly cleaner since `input` is an opt-in field. One thing to verify: if `input` is `undefined` in the original `iss`, the old code would leave it as-is (or delete an undefined property — a no-op). The new code would skip `(full as any).input = input`, which also results in no `input` property. So behavior is consistent.

3. **`inst` still available** — destructured `inst` is correctly used on line 843 for the error message lookup before being discarded. No behavior change.

The 6-8% improvement claim is plausible for error-heavy validation workloads. Would be nice to see the benchmark setup from #5717 linked or inlined for reproducibility, but the change itself is straightforward enough to stand on its own.

colinhacks:
@pullfrog review

colinhacks:
Merged in https://github.com/colinhacks/zod/pull/5718

Issue #5365, opened 2025-10-20
ReDoS checker test fail on Node 22


1knair:
Environment: Windows 11, Node 22.19.0, pnpm 10.12.1

How to reproduce:
(Clone repo and install deps with Node 22)

git clone https://github.com/colinhacks/zod.git
cd zod
pnpm install
pnpm -w vitest --run src/v4/classic/tests/datetime.test.ts -t "redos checker"

Result:
FAIL  src/v4/classic/tests/datetime.test.ts > redos checker
Error: ReDoS issue
    at Object.<anonymous> (src/v4/classic/tests/datetime.test.ts:300:41)

Result: test throws error "ReDoS issue"
Expected: all tests pass

Likely caused by Node 22 regex engine changes. All other 2996 tests pass.
Running the same test on Node 20 LTS passes successfully.
Failure originates from checkSync(schema._zod.pattern.source, "") that uses recheck library for ReDoS detection.
Could stem from updates to V8’s RegExp engine in Node 22 that tightens recheck’s detection .

colinhacks:
Hm can't replicate on Mac using Node 22. I configured CI to run on windows as well and not seeing it there (Node 24). Perhaps you can investigate on your machine and try to put in a fix? Doing some investigation into the exact failing regex would help a lot.

PR #5744, opened 2026-03-04
test: increase timeout for redos checker in datetime.test.ts

### LINKED ISSUE #5743: Performance: Test timeout on node 24 (datetime validation)
i notice that the test `redos checker` in `src/v4/classic/tests/datetime.test.ts` is failing on node 24.12.0 (LTS) due to a timeout. While the test passes on node 22 (~4.9s), it consistently hits ~5.5s on node 24, exceeding the default 5000ms limit.

The bottleneck is isolated to a specific `test()` block containing multiple `z.string().datetime()` validations. The overhead of these regex-heavy checks has increased on the newer V8 engine

#### Quick benchmarks on node 24:
- Original test: ~5.5s (Fail)
- Removing a few datetime variables: (Pass)

i have a fix ready that increases the timeout for this specific test to 10s. Should i open a PR for this?

rishadaufa:
This PR increases the timeout for the `redos checker` test case in `src/v4/classic/tests/datetime.test.ts` to 10s.

#### Why?
As documented in #5743 (and previously discussed in #5365), this test is hitting the 5s limit on node 24.12.0 (LTS). This provides a buffer for the `recheck` library's execution on newer V8 engines.

#### Verified on
- Node 22 (Pass)
- Node 24 (Pass)

Closes #5743

### DIFF
--- a/packages/zod/src/v4/classic/tests/datetime.test.ts
+++ b/packages/zod/src/v4/classic/tests/datetime.test.ts
@@ -299,4 +299,4 @@ test("redos checker", () => {
     const result = checkSync(schema._zod.pattern.source, "");
     if (result.status !== "safe") throw Error("ReDoS issue");
   }
-});
+}, 10000);


colinhacks:
Thanks! 5s was cutting it close under Node 24's V8 — 10s is the right buffer for `recheck` here. 👍

colinhacks:
Landed in Zod 4.4

PR #5745, opened 2026-03-04
Fix `Custom format` example

xgaia:


colinhacks:
Good catch — thanks for the fix 👍

colinhacks:
Landed in Zod 4.4

PR #5759, opened 2026-03-07
fix: strip redundant id from $defs entries in toJSONSchema

### LINKED ISSUE #5731: z.toJSONSchema: .meta({ id }) leaks id into $defs entries — should use $id or omit it
When using `.meta({ id })` to enable `$defs` extraction in `z.toJSONSchema()`, the `id` value is included verbatim as an `id` property in the generated `$defs` entries. This is not a valid JSON Schema keyword in Draft 2020-12 (or Draft 7).

- Draft 4 used `id` (no dollar sign)
- Draft 6+ renamed it to `$id`, and `$id` must be a URI-reference

So the current behavior produces output that is either:
- An unrecognized keyword (Draft 2020-12 / Draft 7), or
- A non-URI value for a URI field (if interpreted as Draft 4 `id`)

## Reproduction

```ts
import * as z from "zod";

const inner = z.object({ name: z.string() }).meta({ id: "Inner" });
const outer = z.object({ a: inner, b: inner });

console.log(JSON.stringify(z.toJSONSchema(outer, { target: "draft-2020-12" }), null, 2));
```

## Actual output

```json
{
  "$defs": {
    "Inner": {
      "type": "object",
      "properties": { "name": { "type": "string" } },
      "required": ["name"],
      "id": "Inner"
    }
  },
  ...
}
```

## Expected

The `id` key should either:
1. Be omitted from the JSON schema output (it's only needed internally for naming the `$defs` key), or
2. Be emitted as `$id` with a proper URI value

Option 1 seems cleanest since the `$defs` key already serves as the identifier via `$ref: "#/$defs/Inner"`.

mibragimov:
## Problem

When a schema is annotated with `.meta({ id: "Name" })`, the `id` is used as the key in `$defs` so that other schemas can reference it via `$ref`. However, the `id` was also being included verbatim in the definition body, producing redundant output:

```json
{
  "$defs": {
    "Inner": {
      "id": "Inner",
      "type": "string"
    }
  }
}
```

The presence of `id` inside `$defs` entries is misleading — it looks like a JSON Schema `$id`, but it's just Zod's internal `id` tag. The `id` is already encoded as the `$defs` key itself, so including it in the value is redundant.

Fixes #5731

## Solution

In `extractToDef()`, after copying the schema into `seen.def`, delete the `id` property from the definition body when it matches `defId` — the id is already encoded as the `$defs` key name.

**Before:**
```json
{ "$defs": { "Inner": { "id": "Inner", "type": "string" } } }
```

**After:**
```json
{ "$defs": { "Inner": { "type": "string" } } }
```

## Tests

Updated 6 existing snapshot tests that documented the old (buggy) behavior to reflect the correct output.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -1219,11 +1219,9 @@ describe("toJSONSchema", () => {
         },
         "definitions": {
           "primary": {
-            "id": "primary",
             "type": "string",
           },
           "rest": {
-            "id": "rest",
             "type": "number",
           },
         },
@@ -2014,11 +2012,9 @@ test("extract schemas with id", () => {
     {
       "$defs": {
         "age": {
-          "id": "age",
           "type": "number",
         },
         "name": {
-          "id": "name",
           "type": "string",
         },
       },
@@ -2106,7 +2102,6 @@ test("describe with id", () => {
     {
       "$defs": {
         "jobId": {
-          "id": "jobId",
           "type": "string",
         },
       },
@@ -2146,7 +2141,6 @@ test("describe with id on wrapper", () => {
     {
       "$defs": {
         "roJobId": {
-          "id": "roJobId",
           "readOnly": true,
           "type": "string",
         },
@@ -2185,12 +2179,10 @@ test("overwrite id", () => {
     {
       "$defs": {
         "aaa": {
-          "id": "aaa",
           "type": "string",
         },
         "bbb": {
           "$ref": "#/$defs/aaa",
-          "id": "bbb",
         },
       },
       "$schema": "https://json-schema.org/draft/2020-12/schema",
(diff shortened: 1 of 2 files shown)


colinhacks:
@pullfrog review

colinhacks:
Hey @mibragimov — thanks for this. I pushed a follow-up commit that extends the fix slightly. Reasoning is that we treat `meta({ id })` as a registration directive for `$defs` extraction, so leaking it into the output body is a leaky abstraction across all targets (and is actively wrong on draft-04, where `id` is a reserved keyword that sets a base URI). The expanded change:

1. **Strip from $defs entries** (your original fix, kept).
2. **Strip from the root schema** when `meta({ id })` is set on the root — same reasoning, the registration tag shouldn't surface there either.
3. **Move the strip into `finalize()`** so that user-authored `override` callbacks still observe `jsonSchema.id` on extracted definitions before it gets cleaned up. Avoids a subtle behavior change to the `override` API.
4. **Tests** for: `$defs` strip (draft-2020-12), `definitions` strip (draft-04), root strip, and override-callback observability.

Snapshot diff for `top-level readonly` was updated accordingly.

Happy to discuss any of this if you'd rather take a different approach.

colinhacks:
Merged 👍 The follow-up moves the `id` strip into `finalize()` inside `toJSONSchema` so it also clears the root schema, not just `$defs` entries — `meta({ id })` on the root was leaking the same way. Closes #5731 — thanks for the fix!

colinhacks:
Landed in Zod 4.4

PR #5763, opened 2026-03-09
docs: update z.custom example for v4 compatibility

### LINKED ISSUE #5605: [Docs] Maybe new example for z.custom

andrewdamelio:
Remove template string literal example since z.templateLiteral is now available in Zod v4. Replace with File instanceof example which better demonstrates z.custom's use case for complex types.

Fixes #5605

### DIFF
--- a/packages/docs-v3/README.md
+++ b/packages/docs-v3/README.md
@@ -2155,12 +2155,10 @@ This returns a `ZodEffects` instance. `ZodEffects` is a wrapper class that conta
 
 ## Custom schemas
 
-You can create a Zod schema for any TypeScript type by using `z.custom()`. This is useful for creating schemas for types that are not supported by Zod out of the box, such as template string literals.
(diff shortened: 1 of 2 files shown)


colinhacks:
Merged, thanks for chasing this down. I swapped the example to a `decimal.js` `Decimal` check on top — the `File instanceof` case overlaps with `z.instanceof()` and `z.file()`, while `z.custom` is best demonstrated on structural or third-party types where no built-in fits. Closes #5605.

colinhacks:
Landed in Zod 4.4

PR #5766, opened 2026-03-13
fix(transform): type standalone z.transform() ctx as $RefinementCtx

### LINKED ISSUE #5678: `transform` callback `ctx.addIssue` not in Typescript

mahmoodhamdi:
## What does this PR do?

Fixes the type of the `ctx` parameter in the standalone `z.transform()` factory function to include `addIssue()`.

## Root cause

The standalone `z.transform((input, ctx) => ...)` factory function typed `ctx` as `ParsePayload`, which only includes `value`, `issues`, and `aborted`. However, at runtime, the classic `ZodTransform` constructor mutates the payload to attach `addIssue()` before invoking the callback — so `ctx.addIssue(...)` works at runtime but TypeScript doesn't see it.

The chained `.transform()` method on `ZodType` already correctly types `ctx` as `$RefinementCtx` (which extends `ParsePayload` with `addIssue`). The standalone factory was just missed.

## Fix

Changed the `ctx` parameter type from `core.ParsePayload` to `core.$RefinementCtx<I>` in the standalone `transform()` function signature.

## Tests

Added two tests:
- Runtime test: standalone `z.transform()` with `ctx.addIssue()` correctly surfaces validation errors
- Type test: verifies `ctx.addIssue` is typed as a function via `expectTypeOf`

## Related issue

Closes #5678

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1814,7 +1814,7 @@ export const ZodTransform: core.$constructor<ZodTransform> = /*@__PURE__*/ core.
 );
 
 export function transform<I = unknown, O = I>(
-  fn: (input: I, ctx: core.ParsePayload) => O
+  fn: (input: I, ctx: core.$RefinementCtx<I>) => O
 ): ZodTransform<Awaited<O>, I> {
   return new ZodTransform({
     type: "transform",
--- a/packages/zod/src/v4/classic/tests/transform.test.ts
+++ b/packages/zod/src/v4/classic/tests/transform.test.ts
@@ -359,3 +359,33 @@ test("encode error", () => {
     `[ZodEncodeError: Encountered unidirectional transform during encode: ZodTransform]`
   );
 });
+
+test("standalone z.transform ctx.addIssue", () => {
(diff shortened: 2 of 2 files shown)


colinhacks:
Thanks for the PR! This was already fixed for the classic standalone `z.transform()` in #5699, which merged earlier today (commit 2f8414bc) — `ctx` is now typed as `$RefinementCtx` on `main`.

PR #5767, opened 2026-03-13
fix(json-schema): use $defs instead of definitions for openapi-3.0 target

### LINKED ISSUE #5693: toJSONSchema with target "openapi-3.0" generates #/definitions/ refs for recursive schemas

mahmoodhamdi:
## What does this PR do?

Fixes `toJSONSchema()` with `target: "openapi-3.0"` to use `$defs` and `#/$defs/` refs instead of `definitions` and `#/definitions/` for recursive schemas.

## Root cause

The `defsSegment` logic in `to-json-schema.ts` and `json-schema-processors.ts` used a binary choice: `$defs` for `draft-2020-12`, `definitions` for everything else. This meant `openapi-3.0` was getting `definitions` — a Swagger 2.0 / JSON Schema Draft 4 concept that is not a recognized keyword in OpenAPI 3.0 Schema Objects.

For a recursive schema like:

```ts
const Recursive = z.lazy(() => z.object({ child: Recursive.optional() }));
const Schema = z.object({ data: Recursive });
z.toJSONSchema(Schema, { target: "openapi-3.0" });
```

The output contained `"$ref": "#/definitions/__schema0"` and a `"definitions"` key, which broke tools like Redoc that strictly validate against the OAS 3.0 spec.

## Fix

Inverted the condition: `definitions` is now only used for `draft-04` and `draft-07`, while everything else (including `openapi-3.0` and `draft-2020-12`) uses `$defs`. This is forward-compatible since OpenAPI 3.1 adopted JSON Schema 2020-12's `$defs` convention.

Changed in three places:
1. `to-json-schema.ts` line 250 — `makeURI` ref path segment
2. `to-json-schema.ts` lines 486-491 — `finalize` defs key placement
3. `json-schema-processors.ts` line 653 — registry `toJSONSchema` defs segment

## Tests

Added a new test `"recursive schema openapi-3.0 uses $defs"` that verifies:
- The output does NOT have a `definitions` key
- The output DOES have a `$defs` key
- Refs use `#/$defs/` prefix
- Full inline snapshot of expected output

All 3577 existing tests pass.

## Related issue

Closes #5693

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -1157,6 +1157,45 @@ describe("toJSONSchema", () => {
     `);
   });
 
+  test("recursive schema openapi-3.0 uses $defs", () => {
+    const Recursive: z.ZodType = z.lazy(() => z.object({ child: Recursive.optional() }));
+    const Schema = z.object({ data: Recursive });
+    const jsonSchema = z.toJSONSchema(Schema, { target: "openapi-3.0" });
+    // OpenAPI 3.0 should not use "definitions" — it is a Swagger 2.0 / JSON Schema
+    // Draft 4 concept not recognized by OpenAPI 3.0 Schema Objects.
+    expect(jsonSchema).not.toHaveProperty("definitions");
+    expect(jsonSchema).toHaveProperty("$defs");
+    expect(jsonSchema).toMatchInlineSnapshot(`
+      {
+        "$defs": {
+          "__schema0": {
+            "additionalProperties": false,
+            "properties": {
+              "child": {
+                "allOf": [
+                  {
+                    "$ref": "#/$defs/__schema0",
+                  },
+                ],
+              },
+            },
+            "type": "object",
+          },
+        },
+        "additionalProperties": false,
+        "properties": {
+          "data": {
+            "$ref": "#/$defs/__schema0",
+          },
+        },
+        "required": [
+          "data",
+        ],
+        "type": "object",
+      }
+    `);
+  });
+
   test("tuple draft-7", () => {
     const schema = z.tuple([z.string(), z.number()]);
     expect(z.toJSONSchema(schema, { target: "draft-7", io: "input" })).toMatchInlineSnapshot(`
--- a/packages/zod/src/v4/core/json-schema-processors.ts
+++ b/packages/zod/src/v4/core/json-schema-processors.ts
@@ -650,7 +650,7 @@ export function toJSONSchema(
     }
 
     if (Object.keys(defs).length > 0) {
-      const defsSegment = ctx.target === "draft-2020-12" ? "$defs" : "definitions";
+      const defsSegment = ctx.target === "draft-04" || ctx.target === "draft-07" ? "definitions" : "$defs";
       schemas.__shared = {
         [defsSegment]: defs,
       };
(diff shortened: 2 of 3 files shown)


colinhacks:
@pullfrog review

colinhacks:

Going to close this. `$defs` isn't valid in OpenAPI 3.0 either — the Schema Object is based on JSON Schema Wright Draft 00 (≈ Draft 5), which predates `$defs` (introduced in Draft 2019-09). Per the [OpenAPI 3.0 keyword reference](https://swagger.io/docs/specification/v3_0/data-models/keywords), neither `definitions` nor `$defs` is a supported keyword on an OpenAPI 3.0 Schema Object — the canonical place for reusable schemas is the document-level `components.schemas`, referenced as `#/components/schemas/Name`. So this swaps one non-spec keyword for another that's arguably less defensible.

The right approach for #5693 is the registry-based `toJSONSchema`. When you pass a registry instead of a single schema, you get back a `{ schemas: { ... } }` map keyed by id — that map is exactly what you drop into `components.schemas`. The `uri` option controls how cross-schema `$ref`s in that map are written, so pointing it at `#/components/schemas/...` makes the output OpenAPI-ready:

```ts
const registry = z.registry();
const Recursive = z.lazy(() => z.object({ child: Recursive.optional() })).register(registry, { id: "Recursive" });
const Schema = z.object({ data: Recursive }).register(registry, { id: "Schema" });

const { schemas } = z.toJSONSchema(registry, {
  target: "openapi-3.0",
  uri: (id) => `#/components/schemas/${id}`,
});

const openapi = {
  openapi: "3.0.3",
  info: { title: "...", version: "1.0.0" },
  paths: {},
  components: { schemas },
};
```

There's no inline def container in the output to argue over — refs go where you tell them, and the schema map is yours to compose into a full document. See https://zod.dev/json-schema#registries. Thanks for the PR.

PR #5768, opened 2026-03-13
fix: cascade inner field defaults through .default({}) on objects

### LINKED ISSUE #5764: .default({}) on objects should cascade inner field defaults
# `.default({})` on objects should cascade inner field defaults

**Minimal repro:**

```ts
import * as z from "zod";

const schema = z.object({
  name: z.string().default("untitled"),
  count: z.number().default(0),
}).default({});

schema.parse({});        // => { name: "untitled", count: 0 } ✅
schema.parse(undefined); // => {}                              ❌
```

`parse({})` and `parse(undefined)` should return the same thing. The schema says "if missing, use `{}`" — and `{}` parsed through this schema produces `{ name: "untitled", count: 0 }`. But `.default()` short-circuits and returns the literal `{}`, skipping inner defaults entirely.

**Why this matters — nested config schemas:**

```ts
const config = z.object({
  revenue: z.object({
    users: z.number().default(1000),
    price: z.number().default(29),
  }).default({}),
  costs: z.object({
    team_size: z.number().default(5),
    salary: z.number().default(10_000),
  }).default({}),
}).default({});

config.parse(undefined);
// Got:      {}
// Expected: { revenue: { users: 1000, price: 29 }, costs: { team_size: 5, salary: 10000 } }
```

This is the most natural way to express "everything has defaults, any level can be omitted." Today, the only workaround is duplicating every default value:

```ts
.default({ users: 1000, price: 29 })  // DRY violation — defaults declared twice
```

...which defeats the purpose of field-level `.default()`.

**The invariant that should hold:**

> For any schema `s` and default value `d`, `s.default(d).parse(undefined)` should equal `s.parse(d)`.

This already holds for every other schema type. Objects are the exception because short-circuiting skips property-level parsing where inner defaults are applied.

**Why this is consistent with Zod v4's own direction:**

Zod v4 already made defaults cascade into optional fields:

```ts
z.object({ a: z.string().default("x").optional() }).parse({});
// Zod 4: { a: "x" }  ← defaults applied inside objects
// Zod 3: {}
```

The changelog says this "aligns better with expectations." The same reasoning applies here when `.default({})` provides an empty object, users expect inner defaults to fill it, just like they would if you passed `{}` directly to `.parse()`.

**Proposed fix:**

When `.default(value)` is called on a `ZodObject` and the default is a plain object (not a function), parse it through the schema at definition time:

```ts
function _default(innerType, defaultValue) {
  if (innerType instanceof ZodObject && isPlainObject(defaultValue)) {
    defaultValue = innerType.parse(defaultValue);
  }
  return new ZodDefault({ innerType, defaultValue });
}
```

This is a one-time cost at schema definition with zero runtime overhead. It preserves the short-circuit optimization while ensuring the stored default already has inner defaults applied.

**Note on `.prefault()`:** Yes, `.prefault({})` works as a workaround since it passes through the parser. But `.default()` is the intuitive API for this `.prefault()` exists for transform pipelines where input/output types differ. Telling users "use `.prefault()` for objects" is a footgun, not a solution.

-----------------

**Patch (for anyone hitting this today):**

In `_default()` inside `v4/classic/schemas.js`, add a check before constructing `ZodDefault`:

```diff
 export function _default(innerType, defaultValue) {
+    // Cascade defaults: when the inner type is a ZodObject and the default is a plain object,
+    // parse the default through the schema so nested field defaults get applied.
+    if (innerType instanceof ZodObject && typeof defaultValue === "object" && defaultValue !== null && !Array.isArray(defaultValue) && typeof defaultValue !== "function") {
+        defaultValue = innerType.parse(defaultValue);
+    }
     return new ZodDefault({
         type: "default",
         innerType: innerType,
```

Works with pnpm `patchedDependencies` / npm `overrides`, we've been running this in production without issues.

mahmoodhamdi:
## Summary

Fixes #5764

When `.default({})` is called on an object schema, the default value is now parsed through the inner schema **at definition time** so that nested field-level defaults are applied. This ensures the invariant:

```
schema.default(d).parse(undefined) === schema.parse(d)
```

holds for object schemas, matching user expectations and Zod v4's own direction on defaults.

### Before

```ts
const schema = z.object({
  name: z.string().default("untitled"),
  count: z.number().default(0),
}).default({});

schema.parse(undefined); // => {} ❌
```

### After

```ts
schema.parse(undefined); // => { name: "untitled", count: 0 } ✅
```

## Approach

In the `_default()` factory (classic, mini, and core), when the inner type is an object schema and the default is a static plain object (not a function), run it through the inner schema's parse at definition time. This:

- Has **zero runtime overhead** — the short-circuit optimization in `$ZodDefault.parse` is preserved
- Handles **deeply nested** objects recursively (inner defaults are already resolved when outer `.default({})` is called)
- Silently falls back to the original default if parsing fails (e.g., invalid default)

### Files changed

- `packages/zod/src/v4/classic/schemas.ts` — classic `_default()` factory
- `packages/zod/src/v4/mini/schemas.ts` — mini `_default()` factory
- `packages/zod/src/v4/core/api.ts` — core `_default()` factory
- `packages/zod/src/v4/classic/tests/default.test.ts` — 4 new test cases

## Test plan

- [x] Existing `default.test.ts` tests still pass (20/20)
- [x] New tests cover: basic cascading, deeply nested objects, partial overrides, shallow clone identity
- [x] Full test suite passes (3583 tests, 0 type errors)

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1924,6 +1924,19 @@ export function _default<T extends core.SomeType>(
   innerType: T,
   defaultValue: util.NoUndefined<core.output<T>> | (() => util.NoUndefined<core.output<T>>)
 ): ZodDefault<T> {
+  // Cascade inner defaults: when the inner type is an object schema and
+  // the default is a static plain object, parse it through the inner schema
+  // so nested field-level defaults are applied at definition time.
+  if (
+    typeof defaultValue !== "function" &&
+    util.isPlainObject(defaultValue) &&
+    (innerType as any)._zod?.def?.type === "object"
+  ) {
+    const result = (innerType as any)._zod.run({ value: { ...defaultValue }, issues: [] }, { async: false });
+    if (!(result instanceof Promise) && result.issues.length === 0) {
+      defaultValue = result.value;
+    }
+  }
   return new ZodDefault({
     type: "default",
     innerType: innerType as any as core.$ZodType,
--- a/packages/zod/src/v4/classic/tests/default.test.ts
+++ b/packages/zod/src/v4/classic/tests/default.test.ts
@@ -363,3 +363,60 @@ test("direction-aware defaults", () => {
     }
   `);
 });
+
+test("default({}) cascades inner field defaults", () => {
+  const inner = z.object({
+    name: z.string().default("untitled"),
+    count: z.number().default(0),
+  });
+  const schema = inner.default({} as z.output<typeof inner>);
+
+  // parse(undefined) should equal parse({})
+  expect(schema.parse(undefined)).toEqual({ name: "untitled", count: 0 });
+  expect(schema.parse({})).toEqual({ name: "untitled", count: 0 });
+  expect(schema.parse(undefined)).toEqual(schema.parse({}));
+});
+
+test("nested objects with cascading defaults", () => {
+  const revenue = z.object({
+    users: z.number().default(1000),
+    price: z.number().default(29),
+  });
+  const costs = z.object({
+    team_size: z.number().default(5),
+    salary: z.number().default(10_000),
+  });
+  const outer = z.object({
+    revenue: revenue.default({} as z.output<typeof revenue>),
+    costs: costs.default({} as z.output<typeof costs>),
+  });
+  const config = outer.default({} as z.output<typeof outer>);
+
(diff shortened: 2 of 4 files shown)


colinhacks:
@pullfrog review

colinhacks:
No, it is just not possible to bubble up default values like this in a sound way.

PR #5770, opened 2026-03-13
feat: add .invert() method to ZodCodec

### LINKED ISSUE #5625: Feature Request: add `.invert()` method to codecs
Hey there, really enjoying the new codecs feature. I would like to request an `.invert()` (AKA `.inverse()`, `.inverted()`,  `.flip()`, `.flipped()` etc) method on Codecs, so that they can be _composed_ into other schema with arbitrary directionality. The method would effectively just flip the inputs and outputs:

```ts
const intToString = z.codec(z.int(), z.string().regex(z.regexes.integer), {
    decode: (num) => num.toString(),
    encode: (str) => Number.parseInt(str, 10),
});

const stringToInt = intToString.invert();
```

### Contrived Use-Case:

```ts
const intToString = z.codec(z.int(), z.string().regex(z.regexes.integer), {
    decode: (num) => num.toString(),
    encode: (str) => Number.parseInt(str, 10),
});

// ✅ Already possible:
const tenAsString = z.number().pipe(intToString).parse(10);

// ❌ Not yet possible:
const tenAsInteger = z.string().pipe(intToString.invert()).parse("10");
```

This example is obviously contrived, because in practice you would just use `.encode`:

```diff
-const tenAsInteger = z.string().pipe(intToString.invert()).parse("10");
+const tenAsInteger = intToString.encode("10");
```

**However**, this method would be useful when composing schema with codecs. Eg:

### Actual Use-Case:

It's common in `react-hook-form` to define validation schema with `zod` that have an `Input` type of `string`, but a transformed `Output` type, eg:

```ts
import { useForm } from "react-hook-form";
import { zodResolver } from "@hookform/resolvers/zod";
import { intToString } from "lib/intToString";

const formValidationSchema = z.object({
  age: intToString().invert(), // `string` as input, `number` as output.
});

export function MyForm() {
  const form = useForm({
    resolver: zodResolver(formValidationSchema),
  });
  
  // …
}
```

mahmoodhamdi:
## Summary

Closes #5625

Adds an `.invert()` method to `ZodCodec` (and `ZodMiniCodec`) that returns a new codec with swapped input/output schemas and swapped decode/encode transforms.

### Usage

```ts
const isoDateCodec = z.codec(z.iso.datetime(), z.date(), {
  decode: (isoString) => new Date(isoString),
  encode: (date) => date.toISOString(),
});

// Inverted: Date → ISO string
const dateToIso = isoDateCodec.invert();

z.decode(dateToIso, new Date("2024-01-15T10:30:00.000Z"));
// → "2024-01-15T10:30:00.000Z"

z.encode(dateToIso, "2024-01-15T10:30:00.000Z");
// → Date object
```

This is useful when composing schemas in pipelines where the direction needs to be flipped — for example, form validation schemas in `react-hook-form` where the input type is `string` but the output should be a parsed value.

## Implementation

The method creates a new codec with:
- `def.in` ↔ `def.out` swapped
- `def.transform` ↔ `def.reverseTransform` swapped

Zero runtime overhead — just a new instance with references swapped.

### Files changed

- `packages/zod/src/v4/classic/schemas.ts` — `ZodCodec` interface + constructor
- `packages/zod/src/v4/mini/schemas.ts` — `ZodMiniCodec` interface + constructor
- `packages/zod/src/v4/classic/tests/codec.test.ts` — 5 new test cases

## Test plan

- [x] Basic invert: decode/encode directions swapped correctly
- [x] Round trip: invert preserves data through decode→encode cycle
- [x] Type inference: `z.input`/`z.output` types correctly swapped
- [x] Double invert: `.invert().invert()` behaves like original
- [x] Invert with pipe: works with piped schemas
- [x] Full test suite passes (3585 tests, 0 type errors)

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -2098,20 +2098,10 @@ export interface ZodCodec<A extends core.SomeType = core.$ZodType, B extends cor
   "~standard": ZodStandardSchemaWithJSON<this>;
   _zod: core.$ZodCodecInternals<A, B>;
   def: core.$ZodCodecDef<A, B>;
-  invert(): ZodCodec<B, A>;
 }
 export const ZodCodec: core.$constructor<ZodCodec> = /*@__PURE__*/ core.$constructor("ZodCodec", (inst, def) => {
   ZodPipe.init(inst, def);
   core.$ZodCodec.init(inst, def);
-
-  inst.invert = () =>
-    new ZodCodec({
-      type: "pipe",
-      in: def.out,
-      out: def.in,
-      transform: def.reverseTransform,
-      reverseTransform: def.transform,
-    }) as any;
 });
 
 export function codec<const A extends core.SomeType, B extends core.SomeType = core.$ZodType>(
@@ -2128,6 +2118,17 @@ export function codec<const A extends core.SomeType, B extends core.SomeType = c
     out: out as any as core.$ZodType,
     transform: params.decode as any,
     reverseTransform: params.encode as any,
+  }) as any;
+}
+
+export function invertCodec<A extends core.SomeType, B extends core.SomeType>(codec: ZodCodec<A, B>): ZodCodec<B, A> {
+  const def = codec._zod.def;
+  return new ZodCodec({
+    type: "pipe",
+    in: def.out as any,
+    out: def.in as any,
+    transform: def.reverseTransform as any,
+    reverseTransform: def.transform as any,
   }) as any;
 }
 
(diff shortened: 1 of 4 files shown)


colinhacks:
Interesting idea, but I don't think this should be a method. I'd be open to a top level `z.invertCodec()` utility function though!

colinhacks:
Pushed a follow-up that drops `.invert()` and replaces it with a top-level `z.invertCodec(codec)` utility (mirrored in `zod/mini`). Also rebased onto `main`. Going to merge once CI is green.

```ts
const dateToIso = z.invertCodec(isoDateCodec);
z.decode(dateToIso, new Date()); // → ISO string
```

colinhacks:
Merged — landed as a top-level `z.invertCodec(codec)` utility (mirrored in `zod/mini`) rather than a `.invert()` method on `ZodCodec`, in line with the project's preference for top-level utilities over methods on schema instances. Closes #5625. Thanks for putting this together!

colinhacks:
Landed in Zod 4.4

PR #5772, opened 2026-03-13
fix(fromJSONSchema): handle circular references in dereferenced schemas

### LINKED ISSUE #5675: z.fromJSONSchema is not supporting circular references (dereferenced schema)
In lot of schema based projects, we have big schema files 9swagger, ...) from where we extract a subpart (objet definition, ...)
As this definition may have reference to original big schema, it is totally useless as it.

That's why lot of project use dereferencing (@apidevtools/json-schema-ref-parser mainly)

But when we use it, then call z.fromSchema() we face a Maximum call stack size exceeded error

you can reproduce with below basic example

```
const testSchema = {
  type: 'object',
  properties: {
    recursiveProp: undefined as any
  }
};
testSchema.properties.recursiveProp = testSchema;
const zSchema = z.fromJSONSchema(testSchema as z.core.JSONSchema.JSONSchema)
```

Here you may create Zot internal references ? or return any for the section that has a circularity ?

mahmoodhamdi:
## Summary

Fixes #5675

`z.fromJSONSchema` crashes with "Maximum call stack size exceeded" when given a **dereferenced** JSON Schema that contains circular object references (e.g., a `Person` schema where `bestFriend` property references the same `Person` schema object).

The existing circular reference handling only tracks `$ref` string paths. When schemas are fully dereferenced (all `$ref` replaced with inline objects), the same schema object can appear at multiple positions. Without object-identity tracking, the converter recurses infinitely.

### Changes

- Added `WeakMap`/`WeakSet`-based object identity tracking to `ConversionContext` alongside the existing string-based `$ref` tracking
- When a schema object is encountered that is already being processed, it is wrapped in `z.lazy()` to break the cycle — matching the same pattern used for `$ref` cycles
- Added 3 tests covering: self-referencing objects, circular arrays, and mutual circular references

### Before
```ts
const person: any = { type: "object", properties: { name: { type: "string" } } };
person.properties.bestFriend = person; // circular by object identity

z.fromJSONSchema(person); // ❌ Maximum call stack size exceeded
```

### After
```ts
z.fromJSONSchema(person); // ✅ Works, circular ref resolved via z.lazy()
schema.parse({ name: "Alice", bestFriend: { name: "Bob" } }); // ✅
```

## Test plan

- [x] Added test for self-referencing circular object schema
- [x] Added test for circular array schema (node with children array of nodes)
- [x] Added test for mutual circular references (parent ↔ child)
- [x] All 3581 existing tests continue to pass
- [x] No type errors

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -26,6 +26,10 @@ interface ConversionContext {
   processing: Set<string>;
   rootSchema: JSONSchema.JSONSchema;
   registry: $ZodRegistry<any>;
+  /** Tracks already-converted schema objects by identity (for dereferenced schemas) */
+  schemaObjects: WeakMap<object, ZodType>;
+  /** Tracks schema objects currently being converted (cycle detection) */
+  processingSchemas: WeakSet<object>;
 }
 
 // Keys that are recognized and handled by the conversion logic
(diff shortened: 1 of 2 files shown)


colinhacks:
Going to close this — took a different approach on main in [`87cf0f93`](https://github.com/colinhacks/zod/commit/87cf0f93cd0f34bdc69f11c9377568e6812841c4). Instead of object-identity tracking that turns cyclic JS object inputs into `z.lazy()`, `fromJSONSchema` now does `JSON.parse(JSON.stringify(input))` at the top. Cyclic input throws a clear error, and class instances, Proxies, and getter-based properties get normalized into plain objects so the converter only ever walks a finite plain-object graph.

The reasoning: `fromJSONSchema`'s contract is JSON Schema in, and JSON has no cycles. A cyclic JS object input isn't really JSON Schema — it's a non-standard graph from some deref tool. Drawing the line there feels cleaner than inventing implicit `z.lazy()` boundaries from JS reference identity, which only works for one specific shape of cyclic input anyway (a getter-based cycle would still stack-overflow under this PR's approach).

Thanks for the PR and especially the repro cases — they made it clear how to think about the contract.

PR #5773, opened 2026-03-13
fix: distinguish ZodCoerced* types from non-coerced via phantom brand

### LINKED ISSUE #5692: [Bug] There is no difference in ZodNumber and ZodCoercedNumber .
Consider the following example:

```ts
import { z } from "zod";

function sample(schema: z.ZodCoercedNumber){
    return schema.parse("123")
}

sample(z.number())
```

No type error is emitted, even though `z.number()` does not coerce strings into numbers (unlike `z.coerce.number()`). As a result, this code compiles successfully but fails at runtime, causing a silent and potentially confusing bug.

Expected behavior

Passing a `ZodNumber` where a `ZodCoercedNumber` is expected should result in a type error, since their runtime behaviors are different.

Suggestion

Introduce a stricter type distinction so that:

- `ZodCoercedNumber` is assignable to `ZodNumber` (since it is effectively a specialized version of it), but

- `ZodNumber` is not assignable to `ZodCoercedNumber`.

This would help TypeScript catch cases where coercion is required but not actually provided, preventing subtle runtime errors.

mahmoodhamdi:
## Summary

Fixes #5692

`ZodCoercedNumber` and `ZodNumber` (and all other coerced/non-coerced pairs) were structurally identical at the TypeScript type level. This allowed passing `z.number()` where `z.coerce.number()` is expected — compiles fine but crashes at runtime:

```ts
function sample(schema: z.ZodCoercedNumber) {
  return schema.parse("123"); // expects coercion
}
sample(z.number()); // No type error, but fails at runtime!
```

### Fix

Add a `readonly "~coerced": true` phantom brand property to all 5 `ZodCoerced*` interfaces (`ZodCoercedString`, `ZodCoercedNumber`, `ZodCoercedBoolean`, `ZodCoercedBigInt`, `ZodCoercedDate`). This follows the same naming convention as the existing `~standard` property.

With the phantom brand:
- `ZodNumber` is **not** assignable to `ZodCoercedNumber` (missing `~coerced` — prevents the bug)
- Runtime behavior is completely unchanged (phantom properties don't exist at runtime)

### After

```ts
sample(z.number());        // TS error: Property '~coerced' is missing
sample(z.coerce.number()); // OK
```

## Test plan

- [x] Added type-level test verifying all 5 non-coerced types are NOT assignable to their coerced counterparts
- [x] All 3577 existing tests pass
- [x] No type errors

### DIFF
--- a/packages/zod/src/v4/classic/coerce.ts
+++ b/packages/zod/src/v4/classic/coerce.ts
@@ -1,27 +1,37 @@
 import * as core from "../core/index.js";
 import * as schemas from "./schemas.js";
 
-export interface ZodCoercedString<T = unknown> extends schemas._ZodString<core.$ZodStringInternals<T>> {}
+export interface ZodCoercedString<T = unknown> extends schemas._ZodString<core.$ZodStringInternals<T>> {
+  readonly "~coerced": true;
+}
 export function string<T = unknown>(params?: string | core.$ZodStringParams): ZodCoercedString<T> {
   return core._coercedString(schemas.ZodString, params) as any;
 }
 
-export interface ZodCoercedNumber<T = unknown> extends schemas._ZodNumber<core.$ZodNumberInternals<T>> {}
+export interface ZodCoercedNumber<T = unknown> extends schemas._ZodNumber<core.$ZodNumberInternals<T>> {
+  readonly "~coerced": true;
+}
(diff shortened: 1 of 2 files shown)


mahmoodhamdi:
Closing — apologies for the noise.

ArthurLobopro:
I am having some trouble because of this behavior. Any previsions for this bugfix?

PR #5774, opened 2026-03-13
fix: apply key transforms in record with enum/literal keys

### LINKED ISSUE #5296: Records do not transform keys, even though the Typescript types says it does.

```typescript
z.record(
  z.string(),
);
```

mahmoodhamdi:
## Summary

Fixes #5296

`z.record()` with a key schema that has known finite values (enum/literal) and a `.transform()` does not apply the key transformation. The output contains the original keys instead of the transformed ones.

This was inconsistent with `z.partialRecord()`, which correctly applies key transforms.

### Before

```ts
const schema = z.record(
  z.string()
);
schema.parse({ a: "hello" }); // { a: "hello" } — key NOT transformed
```

### After

```ts
schema.parse({ a: "hello" }); // { b: "hello" } — key transformed correctly
```

### Root cause

When a record's key schema has known finite values (via `_zod.values`), the parser takes a fast path that iterates those values directly and looks them up in the input object. This fast path never ran the key through `def.keyType._zod.run()`, so any transforms on the key schema were silently ignored.

`z.partialRecord()` worked correctly because it sets `_zod.values = undefined`, forcing the general code path that does run keys through the key schema.

### Fix

Run each key through `def.keyType._zod.run()` in the fast path and use the transformed value (`keyResult.value`) as the output key.

## Test plan

- [x] Added test for record with literal key transform
- [x] Added test for record with enum key transform
- [x] Added test verifying record/partialRecord consistency
- [x] All 3577 existing tests pass
- [x] No type errors

### DIFF
--- a/packages/zod/src/v4/classic/tests/record.test.ts
+++ b/packages/zod/src/v4/classic/tests/record.test.ts
@@ -630,3 +630,29 @@ test("numeric string keys", () => {
   );
   expect(transformedSchema.parse({ 5: "five", 10: "ten" })).toEqual({ 10: "five", 20: "ten" });
 });
+
+test("record with literal key transform applies key transformation", () => {
+  // When a record has finite key values (literal/enum) and a key transform,
+  // the output keys should be transformed (consistent with partialRecord)
+  const schema = z.record(
+    z.literal("a").transform(() => "b" as const),
+    z.string()
+  );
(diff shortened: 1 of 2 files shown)


mahmoodhamdi:
Closing — apologies for the noise.

PR #5791, opened 2026-03-18
fix(docs): heading anchor links now include the hash so it doesnt scoll all the way up, follows navbar logic

### LINKED ISSUE #5788: Buggy navigation on docs
Steps:
1) Navigate to https://zod.dev/api in a new tab
2) Click heading "Coercion" (_not_ the nav link on the right)
3) Address updates to https://zod.dev/api?id=coercion, page aligns with Coercion section
4) Click heading "Coercion" again
5) Page jumps to top instead of staying at the "Coercion" section

Obviously would expect it to stay on Coercion section. The right side nav uses https://zod.dev/api?id=coercion#coercion which works reliably instead.

I also experienced sometimes that the first click on a heading would cause a full page reload and align at the top instead of working the first time, or would trigger a full reload then align correctly after a delay. These are intermittent, I can't tell what causes this to happen sometimes and not other times.

FucciUnavailable:
Fixes #5788

## Problem
Clicking a heading (e.g. "Coercion") on a docs page like `/api` sets the
URL to `?id=coercion`. Clicking the same heading a second time causes the
page to jump to the top instead of staying at the section.

This happens because the `<Link>` only sets a query param (`?id=coercion`).
When the URL hasn't changed, Next.js treats it as a no-op navigation —
no scroll happens, and the page snaps to top via scroll restoration.

The right-side TOC nav already uses `?id=coercion#coercion` (both query
param and hash), which works reliably on every click.

## Fix
Add the hash to the heading anchor href to match what the TOC nav already does:

- Before: `?id=${props.id}`
- After:  `?id=${props.id}#${props.id}`

This means every click is always a real URL change (the hash is always
present), and the browser's native anchor scrolling handles positioning
correctly and idempotently.

### DIFF
--- a/packages/docs/components/heading.tsx
+++ b/packages/docs/components/heading.tsx
@@ -19,23 +19,23 @@ export function Heading<T extends Types = "h1">({ as, className, ...props }: Hea
       <Link
         data-card=""
         // href={`#${props.id}`}
-        href={`?id=${props.id}`}
+        href={`?id=${props.id}#${props.id}`}
         // onclick="event.preventDefault(); history.pushState(null, '', '?asdf=qwer');"
         className="peer"
-        // shallow={true}
-        // onClick={(e) => {
-        //   // function __handleScroll(){
-        //   // if id query parameter is present, scroll to the element with that id
-        //   const params = new URLSearchParams(window.location.search);
-        //   console.dir(params, { depth: null });
-        //   const id = params.get("id");
-        //   console.dir(params, { depth: null });
-        //   if (id) {
-        //     console.dir(document.getElementById(id), { depth: null });
-        //     document.getElementById(id)?.scrollIntoView();
-        //   }
-        //   // }
-        // }}
+      // shallow={true}
+      // onClick={(e) => {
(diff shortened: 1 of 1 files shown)


colinhacks:
@pullfrog review

FucciUnavailable:
I can get the indentation fixed!

colinhacks:
Thanks for the PR! Just wanted to flag that there's already a `Scroller` component (`packages/docs/components/scroller.tsx`) that handles this — it watches `searchParams` and calls `scrollIntoView()` on the element matching `?id=…` whenever it changes. That's why the current `?id=foo` href works on first click without a hash.

Going to close this out for now, but happy to revisit if there's a repro where the `Scroller` path isn't doing the right thing. Thanks again!

FucciUnavailable:
I appreciate your time reviewing this!

colinhacks:
Sorry, I understand the problem a little better now. Thanks for the PR!

FucciUnavailable:
I'm so glad!!
Thank you!! Excited to contribute more!!

PR #5793, opened 2026-03-19
Fix: correct multipleOf validation for floating point values

### LINKED ISSUE #5792: `multipleOf` silently accepts non-multiples for small numbers (scientific notation bug in `floatSafeRemainder`)
`z.number().multipleOf(1e-7).safeParse(2.5e-7)` returns `{ success: true }`,  but `2.5e-7 / 1e-7 = 2.5`  (not a whole number).

Repro:
```
z.number().multipleOf(1e-7).safeParse(2.5e-7)  //  { success: true } — should be false
z.number().multipleOf(1e-7).safeParse(3.3e-7)  // { success: true } — should be false
z.number().multipleOf(1e-7).safeParse(3e-7)    // { success: true } — correct
```

Root cause: `floatSafeRemainder` in util.ts counts decimal places by splitting `val.toString()` on ".". For small numbers, JS uses scientific notation: (2.5e-7).toString() gives "2.5e-7", so the fractional part is "5e-7" (length 4) instead of the actual 8 decimal places. 

The function already handles this for the step parameter (lines 251-255) but not for val.

Suggested fix: Apply the same scientific notation regex to val.toString().

Zod version: 4.3.6

cyphercodes:
## Problem

`z.number().multipleOf(1e-7).safeParse(2.5e-7)` incorrectly returns success. The value `2.5e-7` is **not** an integer multiple of `1e-7` (since `2.5e-7 / 1e-7 = 2.5`), yet validation passes.

### Root Cause

The `floatSafeRemainder` function parses `.toString()` representations to count decimal places. When values use scientific notation (e.g. `2.5e-7`), `split('.')` gives misleading results (`"5e-7"` with length 3), leading to incorrect integer conversions and a false pass.

## Solution

Replace the string-manipulation-based approach with a tolerance-based ratio comparison:

```ts
const ratio = val / step;
const roundedRatio = Math.round(ratio);
const tolerance = Number.EPSILON * Math.max(Math.abs(ratio), 1);
if (Math.abs(ratio - roundedRatio) < tolerance) return 0;
```

This checks whether `val / step` is close to an integer, which is the mathematical definition of "multiple of". The tolerance uses `Number.EPSILON` scaled to the ratio magnitude to correctly handle IEEE 754 representation limits at any scale.

## Changes

- **`packages/zod/src/v4/core/util.ts`** — Rewrote `floatSafeRemainder`
- **`packages/zod/src/v3/types.ts`** — Same fix for v3 backward compatibility
- **`packages/zod/src/v4/classic/tests/number.test.ts`** — Added regression test for #5792

## Test Results

All **3577 tests pass** across 323 test files with 0 type errors.

Closes #5792

### DIFF
--- a/packages/zod/src/v3/types.ts
+++ b/packages/zod/src/v3/types.ts
@@ -1352,12 +1352,12 @@ export type ZodNumberCheck =
 
 // https://stackoverflow.com/questions/3966484/why-does-modulus-operator-return-fractional-number-in-javascript/31711034#31711034
 function floatSafeRemainder(val: number, step: number) {
-  const valDecCount = (val.toString().split(".")[1] || "").length;
-  const stepDecCount = (step.toString().split(".")[1] || "").length;
-  const decCount = valDecCount > stepDecCount ? valDecCount : stepDecCount;
-  const valInt = Number.parseInt(val.toFixed(decCount).replace(".", ""));
-  const stepInt = Number.parseInt(step.toFixed(decCount).replace(".", ""));
-  return (valInt % stepInt) / 10 ** decCount;
+  const ratio = val / step;
+  const roundedRatio = Math.round(ratio);
+  // Use a relative epsilon scaled to the magnitude of the result
+  const tolerance = Number.EPSILON * Math.max(Math.abs(ratio), 1);
+  if (Math.abs(ratio - roundedRatio) < tolerance) return 0;
+  return ratio - roundedRatio;
 }
 
 export interface ZodNumberDef extends ZodTypeDef {
--- a/packages/zod/src/v4/classic/tests/number.test.ts
+++ b/packages/zod/src/v4/classic/tests/number.test.ts
@@ -169,6 +169,20 @@ test(".multipleOf() with scientific notation (multi-digit exponents)", () => {
   expect(schema15.parse(3e-15)).toEqual(3e-15);
 });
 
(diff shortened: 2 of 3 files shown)


cyphercodes:
Good catch — I audited all call sites and confirmed **no code depends on the non-zero magnitude**:

- **v4** (`checks.ts:182`): `util.floatSafeRemainder(payload.value, def.value as number) === 0`
- **v3** (`types.ts:1429`): `floatSafeRemainder(input.data, check.value) !== 0`

These are the only two call sites in the entire codebase (grep confirmed, including tests). Both exclusively check `=== 0` / `!== 0` — the actual non-zero value is never consumed. So the shift from value-space remainder to multiple-space ratio is safe.

colinhacks:
@pullfrog review

colinhacks:
Merged as-is. The tolerance-based ratio approach is the right fix here — the old string-counting algorithm had latent issues beyond scientific notation (`toFixed` range errors past ~100 decimals, `0.1 + 0.1 + 0.1` famously not registering as a multiple of `0.1`) that the ratio comparison handles cleanly, which is also why I closed #5823 in favor of this. Scaling `Number.EPSILON` by `Math.max(|ratio|, 1)` is the correct ULP-aware tolerance. The call-site audit confirming nothing consumes the non-zero remainder magnitude is what made this safe to land as a behavior change — solid investigation.

colinhacks:
Landed in Zod 4.4

PR #5794, opened 2026-03-19
feat: implement type narrowing in z.string().startsWith() and z.string().endsWith()

wojtekmaj:
Related to #1747

The fix has previously been attempted 4 years ago in:
- #1730 - had reservations related to ZodString readability
- #1751 - got closed as stale

Unlike #1730, this PR does not modify `ZodString`, and is instead adding `_ZodStringWithOutput` helper.
Unlike #1751, this PR includes tests and does not use type assertions.

That one required quite a lot of gymnastics to get green and it is not the prettiest PR I've ever made, but hey, it works!

Alternatives considered: making `.startsWith` and `.endsWith` proxy to `.templateLiteral`. Abandoned as `.startsWith` and `.endsWith` may overlap.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -153,6 +153,68 @@ export interface ZodType<
 export interface _ZodType<out Internals extends core.$ZodTypeInternals = core.$ZodTypeInternals>
   extends ZodType<any, any, Internals> {}
 
+type _ZodStringWithOutput<
+  Schema extends _ZodString,
+  Output extends string,
+  Input = Schema["_zod"]["input"],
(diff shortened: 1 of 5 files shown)


wojtekmaj:
@themavik That's widening is very much intentional. Consider the following example:

```
z.string().startsWith('fooba').endsWith('oobar').parse('foobar')
```

```
`${'fooba'}${string}` & `${string}${'oobar'}`
```

simply must be 

```
string
```

!

colinhacks:
This is behaving as intended right now. People can use template literals to achieve this result if they want to achieve this. For `z.string()` methods I am more concerned about keeping the IntelliSense clean. Thanks 👍

PR #5797, opened 2026-03-19
fix(v4): resolve stack overflow in toJSONSchema for recursive lazy with describe

### LINKED ISSUE #5777: z.toJSONSchema() stack overflows on recursive z.lazy() union schemas when recursive branches use .describe()
# Zod Issue Draft

Title:

`z.toJSONSchema()` stack overflows on recursive `z.lazy()` union schemas when recursive branches use `.describe()`

Body:

### Zod version

4.3.6

### What I expected

`z.toJSONSchema()` should serialize a recursive schema using `$ref` when called with:

```ts
z.toJSONSchema(schema, { cycles: "ref", reused: "ref" })
```

### What happened

It throws:

```text
RangeError: Maximum call stack size exceeded
```

This seems related to recursive `z.lazy()` unions when the recursive branches are decorated with `.describe()`.

The same recursive schema shape works if I remove the `.describe()` calls.

### Minimal reproduction

```ts
import { z } from "zod";

type FieldDefinitionValue =
  | { type: "text" | "int"; required?: boolean }
  | { object: Record<string, FieldDefinitionValue>; required?: boolean }
  | { array: FieldDefinitionValue; required?: boolean };

const Scalar = z.enum(["text", "int"]).describe("scalar kind");

const Field: z.ZodType<FieldDefinitionValue> = z.lazy(() =>
  z
    .union([
      z
        .object({
          type: Scalar.describe("scalar field kind"),
          required: z.boolean().optional().describe("required flag"),
        })
        .strict(),
      z
        .object({
          object: z
            .record(z.string(), Field)
            .describe("nested object field map"),
          required: z.boolean().optional().describe("required flag"),
        })
        .strict(),
      z
        .object({
          array: Field.describe("array element field definition"),
          required: z.boolean().optional().describe("required flag"),
        })
        .strict(),
    ])
    .describe("field definition"),
);

const Root = z
  .object({
    fields: z.record(z.string(), Field).describe("field map"),
  })
  .strict()
  .describe("root");

console.log(
  z.toJSONSchema(Root, {
    cycles: "ref",
    reused: "ref",
  }),
);
```

### Notes

- If I remove the `.describe()` calls from the recursive schema branches, `z.toJSONSchema()` succeeds.
- If I keep recursion but only use `.meta(...)`, `z.toJSONSchema()` also succeeds.
- This came up while trying to generate Mermaid class diagrams from Zod contract schemas.
````

### PR DESCRIPTION
**Pull request by Hassad674**
Fixes #5777

When `.describe()` or `.meta()` is used on branches inside a recursive `z.lazy()` schema, `toJSONSchema()` hits a stack overflow. The root cause is that each described clone of a lazy schema re-evaluates the getter function, creating fresh schema instances at every recursion level — none of which are in the `seen` cache.

The fix walks the parent chain in `lazyProcessor` to find the root lazy schema and reuses its cached `innerType`. This ensures all clones of the same recursive lazy share one set of resolved inner schemas, so the `seen` check stops recursion as expected.

Added a test with a recursive tree schema where every branch uses `.describe()` to verify the fix.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2988,3 +2988,18 @@ test("cycle detection - mutual recursion", () => {
     Set the \`cycles\` parameter to \`"ref"\` to resolve cyclical schemas with defs.]
   `);
 });
+
+test("recursive lazy with describe does not stack overflow", () => {
+  const NodeSchema: z.ZodType = z.lazy(() =>
+    z
+      .object({
+        value: z.string().describe("node value"),
+        children: z.array(NodeSchema.describe("child node")).optional().describe("child list"),
+      })
+      .describe("tree node")
+  );
+
+  const result = z.toJSONSchema(NodeSchema, { cycles: "ref", reused: "ref" });
+  expect(result).toBeDefined();
+  expect(result.$defs).toBeDefined();
+});
--- a/packages/zod/src/v4/core/json-schema-processors.ts
+++ b/packages/zod/src/v4/core/json-schema-processors.ts
@@ -556,15 +556,7 @@ export const optionalProcessor: Processor<schemas.$ZodOptional> = (schema, ctx, 
 };
 
 export const lazyProcessor: Processor<schemas.$ZodLazy> = (schema, ctx, _json, params) => {
-  // For clones of lazy schemas (e.g. via .describe()/.meta()), walk up to the
-  // root lazy and reuse its cached innerType. Each clone would otherwise
-  // re-evaluate the getter, producing new schema objects per recursion level
-  // and causing a stack overflow on recursive schemas.
-  let root = schema as schemas.$ZodLazy;
-  while (root._zod.parent && (root._zod.parent as schemas.$ZodType)._zod.def.type === "lazy") {
-    root = root._zod.parent as schemas.$ZodLazy;
-  }
-  const innerType = root._zod.innerType;
+  const innerType = (schema as schemas.$ZodLazy)._zod.innerType;
   process(innerType, ctx as any, params);
   const seen = ctx.seen.get(schema)!;
   seen.ref = innerType;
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by colinhacks** [comment]
Pushed 57f252bf on top, which moves the fix out of `lazyProcessor` and into the `$ZodLazy` constructor. Going to merge as-is.

The root cause is in how lazy clones cache their inner type. `.describe()` / `.meta()` produce a clone via `core.clone(inst)`, which calls `new inst._zod.constr(def)` with the same `def`. The constructor then re-runs `util.defineLazy(inst._zod, "innerType", () => def.getter())` on the clone, so each clone has its own `innerType` cache pointing at the same shared `def.getter`. On a recursive schema, accessing `innerType` through one clone runs `getter()` and constructs a fresh inner subtree (including a fresh `.describe()` clone for any recursive reference), and accessing it through that fresh clone runs `getter()` again — `ctx.seen` never matches, and `toJSONSchema` overflows.

The original commit fixed this in `lazyProcessor` by walking `_zod.parent` up to the root lazy and using its `innerType`. That works for the JSON schema case but it's a symptomatic fix:

- Anywhere else that reads `lazy._zod.innerType` on a clone — `inst._zod.parse`, `pattern`, `propValues`, `optin`/`optout` — still re-evaluates the getter on every access, so a `.describe()`-cloned recursive lazy was rebuilding its inner subtree on every parse.
- The processor had to know about clone semantics and walk parent chains, which is leaky for what should be a generic three-line function.
- The walk-up only follows parents whose `def.type === "lazy"`. `core.clone` only sets `_zod.parent` when called without a `def`, so any change to the cloning path could silently break it.

The follow-up caches the resolved inner type on the shared `def` instead of on `inst._zod`:

```ts
util.defineLazy(inst._zod, "innerType", () => {
  const d = def as $ZodLazyDef & { _cachedInner?: $ZodType };
  if (!d._cachedInner) d._cachedInner = def.getter() as $ZodType;
  return d._cachedInner;
});
```

Since clones share `def`, every clone of the same lazy now resolves to the same inner instance. `lazyProcessor` reverts to the trivial form, and the parse path / computed properties get the same cycle-friendly behavior for free.

The test from the original commit is unchanged. Full suite (3577 tests) passes, no type errors.

colinhacks:
Merged — thanks for tracking this one down 👍. The follow-up I pushed caches `innerType` on the lazy `def` itself so every clone produced by `.describe()` or `.meta()` resolves to the same inner instance, which fixes the recursion at the source instead of having `toJSONSchema` walk the parent chain to find the root lazy. Same upside for `parse` and the other computed properties off `innerType` — they stop rebuilding the subtree on every access through a clone.

colinhacks:
Landed in Zod 4.4

PR #5802, opened 2026-03-24
docs: add documentation for .and() method

### LINKED ISSUE #5798: [Documentation] Add documentation for Zod and method
Based on the code, it looks like `SchemaA.and(SchemaB)` is just a wrapper for `z.intersection(SchemaA, SchemaB)` which is documented [here](https://zod.dev/api?id=intersections). However, `and` itself does not have any documentation.

It is used in the [looseObject docs](https://zod.dev/api?id=intersections):

```tsx
const schema = z
  .object({ name: z.string() })
  .and(z.looseRecord(z.string().regex(/_phone$/), z.e164()));
```

https://github.com/colinhacks/zod/blob/c7805073fef5b6b8857307c3d4b3597a70613bc2/packages/zod/src/v4/classic/schemas.ts#L224

hnshah:
## Fixes #5798

Adds documentation for the `.and()` method which was missing from the API reference.

### Changes
- Added dedicated section after Intersections explaining `.and()` usage
- Included code examples showing it's a shorthand for `z.intersection()`
- Demonstrated practical usage when chaining with other methods

### Context
The `.and()` method exists in the codebase as a convenience wrapper for `z.intersection()` but wasn't documented. It's already used in examples (like in the looseObject docs) but users had no reference documentation for it.

**Location in code:** [`packages/zod/src/v4/classic/schemas.ts#L224`](https://github.com/colinhacks/zod/blob/c7805073fef5b6b8857307c3d4b3597a70613bc2/packages/zod/src/v4/classic/schemas.ts#L224)

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -1829,6 +1829,27 @@ type EmployedPerson = z.infer<typeof EmployedPerson>;
   When merging object schemas, prefer [`A.extend(B)`](#extend) over intersections. Using `.extend()` will give you a new object schema, whereas `z.intersection(A, B)` returns a `ZodIntersection` instance which lacks common object methods like `pick` and `omit`.
 </Callout>
 
+### `.and()`
+
+For convenience, you can use the `.and()` method as shorthand for `z.intersection()`:
+
+```ts
+const Person = z.object({ name: z.string() });
+const Employee = z.object({ role: z.string() });
+
+// These are equivalent:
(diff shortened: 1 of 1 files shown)


colinhacks:
Thanks, but I'm trying to discourage usage of this method. I have loose plans to drop it in a future major version, should there be one. Ultimately the intersection logic is too complex to include on the base class, and dropping this method would let us reduce bundle sizes. Currently it remains undeprecated but I generally think people should be using the functional versions here.

PR #5811, opened 2026-03-27
fix(types): preserve concrete shape type in z.object() return type

### LINKED ISSUE #5810: `z.object` is type widening to `z.ZodObject<util.Writeable<T>, core.$strip>`
Given the following code

```typescript
import { z } from "zod";

export const schema = z.object({
  id: z.string(),
  title: z.string(),
  author: z.string(),
  publishYear: z.number().nullable(),
  isbn: z.string().nullable(),
  description: z.string().nullable(),
  source: z.enum(["OpenLibrary", "Wikidata"]),
});
```

On this version of zod `4.3.6` , the language server in in vscode & cursor on windows 10 is type widening to:

```typescript
const schema: z.ZodObject<util.Writeable<T>, core.$strip>
```

While on `zod@4.3.5` it shows the following correct type.

```typescript
const schema: z.ZodObject<{
    id: z.ZodString;
    title: z.ZodString;
    author: z.ZodString;
    publishYear: z.ZodNullable<z.ZodNumber>;
    isbn: z.ZodNullable<z.ZodString>;
    description: z.ZodNullable<z.ZodString>;
    source: z.ZodEnum<{
        OpenLibrary: "OpenLibrary";
        Wikidata: "Wikidata";
    }>;
}, z.core.$strip>
```

Note, I tried to replicate it in CodeSandbox but I couldn't. I've only been able to see this behavior on my dev environment.

Please let me know what I can help to reproduce this issue on your end as well.

GourangaDasSamrat:
## Problem

`z.object()` widens its return type to `ZodObject<util.Writeable<T>, core.$strip>`
in the language server (VSCode / Cursor), losing all field-level type information.
`strictObject()` and `looseObject()` are unaffected and infer correctly.

Fixes #5810.

## Root Cause

The return type annotation on `object()` used `util.Writeable<T>`:
```ts
): ZodObject<util.Writeable<T>, core.$strip>
```

`Writeable<T>` is a mapped type (`{ -readonly [P in keyof T]: T[P] } & {}`) that
strips `readonly` modifiers. As a side effect, TypeScript cannot resolve it back to
the original concrete `T`, so the language server shows the widened form instead of
the full shape.

## Fix

One-line change — use `T` directly, consistent with `strictObject()` and
`looseObject()`:
```ts
): ZodObject<T, core.$strip>
```

## Verification
```ts
const schema = z.object({
  id: z.string(),
  source: z.enum(["OpenLibrary", "Wikidata"]),
});
// Before: ZodObject<util.Writeable<T>, core.$strip>  
// After:  ZodObject<{ id: ZodString; source: ZodEnum<...> }, $strip>  
```

## CI Note

The CI failure on this PR is pre-existing and unrelated to this change — it is
a `moduleResolution=node10` deprecation error triggered by the "TypeScript latest"
job and exists on `main` as well.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1286,7 +1286,7 @@ export const ZodObject: core.$constructor<ZodObject> = /*@__PURE__*/ core.$const
 export function object<T extends core.$ZodLooseShape = Partial<Record<never, core.SomeType>>>(
   shape?: T,
   params?: string | core.$ZodObjectParams
-): ZodObject<util.Writeable<T>, core.$strip> {
+): ZodObject<T, core.$strip> {
   const def: core.$ZodObjectDef = {
     type: "object",
     shape: shape ?? {},


colinhacks:
Going to close this. The IntelliSense behavior described in #5810 doesn't reproduce in `tsc` on any TS version I tested (5.5–6.0) — the displayed type is always the expanded shape, never `util.Writeable<T>` literally. That's expected: `Writeable<T> = { -readonly [P in keyof T]: T[P] } & {}` is the standard "Prettify" pattern; the trailing `& {}` is what *forces* TS to expand the type inline rather than show the alias. The reporter on #5810 noted they couldn't reproduce outside their local environment either, which points to a stale `tsserver` / cache issue rather than a type-system bug.

Removing `Writeable<T>` from `z.object()` would also regress the displayed shape on getter-defined recursive schemas (which we recommend) and on shapes built from `as const`, where `readonly` modifiers would start leaking into hovers.

The real asymmetry is the other direction: `strictObject` and `looseObject` were missed when `Writeable<T>` was added in #4271, and they currently *do* leak `readonly` for the same inputs. Fixed that in #5882.

Thanks for digging into it.

PR #5818, opened 2026-03-30
Add .semver() string validation

### LINKED ISSUE #5813: Add SemVer string format.
While the string formats are pretty complete, I have been missing one to validate a Semantic Version string.

It's pretty trivial to implement given that the SemVer project itself already suggests a regexp for it [here](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string). But I still think it would be valuable to have it built-in.

I'm not familiar with the codebase myself but if any maintainer find it ok i'm more than happy to go about implementing it myself.

Thanks.

JBR-0100:
This PR adds the highly requested `.semver()` validation to `ZodString`, following the pattern of existing validators like `.email()` and `.uuid()`.

Key changes:
- Added `semver` to `StringValidation` type.
- Implemented `semverRegex` using the official recommended pattern from [semver.org](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string).
- Added `.semver()` method and `isSemver` getter to `ZodString`.
- Comprehensive test coverage in `string.test.ts`.

Closes #5813.

### DIFF
--- a/.configs/tsconfig.base.json
+++ b/.configs/tsconfig.base.json
@@ -29,6 +29,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "isolatedModules": true,
+    "ignoreDeprecations": "5.0",
 
     "pretty": true
   },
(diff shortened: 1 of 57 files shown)


colinhacks:
Thanks for the PR, but I don't think this is something I want to put into Zod core. Semver is a little too niche and I've never had anyone ask me for it. It also has a canonical package on NPM already so it's easy to wire up refinements and transforms that use the utils from `"semver"`.

Issue #5129, opened 2025-08-21
Cannot read properties of undefined (reading '_zod') in 4.0.17


jakobwgnr:
Getting following error using 4.0.17

TypeError: Cannot read properties of undefined (reading '_zod')
    at isTransforming (/home/jakobwgnr/projects/caro-backend/node_modules/zod/v4/core/to-json-schema.cjs:790:24)
    at isTransforming (/home/jakobwgnr/projects/caro-backend/node_modules/zod/v4/core/to-json-schema.cjs:815:21)
    at isTransforming (/home/jakobwgnr/projects/caro-backend/node_modules/zod/v4/core/to-json-schema.cjs:811:20)
    at isTransforming (/home/jakobwgnr/projects/caro-backend/node_modules/zod/v4/core/to-json-schema.cjs:854:20)
    at isTransforming (/home/jakobwgnr/projects/caro-backend/node_modules/zod/v4/core/to-json-schema.cjs:815:21)
    at isTransforming (/home/jakobwgnr/projects/caro-backend/node_modules/zod/v4/core/to-json-schema.cjs:854:20)
    at JSONSchemaGenerator.process (/home/jakobwgnr/projects/caro-backend/node_modules/zod/v4/core/to-json-schema.cjs:536:36)
    at JSONSchemaGenerator.process (/home/jakobwgnr/projects/caro-backend/node_modules/zod/v4/core/to-json-schema.cjs:207:57)
    at JSONSchemaGenerator.process (/home/jakobwgnr/projects/caro-backend/node_modules/zod/v4/core/to-json-schema.cjs:516:30)
    at JSONSchemaGenerator.process (/home/jakobwgnr/projects/caro-backend/node_modules/zod/v4/core/to-json-schema.cjs:510:30)

colinhacks:
Looks like a legit bug but I can't bedug this without knowing the code that produced it. Please open another issue with a complete runnable code snippet that reproduces this 👍

xdannyrobertsx:
FWIW, i just ran into this as well

it seems to be somewhat related to this issue with circular imports:
https://github.com/colinhacks/zod/issues/1193

i say that because moving imports fixed it for me

joshuambg:
For those chasing a solution, what worked for me: Rolling back to **zod@^3.25.76** AND including **"overrides": { "zod": "^3.25.76" }** in **package.json**. Version zod@^4.x.x did not work even with overrides. 

[...]

PR #5822, opened 2026-03-31
fix: add null-checks in toJSONSchema process() and isTransforming()

### LINKED ISSUE #5821: [Bug] toJSONSchema crashes on undefined schema inputs in process() and isTransforming()
**Zod version**: 4.3.6 (also reproduced on 4.3.5)
**Related**: #5129 (closed, but bug still present in 4.3.6)

## Description

`process()` and `isTransforming()` in `v4/core/to-json-schema.js` access `schema._zod.def` without null-checking the input. When a record schema has an undefined `valueType` (e.g., from JSON Schema → Zod conversion generating schemas with optional record types), the `recordProcessor` passes `undefined` to these functions, causing a crash.

## Error

```
TypeError: Cannot read properties of undefined (reading '_zod')
    at process (zod/v4/core/to-json-schema.js:33)
```

And after patching `process()`:

```
TypeError: Cannot read properties of undefined (reading '_zod')
    at isTransforming (zod/v4/core/to-json-schema.js:365)
```

## Crash Sites

**`process()` — line 33:**
```js
export function process(schema, ctx, _params = { path: [], schemaPath: [] }) {
    var _a;
    // ❌ No null-check — crashes if schema is undefined
    const def = schema._zod.def;
```

**`isTransforming()` — line 365:**
```js
function isTransforming(_schema, _ctx) {
    // ❌ No null-check — crashes if _schema is undefined
    const ctx = _ctx ?? { seen: new Set() };
    // ...
    const def = _schema._zod.def;
```

## Root Cause

The `recordProcessor` in `json-schema-processors.js` calls `process(def.valueType, ...)`. For some record schemas (particularly those generated from JSON Schema → Zod conversion via `json-schema-to-zod`), `def.valueType` can be `undefined`. Similarly, `isTransforming()` recursively processes `def.valueType` for record/map types without checking if it's defined first.

## Reproduction Context

This affects anyone using the MCP (Model Context Protocol) ecosystem with Zod v4:

1. `@modelcontextprotocol/sdk` calls `toJSONSchema` internally to serialize tool parameter schemas for `tools/list`
2. `@payloadcms/plugin-mcp` uses `json-schema-to-zod` to convert collection JSON Schemas → Zod schemas
3. Some generated Zod record schemas have `undefined` `valueType` in their `_zod.def`
4. `tools/list` → `toJsonSchemaCompat()` → `toJSONSchema()` → `process(undefined)` → crash

## Proposed Fix

Simple null guards at the entry of both functions:

```diff
 export function process(schema, ctx, _params = { path: [], schemaPath: [] }) {
     var _a;
+    if (!schema || !schema._zod) return {};
     const def = schema._zod.def;
```

```diff
 function isTransforming(_schema, _ctx) {
+    if (!_schema || !_schema._zod) return false;
     const ctx = _ctx ?? { seen: new Set() };
```

These are safe defaults — an undefined schema has no JSON Schema representation (empty object) and is not transforming (false).

## Relationship to #5129

Issue #5129 reported the identical crash on v4.0.17 and was closed. However, we've verified the bug still exists in **v4.3.6** — the `to-json-schema.js` source has no null-checks in either function. The only change between 4.3.5 and 4.3.6 in this file is an unrelated `refSeen.def` guard at line 256.

## Impact

This blocks the entire MCP tool listing for any project using `@payloadcms/plugin-mcp` (or any MCP server that generates Zod schemas from JSON Schema) with Zod v4. The `tools/list` response fails completely, making all MCP tools inaccessible. We're currently working around this with `pnpm patch`.

ShaunMWallace:
## Summary

Adds null-guards to `process()` and `isTransforming()` in `v4/core/to-json-schema.ts` to prevent crashes when these functions receive `undefined` or non-Zod schema inputs.

## Problem

Both functions access `schema._zod.def` without checking if the input is defined:

- `recordProcessor` in `json-schema-processors.ts` calls `process(def.valueType, ...)` — but `def.valueType` can be `undefined` for record schemas generated from JSON Schema → Zod conversion
- `isTransforming()` recursively processes `def.valueType` for record/map/set types without checking if it's defined

This crashes with:
```
TypeError: Cannot read properties of undefined (reading '_zod')
    at process (v4/core/to-json-schema.js:33)
```

## Fix

Simple early-return guards at the entry of both functions:

```typescript
// process() — return empty schema for undefined inputs
if (!schema || !schema._zod) return {};

// isTransforming() — undefined schemas are not transforming
if (!_schema || !(_schema as any)._zod) return false;
```

These are safe defaults — an undefined schema has no JSON Schema representation (empty object) and is not transforming (false).

## Impact

This bug blocks the entire MCP (Model Context Protocol) tool listing for any project using Zod v4 schemas generated from JSON Schema conversion. Specifically:

1. `@modelcontextprotocol/sdk` calls `toJSONSchema` to serialize tool parameter schemas for `tools/list`
2. `@payloadcms/plugin-mcp` uses `json-schema-to-zod` to convert collection JSON Schemas → Zod schemas
3. Some generated record schemas have `undefined` `valueType` in their `_zod.def`
4. `tools/list` → `toJsonSchemaCompat()` → `toJSONSchema()` → `process(undefined)` → crash

The `tools/list` response fails completely, making all MCP tools inaccessible.

## Related Issues

- Fixes #5821
- Related to #5129 (reported same crash on v4.0.17, closed but fix not present in v4.3.6)

### DIFF
--- a/packages/zod/src/v4/core/to-json-schema.ts
+++ b/packages/zod/src/v4/core/to-json-schema.ts
@@ -141,6 +141,11 @@ export function process<T extends schemas.$ZodType>(
   ctx: ToJSONSchemaContext,
   _params: ProcessParams = { path: [], schemaPath: [] }
 ): JSONSchema.BaseSchema {
+  // Guard against undefined or non-Zod schema inputs. This can happen when
+  // recordProcessor passes def.valueType which may be undefined for some
+  // schemas generated from JSON Schema → Zod conversion.
+  if (!schema || !schema._zod) return {};
+
   const def = schema._zod.def as schemas.$ZodTypes["_zod"]["def"];
 
   // check for schema in seens
@@ -521,6 +526,10 @@ function isTransforming(
     seen: Set<schemas.$ZodType>;
   }
 ): boolean {
+  // Guard against undefined or non-Zod schema inputs. Recursive calls from
+  // record/map/set types may pass def.valueType which can be undefined.
+  if (!_schema || !(_schema as any)._zod) return false;
+
   const ctx = _ctx ?? { seen: new Set() };
 
   if (ctx.seen.has(_schema)) return false;


ShaunMWallace:
Note: The CI failures are pre-existing and unrelated to this PR. Both test jobs fail during the build step with:

```
error TS5107: Option 'moduleResolution=node10' is deprecated and will stop functioning in TypeScript 7.0.
```

This is a `tsconfig.build.json` issue in the repo — the `main` branch has the same failure. This PR only modifies `packages/zod/src/v4/core/to-json-schema.ts` (two null-guard additions).

colinhacks:
Closing this—Seems to me like the bigger problem is with the invalid Zod schemas that you're passing in in the first place.  

Zod is full of functions that expect Zod schemas, but we don't always do this check because it would be a huge waste of time in almost all cases. The point is that TypeScript can catch these errors for you if you're passing in some invalid value. `._zod` is a required property on all Zod schemas. Moreover `.def.valueType` is a required property on record schemas. So if it doesn't exist, then you have found a way to create a Zod schema in a way that breaks Zod's own internal contracts. That is not Zod's fault, and I'm not going to add extra checks to work around it. Your problems are upstream.

PR #5823, opened 2026-03-31
fix: handle scientific notation correctly in floatSafeRemainder

### LINKED ISSUE #5792: `multipleOf` silently accepts non-multiples for small numbers (scientific notation bug in `floatSafeRemainder`)

deepesh224-ux:
### Background
Closes #5792.

The `floatSafeRemainder` utility parses values and limits fractions to calculate modulus correctly. Currently, when it processes extremely small numbers outputed via JavaScript's `.toString()` using scientific notation (e.g. `2.5e-7`), it fails to correctly account for the exponent (`e-7`) in evaluating the true decimal length limit on `val`. This wrongly rounds values internally and evaluates `multipleOf` constraints on extremely small floats as `success: true`.

### Changes made
- Addressed `floatSafeRemainder` in both `/src/v4/core/util.ts` and `/src/v3/types.ts`. Extracted a new `getDecCount` helper string handler inside the function that checks BOTH `val` and `step` safely for scientific exponents. 
- Avoided regression in larger precision numbers or unneeded formatting.
- Added regression tests for `#5792` cases in both v3 and v4 `number.test.ts`.

All 3500+ tests pass with `pnpm vitest` and code was verified through Biome.


colinhacks:
Closing — same issue (#5792) was fixed by #5793, which took the tolerance-based ratio approach instead of patching the string-counting algorithm. The string approach still has latent issues (e.g. `toFixed` range error past 100 decimals, `0.1 + 0.1 + 0.1` famously not detected as a multiple of `0.1`) that the ratio approach handles. Thanks for the PR.

PR #5833, opened 2026-04-03
fix(ci): support TypeScript 6+ build and tests

wakita181009:
## Summary

- Patch `tsconfig.base.json` with `ignoreDeprecations=6.0` in CI when TypeScript >= 6
- Align subpackage TypeScript versions (resolution, integration) to avoid version mismatches
- Update test type assertions for TS6 `exactOptionalPropertyTypes` changes

## Problem

`zshy` hardcodes `moduleResolution=node10` for CJS builds, which triggers **TS5107** in TypeScript 6+. Additionally, TS6 now distinguishes `{ a?: T }` from `{ a?: T | undefined
}`, breaking type assertions in 9 test files.

This affects all open PRs — the "Test with TypeScript latest" job fails across the board.

## Approach

**CI workflow:** Detect the installed TypeScript major version. When >= 6:
1. Update TypeScript in subpackages that pin their own version
2. Inject `ignoreDeprecations: "6.0"` into `tsconfig.base.json`

**Tests:** Add explicit `| undefined` to optional properties and tuple elements in expected types, matching what zod's type inference actually produces.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,6 +26,20 @@ jobs:
       - uses: pnpm/action-setup@v4
       - run: pnpm install
       - run: pnpm add typescript@${{ matrix.typescript }} -w
+      - name: Patch for TypeScript 6+ (zshy hardcodes deprecated moduleResolution=node10)
+        run: |
+          TS_MAJOR=$(npx tsc --version | awk '{print $2}' | cut -d. -f1)
+          if [ "$TS_MAJOR" -ge 6 ]; then
+            pnpm add typescript@${{ matrix.typescript }} --filter @zod/resolution --filter @zod/integration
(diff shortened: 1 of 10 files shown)


colinhacks:
Ah, I ended up doing almost this exact same thing before seeing this PR. I appreciate the work on this, though.

colinhacks:

Fixed upstream in `zshy` itself, so zod's CI no longer needs the `ignoreDeprecations` workaround. The matrix's `typescript@latest` leg is currently resolving to 6.0.3 and the suite is green on `main` without any tsconfig patching. The test-assertion changes here landed separately in ddd20a30.

Going to leave this closed. Thanks for the writeup — the root-cause analysis was useful.

PR #5843, opened 2026-04-07
correct logic for validating length

nameearly:
If all eleamnts can be picked -> optStart=0 ->
"tooSmall = input.length < -1" would always be false.

### DIFF
--- a/packages/zod/src/v4/classic/tests/function.test.ts
+++ b/packages/zod/src/v4/classic/tests/function.test.ts
@@ -187,18 +187,11 @@ test("input validation error", () => {
       [
         {
           "code": "too_small",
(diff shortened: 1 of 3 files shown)


colinhacks:
Solid catch — the `optStart - 1` off-by-one was masking the entire under-length path. Pushed a follow-up in `e7c97f1b` that splits the combined early return so `too_big` falls through to element-wise validation (present indices still surface `invalid_type` errors), uses `optStart` directly as the `minimum`, and pins the behavior with snapshot tests for required, trailing-optional, and interior-optional tuples. Thanks for tracking this down 👍

colinhacks:
Landed in Zod 4.4

PR #5847, opened 2026-04-09
feat: add `when` param support to `superRefine`

### LINKED ISSUE #5197: [V3-V4] Superrefine is not triggered if object validation fails
I have some complex form validation where I use superrefine as final check for some non related stuff, so problem is when one of base object fields fail, it prevents superrefine form being triggered. Looks like there is a solution for refine with `where` method. But is there smth similar in zod?

Yihao-G:
Closes #5197

`superRefine` now accepts an optional `when` callback parameter (matching the existing `refine` API), allowing conditional execution of `superRefine` checks based on the validation payload.


colinhacks:
Already merged in #5741. Closing as a duplicate — thanks for the PR!

PR #5852, opened 2026-04-10
Update Italian locale error message for validation

pastorello:
invalid_format default case was left untranslated


colinhacks:
Good catch on the leftover English fallback. Merged 👍

colinhacks:
Landed in Zod 4.4

PR #5855, opened 2026-04-11
fix(v4): clone Map and Set in shallowClone to prevent shared state

### LINKED ISSUE #5826: `.default()` shallow-copies
When using the `default()` functionality, zod appears to return a shallow-copy of the schema's default value. So, when the default value is a Map, or Set, etc. - operating on that default value later on makes for surprising results, because it means that the parser's output is no longer the same for the same input.

If this is desired, please add this caveat to the `default` documentation. Or, even better, allow the `default` function to return only `function` or primitive types. However, I find this to be a very surprising and incorrect result.

Min repro [Node 24]

```ts
import { z } from 'zod';

const itemParser = z.map(z.string(), z.number()).default(new Map());

type Item = { field: Map<string, number> | undefined };

const items: Item[] = [{ field: undefined }, { field: undefined }];

for (const item of items) {
  console.log(item);
  const itemsMap = itemParser.parse(item.field);
  console.log(itemsMap.size);
  itemsMap.set('HELLO', 1);
}
```

outputs

```
{ field: undefined }
0
{ field: undefined }
1
```

artur-seppa:
## Problem

When using `.default()` with mutable values like `Map` or `Set`, every call to
`.parse(undefined)` returns the **same reference**. Mutations on one parse result
leak into subsequent parses:

```ts
const schema = z.map(z.string(), z.number()).default(new Map());

const result1 = schema.parse(undefined);
result1.set("key", 42);

const result2 = schema.parse(undefined);
console.log(result2.size); // 1 — should be 0
```

This already worked correctly for plain objects and arrays via shallowClone,
but Map and Set fell through to return o (same reference).

## Fix
Extended shallowClone in util.ts to cover Map and Set:

```ts
if (o instanceof Map) return new Map(o);
if (o instanceof Set) return new Set(o);
```

## Tests
Added four tests to default.test.ts:

shallow clone returns distinct instances for Map and Set
mutations on one parse result do not affect another (both directions)
Fixes #5826

### DIFF
--- a/packages/zod/src/v4/classic/tests/default.test.ts
+++ b/packages/zod/src/v4/classic/tests/default.test.ts
@@ -332,6 +332,50 @@ test("defaulted array schema returns shallow clone", () => {
   expect(result1).toEqual(result2);
 });
 
+test("defaulted Map schema returns shallow clone", () => {
+  const schema = z.map(z.string(), z.number()).default(new Map([["a", 1]]));
+  const result1 = schema.parse(undefined);
+  const result2 = schema.parse(undefined);
+  expect(result1).not.toBe(result2);
+  expect(result1).toEqual(result2);
+});
+
+test("defaulted Set schema returns shallow clone", () => {
+  const schema = z.set(z.string()).default(new Set(["x"]));
+  const result1 = schema.parse(undefined);
+  const result2 = schema.parse(undefined);
+  expect(result1).not.toBe(result2);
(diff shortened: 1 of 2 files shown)


IsraelGboluwaga (approved):

colinhacks:
@pullfrog review

colinhacks:
Landed in Zod 4.4

PR #5860, opened 2026-04-15
fix(toJSONSchema): emit falsy prefault values as defaults

### LINKED ISSUE #5824: Falsy prefault() values are not emitted by toJsonSchema()
Truthy prefaults work as expected…
```ts
z.boolean().prefault(true).toJSONSchema({ io: "input" });

// {
//   $schema: "https://json-schema.org/draft/2020-12/schema",
//   default: true,
//   type: "boolean"
// }
```

But falsy prefaults do not…
```ts
z.boolean().prefault(false).toJSONSchema({ io: "input" });

// {
//   $schema: "https://json-schema.org/draft/2020-12/schema",
//   type: "boolean"
// }
```

Caused by [this line](https://github.com/colinhacks/zod/blob/c7805073fef5b6b8857307c3d4b3597a70613bc2/packages/zod/src/v4/core/to-json-schema.ts#L208)

Suggested patch:

```diff
- if (ctx.io === "input" && result.schema._prefault) result.schema.default ??= result.schema._prefault;
+ if (ctx.io === "input" && "_prefault" in result.schema) result.schema.default ??= result.schema._prefault;
```

SAY-5:
Closes #5824.

The previous truthy check on \`result.schema._prefault\` skipped prefaults whose value was \`false\`, \`0\` or \`\"\"\`, so they were never propagated to the JSON Schema \`default\`. The reporter included the suggested patch in the issue body.

\`\`\`ts
z.boolean().prefault(false).toJSONSchema({ io: \"input\" })
// before: { type: \"boolean\" }
// after:  { default: false, type: \"boolean\" }
\`\`\`

Switched to an \`in\` check so any explicitly set prefault is emitted. \`_prefault\` is only set in the prefault processor (\`json._prefault = JSON.parse(JSON.stringify(def.defaultValue))\`), so the property is never \`undefined\` when present, and the \`??=\` keeps the existing precedence over an inherited \`default\`.

Added a regression test in \`to-json-schema.test.ts\` covering \`false\`, \`0\` and \`\"\"\` prefaults. Verified it fails on \`main\` and passes with the fix. Full \`to-json-schema\` + \`to-json-schema-methods\` suites both green (322/322).

Note: the unrelated \`datetime > redos checker\` flake (#5743) is failing on \`main\` on Node 24 and is not affected by this change.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2560,6 +2560,33 @@ test("defaults/prefaults", () => {
   `);
 });
 
+test("falsy prefault values are emitted as defaults", () => {
+  // https://github.com/colinhacks/zod/issues/5824
+  expect(z.boolean().prefault(false).toJSONSchema({ io: "input" })).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": false,
+      "type": "boolean",
+    }
+  `);
+
+  expect(z.number().prefault(0).toJSONSchema({ io: "input" })).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": 0,
+      "type": "number",
+    }
+  `);
+
+  expect(z.string().prefault("").toJSONSchema({ io: "input" })).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": "",
+      "type": "string",
+    }
+  `);
+});
+
 test("input type", () => {
   const schema = z.object({
     a: z.string(),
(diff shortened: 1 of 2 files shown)


SAY-5:
Heads up: the failing `Test with TypeScript latest on Node lts/*` CI job is unrelated to this change — it's a build-step failure on `main` as well, triggered by TS 5.7's deprecation of `moduleResolution=node10`:

```
error TS5107: Option 'moduleResolution=node10' is deprecated and will stop functioning in TypeScript 7.0. Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error.
```

Every recent `test` workflow run across branches is hitting the same failure (fix-prefault-falsy, feat/iso-tests, feat/locale-tests, fix/merge-throw-on-refinements, …). I left this PR scoped to the prefault fix — happy to open a separate PR adding `"ignoreDeprecations": "6.0"` to `tsconfig.build.json` if that's the preferred resolution.

PR #5865, opened 2026-04-18
base64 native parser only

drornir:
Hi!

I was looking through the code of the base64 parts and saw something a bit arbitrary there IMO.

There's a check to see if the base64 encoded string length divides by 4.

Looking at the comments and tests, it seemed like it was supposed to verify that the string had padding.

I have two issues with this approach.

The first one is that it doesn't actually do the work it was intended to do. `atob`, the function that parses the base64 string in order to validate it, ignores whitespace characters. This means I can pass `"123 "` and it passes the validation although the string is not padded correctly.

The second one is my opinion - I don't think that requiring padding (without a way to configure it) is actually beneficial to the developer. When I do input validation, what I'm worried about is 'will this crash my program later on?'. In this case, `atob` and its new implemention in `Unit8Array.fromBase64()` both allow unpadded base64 strings. So for me it doesn't make much sense to deny base64 strings that are valid for parsing in JS.

I can see how adding validations on top of what JS supports is beneficial in general, but for now, people can normalize the base64 strings instead of denying them. Later, we can add custom checks with options, but I think that the current state where there is not way to configure the parser, the behavior should match what JS accepts as valid.

At first I was trying to create a backwards compatible implementation and adding options and / or a whole new type (e.g base64js), but IMO it just made things look weird and sloppy. You can see my vibe coded attempt at https://github.com/drornir/zod/pull/1

In my opinion, it's debateble whether this is an API change or a bugfix. So I thought that instead of banging my head agaisnt the wall, I'll do the most minimal change and open this up to conversation.

As you can see from my [first attempt](https://github.com/drornir/zod/pull/1), I have more thoughts and ideas for follow up PR's if you're up for it :)

### DIFF
--- a/packages/zod/src/v4/classic/tests/string.test.ts
+++ b/packages/zod/src/v4/classic/tests/string.test.ts
@@ -194,33 +194,43 @@ test("email validations", () => {
 test("base64 validations", () => {
   const validBase64Strings = [
     "SGVsbG8gV29ybGQ=", // "Hello World"
+    "SGVsbG8gV29ybGQ", // "Hello World" with no padding
+    " SGVsbG8gV29ybGQ\n", // "Hello World" untrimmed
+    "SGVs bG8gV29ybGQ", // "Hello World" with a space in the middle
+    "   SG Vs b G8g V29 y bGQ   ", // "Hello World" with arbitrary spaces
+    "SGVs\nbG8g \fV29y \rbGQ ", // "Hello World" with chars that match /\s/ excluding \v no padding
+    "SGVs\nbG8g \fV29y \rbGQ =", // "Hello World" with chars that match /\s/ excluding \v and padding
+    "SGVsbG8gV29ybGQ", // Missing padding
     "VGhpcyBpcyBhbiBlbmNvZGVkIHN0cmluZw==", // "This is an encoded string"
+    "VGhpcyBpcyBhbiBlbmNvZGVkIHN0cmluZw", // Missing padding
     "TWFueSBoYW5kcyBtYWtlIGxpZ2h0IHdvcms=", // "Many hands make light work"
     "UGF0aWVuY2UgaXMgdGhlIGtleSB0byBzdWNjZXNz", // "Patience is the key to success"
     "QmFzZTY0IGVuY29kaW5nIGlzIGZ1bg==", // "Base64 encoding is fun"
     "MTIzNDU2Nzg5MA==", // "1234567890"
     "YWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXo=", // "abcdefghijklmnopqrstuvwxyz"
     "QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVo=", // "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+    "QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVo", // Missing padding
     "ISIkJSMmJyonKCk=", // "!\"#$%&'()*"
     "", // Empty string is technically a valid base64
   ];
 
   for (const str of validBase64Strings) {
-    expect(str + z.string().base64().safeParse(str).success).toBe(`${str}true`);
+    const escapedStr = JSON.stringify(str).slice(1, -1);
+    expect(escapedStr + z.string().base64().safeParse(str).success).toBe(`${escapedStr}true`);
   }
 
   const invalidBase64Strings = [
-    "12345", // Not padded correctly, not a multiple of 4 characters
-    "SGVsbG8gV29ybGQ", // Missing padding
-    "VGhpcyBpcyBhbiBlbmNvZGVkIHN0cmluZw", // Missing padding
+    "12345", // length % 4 === 1
+    "1 2345", // length % 4 === 1 check ignores whitespace
     "!UGF0aWVuY2UgaXMgdGhlIGtleSB0byBzdWNjZXNz", // Invalid character '!'
     "?QmFzZTY0IGVuY29kaW5nIGlzIGZ1bg==", // Invalid character '?'
     ".MTIzND2Nzg5MC4=", // Invalid character '.'
-    "QUJDREVGR0hJSktMTU5PUFFSU1RVVldYWVo", // Missing padding
+    "SGVsbG8gV29ybGQ\v", // "\v matches /\s/, but is not allowed by atob"
   ];
 
   for (const str of invalidBase64Strings) {
-    expect(str + z.string().base64().safeParse(str).success).toBe(`${str}false`);
+    const escapedStr = JSON.stringify(str).slice(1, -1);
+    expect(escapedStr + z.string().base64().safeParse(str).success).toBe(`${escapedStr}false`);
   }
 });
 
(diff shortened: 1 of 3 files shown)


colinhacks:
@pullfrog review

colinhacks:

Thanks for digging into this. Going to decline the loosening but document the reasoning at length, since the padded-vs-unpadded question is the real decision and it's worth getting on the record. There is also a real bug in here that I want to fix.

## Standards landscape

| Spec / API | Padding behavior |
|---|---|
| RFC 4648 §3.2 (the base64 spec) | Implementations **MUST** include padding by default when emitting. Unpadded is allowed only when "the spec referring to the encoding" explicitly says so. |
| RFC 4648 §3.3 | Implementations SHOULD reject non-alphabet characters (including whitespace) unless the referring spec says otherwise. |
| RFC 2045 (MIME) | Padded, line-broken every 76 chars. Used by email and PEM. |
| RFC 7515 / 7519 (JWS / JWT) | base64url **without** padding. Already covered by `z.base64url()` / `z.jwt()`. |
| `atob()` | Strips ASCII whitespace, accepts unpadded, lenient about trailing chars. |
| `Buffer.from(s, "base64")` | Extremely lenient — silently truncates on invalid chars, accepts whitespace, accepts unpadded. |
| `Uint8Array.fromBase64()` (Stage 4, baseline 2024) | Explicit `{ lastChunkHandling: "loose" \| "strict" \| "stop-before-partial" }`. The platform itself acknowledges both modes are legitimate and forces the caller to pick. |

The directional signal: the spec says padded is canonical, and the newest platform API (`Uint8Array.fromBase64`) treats padded-vs-unpadded as a *deliberate choice the caller has to make*. That's the position a validator should take.

## Use cases that want padding ENFORCED

| Use case | Why padding matters |
|---|---|
| **Cryptographic hashes** (SHA-256 digests, MD5, SRI hashes) | Fixed binary length → fixed canonical base64 length *with* padding (SHA-256 = 44 chars ending in `=`). Lack of padding usually indicates truncation or off-by-one corruption. SRI (`integrity="sha256-..."`) explicitly requires padded base64. |
| **HTTP `Authorization: Basic`** | RFC 7617 mandates padded base64 of `user:pass`. |
| **`data:` URIs** (`data:image/png;base64,...`) | Standard padded base64. |
| **PEM-encoded keys/certs** | MIME base64 with terminal padding. |
| **Email attachments / MIME bodies** | RFC 2045 padded. |
| **Binary data in JSON / config / DB columns** | Almost universally emitted by encoders in canonical padded form. Strict validation catches truncated copy-paste, hex/base64 confusion, encoder misconfiguration, off-by-one bugs. |
| **Round-trip determinism / canonical-form invariants** | If `decode(encode(x)) === canonical_form`, then validating canonical form upstream means downstream code can rely on `input === encode(decode(input))`. This invariant is lost the moment unpadded inputs are accepted. |
| **Cross-system interop where one side will pad and the other won't auto-pad on decode** | Strict validation surfaces the mismatch at the boundary instead of producing decoded-but-wrong bytes deeper in. |
| **Detecting upstream bugs** | Most production systems produce padded base64. An unpadded one in the wild is usually a bug worth raising, not silently accepting. |

## Use cases that want padding OPTIONAL

| Use case | Why unpadded is fine | Already covered? |
|---|---|---|
| **JWT / JWS / JWE** | Spec-mandated unpadded base64url. | Yes — `z.jwt()`, `z.base64url()` |
| **URL-embedded tokens / state params** | `=` is a reserved URI char and gets percent-encoded; convention is to drop it and use the URL-safe alphabet. | Yes — `z.base64url()` |
| **Compact opaque identifiers / short codes** | Bytes saved by dropping `==`. Almost always uses the URL-safe alphabet. | Yes — `z.base64url()` (or `z.nanoid()`) |
| **Legacy systems emitting unpadded standard alphabet** | Rare but exists. | No — would require user-side `.padEnd(…, "=")` or a future `{ padding: "optional" }` opt-in. |
| **Human-typed input where users drop trailing `=`** | Cosmetic. | No — same as above; arguably should be normalized in a `.transform()` rather than relaxed at validation. |

The notable thing about this column: every common unpadded use case is already served by a different zod validator. The remaining "want unpadded standard-alphabet base64" niche is small, and the workaround is one line:

```ts
z.string().transform(s => s.padEnd(Math.ceil(s.length / 4) * 4, "=")).pipe(z.base64())
```

…or a future opt-in like `z.base64({ padding: "optional" })` if demand materializes.

## Cross-validator consistency inside zod

| Validator | Today's behavior |
|---|---|
| `z.base64()` | Strict regex (advertised via `bag.patterns` / `contentEncoding`); lenient runtime. **Inconsistent today.** |
| `z.base64url()` | Strict — disallows padding entirely (post #4386). |
| `z.jwt()` | Strict — uses base64url-no-padding for each segment. |
| `regexes.md5_base64`, `sha1_base64`, `sha256_base64`, `sha384_base64`, `sha512_base64` | Strict — fixed-length, padding required where canonical. |

If `z.base64()` becomes lenient on padding, it becomes the only validator in the family that does. The hash-format validators it composes with would still require padding, so users hand-rolling `z.base64().refine(s => s.length === 44)` get different semantics from `sha256_base64`. The orthogonality breaks.

## Asymmetry of the decision

| | Strict default | Lenient default |
|---|---|---|
| Catches upstream bugs | Yes — surfaces truncation, encoder misconfiguration, malformed input | No — silently passes corrupted-looking input |
| Matches RFC 4648 default | Yes | No |
| Matches advertised pattern (`bag.patterns`, JSON Schema export) | Yes | No (runtime is wider than schema) |
| Consistent with `z.base64url()` and hash validators | Yes | No |
| Covers unpadded niche | No — but `z.base64url()` / `z.jwt()` / `.transform().pipe()` do | Yes |
| Reversible later? | Yes — can add `{ padding: "optional" }` opt-in non-breakingly | No — tightening to require padding is a breaking change |
| Aligns with platform direction | Yes — `Uint8Array.fromBase64({ lastChunkHandling: "strict" })` is an explicit option | Partial — matches `lastChunkHandling: "loose"` default but loses the ability to express "strict" |

The load-bearing argument is the last row: **strict is reversible, lenient is not.** Loosening `z.base64()` now permanently widens the contract; tightening it back later is a breaking change. Keeping it strict preserves the option to add `z.base64({ padding: "optional" })` (or a future `z.base64Loose()`) if real demand shows up, and costs unpadded-input users one `.transform()` line in the meantime.

## The actual bug worth fixing

The legitimate kernel you found is that the `length % 4 !== 0` short-circuit is bypassable because `atob` strips ASCII whitespace internally before validating, so `"123 "` (length 4, mod-4 passes, atob strips to `"123"` and parses) sneaks through today. The minimal fix is a whitespace rejection in front of the existing check:

```ts
export function isValidBase64(data: string): boolean {
  if (data === "") return true;
  if (/\s/.test(data)) return false;
  if (data.length % 4 !== 0) return false;
  try {
    atob(data);
    return true;
  } catch {
    return false;
  }
}
```

Worth noting why we can't just enforce `regexes.base64` at runtime: that's what the validator used to do, and it was replaced with the `atob`-based check specifically for perf on large inputs in #4386 (commit `ec7c89d9`). That same PR added a regression test (`big base64 and base64url` in `packages/zod/src/v4/classic/tests/string.test.ts`) that round-trips 10MB through `z.base64().parse()`. The `/\s/.test()` short-circuit is a single deterministic linear scan with no backtracking, so it fits inside that perf envelope.

Going to close this out in favor of a tightening fix along those lines. Thanks for looking into it!

colinhacks:
This got a little buried in the right up above, but I'm definitely open to a `padding: "optional" | "required"` down the line. Though at this point I haven't really seen much demand for it and as the comment indicated, a bunch of the "no padding" use cases are better handled by other more domain-specific validators (jwt, etc)

PR #5866, opened 2026-04-18
fix ci build step

drornir:
Fixes the build step in test.yml and adds ts v6 as well

As you can see, the tests fail with TS 6+.

In my opinion testing with different TS versions is not needed, it's just a dev dependency for build time.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,10 +13,11 @@ jobs:
   test-node:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest]
         node: ["lts/*"]
-        typescript: ["5.5", "latest"]
+        typescript: ["5.5", "6", "latest"]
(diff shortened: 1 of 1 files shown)


dokson:
Hi @drornir — looks like #5833 (opened slightly earlier) is targeting the exact same issue and goes further:

- same `ignoreDeprecations=6.0` tsconfig patch
- **also** bumps TS in the `@zod/resolution` and `@zod/integration` sub-packages (otherwise `zshy` trips TS5107 on `moduleResolution=node10` under TS 6+)
- **also** updates the type assertions across ~9 test files for the stricter `exactOptionalPropertyTypes` behavior in TS 6+ (e.g. `{ a?: string }` → `{ a?: string | undefined }`)

Without those test updates this PR's CI will keep failing on TS 6/latest (the "latest" job is currently red). Probably cleanest to close this in favor of #5833, or rebase and pull in the test-assertion changes from there.

drornir:
Yep, sure, closing, sorry for not looking at other PRs before opening this

PR #5867, opened 2026-04-20
fix: ZodRecord now applies key schema transforms at runtime

### LINKED ISSUE #5296: Records do not transform keys, even though the Typescript types says it does.

OfekDanny:
## Summary

\`z.record()\` silently ignores transforms on the key schema at runtime, even though Zod's TypeScript types correctly reflect the transformation in the output type. This is a type/runtime mismatch.

**Reproduction:**
\`\`\`ts
const schema = z.record(z.string().transform(k => k.toUpperCase()), z.number());
schema.parse({ hello: 1 }); // returns { hello: 1 } — key NOT transformed
\`\`\`

**Root cause:** \`ZodRecord._parse()\` uses the original input key as the output property name instead of the transformed result from \`keyType._parse()\`.

**Fix:** Use \`keyResult.value\` (the transformed key) instead of the raw input key when building the output object.

## Changes
- \`packages/zod/src/v3/tests/record.test.ts\`: add regression test for key transforms
- \`packages/zod/src/v4/core/tests/record-constructor.test.ts\`: add regression test for key transforms in v4

Fixes #5296

### DIFF
--- a/packages/zod/package.json
+++ b/packages/zod/package.json
@@ -126,7 +126,7 @@
   },
   "scripts": {
     "clean": "git clean -xdf . -e node_modules",
-    "build": "zshy --project tsconfig.build.json",
+    "build": "node -e \"require('child_process').execSync('zshy --project tsconfig.build' + (parseInt(require('typescript').version) >= 6 ? '.ts6' : '') + '.json', {stdio:'inherit'})\"",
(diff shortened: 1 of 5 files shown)


colinhacks:
@pullfrog review

PR #5869, opened 2026-04-20
fix: Handle `z.union([])` and `z.xor([])`

### LINKED ISSUE #5868: `z.union([])` and `z.xor([])` throw internal error on parse
See https://zod-playground.vercel.app?appdata=N4IgzgxgFgpgtgQxALhBA9gOzAFwASSyJ4C8eAXgHQCumAllgBQDaAugJQA6m3A9L3gzZ8heAlIVKAD3QAnFh27dZMHNVmYC0MSAA0IAG4IANtRhgUzEMEwI4MZHk4gAUuig8QAXxCt9BmFkwBkwUEAAWSgBmSgA2PRA6MAAtdAATAFk6ehQAMxMwGC8gA

The types resolve to `never`, so maybe `parse` should treat it as `never` too?

tjenkinson:
The types already resolve to `never`, but `parse` would throw.

Think it makes sense to treat as `never` when parsing?

Fixes https://github.com/colinhacks/zod/issues/5868

### DIFF
--- a/packages/zod/src/v4/classic/tests/union.test.ts
+++ b/packages/zod/src/v4/classic/tests/union.test.ts
@@ -217,3 +217,57 @@ test("z.xor() type inference", () => {
   type Result = z.infer<typeof schema>;
   expectTypeOf<Result>().toEqualTypeOf<string | number | boolean>();
 });
+
+test("z.union([]) constructs and rejects all input", () => {
+  const schema = z.union([]);
+  expectTypeOf<z.infer<typeof schema>>().toEqualTypeOf<never>();
+  const result = schema.safeParse("anything");
+  expect(result.success).toEqual(false);
+  if (!result.success) {
+    expect(result.error.issues).toMatchInlineSnapshot(`
+      [
+        {
+          "code": "invalid_union",
+          "errors": [],
+          "message": "Invalid input",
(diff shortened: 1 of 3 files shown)


colinhacks:
Thanks for the report and the PR. Pushed a different fix on top — wanted to flag what changed and why.

**Approach:** instead of adding a special-case `invalid_type` issue with `expected: "never"`, I just guarded the cached `first` runner so it's only captured when `def.options.length === 1`. The existing `handleUnionResults` / `handleExclusiveUnionResults` helpers already emit `invalid_union` with `errors: []` when given an empty results array, so no new parse branch is needed. Net diff in `core/schemas.ts` is `−6 / +4`.

**Why not `expected: "never"`:** that issue shape collides with what `z.never()` actually emits, so consumers can't distinguish "input rejected by `z.never()`" from "input rejected by an empty union." `invalid_union` with `errors: []` is consistent with how non-empty unions report a no-match.

**Coverage:** also fixes `z.discriminatedUnion("...", [])` (and `zod/mini` equivalents), which crashed for the same reason via `$ZodUnion.init`. Tests added for all six surfaces.

Going to merge as-is.

colinhacks:
Landed in Zod 4.4

PR #5870, opened 2026-04-20
perf(v4): eliminate dictionary-mode overhead via shadow-proto (no breaking changes)

### LINKED ISSUE #5760: Schema instances have ~91 own properties → dictionary mode → ~24x slower property access vs Zod 3
I'm building a form library that introspects Zod schemas to auto-generate form fields, constraints, and validation. After migrating to Zod 4, I hit a significant introspection performance regression. After a lot of profiling, I tracked the root cause down to schema instance property count.

## The problem

Zod 4 schema instances end up with **~91 own properties**. This pushes them past JS engine fast-property thresholds (V8's is ~27), forcing dictionary mode for all property access — including `_def` and `_zod`.

```javascript
const schema = z.string().min(1).max(10).email();
console.log(Object.getOwnPropertyNames(schema).length); // ~91
```

The 91 properties come from:
- **~60** from the core "support prototype modifications" loop that copies prototype methods onto each instance
- **~30** from classic `ZodType` init assigning methods as own properties

Validation perf on valid data is great — this only affects introspection and construction.

## Benchmarks

Property access (same machine, Bun):

| Benchmark | Zod 3 | Zod 4 | Ratio |
|-----------|-------|-------|-------|
| `schema._def` | 298 ps | 7.23 ns | **24x** |
| `schema._zod` | — | 7.2 ns | — |
| Inline extraction | 9.52 ns | 13.08 ns | 1.4x |
| Inline extraction, cached `_zod` | — | 9.2 ns | **parity** |
| `z.string().min(1).max(10).email()` | 2.2 µs | 72.6 µs | **32x** |

Key finding: caching `schema._zod` in a local variable brings per-read performance to Zod 3 parity. The cost is in the initial property lookup on the bloated instance, not in the `_zod` object itself.

<details>
<summary>Real-world impact (form library)</summary>

Cold path (no caching):

| Operation | Zod 3 | Zod 4 | Ratio |
|-----------|-------|-------|-------|
| Field metadata extraction (string+checks) | 7.69M ops/sec | 662K ops/sec | 11.6x |
| Field metadata extraction (number+checks) | 7.26M ops/sec | 961K ops/sec | 7.5x |
| Form mount — simple (5 fields) | 53K ops/sec | 5.8K ops/sec | 9.2x |
| Form mount — complex | 13.1K ops/sec | 1.7K ops/sec | 7.7x |

With WeakMap caching on the warm path, Zod 4 reaches 10.5M ops/sec — actually faster than Zod 3. So it's cacheable, but the cold path and construction cost remain.

</details>

## Reproduction

```javascript
import * as z3 from "zod3"; // zod@3.24.4
import * as z4 from "zod";  // zod@4.x

const v3Schema = z3.string().min(1).max(10).email();
const v4Schema = z4.string().min(1).max(10).email();

// Property count
console.log("Zod 3 own props:", Object.getOwnPropertyNames(v3Schema).length);
console.log("Zod 4 own props:", Object.getOwnPropertyNames(v4Schema).length);

// Property access benchmark
for (let i = 0; i < 10000; i++) v3Schema._def; // warmup
for (let i = 0; i < 10000; i++) v4Schema._def;

console.time("Zod 3 _def × 10M");
for (let i = 0; i < 1e7; i++) v3Schema._def;
console.timeEnd("Zod 3 _def × 10M");

console.time("Zod 4 _def × 10M");
for (let i = 0; i < 1e7; i++) v4Schema._def;
console.timeEnd("Zod 4 _def × 10M");

// Construction benchmark
console.time("Zod 3 construction × 10K");
for (let i = 0; i < 10000; i++) z3.string().min(1).max(10).email();
console.timeEnd("Zod 3 construction × 10K");

console.time("Zod 4 construction × 10K");
for (let i = 0; i < 10000; i++) z4.string().min(1).max(10).email();
console.timeEnd("Zod 4 construction × 10K");
```

## What I've already tried on my side

- `_def.type` string comparison instead of `instanceof` (instanceof is ~36x slower due to the dual-class chain)
- `_zod.bag` reads instead of iterating `_zod.def.checks` (same path `toJSONSchema` uses)
- `_zod.propValues` for discriminated unions, `_zod.values` for enums
- WeakMap caching of extraction results
- Local `_zod` caching within functions

After all that, the remaining gap is the instance property count — that's the one thing consumers can't fix.

## Workarounds for other library authors hitting this

1. Cache `schema._zod` locally: `const zod = schema._zod` then read `zod.def`, `zod.bag`
2. Use `_zod.bag` for constraints instead of iterating checks
3. Use `_def.type` instead of `instanceof`
4. WeakMap cache extraction results — schemas are immutable so this is safe

## Ideas (not sure what's feasible)

The prototype copy loop is the main contributor (~60 of the 91 properties). I tried removing it but it breaks prototype extension (`prototypes.test.ts`, `apply.test.ts`). Some thoughts:

- Could prototype extension work through a different mechanism (Proxy, `_extensions` map, etc.) that doesn't copy methods to instances?
- Could checks be plain objects instead of `$ZodCheck` class instances? Would reduce construction cost.
- Lazy bag population — only compute on first `_zod.bag` access instead of at construction.

Totally understand if the prototype extension design is intentional and this is a known tradeoff. Just wanted to surface the data in case it's useful.

## Environment

- Zod 4.x, compared against 3.24.4
- Bun (JSC) for benchmarks, also reproduced in Node 22 (V8)

gsoldevila:
## Summary

Fixes https://github.com/colinhacks/zod/issues/5760

Own-property count before/after this PR:

| Schema | Before | After |
|--------|--------|-------|
| `z.string()` | ~91 | ~18 |
| `z.string().min(1).max(10).email()` | ~91 | ~18 |
| `z.number()` | ~91 | ~17 |
| `z.object({})` | ~91 | ~18 |

## Problem

Two sources contributed to the ~91 own properties per schema instance:

1. **The copy-loop in `$constructor`** copies every method from `_.prototype` to each new instance, adding ~60 own properties.
2. **Classic `ZodType.init`** assigns ~30 builder methods as per-instance own properties.

V8 switches from fast-property (inline-cache) to dictionary mode at ~20–30 own properties, degrading all property access by up to **24×**.

## Solution: shadow-proto

Introduce a hidden *internal* prototype layer between the user-visible `_.prototype` and the parent prototype. Library methods are placed on `internalProto`; the copy-loop continues to target the empty `_.prototype`.

```
inst
 └── _.prototype           ← user-visible (empty by default; copy-loop targets this)
      └── internalProto    ← library space (_initProto writes here)
           └── Parent / Object.prototype
```

### core.ts

- Export `$internalProto` (a unique Symbol) so `schemas.ts` can locate each constructor's internal prototype.
- Inside `$constructor`, create `internalProto = Object.create(Parent.prototype)`, wire `_.prototype → internalProto` via `Object.setPrototypeOf`, and expose it as `_[$internalProto]`.
- **The copy-loop is unchanged** — `Object.keys(_.prototype)` now returns `[]` by default, making it a no-op unless a user explicitly adds methods to `_.prototype`.

### classic/schemas.ts

- Add `_protoInitMap` (WeakMap) and `_initProto` helper that targets `inst._zod.constr[$internalProto]`.
- All 133 per-instance builder method assignments replaced with `_initProto` calls.
- Parse-family closures (`parse`, `safeParse`, `parseAsync`, `safeParseAsync`, encode/decode variants) remain per-instance — they capture `inst` and must work when detached.

### mini/schemas.ts

- Same `_initProto` infrastructure added.
- The 6 builder methods in `ZodMiniType` (`check`, `with`, `clone`, `brand`, `register`, `apply`) moved to the internal prototype.

## Key advantage: no breaking changes

The original prototype-extension contract is **fully preserved**:

```js
// Still works exactly as before
z.ZodType.prototype.myHelper = function() { return "custom"; };
z.string().myHelper(); // ✓  "custom"
```

Because `_.prototype` is now empty by default, the copy-loop activates only when users add methods there — propagating them to new instances exactly as the original code did.

Both `prototypes.test.ts` suites (classic and mini) pass **without modification**.

## New test: `fast-properties.test.ts`

Asserts for common schemas that:
- Own-property count is < 25
- Builder methods (`optional`, `nullable`, etc.) are NOT own properties
- Parse methods ARE own properties (detached usage must work: `const { parse } = schema; parse("x")`)

All tests passed ✅ 
image

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -8,6 +8,40 @@ import * as checks from "./checks.js";
 import * as iso from "./iso.js";
 import * as parse from "./parse.js";
 
+// Maps (internalProto, key) pairs already initialized — avoids repeated prototype setup
+const _protoInitMap = new WeakMap<object, Set<string>>();
+
+/**
+ * Sets shared methods on the *internal* prototype layer of inst's concrete constructor
+ * (once per concrete type per key). The internal prototype sits one level below the
+ * user-visible `_.prototype`, keeping `Object.keys(_.prototype)` empty by default
+ * and preventing V8 dictionary-mode degradation.
+ *
+ * Falls back to `Object.getPrototypeOf(inst)` for constructors not created with
+ * `$constructor` (e.g., in tests or third-party code).
+ */
+function _initProto(
+  inst: object,
+  key: string,
+  methods: Record<string, Function>,
+  defineProps?: Array<[string, PropertyDescriptor]>
+): void {
+  const proto = (inst as any)._zod?.constr?.[core.$internalProto] ?? Object.getPrototypeOf(inst);
+  let keys = _protoInitMap.get(proto);
+  if (!keys) {
+    keys = new Set();
+    _protoInitMap.set(proto, keys);
+  }
+  if (keys.has(key)) return;
+  keys.add(key);
+  Object.assign(proto as any, methods);
+  if (defineProps) {
+    for (const [k, desc] of defineProps) {
+      Object.defineProperty(proto, k, { ...desc, configurable: true });
+    }
+  }
+}
+
 ///////////////////////////////////////////
 ///////////////////////////////////////////
 ////////////                   ////////////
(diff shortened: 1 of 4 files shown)


colinhacks:
Thanks, Gerard. Unfortunately, despite the efforts here to avoid breaking changes, there are still breaking changes for any users who pull out references to schema methods. Today this works:

```ts
const opt = schema.optional;
opt(); // ZodOptional<typeof schema>
```

After this PR it silently returns a corrupt schema that fails the next time anything tries to parse with it. Parse-family methods are kept as per-instance closures so `const { parse } = schema` still works; nothing else does. 


Zod is now widely used enough that there's a Murphy's law effect with breaking changes like this. Somebody, probably hundreds of somebodies, rely on the ability to pull out methods like this and use them as standalone functions. In fact, I specifically like the fact that there are no gotchas around `this`-binding with Zod's current approach. 

The memory overhead is unfortunate. I'm receptive to anything that could be done to reduce it. I'd even be open to introducing some minor breakage if it meant a significant runtime parsing performance boost. But at this point I'm not convinced that this is a good trade-off. 

Closing for now because I'm going through a backlog here. Perhaps DM me if you'd like to float other approaches or continue this discussion.

gsoldevila:
Thanks for the thorough review, @colinhacks. I totally missed the extracted references scenario.

For us in Kibana, upgrading to 4.x has a huge memory impact (> 100 Mb), so we're exploring non-breaking alternatives.

There is a backward-compatible variant worth considering: **auto-bind getters on the prototype**. Instead of placing a plain shared function on `internalProto`, each builder method could be a getter that returns `_sharedFn.bind(this)`:

```
Object.defineProperty(internalProto, 'optional', {
  get() { return _sharedOptional.bind(this); },
  configurable: true,
});
```

I'm doing some tests locally, but I believe it should preserve the behaviour when doing:

```
const opt = schema.optional;
opt(); // ✓ works — `this` is permanently bound to schema at access time
schema.optional(); // ✓ works as before
```

Happy to submit a new PR if you think the trade-off is acceptable.

colinhacks:
Thanks, Gerard. Seems like you posted this before seeing https://github.com/colinhacks/zod/pull/5897, which is substantially identical to what you suggested. (Good idea!)

gsoldevila:
Hi! yes, I saw your PR after posting.
Both are similar, but your approach is even better as you only pay the allocation overhead once.

FYI we cherry-picked [your commit](https://github.com/colinhacks/zod/commit/8fcb71a5ffa5b32a508ac8bb38b7e1e13c387bf5) and applied it to `4.3.6`.
We did some memory tests and there are substantial memory savings, so I confirm #5760 is correctly addressed.

Thanks a lot for the fix!

PR #5872, opened 2026-04-21
docs: clarify stringFormat usage with z.coerce.string()

### LINKED ISSUE #5815: StringFormat not available when using z.coerce
It appears that StringFormat is not available when coercing a string, either when trying to use `z.coerce.stringFormat` or `z.coerce.string().stringFormat()`

```
Property 'stringFormat' does not exist on type 'typeof import("/Users/reesericci/Library/CloudStorage/MountainDuck-reesericci-1/Developer/protovoters/node_modules/zod/v4/classic/coerce")'. (typescript 2339)
```

iritikyadav123:
This PR adds a clarification to the `z.stringFormat()` documentation to explain that it is not chainable on `z.string()` or `z.coerce.string()`.

It also includes an example demonstrating the correct usage with `.pipe()` when combining `z.coerce.string()` and `z.stringFormat()`.

This helps prevent confusion and improves developer experience when working with custom string formats.

Fixes #5815

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -588,6 +588,15 @@ const coolId = z.stringFormat("cool-id", ()=>{
 // a regex is also accepted
 z.stringFormat("cool-id", /^cool-[a-z0-9]{95}$/);
 ```
+> ⚠️ **Note**
+> `z.stringFormat()` is not chainable on `z.string()` or `z.coerce.string()`.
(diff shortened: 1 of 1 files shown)


colinhacks:
I'm very protective of what gets added to the docs. I'm not interested in adding call outs for every possible misunderstanding. If I merge every PR like this, the docs would be a disaster. Appreciate the thought though.

Issue #3621, opened 2024-07-04
cuid validation


eminden:
parse() function is not throwing an exception for this string `cly63t164000245zw008pggon';select 1;`

```js
const idSchema = z.array(z.string().cuid(), {
    message: "Each ID must be a valid CUID",
}).optional();

try {
    idSchema.parse(params.folders);
} catch (error) {
    throw new Error('Invalid IDs');
}
```

eswarty:
Hi! I was curious and tested your schema, but I wasn't able to repro the issue. It seems to throw an error for me as expected. I tried this [zod playground](https://zod-playground.vercel.app?appdata=N4IgzgxgFgpgtgQxALhALwHQIE7YQTwApMwAXbASwDsBzQgSgwgFcKATBgGgAJgAdKtyHc4MMGAQ0YybnxABRBNG4BJACIjmZbgCMY3BNwBuCADbtuAYQCq6uZwEBfRgHsADqQouqZhiE4gJqbMYigA2iBhchCm%2BABsAMykAIxxACwADFkATGkArGgA7lkAHG40NN4A5ADcYDCmMBCk3Mk1cgC6IB0BRjDYYF5UKCAJGNljJSCOQA) and also tried running your code locally and both throw the expected error for me. Sorry that's not super helpful 🙁

Issue #3053, opened 2023-12-14
cuid() considers the word corrigendum as a cuid


UdittLamba:
![image](https://github.com/colinhacks/zod/assets/23259680/1669da9c-c904-4726-945e-50fdbd65c590)

mass2527:
It's because the current implementation of `cuid` method is checking string using `/^c[^\s-]{8,}$/i`
1. The string must start with the letter 'c'.
2. Match at least 8 characters that are not white space or "-" until the end of the string

This happens since there's no perfect way to find only cuid from the string.

UdittLamba:
> It's because the current implementation of `cuid` method is checking string using `/^c[^\s-]{8,}$/i`
> 
>     1. The string must start with the letter 'c'.
> 
>     2. Match at least 8 characters that are not white space or "-" until the end of the string
> 
> 
> This happens since there's no perfect way to find only cuid from the string.

For now, I am using a regex check and it seems to work so far. But would be great if this is handled better.
If I get time from work, I would like to handle this and open pr myself.

PR #5880, opened 2026-04-27
Tighten cuid() regex and deprecate CUID v1

colinhacks:
## Summary

- Changes the CUID v1 regex from `/^[cC][^\s-]{8,}$/` to `/^[cC][0-9a-z]{6,}$/` in both v3 and v4. The body is now restricted to base36 — the actual CUID v1 charset — and the length floor is aligned with Eric Elliott's documented guidance (["starts with c, contains at least 7 chars"](https://github.com/paralleldrive/cuid/issues/88#issuecomment-339848922)). Brings the regex stylistically in line with the other ID validators (`ulid`, `xid`, `ksuid`, `nanoid`, `cuid2`), all of which use allowlist character classes.
- Marks every CUID v1 surface (`z.cuid()`, `.cuid()` chain method, `ZodCUID` / `ZodMiniCUID`, `$ZodCUID*`, `_cuid`, the `cuid` regex export) as `@deprecated`, pointing at `z.cuid2()`. The upstream `paralleldrive/cuid` repo itself reads: *"Deprecated collision-resistant id spec. Insecure because it leaks timestamps. Use cuid2 instead."*

## Why

The `{8,}` length floor was introduced in #438 (May 2021) without documented rationale and has been frozen since. The original PR made an arbitrary choice that diverged from the spec author's "≥7 chars" guidance by accident.

The denylist body (`[^\s-]`) was also unprincipled — it allowed quotes, angle brackets, curly braces, semicolons, and arbitrary Unicode in a string nominally identifying itself as a CUID. Real CUID v1 output is always `c` followed by 24 base36 characters; the validator should reflect that. This change rejects the punctuation-shaped false positives reported in #3621 (`cly63t164000245zw008pggon';select1;` and similar) while remaining compatible with every real CUID v1 ever generated.

The change is **strictly more permissive in length** (`{8,}` → `{6,}`) and **strictly stricter in charset** (`[^\s-]` → `[0-9a-z]`). No real CUID is rejected; some non-CUID strings that previously passed now correctly fail.

This is **not** a security fix and should not be framed as one. CUID format validation is not — and cannot be — a substitute for output encoding (XSS prevention) or parameterized queries (SQLi prevention). It's a routine correctness/consistency improvement.

## Known limitation (not addressed)

#3053 (`corrigendum` validates as a cuid) is not fully fixed. Any all-lowercase English word starting with `c` will still pass — that's an inherent limitation of regex-based CUID validation without a fixed length, not a bug we can solve here. Deprecating `z.cuid()` in favor of `z.cuid2()` is the proper long-term resolution.

## Test plan

- [x] `pnpm vitest run` — 3575 tests pass
- [x] No lint errors (`pnpm lint` clean via lint-staged on commit)
- [x] No type errors (`Type Errors no errors` in vitest output)
- [x] Existing `cuid` tests in v3 and v4 still pass with updated snapshots
- [x] Added regression tests asserting previously-accepted non-base36 payloads (SQLi-shaped, XSS-shaped, bracket/curly chars, underscores, uppercase body) are now rejected
- [x] Updated one template-literal test that was implicitly relying on the old `{8,}` length floor

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -342,7 +342,11 @@ export interface ZodString extends _ZodString<core.$ZodStringInternals<string>> 
   nanoid(params?: string | core.$ZodCheckNanoIDParams): this;
   /** @deprecated Use `z.guid()` instead. */
   guid(params?: string | core.$ZodCheckGUIDParams): this;
-  /** @deprecated Use `z.cuid()` instead. */
+  /**
+   * @deprecated CUID v1 is deprecated by its authors due to information leakage
+   * (timestamps embedded in the id). Use `z.cuid2()` instead.
+   * See https://github.com/paralleldrive/cuid.
+   */
   cuid(params?: string | core.$ZodCheckCUIDParams): this;
   /** @deprecated Use `z.cuid2()` instead. */
   cuid2(params?: string | core.$ZodCheckCUID2Params): this;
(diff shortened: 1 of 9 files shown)


colinhacks:
@pullfrog review

colinhacks:
Landed in Zod 4.4

PR #5883, opened 2026-04-28
docs(agents): document push-to-main footgun and version-bump rule

colinhacks:
## Summary

- Document the `git worktree add -b <branch> origin/main` upstream-config footgun that has caused two accidental pushes to `refs/heads/main`. The fix is an explicit refspec on the first push: `git push -u origin <branch>:refs/heads/<branch>`, then verify the `local -> remote` line in the output.
- Add the only release-relevant rule: never bump `package.json` versions. Direct pushes to `main` are recoverable until that happens.

## Test plan

- [x] No code changes; AGENTS.md only.

PR #5884, opened 2026-04-28
feat(locales/en): handle instanceof and add comprehensive locale tests

colinhacks:
## Summary

The English locale is the reference for every translated locale, so it should expose every behavior translators need to mirror. Two gaps:

1. **`invalid_type` did not special-case class-name expectations** the way 46 of the 51 other locales do. Without it, `z.instanceof(Date)` produced `"Invalid input: expected Date, received ..."` instead of the `"instanceof Date"` form used by `de`, `es`, `fr`, `it`, `ru`, etc. Added the `/^[A-Z]/.test(issue.expected)` branch for parity.

2. **`en.test.ts` only exercised `parsedType`** — the locale itself was untested. Added a comprehensive `describe`-block suite covering every `issue.code` branch:
   - `invalid_type` (incl. `instanceof Date` + `instanceof Foo`)
   - `invalid_value` (literal + enum)
   - `too_small` / `too_big` (string, number inclusive + exclusive, array, set)
   - `invalid_format` (`regex`, `starts_with`, `ends_with`, `includes`, plus FormatDictionary lookups for `email` / `url` / `uuid`)
   - `not_multiple_of`
   - `unrecognized_keys` (single + multiple)
   - `invalid_key` (record)
   - `invalid_element` (map)
   - `invalid_union`
   - `custom` fallback

   The existing `parsedType` test is preserved.

These tests double as a regression suite and document the canonical message format that translations should mirror.

## Background

This PR is a re-land of commit `3c1f32bd`, which I accidentally pushed directly to `main` and then reverted in `bf6d99ed` so it could go through normal review. AGENTS.md was updated separately in #5883 to prevent recurrence. No code differences from the original commit.

## Test plan

- [x] All 33 new tests pass
- [x] Full suite (3741 tests across 327 files) passes
- [x] No type errors
- [x] biome lint clean

### DIFF
--- a/packages/zod/src/v4/core/tests/locales/en.test.ts
+++ b/packages/zod/src/v4/core/tests/locales/en.test.ts
@@ -1,5 +1,11 @@
-import { expect, test } from "vitest";
+import { beforeEach, describe, expect, test } from "vitest";
+import { z } from "../../../../index.js";
+import en from "../../../locales/en.js";
 import { parsedType } from "../../util.js";
+
+// `en` is the default locale, so its messages are also Zod's reference output.
+// These tests double as a regression suite for any locale change and document
+// the canonical message format that translations should mirror.
 
 test("parsedType", () => {
   expect(parsedType("string")).toBe("string");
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing — turns out the instanceof branch was deliberately removed from `en.ts` in 15cc73fb (one hour after being introduced in 5c44d2e8). The 46 other locales still carry it because they didn't get the same follow-up cleanup, and the pattern has been cargo-culted into every new locale since. `en.ts` is correct as-is.

PR #5887, opened 2026-04-28
fix(email): reject addresses longer than 254 chars (RFC 5321)

### LINKED ISSUE #5878: z.email() should restrict length to 254 characters
Admittedly this is in the dark recesses of IETF errata but, in accepted errata for RFC 3696, emails should not exceed 254 characters: https://www.rfc-editor.org/errata_search.php?rfc=3696

> In addition to restrictions on syntax, there is a length limit on
   email addresses.  That limit is a maximum of 64 characters (octets)
   in the "local part" (before the "@") and a maximum of 255 characters
   (octets) in the domain part (after the "@") for a total length of 320
   characters. However, there is a restriction in RFC 2821 on the length of an
   address in MAIL and RCPT commands of 254 characters.  Since addresses
   that do not fit in those fields are not normally useful, the upper
   limit on address lengths should normally be considered to be 254.

RFC 3696 is for the [checking and transformation of names](https://www.rfc-editor.org/rfc/rfc3696) and includes a section on [email validation](https://www.rfc-editor.org/rfc/rfc3696#section-3) which in the non-corrected version says:

>   In addition to restrictions on syntax, there is a length limit on
   email addresses.  That limit is a maximum of 64 characters (octets)
   in the "local part" (before the "@") and a maximum of 255 characters
   (octets) in the domain part (after the "@") for a total length of 320
   characters.  Systems that handle email should be prepared to process
   addresses which are that long, even though they are rarely
   encountered.

So: the original limit was 320, the corrected limit is 254. Right now ZodEmail is defined as a [regex](https://github.com/colinhacks/zod/blob/c7805073fef5b6b8857307c3d4b3597a70613bc2/packages/zod/src/v4/core/schemas.ts#L452-L458) so (afaict) we'd want a fast length check before running that regex.

DORI2001:
## Problem

`z.email()` accepts email addresses of any length. RFC 5321 §4.5.3.1.3 caps the maximum total length at **254 characters**, but a purely regex-based check can't enforce this efficiently.

```ts
const schema = z.email();
const tooLong = "a".repeat(243) + "@example.com"; // 255 chars
schema.parse(tooLong); // currently passes — should fail
```

## Fix

Add an explicit length guard at the top of `$ZodEmail`'s check function. If the input exceeds 254 characters the check pushes a `too_big` issue and returns early — the regex never runs, keeping this path fast.

```ts
inst._zod.check = (payload) => {
  if (payload.value.length > 254) {
    payload.issues.push({
      origin: "string",
      code: "too_big",
      maximum: 254,
      inclusive: true,
      input: payload.value,
      inst,
      continue: !def.abort,
    });
    return;
  }
  patternCheck(payload);
};
```

The `too_big` code gives callers a structured way to distinguish this failure from a format error.

## Test

A new test in `string-formats.test.ts` verifies that a 254-char address passes and a 255-char address returns `{ code: "too_big" }`.

## References

- RFC 5321 §4.5.3.1.3: https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1.3
- Closes #5878

### DIFF
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -449,11 +449,30 @@ export interface $ZodEmail extends $ZodType {
   _zod: $ZodEmailInternals;
 }
 
+// RFC 5321 §4.5.3.1.3 limits the maximum total email address length to 254 chars.
+const EMAIL_MAX_LENGTH = 254;
+
 export const $ZodEmail: core.$constructor<$ZodEmail> = /*@__PURE__*/ core.$constructor(
   "$ZodEmail",
   (inst, def): void => {
     def.pattern ??= regexes.email;
     $ZodStringFormat.init(inst, def);
+    const patternCheck = inst._zod.check!;
+    inst._zod.check = (payload) => {
+      if (payload.value.length > EMAIL_MAX_LENGTH) {
+        payload.issues.push({
+          origin: "string",
+          code: "too_big",
+          maximum: EMAIL_MAX_LENGTH,
+          inclusive: true,
+          input: payload.value,
+          inst,
+          continue: !def.abort,
+        });
+        return;
+      }
+      patternCheck(payload);
+    };
   }
 );
 
--- a/packages/zod/src/v4/classic/tests/string-formats.test.ts
+++ b/packages/zod/src/v4/classic/tests/string-formats.test.ts
@@ -30,6 +30,25 @@ test("string format methods", () => {
   expect(e.safeParse("EMAIL@EXAMPLE.COM").success).toBe(false); // Negative
 });
 
+test("z.email() rejects addresses longer than 254 characters (RFC 5321)", () => {
+  const schema = z.email();
+  // "@example.com" = 12 chars; fill the remaining budget with the local part
+  const domain = "@example.com";
+  const addr254 = "a".repeat(254 - domain.length) + domain; // exactly 254
+  const addr255 = "a".repeat(255 - domain.length) + domain; // exactly 255
+
+  expect(addr254.length).toBe(254);
+  expect(addr255.length).toBe(255);
+
+  expect(schema.safeParse(addr254).success).toBe(true);
+
+  const result = schema.safeParse(addr255);
+  expect(result.success).toBe(false);
+  if (!result.success) {
+    expect(result.error.issues[0].code).toBe("too_big");
+  }
+});
+
 test("z.stringFormat", () => {
   const ccRegex = /^(?:\d{14,19}|\d{4}(?: \d{3,6}){2,4}|\d{4}(?:-\d{3,6}){2,4})$/u;
 


packages/zod/src/v4/core/schemas.ts
```diff
@@ -449,11 +449,30 @@ export interface $ZodEmail extends $ZodType {
...
+        payload.issues.push({
+          origin: "string",
+          code: "too_big",
+          maximum: EMAIL_MAX_LENGTH,
+          inclusive: true,
+          input: payload.value,
+          inst,
+          continue: !def.abort,
+        });
+        return;
```
FucciUnavailable, line 472: return; // early return means malformed + too-long addresses only get too_big, never a format error

packages/zod/src/v4/core/schemas.ts
```diff
@@ -449,11 +449,30 @@ export interface $ZodEmail extends $ZodType {
...
+      }
+      patternCheck(payload);
+    };
   }
```
FucciUnavailable, line 476: This looks like a breaking change, but if we are gonna follow the RFC we may as well look at all the rules!

### First; Local-Part issue [RFC 5321 §4.5.3.1.1](https://www.rfc-editor.org/rfc/rfc5321.html#section-4.5.3.1.1): 
=> "a".repeat(65) + "@mail.com" // 65 chars before @, passes PR check, violates RFC 

### Second; Domain [RFC 5321 §4.5.3.1.2](https://www.rfc-editor.org/rfc/rfc5321.html#section-4.5.3.1.2):
=> "a@" + "b".repeat(256) + ".com" // domain > 255, passes PR check, violates RFC

### Third; Path  [RFC 5321 §4.5.3.1.3](https://www.rfc-editor.org/rfc/rfc5321.html#section-4.5.3.1.3):

These limits are mainly set for sending and receiving, I am not sure if this should affect users who use zod for other purposes!

## Question:
RFC 5321 path limits are an SMTP transport concern. Someone using Zod to validate form input, or building an app that doesn't send email via SMTP, might legitimately accept longer addresses. Worth asking explicitly: should Zod validate to SMTP transport rules, or just syntactic format?

FucciUnavailable (commented):
This PR introduces a length check that didn't exist before, which could be a breaking change for some users who don't rely on SMTP to send mail and might accept very long addresses.

However, I totally understand where this is coming from, I double checked Twilio's SendGrid docs for reference and they do set the hard limit proposed (to some extent)

> [RFC 5321, section 4.5.3.1](https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1)
 defines the parts and lengths of an email address.
Email addresses consist of two parts: the local-part that comes before the @ sign and the domain part that comes after.
The [local-part](https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1.1)
 can't exceed 64 octets in length.
The [domain part](https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1.2)
 can't exceed 255 octets in length.
Taken together, the whole email address can't exceed 254 octets in length

[Twilio Sendgrid Docs](https://www.twilio.com/docs/sendgrid/ui/sending-email/invalid-emails)


The maintainers would have to respond to the question I already added to the second comment of this review
=>  should Zod validate to SMTP transport rules, or just syntactic format?

colinhacks:
Closing this out. `z.email()` is intentionally not trying to match any email RFC — I wrote up the reasoning here: [An email regex for reasonable people](https://colinhacks.com/essays/reasonable-email-regex). Once you start enforcing parts of RFC 5321 you end up debating IPv6 literals and quoted local parts forever, and that's not the line `z.email()` is drawing.

If you need the 254-char cap specifically, `z.email().max(254)` covers it without changing the default for everyone else.

DORI2001:
Makes total sense — I read the essay and the tradeoffs are well-reasoned. `z.email().max(254)` is the right escape hatch for those who need it. Thanks for the quick response and for the context.

Issue #5789, opened 2026-03-18
CSP error triggered in production - `z.config({ jitless: true })` is not a great workaround


bjornpijnacker:
We see CSP errors in production builds because Zod checks `new Function("")` to determine whether unsafe-eval is possible. I don't really care that Zod _tries_ to use unsafe-eval, because our policy will disallow it. It being logged is the annoying part. Since these are reported to our violation report endpoint and clutter up (very important) security logs, we consider it a security regression.

In several issues, among which:
- https://github.com/colinhacks/zod/issues/4461
- https://github.com/colinhacks/zod/issues/5414

the recommended solution is to set `z.config({ jitless: true })`. This needs to happen _before_ any other imports occur that use Zod, which is problematic in our SPA-case, because this forces me to load and configure Zod before anything else can happen, leading to unnecessary slowdown in startup.

Other solutions, such as re-exporting Zod, are unstable, when considering third party libraries or generated code using regular Zod imports.

The above workaround is also undesirable, because it creates a forced bootstrap/import-order workaround, it introduces an extra loading step / waterfall in app startup, which make it not feel like a long-term solution to this problem.

Instead we would like a way to disable unsafe-eval behavior without relying on import order. Ideally this is build-time or package-level flag to disable JIT outright. Alternatively, this behavior should be opt-in. I don't think anyone is in good conscience enabling unsafe-eval on production environments. There have been similar proposals in earlier issues, but those were closed without a clear path forward. Could this be reconsidered?

reilem:
Sharing my own journey with this. I tried using `z.config({ jitless: true })` and at first it didn't work at all in my current React + Vite project. I tried putting it in a separate module script before my main `index.tsx` was loaded and I also tried putting it as the first thing in the `index.tsx` file, neither worked. I kept getting the violation right as the `index.tsx` was loaded in.

I could see in my minified JS bundle code that the `allowsEval` helper function was somehow being bundled and executed before my `z.config({ jitless: true })` line was being given the chance to run. It would simply always result in this error: 

<img width="840" height="379" alt="Image" src="https://github.com/user-attachments/assets/bd0ce94e-e22e-4408-bd27-5e16a7f61f67" />

Pointing to this code in the bundle:
```javascript
    util$a.allowsEval = Oa( () => {
        if (typeof navigator < "u" && navigator?.userAgent?.includes("Cloudflare"))
            return !1;
        try {
            const Gl = Function;
            return new Gl(""),
            !0
        } catch {
[...]

PR #5889, opened 2026-04-28
feat(core): share `globalConfig` across module systems via `globalThis`

colinhacks:
> _PR drafted with AI assistance._

## Summary

Promotes `globalConfig` to a `globalThis` singleton, mirroring the dual-package fix already applied to `globalRegistry` in #5452. After this change, `globalThis.__zod_globalConfig` is the single source of truth for `customError` / `localeError` / `jitless` regardless of how Zod was loaded (CJS, ESM, multiple bundles in a monorepo).

```ts
// packages/zod/src/v4/core/core.ts
(globalThis as GlobalThisWithConfig).__zod_globalConfig ??= {};
export const globalConfig: $ZodConfig = (globalThis as GlobalThisWithConfig).__zod_globalConfig!;
```

## Why

Today `globalConfig` is a fresh module-scope object on every load of Zod. In any project where some packages resolve `zod` as ESM and others as CJS — or where multiple bundles each carry their own copy — a single `z.config({ jitless: true })` only mutates one of them.

#5789 is exactly this footgun in production:

> After some more investigation I discovered that my monorepo was using zod in a few different places. Some still used CJS and some used ESM. This caused two instances of zod to be loaded in and so the initial config call was only being applied to one of the instances. **I fixed this by calling `z.config` twice, once in the index of the main application to configure the ESM instances and once in one of the CJS package index files to configure the CJS instance.**

`globalRegistry` already solved the same dual-package hazard in #5452. There's no good reason for `globalConfig` to behave differently — and it's surprising that it does, since they live in the same `core/` directory.

## Interaction with #5864 (`allowsEval` honours `jitless`)

Combined with #5864, this gives strict-CSP users a clean pre-import opt-out without any new API surface:

```html
<script>globalThis.__zod_globalConfig = { jitless: true };</script>
<script type="module" src="app.js"></script>
```

That's strictly more general than the bespoke `window.ZOD_NO_EVAL` flag proposed in #5862 — it works in Node/workers/edge/browser, it covers every config field (not just eval), and it falls out of an existing pattern instead of introducing a new one. Closing #5862 in favor of this.

## Compatibility

- **No default-behaviour change.** `??=` preserves any pre-populated object on `globalThis`, and `Object.assign(globalConfig, …)` keeps mutating the same identity that `z.config()` and external pre-set scripts both observe. `z.config()` returns the same object identity for a given realm.
- **Symmetric with the registry pattern.** No versioned key (`__zod_globalConfig`, not `__zod_globalConfig_v4`), so cross-major loads share state. That's the same trade-off #5452 already accepted; if it's wrong for config it's also wrong for the registry. Easy to revisit if it bites someone.
- **No public API change.**

Refs #5789, #5862, #5452, #5864, #4461, #5414.

## Test plan

- [x] `pnpm vitest run packages/zod/src/v4/classic/tests/global-config.test.ts` — 3 new tests pass:
  - `globalConfig` is referentially equal to `globalThis.__zod_globalConfig`
  - `z.config(...)` writes are observed via `globalThis.__zod_globalConfig`
  - Pre-set `globalThis.__zod_globalConfig` is preserved on import
- [x] `pnpm vitest run` — full suite, 3703/3703 pass (pre-push hook)
- [x] `pnpm lint` — clean
- [x] Existing `jitless-allows-eval.test.ts` (#5864) still passes — confirms the `globalConfig.jitless` short-circuit still works through the new globalThis-backed object

### DIFF
--- a/packages/zod/src/v4/classic/tests/global-config.test.ts
+++ b/packages/zod/src/v4/classic/tests/global-config.test.ts
@@ -1,0 +1,39 @@
+import { afterEach, expect, test } from "vitest";
+import * as z from "zod/v4";
+import * as core from "zod/v4/core";
+
+// `globalConfig` is attached to `globalThis.__zod_globalConfig` so that a
+// single config object is shared across CJS/ESM builds and across multiple
+// bundled copies of Zod in a monorepo. This mirrors the existing
+// `globalRegistry` -> `globalThis.__zod_globalRegistry` treatment.
+//
+// See #5789 for the footgun this fixes: users with both CJS and ESM
+// instances of Zod loaded had to call `z.config({ jitless: true })`
+// twice — once per instance — because each module-scope `globalConfig`
+// was a different object. With this change, one call updates state seen
+// by every loaded copy.
+
+afterEach(() => {
+  // Don't leak config mutations into other test files.
+  delete core.globalConfig.jitless;
+});
+
+test("globalConfig is singleton and attached to globalThis", () => {
+  expect(core.globalConfig).toBe((globalThis as any).__zod_globalConfig);
+});
+
+test("z.config writes are observed via globalThis.__zod_globalConfig", () => {
+  z.config({ jitless: true });
+  expect((globalThis as any).__zod_globalConfig.jitless).toBe(true);
+});
+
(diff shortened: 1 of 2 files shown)


colinhacks:
Landed in Zod 4.4

PR #5891, opened 2026-04-28
fix(v4): apply key schema transforms in z.record()

### LINKED ISSUE #5296: Records do not transform keys, even though the Typescript types says it does.

colinhacks:
## Summary

`z.record()` with a finite-key keyType (literal, enum, union of literals, etc.) skipped the keyType entirely on its fast path and used the original input key as the output property name. Any `.transform()` chained onto the key schema was therefore dropped at runtime, contradicting the inferred output type.

`z.partialRecord()` already did the right thing — it works by clearing the keyType's `values` to force the slow path, where the keyType is parsed per input key.

```ts
z.record(z.literal("a").transform(() => "b" as const), z.string())
  .parse({ a: "John" });
// before: { a: "John" }   ← key not transformed, type lied
// after:  { b: "John" }
```

## Approach

Run the keyType once per known input value to derive the output property name. The fast path's job — enforcing required-key semantics for finite-key records (`Record<K, V>` vs `Partial<Record<K, V>>`) and detecting unrecognized keys — is preserved. Lookup into the input object still uses the input-side key.

This is the same key-resolution work the slow path already does and that `partialRecord` relies on; the fast path was the lone code path skipping it.

## Behavior

- `z.record(z.literal("a").transform(() => "b"), v).parse({ a: 1 })` → `{ b: 1 }` (was `{ a: 1 }`).
- `z.record(z.literal(["a","b"]), v).parse({ a: 1 })` — still fails on missing required key `b`.
- `z.record(z.literal(["a","b"]), v).parse({ a: 1, c: 2 })` — still flags `c` as unrecognized.
- `z.partialRecord(...)` — unchanged.
- Discriminated unions over codec/transform discriminators — unchanged (this fix doesn't touch `values`/`propValues` propagation).

Fixes #5296. Supersedes #5867 (which added the same intent but without the underlying source change and bundled unrelated build-config edits).

## Test plan

- [x] Added regression test covering `z.literal(...).transform(...)`, multi-value literal+transform, enum+transform, required-key semantics, and `partialRecord` parity.
- [x] Existing record + discriminated-union test files pass.
- [x] Full vitest suite (3713 tests) passes locally.

### DIFF
--- a/packages/zod/src/v4/classic/tests/record.test.ts
+++ b/packages/zod/src/v4/classic/tests/record.test.ts
@@ -273,37 +273,64 @@ test("union exhaustiveness", () => {
 });
 
 test("applies transforms on the key schema (#5296)", () => {
-  // Single literal with transform — was returning the original key before the fix.
   const single = z.record(
     z.literal("a").transform(() => "b" as const),
     z.string()
   );
   expect(single.parse({ a: "John" })).toEqual({ b: "John" });
 
-  // Multi-value literal with transform — each known input value is run through
-  // the keyType to derive the output property name.
   const multi = z.record(
     z.literal(["a", "b"]).transform((k) => k.toUpperCase()),
     z.number()
   );
   expect(multi.parse({ a: 1, b: 2 })).toEqual({ A: 1, B: 2 });
 
-  // Required-key semantics are preserved: missing input keys still error.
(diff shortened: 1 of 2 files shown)


colinhacks:
Landed in Zod 4.4

PR #5894, opened 2026-04-29
feat(v4): add z.script() for unicode script validation

### LINKED ISSUE #5804: Feature Request: Built-in Unicode / Script Validators
### Problem

Validating non-Latin input (e.g., Arabic) in **Zod** requires custom regex like:

```ts
z.string().regex(/^[\u0600-\u06FF...]+$/)
```

This is repetitive, hard to read, and error-prone.

---

### Proposal

Add built-in helpers for common scripts:

```ts
z.string().arabic()
z.string().latin()
z.string().script("Arabic")
```

---

### Benefit

* Better DX
* Less repetition
* Safer, standardized Unicode validation

---

### Note

Could leverage Unicode property escapes:

```ts
/^\p{Script=Arabic}+$/u
```

FucciUnavailable:
Will eventually close #5804 

This PR adds z.script(name) which validates that every character in the string belong to the given unicode script, using the \p{Script=lang} property escape built into the JS engine

### Expected behavior:

z.script("Arabic").parse("حفلة");   // passes
z.script("Latin").parse("hello");    // passes
z.script("Cyrillic").parse("привет"); // passes

z.string().script("Cyrillic") // the method form also available and passed the tests

z.script("NotAScript") // throws syntax error at Schema creation; The JS Engine rejects unknowin script identifiers.

### Notes:
 
 - Invalid script names fail fast at schema creation time (not parse time) => the JS engine rejects unknown \p{Script=...} identifiers immediately
  
- Strict by design: spaces, punctuation, and diacritics outside the script's own code points will fail (suitable for validating individual words/tokens, not full sentences) => Support can be extended with Script_Extensions instead of Script `^\p{Script_Extensions=Arabic}+$`
 
 - Script names are the official Unicode identifiers (Old_Persian, Linear_B..etc) => casing matters

  - New scripts are supported automatically as Node.js/V8 updates their ICU Unicode tables => nothing to maintain in Zod

### Motivation:
Making zod even more accessible worldwide! :D

## Important:

Named aliases (z.arabic(), z.latin(), etc.) are intentionally left out each would be a one-liner in classic/schemas.ts, but picking which of the 150+ Unicode scripts deserve a named helper is an arbitrary call.
z.script("Arabic") is already readable, and the escape hatch is there if we want to ship named aliases later.

Reference Issue: #5804

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -389,6 +389,7 @@ export interface ZodString extends _ZodString<core.$ZodStringInternals<string>> 
   ): this;
   /** @deprecated Use `z.iso.duration()` instead. */
   duration(params?: string | core.$ZodCheckISODurationParams): this;
+  script(name: string, params?: string | core.$ZodCheckScriptParams): this;
 }
 
 export const ZodString: core.$constructor<ZodString> = /*@__PURE__*/ core.$constructor("ZodString", (inst, def) => {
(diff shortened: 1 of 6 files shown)


colinhacks:
Closing — I declined this in #5804. Regex covers the use case cleanly, and I want to be careful about what earns a dedicated API. Thanks for the idea!

PR #5895, opened 2026-04-29
fix(v4): apply util.Writeable in mini constructors and extend/safeExtend/partial/required

colinhacks:
## Summary

Follow-up to #5882. After fixing `strictObject`/`looseObject` in classic, audited the rest of the surface that takes (or constructs) a `$ZodLooseShape` and found four more places where `readonly` modifiers were leaking into the displayed shape. This brings classic and mini to parity.

## What was leaking, before

```ts
// classic
Base.extend({ get sub() { return z.array(Extended); } })
// ZodObject<{ name: ZodString; readonly sub: ZodArray<...>; }, $strip>
//                              ^^^^^^^^

Base.extend({ a: z.string(), b: z.number() } as const)
// ZodObject<{ name: ZodString; readonly a: ZodString; readonly b: ZodNumber; }, $strip>

Extended.partial()
// preserves readonly from Shape

// mini
zm.object({ name: zm.string(), get sub() {...} })
zm.strictObject({ name: zm.string(), get sub() {...} })
zm.looseObject({ name: zm.string(), get sub() {...} })
zm.extend(Base, { get sub() {...} })
// All four leak readonly — mini was missing Writeable everywhere.
```

## Changes

**Mini** (`packages/zod/src/v4/mini/schemas.ts`):
- `object`, `strictObject`, `looseObject` wrap `T` in `util.Writeable<T>` (mini was missing it on all three constructors).
- `extend`, `safeExtend` wrap `U` in `util.Writeable<U>`.
- `partial`, `required` use `-readonly` mapped types.

**Classic** (`packages/zod/src/v4/classic/schemas.ts`):
- `extend`, `safeExtend` wrap `U` in `util.Writeable<U>`.
- `partial`, `required` use `-readonly` mapped types.

## Why partial/required need `-readonly`

In normal flows the input `Shape` arrives prettified after this PR, so the modifier is defensive. But `partial()` and `required()` operate on a `Shape` from `._zod.config`, and if any future entry point introduces a readonly shape (e.g. someone reads `.shape` and pipes it manually), the inline mapped types `{ [k in keyof Shape]: ZodOptional<Shape[k]> }` would preserve readonly. The `-readonly` modifier costs nothing and closes that hole.

## Behavior unchanged

- `z.infer` output is identical — `$InferObjectOutput` already strips readonly via `-readonly`.
- No runtime behavior change.
- This is a display/shape-typing alignment fix only.

## Test plan

- [x] Added `expectTypeOf<(typeof X)["shape"]>` assertions in both `packages/zod/src/v4/classic/tests/recursive-types.test.ts` and `packages/zod/src/v4/mini/tests/recursive-types.test.ts`. Covers object/strictObject/looseObject with getters, extend/safeExtend, `as const` shapes, partial(), required().
- [x] `pnpm vitest run` — 333 files / 3741 tests passing
- [x] Confirmed via `tsc --noErrorTruncation` that hovers no longer surface `readonly` on any of the audited paths
- [x] Biome format/lint clean

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1220,11 +1220,11 @@ export interface ZodObject<
   /** This is the default behavior. This method call is likely unnecessary. */
   strip(): ZodObject<Shape, core.$strip>;
 
-  extend<U extends core.$ZodLooseShape>(shape: U): ZodObject<util.Extend<Shape, U>, Config>;
+  extend<U extends core.$ZodLooseShape>(shape: U): ZodObject<util.Extend<Shape, util.Writeable<U>>, Config>;
 
   safeExtend<U extends core.$ZodLooseShape>(
     shape: SafeExtendShape<Shape, U> & Partial<Record<keyof Shape, core.SomeType>>
-  ): ZodObject<util.Extend<Shape, U>, Config>;
+  ): ZodObject<util.Extend<Shape, util.Writeable<U>>, Config>;
 
   /**
    * @deprecated Use [`A.extend(B.shape)`](https://zod.dev/api?id=extend) instead.
@@ -1241,7 +1241,7 @@ export interface ZodObject<
 
   partial(): ZodObject<
     {
-      [k in keyof Shape]: ZodOptional<Shape[k]>;
+      -readonly [k in keyof Shape]: ZodOptional<Shape[k]>;
     },
     Config
   >;
@@ -1249,7 +1249,7 @@ export interface ZodObject<
     mask: M & Record<Exclude<keyof M, keyof Shape>, never>
   ): ZodObject<
     {
-      [k in keyof Shape]: k extends keyof M
+      -readonly [k in keyof Shape]: k extends keyof M
         ? // Shape[k] extends OptionalInSchema
           //   ? Shape[k]
           //   :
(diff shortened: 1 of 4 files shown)


colinhacks:
## Perf

Reasonable concern given this is purely an aesthetic fix — measured before/after on a worst-case bench (15-key ZodObject, 5 chained `.extend()`s, full recursive-types fixture exercising every wrapper, plus `partial()`/`required()`/`merge()`/`safeExtend()`/`pick()`/`omit()`):

| | main | this PR | delta |
|---|---|---|---|
| Types | 121,800 | 121,993 | +193 (+0.16%) |
| Instantiations | 373,014 | 373,459 | +445 (+0.12%) |
| Check time (warm, n=3) | 1.36–1.50s | 1.36–1.40s | indistinguishable |

Recursive-only bench (every `ZodLazy`/`ZodArray`/`ZodOptional`/etc. variant referencing a self-recursive `A`):

| | main | this PR | delta |
|---|---|---|---|
| Types | 109,166 | 109,237 | +71 (+0.065%) |
| Instantiations | 328,657 | 328,809 | +152 (+0.046%) |
| Check time (warm, n=3) | 1.26s | 1.26s | identical |

Full `packages/zod/src/v4` vitest typecheck (steady state, n=3): main 924–934ms, this PR 824–883ms — within noise, slightly faster on the PR if anything.

Why so cheap: `Writeable<T>` is a homomorphic mapped type with `& {}` tail. TS has fast paths for these, and the result is structurally identical to the input, so it doesn't compound through downstream operations. The added `-readonly` modifier on `partial()`/`required()` mapped types replaces existing inline mapped types — net zero.

I wouldn't ship this if it cost wall time, but it's free.

---

colinhacks:
Landed in Zod 4.4

PR #5897, opened 2026-04-29
perf(v4): lazy-bind builder methods to shared internal prototype

colinhacks:
## Summary

Builder methods on Zod schemas (`.optional`, `.nullable`, `.array`, `.pick`, `.partial`, `.regex`, `.min`, etc.) are currently allocated as fresh per-instance closures during `init`. On a plain `z.string()` that's ~50 closures up front — dominated by V8 `JSFunction` + `Context` bytes — even when the user never calls them.

This PR moves them onto a hidden prototype layer inserted between `_.prototype` (user-extension space) and the parent. On first access from any instance the lazy getter allocates `fn.bind(this)` and caches it as an own enumerable property; subsequent accesses skip the getter entirely.

Detached usage keeps working — that was the dealbreaker for #5870, and it's why this PR adds a regression test guarding it:

```ts
const opt = z.string().optional;
const detached = opt(); // works: ZodOptional<ZodString>
detached.parse(undefined); // ok
```

The cached value is a bound function with `this` already resolved on first access, so it's safe to detach.

Parse-family methods (`parse`, `safeParse`, `parseAsync`, `encode`, `decode`, etc.) intentionally stay as eager per-instance closures — they're the hot path AND the most-detached methods (`arr.map(schema.parse)`, `const { parse } = schema`), so paying the allocation up front is worth the monomorphic call site and predictable behavior.

## Benchmarks

Run with `node --expose-gc` + `tsx`, 10k instances each, `process.memoryUsage().heapUsed` delta after two `global.gc()` calls. Numbers are stable across runs (variance < 5% on heap, ~10% on parse times).

### Heap per instance (lower is better)

| Schema                              |    main |  this PR |  Δ      |
| ----------------------------------- | ------: | -------: | ------: |
| `z.string()`                        | 12.8 KB |  7.5 KB  | **−41%** |
| `z.string().min(1).max(10).email()` | 64.5 KB | 40.1 KB  | **−38%** |
| `z.number()`                        |  8.2 KB |  4.5 KB  | **−45%** |
| `z.boolean()`                       |  7.3 KB |  3.7 KB  | **−49%** |
| `z.array(z.string())`               | 20.4 KB | 11.2 KB  | **−45%** |
| `z.object({ a, b })`                | 31.7 KB | 18.4 KB  | **−42%** |
| `z.string().optional()`             | 21.4 KB | 12.5 KB  | **−42%** |

The savings come from not allocating `JSFunction` + `Context` pairs for the ~50 builder methods up front. Methods are still bound and cached on the instance _after first access_, so a schema you actually `.optional()` ends up paying close to today's cost on that one method, but nothing for the other ~49 you didn't touch.

### Construction time (lower is better)

| Schema                              |    main |  this PR |     Δ    |
| ----------------------------------- | ------: | -------: | -------: |
| `z.string()`                        | 6.24 µs | 4.41 µs  | **−29%** |
| `z.string().min(1).max(10).email()` | 45.3 µs | 35.3 µs  | **−22%** |
| `z.number()`                        | 4.09 µs | 3.30 µs  | **−19%** |
| `z.array(z.string())`               | 10.5 µs | 7.35 µs  | **−30%** |
| `z.object({ a, b })`                | 16.2 µs | 12.0 µs  | **−26%** |

Construction is faster because the per-instance loop that used to assign ~50 closures now installs them on the prototype once per constructor and skips the work on every subsequent instance.

### Parse hot path (lower is better, 100k iterations, 3-run avg)

| Workload                                         |    main | this PR  |   Δ |
| ------------------------------------------------ | ------: | -------: | ---: |
| `z.string().min(1).max(50).email()` × 100k       | 17.8 ms | ~17 ms   | ~neutral |
| `z.object({ name, age, tags }).safeParse` × 100k | 32.1 ms | ~26 ms   | ~neutral / slightly faster |

Parse-family methods are unchanged (eager per-instance closures), so any movement here is JIT/GC noise. Run-to-run variance was higher than the delta in both directions; we treat this as effectively unchanged.

### Property access throughput

| Access            |    main  | this PR  |
| ----------------- | -------: | -------: |
| `schema._zod`     | 6.0 ns  | 6.5 ns   |
| `schema.parse`    | 5.8 ns  | 5.9 ns   |
| `schema.optional` | 6.3 ns  | 6.0 ns   |

After first access the lazy methods are cached as own properties, so steady-state access cost is identical to before (the getter only runs once per instance per method).

### Bundle size

The new infrastructure is small fixed overhead and lives entirely in classic — mini bundles are bytes-for-bytes identical to `main`. Bundled with `esbuild --bundle --minify --format=esm`, gzipped:

**Classic — `import * as z from "zod"; z.string().parse("")`**

|         |  raw  |  gzip  |
| ------- | ----: | -----: |
| main    | 56,015 B | 15,677 B |
| this PR | 56,875 B | 15,852 B |
| Δ       | **+860 B (+1.5%)** | **+175 B (+1.1%)** |

**Classic — rich object schema (`z.object({ email, age, tags, role }).strict()` with checks)**

|         |  raw  |  gzip  |
| ------- | ----: | -----: |
| main    | 68,448 B | 19,100 B |
| this PR | 69,583 B | 19,321 B |
| Δ       | **+1,135 B (+1.7%)** | **+221 B (+1.2%)** |

**Mini — `import * as z from "zod/mini"; z.parse(z.string(), "")`**

|         |  raw  |  gzip  |
| ------- | ----: | -----: |
| main    | 6,664 B  | 2,707 B  |
| this PR | 6,664 B  | 2,705 B  |
| Δ       | **0 B (identical)** | **−2 B (gzip noise; identical)** |

**Mini — rich object schema**

|         |  raw  |  gzip  |
| ------- | ----: | -----: |
| main    | 15,389 B | 5,367 B  |
| this PR | 15,389 B | 5,365 B  |
| Δ       | **0 B (identical)** | **−2 B (gzip noise; identical)** |

Notes:
- Mini bundles are byte-for-byte identical to `main`. The lazy-bind layer lives entirely in `classic/schemas.ts`; nothing was added to `core/core.ts`.
- Classic's overhead is the helper itself plus the inline method bodies — small fixed cost dominated by infrastructure rather than per-method weight.
- This is the cost paid for ~40% per-instance heap savings and ~20–30% faster construction. For apps that build many schemas (typical Zod usage), the runtime savings dwarf the bundle delta within a few hundred instances.

## Implementation

- `core/core.ts` — **no changes**. The lazy-bind layer is entirely classic-side.

- `classic/schemas.ts`:
  - One ~30-line `_installLazyMethods(inst, group, methods)` helper. Installs each method as a non-enumerable getter directly on the instance's prototype (idempotent per (prototype, group) via `WeakMap`). The getter, on first access from an instance, allocates `fn.bind(this)` and stores it as an own property on the instance — subsequent accesses skip the getter entirely.
  - All builder methods on `ZodType`, `_ZodString`, `ZodNumber`, `ZodArray`, `ZodObject` are passed in as inline method-shorthand bodies — no top-level `_sharedXxx` named functions, no `_shared` prefix.

## Compatibility

- All 3,743 tests pass (full repo, including v3, v4 classic, v4 mini).
- No public type changes. Interface declarations on `ZodType`/`ZodObject`/etc. are unchanged.
- Detached method usage preserved (new `detached-methods.test.ts` covers ~30 such patterns).
- `z.ZodType.prototype.x = ...` user extensions still work. The lazy methods are installed as **non-enumerable** getters on `_.prototype`, so the existing copy-loop in `core.$constructor` (which uses `Object.keys`, enumerable own only) doesn't see them, and any enumerable user extension on the same prototype shadows the lazy method as before.
- `Object.keys(z.ZodString.prototype)` is unchanged (still empty by default) and `Object.keys(schema)` enumerates the same keys as before once methods have been touched.
- One observable change: `Object.getOwnPropertyNames(z.ZodString.prototype)` and `'optional' in z.ZodString.prototype` now reflect the lazy methods (was previously empty / `false`). This is arguably more intuitive — the methods conceptually do live on the prototype.

## Test plan

- [x] `pnpm vitest run` — 3,743 passed
- [x] `pnpm tsc --noEmit -p packages/zod/tsconfig.json` — same 13 pre-existing unused-variable errors as `main`, no new errors
- [x] Sanity: detached methods produce working schemas
- [x] Sanity: lazy bind caches on instance after first access; layer is not enumerated by `Object.keys`

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -253,13 +253,13 @@ export const ZodType: core.$constructor<ZodType> = /*@__PURE__*/ core.$construct
   // detachability preserved (`const m = schema.optional; m()` works), and
   // shared underlying function references across all instances.
   _installLazyMethods(inst, "ZodType", {
-    check(...chks: any[]) {
+    check(...chks) {
       const def = this.def;
       return this.clone(
         util.mergeDefs(def, {
           checks: [
             ...(def.checks ?? []),
-            ...chks.map((ch: any) =>
+            ...chks.map((ch) =>
               typeof ch === "function" ? { _zod: { check: ch, def: { check: "custom" }, onattach: [] } } : ch
             ),
           ],
@@ -267,26 +267,26 @@ export const ZodType: core.$constructor<ZodType> = /*@__PURE__*/ core.$construct
         { parent: true }
       );
     },
-    with(...chks: any[]) {
+    with(...chks) {
       return this.check(...chks);
     },
-    clone(def?: any, params?: any) {
+    clone(def, params) {
       return core.clone(this, def, params);
     },
     brand() {
       return this;
     },
-    register(reg: any, meta: any) {
+    register(reg, meta) {
       reg.add(this, meta);
       return this;
     },
-    refine(check: any, params?: any) {
+    refine(check, params) {
       return this.check(refine(check, params));
     },
-    superRefine(refinement: any, params?: any) {
+    superRefine(refinement, params) {
       return this.check(superRefine(refinement, params));
     },
-    overwrite(fn: any) {
+    overwrite(fn) {
       return this.check(checks.overwrite(fn));
     },
     optional() {
@@ -301,37 +301,37 @@ export const ZodType: core.$constructor<ZodType> = /*@__PURE__*/ core.$construct
     nullish() {
       return optional(nullable(this));
     },
-    nonoptional(params?: any) {
+    nonoptional(params) {
       return nonoptional(this, params);
     },
     array() {
       return array(this);
     },
-    or(arg: any) {
+    or(arg) {
       return union([this, arg]);
     },
-    and(arg: any) {
+    and(arg) {
       return intersection(this, arg);
     },
-    transform(tx: any) {
+    transform(tx) {
       return pipe(this, transform(tx));
     },
-    default(d: any) {
+    default(d) {
       return _default(this, d);
     },
-    prefault(d: any) {
+    prefault(d) {
       return prefault(this, d);
     },
-    catch(params: any) {
+    catch(params) {
       return _catch(this, params);
     },
-    pipe(target: any) {
+    pipe(target) {
       return pipe(this, target);
     },
     readonly() {
       return readonly(this);
     },
-    describe(description: string) {
+    describe(description) {
       const cl = this.clone();
       core.globalRegistry.add(cl, { description });
       return cl;
@@ -339,7 +339,8 @@ export const ZodType: core.$constructor<ZodType> = /*@__PURE__*/ core.$construct
     meta(...args: any[]): any {
       // overloaded: meta() returns the registered metadata, meta(data)
       // returns a clone with `data` registered. The mapped type picks
-      // up the second overload, so we return `any` to satisfy both.
+      // up the second overload, so we accept variadic any-args and
+      // return `any` to satisfy both at runtime.
       if (args.length === 0) return core.globalRegistry.get(this);
       const cl = this.clone();
       core.globalRegistry.add(cl, args[0]);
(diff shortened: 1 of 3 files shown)


colinhacks:
Landed in Zod 4.4

PR #5898, opened 2026-04-29
fix(v4): skip __proto__ key in object catchall

colinhacks:
## Summary

In `handleCatchall()`, an unrecognized `__proto__` key in the input is currently assigned directly to a fresh `{}` result object via `(final.value)[key] = result.value`. Because that goes through the `__proto__` setter on a plain object, the assignment **replaces the prototype of the parsed result**, giving the validated object attacker-controlled inherited properties. `Object.prototype` itself is not affected.

PoC on `4.3.6`:

```ts
const schema = z.looseObject({ name: z.string() });
const input = JSON.parse('{"__proto__":{"isAdmin":true},"name":"alice"}');
const parsed = schema.parse(input);
parsed.isAdmin // true  ← inherited from injected prototype
```

Same shape reproduces with `passthrough()` and `catchall(...)`, sync/async, and `{ jitless: true }` (the JIT fastpass falls through to the same `handleCatchall` for unknown keys).

## Fix

One-line guard in the catchall loop:

```ts
for (const key in input) {
  if (key === "__proto__") continue;
  ...
}
```

Mirrors the existing `__proto__` skip in `$ZodRecord` (`schemas.ts:2867`) and the v3 object-assembly guard at `parseUtil.ts:144`. Deliberately conservative — `for..in` iteration of input is preserved, so behavior for any input *other than* one with a `__proto__` own key is unchanged. Inherited prototype-chain enumeration (the separate concern in GHSA-9hmj-h52h-mxcf) is not addressed here; that's a narrower attack model and a noisier behavior change worth handling separately.

Reported in:

- GHSA-r34p-xfmx-58wv (critical)
- GHSA-84jv-fqfx-wxhr

## Test plan

- [x] 5 new tests in `object.test.ts` covering `looseObject` / `passthrough` / `catchall(unknown)` / `safeParseAsync` + `{ jitless: true }` / `strict()`, asserting `Object.getPrototypeOf(parsed) === Object.prototype` and that the injected key is not visible on the result.
- [x] `pnpm vitest run` — 3753 passed across 333 files, no regressions.

### DIFF
--- a/packages/zod/src/v4/classic/tests/object.test.ts
+++ b/packages/zod/src/v4/classic/tests/object.test.ts
@@ -1,4 +1,4 @@
-import { expect, expectTypeOf, test } from "vitest";
+import { describe, expect, expectTypeOf, test } from "vitest";
 import * as z from "zod/v4";
 import * as core from "zod/v4/core";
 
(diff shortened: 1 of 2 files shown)


colinhacks:
Landed in Zod 4.4

PR #5900, opened 2026-04-29
fix(v4): reject tuple holes before required defaults

colinhacks:
## Summary
- Reject absent optional-output tuple slots when a later slot has required output, such as a defaulted element.
- Add regression coverage for `exactOptional()` before `.default()` plus refined optional async/sync parity.

## Test plan
- `pnpm vitest run packages/zod/src/v4/classic/tests/tuple.test.ts`
- `pnpm vitest run packages/zod/src/v4/classic/tests/optional.test.ts`


colinhacks:
Landed in Zod 4.4

Issue #5663, opened 2026-01-26
intersection (.and(...)) does not run check, superRefine or transform functions (and probably others)


mrft:
When reusing a schema in another schema, I am assuming that the entire schema will be reused, including the checks, transforms and other hooks. This is also suggested by the types (the output types reflect the transformed type, also when that schema is reused in an intersection).

In a more specific case, not even the default ran (cfr. latest test case)

I added these tests to packages/zod/src/v4/core/tests/extend.test.ts which I think should work (but that fail in zod 4.3.6).

```typescript
test("intersection check fn should run ({ x: string; } & { a: string; })", () => {
  const spy = vi.fn();
  const alwaysFailingCheckFn = (ctx: any) => {
    spy();
    ctx.addIssue({
      code: "custom",
      message: "This check always fails",
    });
  };

  const xSubSchema = z
    .strictObject({ x: z.string() })
    .check(alwaysFailingCheckFn);

  const problemSchema = z.intersection(
    xSubSchema,
    z.strictObject({ a: z.string() }).check(alwaysFailingCheckFn),
  );

  type TInput = z.input<typeof problemSchema>; // type TProblemInput = ({ x: string; } & { a: string; })  | ({ x: string; } & { b: boolean; })

  const parseResult = problemSchema.safeParse({
    x: "test",
    a: "hello",
  } satisfies TInput);
  expect(spy).toHaveBeenCalledTimes(2);
  expect(parseResult.error).not.toBe(undefined);
});

test("intersection superRefine should run ({ x: string; } & { a: string; })", () => {
  const spy = vi.fn();
  const alwaysFailingSuperRefine = (_data: any, ctx: any) => {
    spy();
    ctx.addIssue({
      code: "custom",
      message: "This check always fails",
    });
  };

  const xSubSchema = z
    .strictObject({ x: z.string() })
    .superRefine(alwaysFailingSuperRefine);

  const problemSchema = z.intersection(
    xSubSchema,
    z.strictObject({ a: z.string() }).superRefine(alwaysFailingSuperRefine),
  );
  type TInput = z.input<typeof problemSchema>; // type TProblemInput = ({ x: string; } & { a: string; })  | ({ x: string; } & { b: boolean; })

  const parseResult = problemSchema.safeParse({
    x: "test",
    a: "hello",
  } satisfies TInput);
  expect(spy).toHaveBeenCalledTimes(2);
  expect(parseResult.error).not.toBe(undefined);
});

[...]

PR #5903, opened 2026-04-30
Preserve merged values in nested intersections

colinhacks:
## Summary
- Preserve a valid merged intersection output even when validation issues remain on the payload.
- Add a regression for nested strict object intersections preserving defaults from an inner intersection.

Refs #5663

## Test plan
- pnpm vitest run packages/zod/src/v4/classic/tests/intersection.test.ts
- pnpm dev --eval 'import * as z from "zod/v4"; const inner = z.intersection(z.strictObject({ x: z.string().default("X default"), y: z.number() }), z.strictObject({ z: z.boolean() })); const schema = z.intersection(inner, z.strictObject({ a: z.string() })); const result = schema.safeParse({ y: 34, z: true, a: "hello" }); if (!result.success || result.data.x !== "X default") process.exit(1);'
- pnpm biome check packages/zod/src/v4/core/schemas.ts packages/zod/src/v4/classic/tests/intersection.test.ts
- git push pre-push hook: pnpm test

### DIFF
--- a/packages/zod/src/v4/classic/tests/intersection.test.ts
+++ b/packages/zod/src/v4/classic/tests/intersection.test.ts
@@ -67,6 +67,29 @@ test("object intersection: strict + strict", () => {
   `);
 });
 
+test("nested strict object intersection preserves merged value with defaults", () => {
+  const inner = z.intersection(
+    z.strictObject({
+      x: z.string().default("X default"),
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing this narrower version in favor of #5905, which handles the full strict intersection issue.

Issue #2573, opened 2023-07-12
Intersection between `z.object` and `z.record` incorrectly merges validators


lars-berger:
When intersecting a `z.object` and a `z.record`, zod applies validators for the keys and values of the record to the rest of the object. 

```typescript
const ConfigSchema = z
  .object({
    name: z.object({
      first: z.string(),
    }),
  })
  .and(
    z.record(
      z.enum(['a', 'b']),
      z.object({
        sub: z.string(),
      })
    )
  );

const validInput: z.infer<typeof ConfigSchema> = { // typescript type is correct and doesn't complain here 👍
  name: {
    first: '',
  },
  a: {
    sub: '',
  },
};

ConfigSchema.parse(validInput);
// ^ this throws 2 errors because:
// 1. 'name' doesn't match ['a', 'b']
// 2. 'name.sub' is required (even though this will never exist)
```
<img width="185" alt="image" src="https://github.com/colinhacks/zod/assets/34844898/90ce3c93-ec39-4357-a782-a0f0928f3e2e">

**[Reproduction URL: stackblitz.com](https://stackblitz.com/edit/typescript-dauayd?file=index.ts,package.json)**

The same problem occurs with intersections of records that have native enum keys and template literal keys (eg. `[key: dynamic/${string}]`).

GauBen:
Related to #2195

yovanoc:
is there any alternative?

Aurailus:
Just wondering if a fix for this is planned?

marioparaschiv:
Will this ever be fixed? Is it a limitation of how zod is built?

Teages:
I think it is fixed via `looseRecord` in v4.3.0
[...]

PR #5904, opened 2026-04-30
Add strictRecord and strip record extras by default

### LINKED ISSUE #2200: `z.object` to be mergeable with `z.record`, or ability to define keys by a custom type
I have a type that's basically like this

```ts
type YKeys = `y${number}`;
type ZKeys = `z${number}`;
type DataKeys = `d${number}`;

type Point = {
  x: string;
  id: string;
} & Record<YKeys, number> & Record<ZKeys, number> & Record<DataKeys, { stuff: string }>

```

I have the schemas for the special keys
```ts 
const yKeysSchema = z.custom<`i${number}`>((val) => /^y(\d+)$/.test(val as string));
const zKeysSchema = z.custom<`i${number}`>((val) => /^z(\d+)$/.test(val as string));
const dataKeysSchema = z.custom<`i${number}`>((val) => /^d(\d+)$/.test(val as string));
```

and I used to be able to get by with just a superRefine call to a `pointValidityRefinement` function that will check the keys, with something like

```ts
const pointSchema = z.object({ x: z.string(), id: z.string() }).passthrough().superRefine(pointValidityRefinement);
```

But since the last update I get errors like 

```
Type 'objectOutputType<{ x: ZodNumber; id: ZodString; }, ZodTypeAny, "passthrough">' is not assignable to type 'Point'.
    'string' and '`y${number}`' index signatures are incompatible.
      Type 'unknown' is not assignable to type 'number | undefined'.
```

I tried to change the schema of the point to something like

```ts
const pointSchema = z.object({ x: z.string(), id: z.string() })
  .merge(z.record(yKeysSchema, z.number())
  .merge(z.record(zKeysSchema, z.number())
  .merge(z.record(dataKeysSchema, z.object({ stuff: z.string() }))
```
but that doesn't work.

Is there any way I can do this?

btw, I'm using TypeScript 4.8 and latest zod (3.21.4)

### LINKED ISSUE #5666: Stripping Unrecognised keys from enum record.
I want to validate a record created by a zod enum. I'd like to replicate the standard .object functionality where unrecognised keys are stripped from the output but do not produce a validation error. `.looseRecord` stops it erroring but also allows unrecognised values to pass through.

```ts
const supportedLanguagesSchema = z.enum(['en', 'fr']);

const translationsSchema = z.record(supportedLanguagesSchema, z.string());

translationsSchema.parse({
  en: 'Hello',
  fr: 'Bonjour',
  de: 'Guten Tag',
}); // => { en: 'Hello', fr: 'Bonjour' }
```

Is there an option for this or way of doing this that I am missing?

colinhacks:
Make `z.record()` and `z.partialRecord()` strip properties whose keys do not match the key schema by default, matching `z.object()`'s default strip behavior. This also makes records easier to compose in intersections: a record branch can validate the keys it understands without rejecting unrelated properties owned by another branch.

Add `z.strictRecord()` for the previous invalid-key behavior in Classic and Mini, and update finite-key `z.looseRecord()` so unrecognized keys pass through consistently. JSON Schema import keeps strict `propertyNames` behavior by producing `z.strictRecord()`.

Validated with the focused record/JSON Schema tests and the full pre-push `pnpm test` suite.

Fixes #5666.
Fixes #2200.

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -1853,13 +1853,15 @@ const AnyObject = z.record(Keys, z.unknown());
 // Record<string | number | symbol, unknown>
 ```
 
-To create an object schemas containing keys defined by an enum:
+To create an object schema containing keys defined by an enum:
 
 ```ts
 const Keys = z.enum(["id", "name", "email"]);
 const Person = z.record(Keys, z.string());
 // { id: string; name: string; email: string }
 ```
+
+By default, keys that don't match the key schema are stripped from the parsed result.
 
 Zod supports numeric keys inside records in a way that closely mirrors TypeScript itself. A `number` schema, when used as a record key, will validate that the key is a valid "numeric string". Additional numerical constraints (min, max, step, etc.) will also be validated.
 
@@ -1870,7 +1872,7 @@ numberKeys.parse({
   2: "two", // ✅
   "1.5": "one", // ✅
   "-3": "two", // ✅
-  abc: "one" // ❌
+  abc: "one" // stripped
 });
 
 // further validation is also supported
@@ -1879,8 +1881,8 @@ intKeys.parse({
   0: "zero", // ✅
   1: "one", // ✅
   2: "two", // ✅
-  12: "twelve", // ❌
-  abc: "one" // ❌
+  12: "twelve", // stripped
+  abc: "one" // stripped
 });
 ```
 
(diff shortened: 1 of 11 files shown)


colinhacks:

This should cover #2573 as well. I verified the v4 equivalent of that repro with `z.partialRecord(z.enum(["a", "b"]), ...)` on this branch, and the object-plus-record intersection now parses successfully because the record branch strips sibling keys it does not own. The template-literal key variant mentioned in the issue passes for the same reason.

colinhacks:
Closing this. The motivation was composability — records not intersecting cleanly with objects — and that turned out to be a separate bug: a record reports a key it doesn't own as `invalid_key` instead of `unrecognized_keys`, so the intersection never reconciles it against the other operand. Fixing that classification gets #2200 and #2573 without touching what `z.record` does on its own.

Changing the default to strip also costs something I'd rather keep: once a record strips instead of flagging, an intersection can no longer tell that a key belongs to neither operand, so `z.strictObject` composed with a record stops catching unknown keys entirely.

Not planning to add a separate spelling for the strip behavior for now.

Issue #5733, opened 2026-02-26
`preprocess` doesn't constrain return type of `fn` to the type of the output schema


ericnorris:
[`preprocess`](https://zod.dev/api?id=preprocess), roughly speaking, takes a `fn: (arg: B) => A` and a `schema: U`, but I would expect the return type of `fn` to have to match `z.infer<U>`. For example, in the following code, it is surprising that the former is not equivalent to the latter:

```typescript
// ❌ not caught at compile time, only at runtime
z.preprocess((input) => {
  return 4;
}, z.string());

// ✅ caught at compile time
z.pipe(
  z.transform((input) => 4),
  z.string(),
)
```

As a proof-of-concept, I implemented an alternative `preprocess` locally, which works as I would expect:

```typescript
// equivalent to preprocess and z.pipe(z.transform(...)
function preprocess<TContext, T extends z.ZodType<any, any>>(
  fn: (ctx: TContext) => z.infer<T>,
  schema: T
) {
  return z.transform(fn).pipe(schema);
}

// ✅ caught at compile time, just like `z.pipe(z.transform(...`
preprocess((input) => {
  return 4;
}, z.string())
```

Typescript playground link: https://www.typescriptlang.org/play/?noImplicitAny=false#code/JYWwDg9gTgLgBAbzgLzgXznAZlCI4DkyEAJgQNwBQlA9DXIDLkcAdhPAMYCGArgOYAW8ThzxhgAGwCmcGKEmVkAOjBRJKiO0kBnLQApdwZmG4wAlHAC8APkSVMqmNyjM4AFipoANCkVaYUQ15dU1MqWnpAUHI4Lj5BOGFo0QlpWRB5JTEwSV07H39OZi0saBB9Q2MzSxtXU09cpT8A5iDaylNqOjhJAEduYAA3TilmeBgIOBU1XE0deOYSH0zspXzC4qhS2rhFHfasbmZ2WQgXSfUZrQAeABUAYROYSQAPGG9rrpfJea0fAC1Sa4ATyylwKgO8YKsVhymCwzAAXHBdEcnoi7g9npVrD5DFhJFAblY6pgtOx+JIQJw0W1bPZJI5nHkoAUiiVdHDTMpgFldKTyZTQpQ0B1ItEeAIhCJwMkZHJKGdpto9AYjCZzNiELkHE4XO4hd4Gv5AsF2kA

[...]

PR #5908, opened 2026-04-30
Tighten preprocess return types

colinhacks:
Adds a type-level compatibility check for `z.preprocess()` callback returns so clear mismatches are caught before the value reaches the target schema, while preserving the existing `<A, U, B>` generic shape and validation-boundary patterns like `unknown`, `unknown[]`, object wrappers, `void`, and `z.NEVER`.

Covers the regression reported in #5733 with focused type assertions for sync, async, union, array, object, and default-schema cases.

Validated with `pnpm exec tsc -p packages/zod/tsconfig.test.json --noEmit --extendedDiagnostics` and `pnpm vitest run packages/zod/src/v4/classic/tests/preprocess.test.ts`.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -2642,8 +2642,45 @@ export function json(params?: string | core.$ZodCustomParams): ZodJSONSchema {
 
 // preprocess
 
+type _IsUnknown<T> = util.IsAny<T> extends true
+  ? false
+  : unknown extends T
+    ? [T] extends [unknown]
+      ? true
+      : false
+    : false;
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing this out. The callback of `z.preprocess()` is typed loosely on purpose — a preprocessor is expected to take arbitrary input, and normalizing into a literal or enum target can never satisfy an assignability check:

```ts
z.preprocess((v) => String(v).toUpperCase(), z.literal("A"));
z.preprocess((v) => String(v).toLowerCase(), z.enum(["a", "b"]));
```

Both of those are in our own test suite today and both fail under this constraint, and the same holds for any assignability-based rule — rejecting `() => 4` for `z.string()` necessarily rejects `String(v)` for `z.literal("A")`.

If you want the return type checked, `z.pipe(z.transform(fn), schema)` already does exactly that.

PR #5913, opened 2026-04-30
Add z.toZod helper

### LINKED ISSUE #372: Typecheck schemas against existing types
### TL;DR
I would love to have a way to ensure that a Zod schema matches an existing type definition using the normal Typescript type checker. Below is an example of the desired functionality using one possible implementation stolen from [this comment](https://github.com/colinhacks/zod/issues/53#issuecomment-655879215).

```typescript
type Dog = {
  name: string
  neutered: boolean
}

//Passing "Dog" as a generic type parameter tells Typescript what the schema should look like
const dogSchema = z.object<Dog>({
  name: z.string().min(3),
  neutered: z.string(), //Error: string can't be assigned to boolean
});
```

### Introduction

I originally raised this suggestion in #53, but decided it probably needs its own issue. See my original comments [here](https://github.com/colinhacks/zod/issues/53#issuecomment-737550649) and [here](https://github.com/colinhacks/zod/issues/53#issuecomment-738110750). Based on the reactions to these comments, there are at least a few other people thinking along the same lines as I am. I will restate my thoughts below.

I want to start by saying that Zod is a really, really cool project. It is the best runtime validation system for typescript by far. I hope the following critiques are constructive. 

### My runtime validation requirements

I started implementing Zod in my project, and I went into this implementation assuming that Zod would meet the following two requirements:

1.  Zod schemas would provide run time checks of my data types. **[true]**
2.  Zod schemas would conform to my existing types, so that it is impossible to change the type without also changing the associated schema (and vice versa) . **[only sort of true]**

In order to get the effect of my second requirement, I discovered that I need to replace my existing code, eg...

```typescript
export interface A {
   readonly ID: number;
   delayEnd: number;
   userID: number;
   reason: string;
   taskID: number;
   initiationDate: number;
   days?: number;
   userName?: string;
}
```

...with something like this...

```typescript
export const aSchema = z.object({
   ID: z.number(), //Note that I've lost the functionality of `readonly` in this conversion
   delayEnd: z.number(),
   userID: z.number(),
   reason: z.string(),
   taskID: z.number(),
   initiationDate: z.number(),
   days: z.number().optional(),
   userName: z.string().optional()
});

//"A" is generated from the schema
export type A = z.infer<typeof aSchema>;
```

This makes it so that if I change `aSchema`, `A` will automatically update to match, which gives me most of what I was looking for. But there are some serious problems with this solution.

### The Problems

The most obvious problem with the above code example is that it removes some really valuable typescript features: As just one example, the functionality of `readonly` has been lost in the conversion to `aSchema`. Perhaps it is possible to reintroduce that functionality with some fancy Typescript manipulation, but even if that is the case it is still not ideal.

Perhaps a more central problem, though, is that I need to strip out pretty much all of my current type definitions and replace them with Zod schemas. There are some tools out there that will do this work for you (issue #53 was originally and ultimately about building these sorts of tools), but the real issue for me isn't the work of refactoring: The real problem is that such a refactor puts Zod in charge of my type system, which is very undesirable. In my opinion, Typescript should be the master of typing, and Zod should be the master of validation. In the current system, Typescript is subordinated to Zod rather than the other way around.

To make sure my intent is clear, here are a few re-statements of this idea:
- I want to keep all my types as they are and create schemas that conform to them.
- I do not want to replace my existing type definitions with schemas; instead I want to create schemas that match my existing types. 
- I want to keep my type system in typescript and only use Zod to validate that objects fit my existing types. 

To put it a different way, Zod is advertised as "Typescript first", but right now it feels more like "Zod first with Typescript as a close second". I say that because, currently, if I want to maintain type consistency I have to write the Zod schemas first, then use them to generate types. To be truly "Typescript first", the schemas should conform to the types instead of the types being generated from the schemas.

### The `tozod` solution

A great idea that addresses these issues was introduced in [this comment](https://github.com/colinhacks/zod/issues/53#issuecomment-655879215), discussed in [this comment](https://github.com/colinhacks/zod/issues/53#issuecomment-655937376), then partially implemented in the  `tozod` library (see [this comment](https://github.com/colinhacks/zod/issues/53#issuecomment-664193624); the library can be found [here](https://www.npmjs.com/package/tozod)). The `tozod` utility allows me to write the following in place of the above code example:

```typescript
//My interface does not change
export interface A {
   readonly ID: number;
   delayEnd: number;
   userID: number;
   reason: string;
   taskID: number;
   initiationDate: number;
   days?: number;
   userName?: string;
}

//The use of toZod ensures that the schema matches the interface
export const aSchema: toZod<A> = z.object({
   ID: z.number(),
   delayEnd: z.number(),
   userID: z.number(),
   reason: z.string(),
   taskID: z.number(),
   initiationDate: z.number(),
   days: z.number().optional(),
   userName: z.string().optional()
});
```

This meets my requirements perfectly. It preserves my original types and has a schema that conforms to those types. It gives me the same strong typing as using `z.infer` would have, but it leaves Typescript in charge of defining my type system and still gives me all the benefits of runtime validation. It also preserves certain Typescript functionality that can't be modeled in Zod, like the `readonly` in `A`. I think this scenario gives the best of all worlds and is truly "Typescript-first". I realize that it is slightly more verbose than just having a schema, but I think the benefits are well worth it. It fits much better into the established Typescript paradigm. I could go on and on about why this is a better solution.

### The problem with `tozod`

There is just one problem with the `tozod` utility. I quickly discovered, and the author of `tozod` admits, that it is "pretty fragile" and can only handle the most basic types (see [this comment](https://github.com/colinhacks/zod/issues/53#issuecomment-664753908)). Even a simple union type will cause it to break. The `tozod` library was a great step in the right direction, but in its current state it is virtually unusable in real application environments.

### My suggestion

My suggestion is that we need something like `tozod`, preferably built into `Zod`, that allows schemas to be type checked against existing types in a real application environment. I don't know if this is feasible -- I'm not a Typescript expert, so it might not even be possible; but if it is possible, I think this change would be extremely beneficial.

### LINKED ISSUE #1917: Make Zod be truly TypeScript first
Right now Zod is useful when used from the start of a project, no doubt. 

However, I have some issues when it comes to the real-world usage of Zod when it comes to third-party libraries and existing codebases.

For example, it's currently impossible to use types from third-party packages in your schemas without first redefining those types in your own code as Zod schemas. This makes for a horrible developer experience in the first instance as now I have to inspect the types in the third-party package in much more detail than I otherwise would and I have to write a bunch of additional boilerplate.

However, it gets worse if/when that third-party package changes the type definition. Maybe they add new properties. Maybe they rename or remove something. Either way, even if I wasn't using/relying on those properties I now have to change my code to match.

My ideal developer experience here would be something like this:

```ts
import { z } from 'zod'
import { SomeType } from 'some-third-party-package'

const MySchema = z.object({
  myProperty: z.string(),
  otherProperty: z.deduce<typeof SomeType>() // Zod would inspect the type and deduce a schema or object shape automatically
})

export type MyType = z.infer<typeof MySchema>
```

There is also the issue of retrofitting Zod into existing TypeScript projects. Typically dev teams will find it difficult to get buy-in for additional work that the people holding the purse strings will perceive as "rework" that's not going to produce any tangible benefit to the value of the software or the revenue the company makes. Basically, why would they pay for that when they could be paying for new features to be added that they can charge their customers more money for?

From the suggestion above, the same idea could then also be used to give true TypeScript first capabilities to Zod **and** allow a much easier uptake on older projects with little to no additional developer overhead:

```ts
import { z } from 'zod' // added in retrofit

// Existing type in the project
export type User {
  name: string
  age: number
}

// To retrofit Zod into an older project the developer could use the `deduce<T>()` function 
// to produce schemas directly from their existing types
export const UserSchema = z.deduce<typeof User>()
```

Now the developer can use the Schema to achieve the runtime type protection they want, and the bosses are happy because the developer didn't take time to rewrite old code rather than build new features.

### LINKED ISSUE #2084: Feature Request : A "create a schema according to this type" helper
Hello there,

What do you think about this discussion?

If this is something you are interested in, what's the direction to add this feature?

### Discussed in https://github.com/colinhacks/zod/discussions/1928

<sup>Originally posted by **rphlmr** January 24, 2023</sup>
Hello,

What about adding a helper that takes a type and "forces" you to create a schema according to this type?

Can be useful, for example, to create a schema derived from a Prisma Model.

```ts
type Implements<Model> = {
  [key in keyof Model]-?: undefined extends Model[key]
    ? null extends Model[key]
      ? z.ZodNullableType<z.ZodOptionalType<z.ZodType<Model[key]>>>
      : z.ZodOptionalType<z.ZodType<Model[key]>>
    : null extends Model[key]
    ? z.ZodNullableType<z.ZodType<Model[key]>>
    : z.ZodType<Model[key]>;
};

export function implement<Model = never>() {
  return {
    with: <
      Schema extends Implements<Model> & {
        [unknownKey in Exclude<keyof Schema, keyof Model>]: never;
      }
    >(
      schema: Schema
    ) => z.object(schema),
  };
}

// usage
export type UserModel = {
  id: string
  email: string | null
  name: string
  firstName: string
  createdAt: Date
}

export const UserCreateSchema = implement<Omit<UserModel, "createdAt" | "id">>().with({
  email: z.string().email().nullable(),
  name: z.string(),
  firstName: z.string(),
});

export type UserCreatePayload = z.infer<typeof UserCreateSchema>
```

This is maybe not perfect, but I use this on my projects and I'm pretty happy with It.

Force to put `nullable` `optional` or `nullish` when Model requires it.
Unknown properties are forbidden and cause a TS error.

Discussed here : https://github.com/colinhacks/zod/issues/372#issuecomment-830094773</div>

### LINKED ISSUE #2807: How to use Zod schema with existing TypeScript interface?
### Discussed in https://github.com/colinhacks/zod/discussions/2796

<sup>Originally posted by **adamerose** September 24, 2023</sup>
I see Zod described as a "TypeScript-first" library but am struggling to figure out how to actually integrate it with my TypeScript definitions. Is there a way to pass a TypeScript interface to my Zod schema as a generic and have my linter tell me if they're not compatible? 

[This](https://github.com/jquense/yup#ensuring-a-schema-matches-an-existing-type) is how it looks in Yup:
```typescript
import { object, number, string, ObjectSchema } from 'yup';

interface Person {
  name: string;
  age?: number;
  sex: 'male' | 'female' | 'other' | null;
}

// will raise a compile-time type error if the schema does not produce a valid Person
const schema: ObjectSchema<Person> = object({
  name: string().defined(),
  age: number().optional(),
  sex: string<'male' | 'female' | 'other'>().nullable().defined(),
});

// ❌ errors:
// "Type 'number | undefined' is not assignable to type 'string'."
const badSchema: ObjectSchema<Person> = object({
  name: number(),
});
```

I found discussion [here](https://github.com/colinhacks/zod/issues/53) and [here](https://github.com/colinhacks/zod/issues/2086) but didn't see any solution I could understand, or relied on codegen and external libraries.

I think something like this might work? But it's very boilerplatey and I don't see anything like this in the docs:
```typescript
interface Person {
  name: string;
  age?: number;
  sex: "male" | "female" | "other" | null;
}

const zodSchema = z.object({
  name: z.string(),
  age: z.number().optional(),
  sex: z.enum(["male", "female", "other"]).nullable(),
});

type InferredPerson = z.infer<typeof zodSchema>;
function assertType<T>(_value: T) {}
assertType<Person>({} as InferredPerson);
```</div>

### LINKED ISSUE #5418: “Proposal: Typed schema helper (makeSchema) for safe shape → type mapping”
## Summary

In many real-world projects, schemas are generated from existing TypeScript types.  
Today, Zod allows schema → type inference, but not the opposite direction (type → schema shape typing).

I'm proposing a small helper that improves safety when building a schema based on a TypeScript type via generics, ensuring that the shape fully aligns with the TypeScript type.

## Problem

When defining a schema based on a given TypeScript type:

```ts
type Company = {
  id: string;
  name: string;
  webAddress?: string | null;
};
```

We typically write:

```ts
const CompanySchema = z.object({
  id: z.string(),
  name: z.string(),
  webAddress: z.string().url().nullish(),
});
```

This works, but **type → schema mismatches are not checked at compile time**:

- missing keys  
- adding extra keys  
- wrong Zod types (e.g., `z.number()` for a `string`)  

These mistakes become runtime-only errors.

## Proposed Helper

A small factory function that enforces compile-time parity between a TypeScript type `T` and the provided Zod object shape.

### Implementation

```ts
import { z } from "zod";

export function makeSchema<T extends Record<string, any>>() {
  return <S extends { [K in keyof T]-?: z.ZodType<T[K]> }>(shape: S) => {
    return z.object(shape).strict() as z.ZodObject<S, "strict", z.ZodTypeAny, T>;
  };
}
```

### Usage

```ts
type Company = {
  id: string;
  name: string;
  webAddress?: string | null;
};

const CompanySchema = makeSchema<Company>()({
  id: z.string(),
  name: z.string(),
  webAddress: z.string().url().nullish(),
});
```

## Benefits

✅ Compile-time validation  
✅ Ensures full type–schema parity  
✅ Zero runtime overhead  
✅ Purely additive; does not modify existing API  
✅ Already extremely helpful in form-heavy or DTO-heavy projects  

Would the maintainers be open to a PR with implementation + tests + documentation?

colinhacks:
Adds `z.toZod<T>()`, a no-op helper that checks a hand-written schema against an existing TypeScript type and returns it unchanged, so `.shape` and chaining keep working. Exported from Zod, Zod Mini, and Zod Core, and documented under "Matching an existing type" in `basics.mdx`.

The check is exact type equality rather than assignability, which is the point — `satisfies z.ZodType<T>`, the workaround people reach for today, lets extra keys, omitted optional keys, and a bare `z.any()` through.

Exact equality is the only mode. A loose bidirectional-assignability mode was proposed so that an intersection target would accept `.safeExtend()`, but assignability structurally cannot see either hole above: `any` is assignable in both directions by construction, and `readonly` is not part of assignability. Both are `@ts-expect-error` assertions in this PR's own tests, so a loose mode would silently break its own guarantees. The cost is an asymmetry: an intersection target matches `.and()` and a flat target matches `.safeExtend()`. Flattening the target switches which spelling matches rather than accepting both, and every case is pinned by a test so it reads as deliberate.

The helper is curried because TypeScript has no partial type-argument inference — supplying `T` explicitly stops `S` inferring from the argument, so the target type and the schema cannot be passed in one call.

Closes #372. Closes #2084. Closes #2807. Closes #5418. Not #1917 — that one asks for generating a schema from a type, which a compile-time checker can't do.


Anoesj:
Cool, thanks for considering adding this to Zod itself!

In my own version (based on yours), I added the option of setting `Strict` to `false`, making the the helper check bidirectional assignability instead of TS type tree equivalence. In a few exotic cases where complex types are assembled differently than how the Zod schema is assembled, this saved the day. It also makes it a bit easier to discover whether you've already achieved equality, but accidentally used `.and()` instead of `.safeExtend()` or vice versa (examples further down below).

So my own full version looks like this:
```ts
// "AssertEqual" renamed
type IsStrictEqual<T, U> = (
  <V>() => V extends T ? 1 : 2
) extends <V>() => V extends U ? 1 : 2
  ? true
  : false;

type IsLooseEqual<T, U> = [T] extends [U]
  ? [U] extends [T]
    ? true
    : false
  : false;

function createSchema<
  TargetType,
  Strict extends boolean = true,
> () {
  return <S extends z.ZodType<TargetType, unknown>>(
    schema: (
      Strict extends true
        ? IsStrictEqual<S['_output'], TargetType>
        : IsLooseEqual<S['_output'], TargetType>
    ) extends true
      ? S
      : S & {
        'Error: Model and validation types do not match': {
          expected: TargetType;
          received: S['_output'];
        };
      },
  ): S => {
    return schema;
  };
}
```

<details>

<summary><strong>Examples of the behavioral differences here (expand to view)</strong></summary>

```ts
type XIntersection = { a: number; } & { b: string; };
const xIntersection: XIntersection = { a: 1, b: 'two' };
// Strict equal -> Perfect (trees match)
const xIntersectionSchema1 = createSchema<XIntersection>()(
  z.object({ a: z.number() }).and(z.object({ b: z.string() })),
);
// Non-strict equal -> Good too (bidirectional assignability holds and trees match)
const xIntersectionSchema2 = createSchema<XIntersection, false>()(
  z.object({ a: z.number() }).and(z.object({ b: z.string() })),
);
// Strict equal -> Error (trees don't match)
const xIntersectionSchema3 = createSchema<XIntersection>()(
  // @ts-expect-error
  z.object({ a: z.number() }).safeExtend({ b: z.string() }),
);
// Non-strict equal -> Okay (bidirectional assignability holds, even though trees don't match)
const xIntersectionSchema4 = createSchema<XIntersection, false>()(
  z.object({ a: z.number() }).safeExtend({ b: z.string() }),
);

type XMerged = { a: number; b: string; }; // or: Prettify<{ a: number; } & { b: string; }>
const xMerged: XMerged = { a: 1, b: 'two' };
// Strict equal -> Error (trees don't match)
const xMergedSchema1 = createSchema<XMerged>()(
  // @ts-expect-error
  z.object({ a: z.number() }).and(z.object({ b: z.string() })),
);
// Non-strict equal -> Okay (bidirectional assignability holds, even though trees don't match)
const xMergedSchema2 = createSchema<XMerged, false>()(
  z.object({ a: z.number() }).and(z.object({ b: z.string() })),
);
// Strict equal -> Perfect (trees match)
const xMergedSchema3 = createSchema<XMerged>()(
  z.object({ a: z.number() }).safeExtend({ b: z.string() }),
);
// Non-strict equal -> Good too (bidirectional assignability holds and trees match)
const xMergedSchema4 = createSchema<XMerged, false>()(
  z.object({ a: z.number() }).safeExtend({ b: z.string() }),
);
```

</details>

karlhorky:
@colinhacks Just revisiting this after [the Twitter thread](https://x.com/karlhorky/status/2086759685294571972) with @arackaf.

Is there much left to get this one over the finish line? Would be great to try this out in Zod!

capaj:
@colinhacks

arackaf:
So like a Zod-specific version of the satisfies keyword? Very cool!

karlhorky:
@arackaf if I remember correctly, `satisfies` is not a robust solution for exact object checks:

- https://github.com/colinhacks/zod/issues/372#issuecomment-1435670103

arackaf:
Hey love seeing this merge. I can't wait to remove some satisfies and try this out - any eta on a new version on npm? :D

manuel-mauky:
Did you consider the "enum" handling mentioned [here](https://github.com/colinhacks/zod/pull/5913#pullrequestreview-4974303752)?

I've just tried with version "4.5.0-canary.20260820T155656" and wasn't able to use enums.

```ts
enum SkillLevel {
  NOOB= "noob",
  PRO= "pro"
}

type Player = {
  username: string
  xp: number
  skillLevel: SkillLevel
}

const Player = z.toZod<Player>()(
  z.object({
    username: z.string(),
    xp: z.number(),
    skillLevel: z.enum(SkillLevel)
  }),
)
```

This doesn't work. It raises this compile error:

```
Types of property skillLevel are incompatible.
Type ZodEnum<typeof SkillLevel> is not assignable to type ToZodKeyMismatch<SkillLevel, SkillLevel>
Property "types do not match" is missing in type ZodEnum<typeof SkillLevel> but required in type
{
  "types do not match": {
    expected: SkillLevel;
    received: SkillLevel;
  };
}
```

karlhorky:
@arackaf it seems that @colinhacks released this in [`zod@4.5.0`](https://github.com/colinhacks/zod/releases/tag/v4.5.0), although I did not see any Twitter posts about it yet...

> ## `z.toZod<T>()`
>
> A utility to define a Zod schema that agrees exactly with a static type, often one that is handwritten or externally defined. ([#5913](https://github.com/colinhacks/zod/pull/5913))

Source: https://github.com/colinhacks/zod/releases/tag/v4.5.0#ztozodt

arackaf:
I see it now - thanks!

arackaf:
If anyone winds up here looking for the api, it's slightly different than you may have surmised from this thread. 

This is what you want:

```ts
type Person = {
  name: string;
  age: number;
};

export const PersonSchema = z.object({
  name: z.string(),
  age: z.number(),
});

const PersonValidator = z.toZod<Person>()(PersonSchema);
```

and this 

```ts
type Person = {
  name: string;
  age: number;
};

export const PersonSchema = z.object({
  nameXXXXXXXXX: z.string(),
  age: z.number(),
});

const PersonValidator = z.toZod<Person>()(PersonSchema);
```

correctly produces a TS error of 

> Argument of type 'ZodObject<{ nameXXXXXXXXX: ZodString; age: ZodNumber; }, $strip>' is not assignable to parameter of type '$ZodObject<{ nameXXXXXXXXX: ToZodKeyMismatch<never, string>; age: ZodNumber; name: ToZodKeyMismatch<string, never>; }, $strip>'.
  Types of property '_zod' are incompatible.
    Type '$ZodObjectInternals<{ nameXXXXXXXXX: ZodString; age: ZodNumber; }, $strip>' is not assignable to type '$ZodObjectInternals<{ nameXXXXXXXXX: ToZodKeyMismatch<never, string>; age: ZodNumber; name: ToZodKeyMismatch<string, never>; }, $strip>'.
      Property 'name' is missing in type '{ nameXXXXXXXXX: ZodString; age: ZodNumber; }' but required in type '{ nameXXXXXXXXX: ToZodKeyMismatch<never, string>; age: ZodNumber; name: ToZodKeyMismatch<string, never>; }'.

Great work - thank you!!! 🚀

Issue #5694, opened 2026-02-06
[V4]: Cannot pipe to less restrictive schema


sybereal:
It doesn't seem possible to use `.pipe()` with a target schema that is less restrictive on the type level, yet fully compatible semantically.
This is somewhat related to #4778, but can be triggered without any transforms whatsoever, using only runtime-idempotent schemas.

The simplest reproduction I could come up with:[^tsplay]

```typescript
import {z} from "zod";

const maybeNumber = z.number().optional()
const definitelyNumber = z.number().pipe(maybeNumber)
```

This results in the following error:

```
Argument of type 'ZodOptional<ZodNumber>' is not assignable to parameter of type '$ZodType<any, number, $ZodTypeInternals<any, number>>'.
  The types of '_zod.input' are incompatible between these types.
    Type 'number | undefined' is not assignable to type 'number'.
      Type 'undefined' is not assignable to type 'number'.
```

[^tsplay]: https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbwF4F84DMoRHAREiAE1wG4AoMgYwgDsBneEAQwE8AjAUwDkBXETqHAC8cJADoafAQAoAlGIhgYwWkwA2cqrQZxCHdMBrAYHNS178OgkeMmWocsWGBgO05u25SrsoA

Our use case for this is as follows: We have two schemas. One validates and processes data from an external source and brings it into a format suitable for persistence. Another schema is used to type check the data again once loaded from the untyped persistence layer. Because of the need to accommodate historical data, the latter schema has a lot of fields marked as optional that are mandatory for new data and therefore not optional in the former schema.

To ensure that the data from the external source was compatible with what would eventually be read from the persistence layer and we wouldn't notice bad data only after reading it back, we piped these schemas together on Zod v3, as illustrated by the code snippet. However, on Zod v4, attempting to do the same results in the quoted error message.

As a personal opinion, the addition of input type checking to `.pipe()` in v4 is a very welcome change. It prevents a lot of mistakes that would completely invalid or silently broken schemas in v3. However, the strictness of v4 seems a bit counterintuitive to Zod's modular and reusable approach to schema definition, as it often requires redefining schemas with slight differences to exactly match the required types. Those multiple definitions then need to be kept in sync, increasing maintenance effort.

ryanagillie:
Having the same issue with our upgrade to v4

[...]

Issue #4778, opened 2025-06-27
`.pipe` seems arbitriarily restricted in V4


ConnorSinnott:
Hello! I'm in the process of migrating to v4 and the largest headache so far has been the changes to `.pipe`. From what I'm gathering from errors, whatever schema is being piped to must have the input match the output of the source schema. 

For my specific example, I had a JSON parsing schema which accepts a string and parses the result as JSON returning an object. I'd then pipe that into whatever schema I needed that object to match. This doesn't appear to be supported anymore since the input requirement for pipe can not be determined.

```typescript
import { z } from 'zod/v4';

export const someStringSchema = z.string();

const someObjectSchema: z.ZodSchema = z.object({ // Arbitrarily erasing the inferred type to be just ZodSchema
  name: z.string(),
})

someStringSchema.pipe(someObjectSchema).parse({}) 
// Types of property _zod are incompatible.
// Type $ZodTypeInternals<unknown, unknown> is not assignable to type $ZodTypeInternals<any, string>
// Type unknown is not assignable to type string
```

I can apparently calm the linter by defining `z.ZodSchema<any, any>`

```typescript
import { z } from 'zod/v4';

export const someStringSchema = z.string();

const someObjectSchema: z.ZodSchema<any, any> = z.object({
  name: z.string(),
})

someStringSchema.pipe(someObjectSchema).parse({}) // No issues
```

I guess I can see why perhaps a developer would want to be warned that they are piping an incompatible type, but this seems to hinder abstract Zod work. 

Any suggestions?

ConnorSinnott:
@dosu maybe I'm missing that the answer truly is to use `any`s. What is the recommended way to solve this pattern? 

```typescript
export const stringToJSONSchema = z.string().transform((str, ctx): Json => {
  try {
    return JSON.parse(str) as Json;
  } catch {
    ctx.addIssue({ code: 'custom', message: 'Invalid JSON' });
    return z.NEVER;
  }
});

export interface IEventHandler<Params> {
  get eventSchema(): z.ZodSchema<Params>;

  handleEvent(params: Params): Promise<...>;
}

type TestObj = {
  name: string;
};

class Test implements IEventHandler<TestObj> {
  get eventSchema(): z.ZodSchema<TestObj> {
    return z.object({
      name: z.string(),
    });
  }
}

const handler = new Test();

const data = stringToJSONSchema
  .pipe(handler.eventSchema) // Types of property _zod are incompatible.
[...]

PR #5914, opened 2026-04-30
Fix classic pipe compatibility for branded outputs

### LINKED ISSUE #5648: Branding gets dropped in zod v4 from transforms and pipes
```ts
import z from 'zod';

const zodBrand = z.string().brand('myBrand');

const inputSchema = z.object({
  a: z.number(),
  c: zodBrand,
});

// Exact same schema
const validateSchema = z.object({
  a: z.number(),
  c: zodBrand,
});

// Lint errors
const testSchemaPipeline = inputSchema.transform((input) => input).pipe(validateSchema);
const testSchemaPipeline2 = inputSchema.pipe(validateSchema);
const testSchemaPipeline3 = inputSchema.pipe(inputSchema);

// Recommended fix (?)
const testSchemaPipeline4 = inputSchema.transform((input) => input as unknown).pipe(validateSchema);
```

With the zod 4 changes in pipe input 'strictness', we are getting lint errors because the branding isn't preserved:
```
Argument of type 'ZodObject<{ a: ZodNumber; c: $ZodBranded<ZodString, "myBrand", "out">; }, $strip>' is not assignable to parameter of type '$ZodType<any, { a: number; c: string & $brand<"myBrand">; }, $ZodTypeInternals<any, { a: number; c: string & $brand<"myBrand">; }>>'.
  The types of '_zod.input.c' are incompatible between these types.
    Type 'string' is not assignable to type 'string & $brand<"myBrand">'.
      Type 'string' is not assignable to type '$brand<"myBrand">'.
```

colinhacks:
Fixes #5648 and #5694. #4778 covers the same ground.

The strict `.pipe()` constraint is written backwards: it requires the target's input to be assignable to the source's output, while the runtime feeds the source's output into the target's parse. So correct pipes get rejected — a branded output into a schema accepting the corresponding unbranded input, or a stricter source into a looser target like `z.number().pipe(z.number().optional())`.

This keeps the strict overload first, so contextual typing for transforms is unchanged, and adds a fallback that checks the relation in the direction the runtime actually uses. The fallback declares a required rest tuple when the two are incompatible, which arity-filters it out of the candidate set — an incompatible target still reports the original `TS2345` rather than degrading to "No overload matches this call". An overload set only ever accepts the union of what its members accept, so this loosens and swallows nothing.

Applied to all three spellings, not just the method. `z.pipe(a, b)` is documented and used throughout the tests, so fixing only `.pipe()` would leave `z.number().pipe(x)` compiling while `z.pipe(z.number(), x)` errored. `zm.pipe()` gets the same treatment; splitting it into overloads moves `@__NO_SIDE_EFFECTS__` onto the implementation signature, since overload signatures are erased and the pragma has to survive into the emitted JS.

Types-only — the transpiled output of both changed files is byte-identical, so runtime, memory and bundle size are unaffected. `tsc --extendedDiagnostics` over `packages/zod/tsconfig.test.json` moves instantiations 1,865,461 → 1,867,725 (+0.12%) and types 488,057 → 488,541 (+0.10%).

Not addressed here: the same reversed constraint also admits false negatives, so `z.number().optional().pipe(z.number())` type-checks and throws at runtime. Fixing that means replacing the strict overload rather than adding to it, which is breaking and deserves its own issue.


dokson:
Heads-up: this PR also resolves #5694 (and effectively #4778). The new overload's `core.output<this> extends core.input<T>` check covers both the "branded output → unbranded input" case in your tests and the "stricter source → looser target" case (`z.number().pipe(z.number().optional())`, `.transform().pipe(nullable_schema)`) reported there.

Verified locally: both #5694 reproductions fail on `main` with the documented `TypeCheckError`, and pass on this branch with no extra changes. If it'd help reviewer confidence to lock in those cases too, here's a drop-in test file you can include — feel free to ignore if you'd rather keep the PR scope narrow.

```ts
// packages/zod/src/v4/classic/tests/pipe-compat.test.ts (or merge into pipe.test.ts)
import { expectTypeOf, test } from "vitest";
import * as z from "zod/v4";

test("pipe stricter source into looser target (issue #5694)", () => {
  const maybeNumber = z.number().optional();
  const out = z.number().pipe(maybeNumber).parse(42);
  expectTypeOf(out).toEqualTypeOf<number | undefined>();
});

test("pipe transform output into nullable target (issue #5694)", () => {
  const backEnd = z.object({ field: z.number().min(1).max(100).nullable() });
  backEnd.extend({
    field: z.string().nonempty().transform(Number).pipe(backEnd.shape.field),
  });
});
```

Either way, two more issues fall behind this PR — worth flagging in the description if you'd like.

viceice:
any reason why this is still not merged?

colinhacks:
Holding this one. It breaks recursive schemas on TypeScript 7 — a schema whose getter pipes to itself now errors with `'pipe' implicitly has return type 'any'`. TypeScript 5.5 and 6 both pass, so it only surfaced when the matrix picked up 7.

That's the approach rather than the diff: every formulation of the second overload I tried reproduces it, including one carrying no conditional type at all, and the only shape that keeps recursion working can't express the compatibility check. Leaving #5648 and #5694 open.

wahajahmed010:
## Code Review

Reviewed the diff against the linked issues (`#5648`, `#5694`, `#4778`). The structural fix is correct: the strict overload stays first for transforms' contextual typing, and the new fallback overload uses `core.output<A> extends core.input<B>` to relax the constraint in the runtime-feeding direction. The branded/unbranded and stricter-source/looser-target cases from both issues are now accepted, and the `@ts-expect-error` tests in `pipe.test.ts` and `mini/tests/index.test.ts` confirm the new overload still rejects truly incompatible targets.

A few observations worth weighing before merge:

1. **TypeScript 7 regression on recursive schemas (blocker per your own comment).** Your Aug 24 message flags this — a schema whose getter pipes to itself now errors with `'pipe' implicitly has return type 'any'` on TS 7 (passes on 5.5 and 6). Before this lands, the CI matrix needs a TS 7 entry that exercises at least one self-referential schema using `.pipe`, and the failure mode needs a reproducer in the test suite. Otherwise the matrix you cited will pass green once a TS 7 bump is queued but the project itself breaks for users. The matrix that caught this needs to be added back before the hold is released, not after.

2. **`as core.$ZodType` cast in `pipe(target)` (classic line 276).** Reasonable given the overloads, but worth a comment at the call site explaining why the cast is safe — future maintainers won't have the issue thread to anchor the reasoning.

3. **`@__NO_SIDE_EFFECTS__` moved below the overloads in mini/schemas.ts.** Correct for tree-shaking — the marker now sits next to the implementation signature rather than the first declaration. No issue, just confirming the placement change matches the intent.

4. **Branding test asymmetry.** `pipe.test.ts` covers `branded -> branded` and `inputSchema -> validateSchema` where both schemas are identical. The realistic case from `#5648` is `inputSchema.transform(...).pipe(validateSchema)` — the `transform` step is what historically dropped the brand. The new test "pipe transform output into nullable target" covers `transform + pipe` with a nullable target, but not with a brand on the target. If the goal is to lock in the brand fix from `#5648`, a dedicated `transform(brand).pipe(unbranded)` round-trip assertion would make the regression target explicit.

5. **No new test for the `z.pipe` function-form brand case in classic.** The mini suite has `z.pipe(branded, branded)`, but classic only tests the method form. Worth a one-liner for parity since both surfaces are public.

Overall direction is right and the diff is small and targeted. The TS 7 blocker is the only thing standing between this and merge — everything else is polish.

Issue #5240, opened 2025-09-15
Support for Schema Instance Parameter in Zod Global customError and localeError?


nuintun:
In Zod v4, the global configuration method z.config allows customizing error messages via the customError and localeError function, which currently only receives the issue parameter. The requirement is for customError localeError to support a second parameter, which provides the actual schema instance responsible for the validation error. This enables direct access to the meta information (such as title, label, etc.) attached to that schema instance and allows developers to implement consistent localized error handling logic.

Key Points:

customError(issue, schemaInstance): the second parameter is the schema instance where the error occurred, localeError accepts the same parameters.
Works for both base and chained schemas (e.g., min/max/refine), always providing the schema with its associated meta.
Compatible with localization and meta-driven error formatting, enabling easier access to meta fields like title or label for custom error messages.

```ts
z.config({
  localeError,
  customError(issue, schemaInstance) {
    const title = schemaInstance.meta()?.title;
    const message = localeError(issue, schemaInstance);
    // ...
    
    return title ? `${title} - ${message}` : message;
  }
})
```

Expected Benefits:

Simplifies multilingual and meta-driven error handling.
Resolves the problem of losing meta information in chained validations.
Makes Zod’s error handling API more consistent and easier to use for localization and customization scenarios.

Using `issue.inst` is not safe, as it may not necessarily point to `schemaInstance`.

tmcw:
Ran into the same issue today - we currently ship Zod errors to users, but getting an error about a "string" being too long when it's the "title" that's too long is confusing to them. There aren't very good alternative routes to making this happen:

- Adding custom formatters to each `.min()` and `.max()` check is repetitive
- It would be possible to look through `z.globalRegistry` and create a map like `Map<ZodCheck, schema>` and be able to lookup the schema from the check and then the meta from the schema, but `z.globalRegistry` is a WeakMap so that isn't possible.

cgatian:
I'm attempting to convert from yup to zod, and really struggling with localization and custom errors. I simply want to have a localized error message with a label in it. I'm probably just missing something because feels like a common scenario.

For example:

```ts
const schema = zod.string()
const output = schema.safeParse(undefined);
// Invalid input: expected string, received undefined
```

I get that the error message is localized, but since this is in a form, the error would be better as `Name is a required field`. 
The workaround I have is to attach meta to the schema:

```ts
const schema = zod.string().meta({label: 'Name'});
const output = schema.safeParse(undefined);
```

```ts
[...]

PR #5915, opened 2026-04-30
Add schema metadata to raw issues

colinhacks:
## Summary
Expose the owning schema as the second argument to error maps so metadata-driven messages can work for checks and refinements.

This keeps `inst` as the immediate issue origin, while passing `schema` directly to `customError`, `localeError`, contextual error maps, and schema/check-bound error maps. The schema reference remains off finalized public issues.

Related: #5240

## Validation
Focused error test: `pnpm vitest run packages/zod/src/v4/classic/tests/error.test.ts` passed with 56 tests and no type errors.

Full suite via pre-push: `pnpm vitest run` passed with 335 files, 3,789 tests, and no type errors.

Runtime spot check for valid checked string parses showed no meaningful happy-path difference: main median 174.37ms vs branch median 174.76ms for 2M `safeParse` calls. The no-error sync path still exits before schema stamping runs.


wahajahmed010 (commented):
Solid change. A few things to consider before merge:

1. The new `for` loop mutating `(payload.issues[i]! as any).schema ??= inst` runs on every failed check. The current placement (after `nextLen > currLen` check) means it only executes on the error path, which is fine — but worth confirming with a quick benchmark that this doesnt regress the failing-parse path. The new `AGENTS.md` guideline you added ("allocation-free parsing fast paths") applies to the *success* path, which is correctly preserved.

2. The `as any` casts for the `schema` property are unavoidable here since `$ZodRawIssue` doesnt include it by design, but consider a small typed helper like `attachSchema(issue, inst)` to centralize the cast and make the intent explicit. It also makes it trivial to add a debug assertion or rename later.

3. Schema propagation through `def.error?.(iss, schema)` and `ctx?.error?.(iss, schema)` — these are user-facing error functions defined in `$ZodErrorMap`. The breaking-change surface is: any third-party type augmenting `$ZodErrorMap` now sees a second parameter. Since the parameter is optional and unused, this is backwards compatible. Worth a line in the changelog though.

4. The docs example in `error-customization.mdx` is good but uses `.meta({ title: "Name" })` on a `z.string()`. The error path fires for both `invalid_type` and `too_small` — consider showing the `refine` case too so users see the most useful application (which is exactly what the implementation targets).

5. The test correctly asserts that `schema` is *not* present on the finalized public issue. Good. One more thing worth testing: when an inner schema (e.g. `z.string().min(5).refine(...)`) fails, does `schema` point to the inner check schema or the outer `z.string()`? The current code uses `inst` which I believe is the immediate owning schema at that point — worth a clarifying test or comment.

6. The `finalizeIssue` change reads the schema once at the top (`const schema = ((iss as any).schema ?? iss.inst)`) and reuses it across all four error-map calls. Nice — avoids inconsistent schema context if some user mutates between calls.

colinhacks:
Superseded by #6420 (79cfedea), which landed this as an optional `schema` on the raw issue rather than a second argument to the error map.

Same coverage in practice — it resolves before the error-map precedence chain, so `customError`, `localeError`, per-parse, schema-bound and check-bound maps all see it, and it's stripped in `finalizeIssue` so it never reaches `$ZodIssue`. #6421 pins each of those.

Closing this out.

PR #5916, opened 2026-04-30
fix(v4): avoid recursive core pipe metadata types

### LINKED ISSUE #4611: $ZodPipe: Type instantiation is excessively deep
I’m using `3.25.50` and get a `Type instantiation is excessively deep and possibly infinite.` error when defining a recursive type using `$ZodPipe`.

```ts
type Foo = $ZodPipe<Bar, $ZodTransform>; // Type instantiation is excessively deep and possibly infinite.
type Bar = $ZodString | Foo;
```

with other zod types this works fine, e.g. with `$ZodArray`

```ts
type Foo2 = $ZodArray<Bar2>; // no error
type Bar2 = $ZodString | Foo2;
```

colinhacks:
## Summary
- Fixes #4611 by broadening core pipe/codec metadata internals so recursive `$ZodPipe` aliases do not force TypeScript into infinite instantiation.
- Preserves precise bubbled metadata on concrete classic/mini `ZodPipe` and `ZodCodec` types, with coverage for nested pipe optionality.

## Test plan
- `pnpm test`

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -2295,9 +2295,16 @@ export interface ZodPipe<A extends core.SomeType = core.$ZodType, B extends core
   extends _ZodType<core.$ZodPipeInternals<A, B>>,
     core.$ZodPipe<A, B> {
   "~standard": ZodStandardSchemaWithJSON<this>;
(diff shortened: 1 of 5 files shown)


colinhacks:
Closing this out. This fixes the narrow repro, but the approach weakens @zod/core as the generic substrate for classic and mini, so it is not the right direction.

PR #5919, opened 2026-05-01
fix(test-setup): resolve `recheck` backend path cross-platform

### LINKED ISSUE #5933: redos checker test fails on Windows: recheck-jar path uses forward-slash regex on backslash paths
## Summary

`src/v4/classic/tests/datetime.test.ts > redos checker` fails on every Windows contributor's machine with:

```
Error: Invalid or corrupt jarfile C:\Users\<…>\node_modules\recheck-jar\package.json
```

Linux CI is green; the bug is purely platform-specific path handling in `recheck`'s self-resolution logic.

## Reproduction

- Windows 10 / 11
- Any Node version
- Clean clone of zod, `pnpm install`, then:

```sh
pnpm vitest run src/v4/classic/tests/datetime.test.ts -t "redos checker"
```

Result: test fails (or times out) with the "Invalid or corrupt jarfile" error above.

## Root cause

`recheck` (the library used to assert no ReDoS-vulnerable patterns) resolves both its native binary and its bundled jar using:

```ts
require.resolve(pkg).replace(/\/package\.json$/, "/<bin>")
```

The forward-slash regex silently no-ops on Windows, where `require.resolve` returns a backslash path. `recheck` then spawns Java/exe with `package.json` itself as the executable argument — hence the corrupt-jar error.

The test itself is correct; the failure is in the optional-dependency resolution shape inside `recheck`.

## Proposed fix

[#5919](https://github.com/colinhacks/zod/pull/5919) registers a small `scripts/recheck-resolve.ts` ahead of `fail-on-console` in the vitest configs. It uses `path.join` (separator-correct on every platform) to resolve:

- `recheck-{platform}-{arch}/recheck[.exe]` → exported as `RECHECK_BIN` (the native-backend override)
- `recheck-jar/recheck.jar` → exported as `RECHECK_JAR` (JVM fallback)

`recheck` short-circuits its own (buggy) resolution when either env var is set, so this is a transparent override. Pure test-infra change — no production code touched. Skips on platforms without an installed `recheck-{platform}-{arch}` or `recheck-jar` package (e.g. s390x, riscv64, FreeBSD).

## Related (different cause, same test)

These touch the same `redos checker` test but at different layers, so this issue is not a duplicate:

- #5365 — Node 22 regex engine throws a ReDoS issue inside `recheck` (closed not-planned).
- #5743 — Node 24 V8 regex perf makes the test exceed 5s timeout (closed via timeout bump).

This issue is the third axis: Windows-specific path resolution inside `recheck`'s optional-dep loader.

dokson:
Closes #5933.

## Summary

Fixes the `redos checker` test (`packages/zod/src/v4/classic/tests/datetime.test.ts`) on Windows, where it has been silently broken for every Windows contributor since `recheck` was wired up. The fix is purely test-infrastructure — no production code is touched. As a bonus, the suite now uses `recheck`'s native binary instead of the JVM, so the redos check runs ~5× faster on every platform and is no longer flaky under parallel-file load.

## Problem

`recheck` (the ReDoS-checking library called from the redos test) resolves both its native binary and its bundled jar with a single line in `node_modules/recheck/lib/main.js`:

```js
require.resolve(pkg).replace(/\/package\.json$/, "/<bin>")
```

That regex hard-codes a forward slash. On Windows, `require.resolve` returns a backslash path (`...\recheck-jar\package.json`), so the replace silently no-ops and `recheck` ends up invoking Java (or the native exe) with `package.json` itself as the executable argument:

```
Error: Invalid or corrupt jarfile C:\...\recheck-jar\package.json
```

The `redos checker` test then fails on every Windows contributor's machine even though the test is correct and CI on Linux is green. This is invisible to maintainers because Linux/macOS paths use forward slashes, so the upstream regex matches and the bug never surfaces there.

The same regex is used for the native-binary lookup (`recheck-{platform}-{arch}/package.json` → `/recheck[.exe]`), so even `recheck`'s own JVM-fallback chain is broken on Windows: `auto` mode tries native first, fails resolution, falls back to the jar, which also fails resolution. Both paths are dead.

## Fix

Add `scripts/recheck-resolve.ts`, registered ahead of `fail-on-console.ts` in both `vitest.config.ts` and `vitest.root.mjs`. It uses `path.join` (separator-correct on every platform) to resolve:

- `recheck-{platform}-{arch}/recheck[.exe]` → exported as `RECHECK_BIN`, the env var `recheck` already honours as a native-backend override.
- `recheck-jar/recheck.jar` → exported as `RECHECK_JAR`, the JVM fallback for platforms without a pre-compiled native package.

`recheck` short-circuits its own (buggy) resolution when either env var is set, so this is a transparent override of one specific code path. Linux/macOS already worked: there we just hand `recheck` the same paths it would have computed anyway, and gain the ~5× native-binary speedup as a side benefit.

```ts
function resolveOptional(specifier: string, file: string): string | undefined {
  try {
    const path = join(dirname(requireFromHere.resolve(specifier)), file);
    return existsSync(path) ? path : undefined;
  } catch (err) {
    if (isModuleNotFound(err)) return undefined;
    throw err;
  }
}
```

Three guarantees built in:
1. **Respects user overrides.** We never overwrite a pre-existing `RECHECK_BIN` / `RECHECK_JAR`.
2. **No-op on unsupported configurations.** The optional packages are matched via `Partial<Record<NodeJS.Platform, string>>` / `Partial<Record<NodeJS.Architecture, string>>` lookups, so an unknown platform/arch combo simply leaves the env var unset; `recheck` then runs its own resolution as today.
3. **Narrow error handling.** Only `MODULE_NOT_FOUND` (the expected "optional dep missing" path) is swallowed. Any other resolver error — permissions, broken stub `package.json`, etc. — re-throws so real bugs aren't masked.

## Performance

The bigger win is moving the redos check to the native binary backend:

- **Before** (JVM via `recheck-jar`): cold JVM startup of 8–13s per worker on Windows, frequently breaching the test's 10s `}, 10000)` timeout under parallel-file load. The test was effectively flaky, not just broken.
- **After** (native binary via `RECHECK_BIN`): sub-second startup. Local full-suite run on Windows: 16.16s, all 3791 tests pass deterministically.

Linux and macOS get the same native-binary path — no JVM is ever spawned.

## Tests

- **New `packages/zod/src/v4/classic/tests/recheck-resolve.test.ts`** locks in the setup-file contract: at least one of `RECHECK_BIN` / `RECHECK_JAR` must be set after vitest setup, and whichever is set must point to a real file. This catches future regressions where, say, an `optionalDependencies` rename or a pnpm-lockfile churn breaks resolution silently.
- **The existing `redos checker` (datetime.test.ts) is the integration test.** It exercises `checkSync` end-to-end, which is exactly the path that was broken on Windows. With this PR it passes there for the first time.

```
✓ src/v4/classic/tests/recheck-resolve.test.ts (1 test) 8ms
✓ src/v4/classic/tests/datetime.test.ts (14 tests) 1.8s
  ...
Test Files  339 passed (339)
     Tests  3791 passed (3791)
```

## API / behaviour

- **No public API surface change.** `scripts/` and the vitest config are dev-only; `packages/zod` runtime code is untouched.
- **No new dependencies.** `recheck-jar` and `recheck-{platform}-{arch}` are already declared as `optionalDependencies` of `recheck`, which is already in `devDependencies`. We just resolve them with the right separator.
- **Backwards-compatible with manual configuration.** Anyone who has already set `RECHECK_BIN` or `RECHECK_JAR` (e.g. as a shell-level workaround for the bug this PR fixes) keeps that exact value — we never overwrite.

## Scope / risk

- 4 files touched: `scripts/recheck-resolve.ts` (new, 49 lines), `packages/zod/src/v4/classic/tests/recheck-resolve.test.ts` (new, 17 lines), `vitest.config.ts` (+4/-1), `vitest.root.mjs` (+4/-1).
- Test-infra only. Zero changes to `packages/zod/src/**` runtime code.
- No type signature changes, no exports added, no bundler config changes.

I considered filing the regex bug upstream against `makenowjust-labs/recheck` instead — it's the right long-term fix — but the round-trip on a third-party fix can be months, this PR is a self-contained workaround that uses `recheck`'s own documented escape hatch (`RECHECK_*` env vars), and the performance bonus is real on every platform. Happy to drop or reshape if you'd prefer to wait on upstream.

### DIFF
--- a/packages/zod/src/v4/classic/tests/recheck-resolve.test.ts
+++ b/packages/zod/src/v4/classic/tests/recheck-resolve.test.ts
@@ -1,0 +1,42 @@
+import { statSync } from "node:fs";
+import { createRequire } from "node:module";
+import { expect, test } from "vitest";
+
+// Locks in the contract of scripts/recheck-resolve.ts: when an optional
+// `recheck` backend package is installed, the corresponding env var
+// must point to a real file after vitest setup runs. On platforms/archs
+// without any installed `recheck` backend (e.g. s390x, riscv64,
+// FreeBSD), the setup is a no-op and this test skips.
+const requireFromHere = createRequire(import.meta.url);
+
+function isResolvable(specifier: string): boolean {
+  try {
+    requireFromHere.resolve(specifier);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+const PLATFORM: Record<string, string> = { darwin: "macos", linux: "linux", win32: "windows" };
+const ARCH: Record<string, string> = { x64: "x64", arm64: "arm64" };
+const nativePackage =
+  PLATFORM[process.platform] && ARCH[process.arch]
+    ? `recheck-${PLATFORM[process.platform]}-${ARCH[process.arch]}/package.json`
+    : null;
+
+const hasAnyBackend =
+  isResolvable("recheck-jar/package.json") || (nativePackage !== null && isResolvable(nativePackage));
+
+test.skipIf(!hasAnyBackend)("recheck-resolve setup file points recheck at the installed backend(s)", () => {
+  const bin = process.env.RECHECK_BIN;
+  const jar = process.env.RECHECK_JAR;
+
+  expect(bin || jar, "neither RECHECK_BIN nor RECHECK_JAR was set by scripts/recheck-resolve.ts").toBeTruthy();
+
+  if (bin) expect(statSync(bin).isFile(), `RECHECK_BIN missing: ${bin}`).toBe(true);
+  if (jar) {
+    expect(jar).toMatch(/recheck\.jar$/);
+    expect(statSync(jar).isFile(), `RECHECK_JAR missing: ${jar}`).toBe(true);
+  }
+});
--- a/scripts/recheck-resolve.ts
+++ b/scripts/recheck-resolve.ts
@@ -1,0 +1,52 @@
+import { existsSync } from "node:fs";
+import { createRequire } from "node:module";
+import { dirname, join } from "node:path";
+
+// `recheck`'s backend resolution does
+// `require.resolve(pkg).replace(/\/package\.json$/, "/<bin>")` for both
+// the JVM jar and the native binary. The forward-slash regex no-ops on
+// Windows backslash paths, so `recheck` invokes Java/exe with
+// `package.json` as the executable. We resolve both ourselves and set
+// the env vars `recheck` already honours as overrides.
+const requireFromHere = createRequire(import.meta.url);
+
+const RECHECK_PLATFORM: Partial<Record<NodeJS.Platform, string>> = {
+  darwin: "macos",
+  linux: "linux",
+  win32: "windows",
+};
+const RECHECK_ARCH: Partial<Record<NodeJS.Architecture, string>> = {
+  x64: "x64",
+  arm64: "arm64",
+};
+
+function isModuleNotFound(err: unknown): boolean {
+  return typeof err === "object" && err !== null && "code" in err && err.code === "MODULE_NOT_FOUND";
+}
+
+function resolveOptional(specifier: string, file: string): string | undefined {
+  try {
+    const path = join(dirname(requireFromHere.resolve(specifier)), file);
+    return existsSync(path) ? path : undefined;
+  } catch (err) {
+    if (isModuleNotFound(err)) return undefined;
+    throw err;
+  }
+}
+
+if (!process.env.RECHECK_BIN) {
+  const platform = RECHECK_PLATFORM[process.platform];
+  const arch = RECHECK_ARCH[process.arch];
+  if (platform && arch) {
+    const bin = resolveOptional(
+      `recheck-${platform}-${arch}/package.json`,
+      platform === "windows" ? "recheck.exe" : "recheck"
+    );
+    if (bin) process.env.RECHECK_BIN = bin;
+  }
+}
+
+if (!process.env.RECHECK_JAR) {
+  const jar = resolveOptional("recheck-jar/package.json", "recheck.jar");
+  if (jar) process.env.RECHECK_JAR = jar;
+}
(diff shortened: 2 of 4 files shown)


dokson:
Quick follow-up on the "filing upstream is the right long-term path" line in the PR body — turns out it's already been done, and merged:

- Issue: [makenowjust-labs/recheck#1616](https://github.com/makenowjust-labs/recheck/issues/1616) — same error signature ("Invalid or corrupt jarfile node_modules\\recheck-jar\\package.json")
- Fix: [makenowjust-labs/recheck#1620](https://github.com/makenowjust-labs/recheck/pull/1620) — merged 2025-11-09

The catch: the fix only ships in the **pre-release** line so far (`v4.6.0-beta.2` and later). The latest stable is `v4.5.0` (2025-03-02) — predates the fix — and that's what this repo currently consumes.

So the workaround in this PR still has standalone value:
1. Stable `recheck` is still broken on Windows until `v4.6.0` ships.
2. Even after a `recheck` bump, the explicit `RECHECK_BIN` override gives the redos test the native-binary backend on every platform — ~5× faster than the JVM path that `recheck`'s own auto-detection picks today.

Happy to drop the PR if you'd rather wait for `recheck` v4.6.0 and bump the dep instead — that path closes the bug at the source. Just figured the speedup + the not-yet-released-stable-fix gap was worth flagging.

dokson:
Hi @colinhacks — friendly bump on this one if you have a moment. The Windows ReDoS check flake is currently the only thing failing locally on a clean tree for me, which is making pre-push hooks unreliable and slowing down my iteration on other PRs (had to bypass with `--no-verify` on #5938 just now, which I'd rather not do as a habit). Happy to help if there's anything actionable on my end — but if it's just waiting on a review, no rush, thanks for considering!

colinhacks:
Fixed upstream already—pinning to the beta version of recheck 4.6 that contains the fix.

Issue #35, opened 2020-04-20
Support for other primitives


ksmithut:
Are there plans to support other primitives such as Promise and BigInt?

vriad:
I think it's a great idea - I'll look into this in the next couple days. Promise especially is important. The combination of function schemas and Promise types is gonna be 🔥

vriad:
@ksmithut  Just implemented support for `bigint` and Promises in zod@1.3.0 

Take a look at the README to see how I'm handling Promises. Let me know if this approach works for your use case. 

What other primitive types would you want to see? If you think there are other types that should be implemented, I'll create additional issues for each of those. In the meantime, closing this.

ksmithut:
That's awesome, love the promise approach you implemented. This will work wonderfully

PR #5922, opened 2026-05-01
fix(v4): type `safeParse` errors by input, not output (#5195)

### LINKED ISSUE #5195: [V4] ZodSafeParseResult only keeps the output type of the schema
When using `safeParse`, the returned `ZodSafeParseResult` only keeps track of the output type of the schema. As a result, when trying to format errors with `treeifyError`, the resulting `ZodErrorTree` uses the type of the schema output instead of the input. This leads to incorrect typing when piping a schema to a transform:

```typescript
import { z, treeifyError } from 'zod';

type A = { a: number };
type B = { b: string };

const baseSchema = z.object({
  a: z.number().max(9),
});

const transform = z.transform<A, B>(({ a }) => ({
  b: a.toString(),
}));

const pipe = baseSchema.pipe(transform);

const errors = pipe.safeParse({ a: 10 })?.error;

if (errors) {
  console.log(treeifyError(errors).properties?.a); // Typescript error but the property exists
  console.log(treeifyError(errors).properties?.b); // No Typescript error but the property does not exist
}

```

dokson:
Fixes #5195. Related ecosystem evidence: [`nicoespeon/zod-v3-to-v4#30`](https://github.com/nicoespeon/zod-v3-to-v4/issues/30) and [`#35`](https://github.com/nicoespeon/zod-v3-to-v4/pull/35) — the official v3→v4 migration tool had to drop v3's input generic when rewriting `SafeParseReturnType<I, O>` to `ZodSafeParseResult<O>`, and explicitly flagged this issue as the upstream gap.

Discussion seed in https://github.com/colinhacks/zod/issues/5195#issuecomment-4359672000. Opening the PR for concrete review surface; happy to close it if the maintainer would rather keep the current shape.

## Problem

When a schema's input differs from its output (pipe / transform / codec), `safeParse(...).error` is currently typed as `ZodError<output>` even though the issues it carries are paths in the **input** space (validation runs there). `z.treeifyError(result.error)` therefore infers a tree keyed by output properties that don't exist at runtime, while the input-keyed properties that DO exist are flagged as type errors.

OP's reproduction:

```ts
type A = { a: number };
type B = { b: string };
const pipe = z.object({ a: z.number().max(9) })
  .pipe(z.transform<A, B>(({ a }) => ({ b: a.toString() })));

const errors = pipe.safeParse({ a: 10 })?.error;
if (errors) {
  z.treeifyError(errors).properties?.a; // ← flagged as a type error, but real
  z.treeifyError(errors).properties?.b; // ← accepted, but does not exist at runtime
}
```

The maintainer-suggested workaround `z.treeifyError<unknown>(errors)` is fine for one-off code but doesn't scale into codemods or library APIs that need typed error trees per schema.

## Fix

Generalise the success/error split so error carries the **input** generic, defaulting to the output for backward compatibility:

```ts
// before
export type SafeParseResult<T> = SafeParseSuccess<T> | SafeParseError<T>;

// after
export type SafeParseResult<O, I = O> = SafeParseSuccess<O> | SafeParseError<I>;
```

Then thread `core.input<this>` through:

- `safeParse` / `safeParseAsync` / `spa` overloads on `ZodType` (classic) and `ZodMiniType` (mini)
- `safeEncode` / `safeDecode` (and async variants) — error keyed by the value being parsed
- `inferFlattenedErrors` / `inferFormattedError` v3-compat aliases so `error.flatten()` / `error.format()` typings agree

The two-generic shape mirrors v3's `SafeParseReturnType<I, O>` exactly, which v4 collapsed. Codecs and migrations that used to work in v3 keep working with the same intent.

## Backward compatibility

- `ZodSafeParseResult<X>` (single arg) defaults to `<X, X>` — identical typing for the 98% of schemas where `input === output`.
- `result.success` / `result.data` typings unchanged.
- `result.error` typing only changes for pipe / transform / codec schemas, where the previous typing was unsound: any user code that relied on it was already broken at runtime (the OP's reproduction is exactly that case).

## Tests

- `packages/zod/src/v4/classic/tests/safeparse-error-input.test.ts` — locks in the OP reproduction (tree keyed by input, output keys rejected) and verifies the one-arg form still types correctly for simple schemas.
- Existing `inferFlattenedErrors` test in `error.test.ts` now compiles cleanly (it was already asserting the correct runtime shape — only the inferred type was lying).

## Local results

```
Test Files  1 failed | 338 passed (339)
     Tests  1 failed | 3792 passed (3793)
Type Errors no errors
```

The single failing test (`src/v4/classic/tests/datetime.test.ts > redos checker`) is **not** caused by this PR. It's a `recheck` library bug on Windows that fires for every contributor on Windows, regardless of branch — `recheck`'s jar resolution uses a forward-slash regex that no-ops on backslash paths. This is independent of `safeParse` typings and is fixed by **#5919**. Same test passes on Linux CI today.

Diff stats: 7 files, +62 / -25 (most of which is overload signatures gaining a second generic).

## Note about CI

CI on this PR will also show a red `Test with TypeScript latest` job, also unrelated to this change: it's the repo-wide `baseUrl` deprecation that surfaces under TS 6 since [`e58ea4d`](https://github.com/colinhacks/zod/commit/e58ea4d91b1dfe8194b73508203213cbc7e9c936). Every PR opened since that commit fails the same way. **#5921** (small workflow patch) addresses it; once that lands the noise on this PR will go away.

Sanity-checked locally on TS 6.0.3 with #5919 + #5921 patches applied: full suite green, `safeparse-error-input.test.ts` passes, `inferFlattenedErrors` test passes, no type errors anywhere in the repo.

### DIFF
--- a/packages/zod/src/v4/classic/compat.ts
+++ b/packages/zod/src/v4/classic/compat.ts
@@ -28,11 +28,11 @@ export const ZodIssueCode = {
 } as const;
 
 /** @deprecated Use `z.$ZodFlattenedError` */
-export type inferFlattenedErrors<T extends core.$ZodType, U = string> = core.$ZodFlattenedError<core.output<T>, U>;
+export type inferFlattenedErrors<T extends core.$ZodType, U = string> = core.$ZodFlattenedError<core.input<T>, U>;
 
 /** @deprecated Use `z.$ZodFormattedError` */
 export type inferFormattedError<T extends core.$ZodType<any, any>, U = string> = core.$ZodFormattedError<
-  core.output<T>,
+  core.input<T>,
   U
 >;
 
--- a/packages/zod/src/v4/classic/parse.ts
+++ b/packages/zod/src/v4/classic/parse.ts
@@ -1,7 +1,7 @@
 import * as core from "../core/index.js";
 import { type ZodError, ZodRealError } from "./errors.js";
 
-export type ZodSafeParseResult<T> = ZodSafeParseSuccess<T> | ZodSafeParseError<T>;
+export type ZodSafeParseResult<O, I = O> = ZodSafeParseSuccess<O> | ZodSafeParseError<I>;
 export type ZodSafeParseSuccess<T> = { success: true; data: T; error?: never };
 export type ZodSafeParseError<T> = { success: false; data?: never; error: ZodError<T> };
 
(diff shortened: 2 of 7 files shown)


colinhacks:
Too breaky and not what most people want. Zod had to make a call here. One counterexample doesn't merit a change in behavior. And I suspect it's more common for the input to be loose (z.coerce, z.preprocess) than the output (piping into transforms).

Issue #5002, opened 2025-07-29
Breaking change in v4.0.6?


jaswilli:
Hi, 

I recently upgraded a project from version `3.25.76` to `4.0.5` and on versions `4.0.6` and later I'm seeing a change in how a schema I'm using is being validated [(link to playground)](https://zod-playground.vercel.app?appdata=N4IgzgxgFgpgtgQxALhALwHQHsBGArGCAFwApgAdAOwAJqBLAE2WswFdK6tKSrbaBtXn1qZKrODhgAnEgEoANEOGYwRKXUoBzORjEAbPXTBQ5S6gF1ZZjGoSUwAMyxS4JEgDcEe2dQC8APmoPLz9fX2p9PWoAH2jqTyiw8PYGGAcNGAZqAH54kOYAOXFJGQTZcusABzpKmBJRYuk5Kxo%2BbEqiTkovU0oAXxaQeRAE1hgwFH4QYEZmFLSMhj6qIenZ6gBGPpBzYfdpMC6UEABmDAAmAFYMAHYANlWjAC0sBgBZDToUBy8wGD6gA).

Prior to `4.0.6` when using the following schema:

```
z.object({
  id: z.union(
    [
      z.number(),
      z.string().nullish()
    ])
    .transform((val) => (val === null || val === undefined ? val : Number(val)))
    .pipe(z.number())
    .optional()
})
```

An input that didn't contain an `id` property would validate but now throws an error ("id: invalid_type - Invalid input: expected number, received undefined").

It seems like maybe this was the commit that introduced the change in behavior: https://github.com/colinhacks/zod/pull/4941/commits/ede96295f0b6648be5a5c225aff2aa5c30b822a4#diff-22c8e6df681ac92b8f6c73108cc92d6f75c13232421cb54f7024195a9b631980

Is this something that can or should be reverted back to previous v3 and early v4 behavior or do I need to work around it?

Thanks!

colinhacks:
Apologies. It's always shocking when something that seems so simple isn't in the test suite. Fixed in `zod@4.0.13`.

PR #5923, opened 2026-05-01
fix(v4): propagate inner-schema `optin` through `pipe(transform, …)` (#5917)

### LINKED ISSUE #5917: The results vary depending on the position of the “optional” in the schema with preprocess since v4.4.0
The results vary depending on the position of the “optional" in the schema with preprocess since v4.4.0.

```ts
// If “optional” is placed outside the preprocess block, `success = true`.
console.log(
  z.object({
    optionalNumber: z.preprocess(v => v, z.number()).optional()
  })
    .safeParse({})
    .success
); // true

// If “optional” is placed within the preprocess block, `success = false`.
console.log(
  z.object({
    optionalNumber: z.preprocess(v => v, z.number().optional())
  })
    .safeParse({})
    .success
); // false
```

The error message for the latter.
```
{
  "code": "invalid_type",
  "expected": "nonoptional",
  "path": [
    "optionalNumber"
  ],
  "message": "Invalid input: expected nonoptional, received undefined"
}
```

For version 4.3.6 or earlier, both will return `success = true`.

dokson:
Fixes #5917. Same family as #5002 (fixed in v4.0.13 via `362eb33`), now affecting the inner-`.optional()` placement under `z.preprocess`.

## Problem

Since v4.4.0, the position of `.optional()` inside `z.preprocess` matters:

```ts
// Outside — works (v4.3.6 and v4.4.0+)
z.object({
  optionalNumber: z.preprocess((v) => v, z.number()).optional(),
}).safeParse({}); // success: true

// Inside — works pre-v4.4.0, fails on main
z.object({
  optionalNumber: z.preprocess((v) => v, z.number().optional()),
}).safeParse({}); // success: false  ← regression
// → "Invalid input: expected nonoptional, received undefined"
```

`z.preprocess(fn, schema)` desugars to `pipe(transform(fn), schema)`. `$ZodPipe` reads its input optionality from the **left** side:

```ts
util.defineLazy(inst._zod, "optin", () => def.in._zod.optin);
```

The left side is a `$ZodTransform`, which has no `optin` (transforms accept any value, including `undefined`). So the pipe reports `optin === undefined` → required → the parent object rejects the missing key before ever invoking the inner schema.

## Fix

When the left side of a pipe is a transform, defer `optin` to the right side. The transform is input-transparent (it passes through whatever it receives), so the schema downstream is what decides whether `undefined` is a valid input.

```diff
- util.defineLazy(inst._zod, "optin", () => def.in._zod.optin);
+ util.defineLazy(inst._zod, "optin", () =>
+   def.in._zod.def.type === "transform" ? def.out._zod.optin : def.in._zod.optin
+ );
```

Surgical: the special case fires only on `transform → schema` pipes (the `z.preprocess` shape). Ordinary `pipe(realSchema, …)` keeps its existing left-side-driven `optin` — so e.g. `z.string().pipe(z.number())` continues to be required as before.

## Tests

`packages/zod/src/v4/classic/tests/preprocess-inner-optional.test.ts` covers:

1. `.optional()` outside preprocess — accepts missing key (was already passing, locked in as non-regression).
2. `.optional()` inside preprocess — accepts missing key (the main bug from #5917).
3. The exact "expected nonoptional" failure mode no longer fires.
4. Both placements parse a present value identically.
5. `z.preprocess` without `.optional()` still rejects missing keys.
6. `pipe(realSchema, transform)` keeps its left-side-driven `optin` (non-regression for ordinary pipes).

Local results:

```
Test Files  338 passed (339)   [redos checker is the unrelated Windows flake — fixed by #5919]
     Tests  3794 passed (3795)
Type Errors no errors
```

## Note about CI

The `Test with TypeScript latest` job will likely show red because of the repo-wide `baseUrl` deprecation surfacing under TS 6 — every PR opened since `e58ea4d` fails the same way; **#5921** addresses it.

### DIFF
--- a/packages/zod/src/v4/classic/tests/preprocess-inner-optional.test.ts
+++ b/packages/zod/src/v4/classic/tests/preprocess-inner-optional.test.ts
@@ -1,0 +1,57 @@
+import { expect, test } from "vitest";
+import * as z from "zod/v4";
+
+// Regression tests for #5917.
+
+// OP's first case: `.optional()` outside preprocess. Worked before, must keep working.
+test("optional outside z.preprocess accepts missing key", () => {
+  const schema = z.object({
+    optionalNumber: z.preprocess((v) => v, z.number()).optional(),
+  });
+  expect(schema.safeParse({}).success).toBe(true);
+});
+
+// OP's second case: `.optional()` inside preprocess. Regressed in v4.4.0 → main bug.
+test("optional inside z.preprocess accepts missing key (#5917 main bug)", () => {
+  const schema = z.object({
+    optionalNumber: z.preprocess((v) => v, z.number().optional()),
+  });
+  const result = schema.safeParse({});
+  expect(result.success).toBe(true);
+});
+
+// Before the fix the failure mode was a specific error message — keep it pinned
+// so we know what to look for if the pipe optin logic regresses again.
+test("optional inside z.preprocess no longer reports `expected nonoptional`", () => {
+  const schema = z.object({
+    optionalNumber: z.preprocess((v) => v, z.number().optional()),
+  });
+  const result = schema.safeParse({});
+  if (!result.success) {
+    throw new Error(JSON.stringify(result.error.issues));
+  }
+});
+
+// Both placements should produce the same parsed shape for present values too.
+test("both placements parse a present value identically", () => {
+  const inside = z.object({ n: z.preprocess((v) => v, z.number().optional()) });
+  const outside = z.object({ n: z.preprocess((v) => v, z.number()).optional() });
+  expect(inside.safeParse({ n: 7 })).toEqual(outside.safeParse({ n: 7 }));
+});
+
+// Sanity: removing `.optional()` entirely must still reject missing keys.
+test("z.preprocess without optional rejects missing keys", () => {
+  const schema = z.object({
+    requiredNumber: z.preprocess((v) => v, z.number()),
+  });
+  expect(schema.safeParse({}).success).toBe(false);
+});
+
+// Sanity: regular `pipe(realSchema, transform)` (NOT preprocess) keeps the old
+// left-side optin behaviour — only the transform-on-the-left case shifts.
+test("pipe with a real schema on the left keeps left-side optin", () => {
+  const schema = z.object({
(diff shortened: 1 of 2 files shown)


dokson:
Closing in favor of #5929 by @colinhacks (branch `zod-preprocess-codec`), which fixes #5917 with a more isolated approach: a new `ZodPreprocess` structural subtype of `ZodPipe` that re-installs the lazy `optin`/`optout`/`values`/`propValues` to defer to the inner schema, without touching the general `$ZodPipe` propagation algorithm.

This PR took the opposite scope — patching `$ZodPipe.optin` directly so the fix applied to all pipes (incl. manual `z.pipe(z.transform(fn), x.optional())`). #5929 is intentionally narrower and matches what was reported in #5917, while leaving the general pipe internals untouched — consistent with the maintainer's stance in #4611 about preserving the current `optin`/`optout` propagation structure.

Thanks for the alternate approach 👍

colinhacks:
Still weighing the relative merits myself! Tricky one. Your PR is very much still in the running.

dokson:
Reopened — sorry for the premature close, didn't see your message before.

Quick framing of the trade-off as I see it, in case it helps your decision:

| | This PR (#5923) | #5929 (`ZodPreprocess`) |
|---|---|---|
| Where the fix lives | `$ZodPipe.optin` propagation | New structural subtype of `ZodPipe` |
| Coverage | Any `pipe(transform, X)` — incl. manual `z.pipe(z.transform(fn), z.string().optional())` | Only `z.preprocess(...)` |
| Touches `$ZodPipe` internals | Yes (the `optin`/`optout` propagation you flagged in #4611) | No |
| New surface area | None | A new schema class + JSON-schema special-cases |

If the right semantic is "preprocess is just sugar over `pipe(transform, schema)`", the broader fix matches user intuition (manual pipe behaves the same as preprocess). If the goal is to keep `$ZodPipe` internals frozen for the type-recursion reasons in #4611, the narrower fix is the right call and the manual `pipe(transform, optional())` case stays as-is.

Happy either way. If you decide on #5929, will close this again.

colinhacks:
Went with the other approach, with some tweaks. I think it encapsulates its own complexity a little better.

PR #5925, opened 2026-05-01
fix(v4): allow dynamic `.catch()` under `unrepresentable: "any"` (#5273)

### LINKED ISSUE #5273: Workaround for "Dynamic catch values are not supported in JSON Schema"
When using `z.toJSONSchema`, I noticed that `.catch((ctx) => somethingCustom(ctx))` was causing a crash. There was not much info on this edge case in docs and discussions, so I decided to create this issue just in case.

Relevant source:

https://github.com/colinhacks/zod/blob/62bf4e439e287e55c843245b49f8d34b1ad024ee/packages/zod/src/v4/core/to-json-schema.ts#L561-L573

Relevant test:

https://github.com/colinhacks/zod/blob/62bf4e439e287e55c843245b49f8d34b1ad024ee/packages/zod/src/v4/classic/tests/to-json-schema.test.ts#L281-L284

`z.toJSONSchema` supports `unrepresentable: "any"` which could potentially used for dynamic catch, similarly to how it's done for `transform()` and others:

https://github.com/colinhacks/zod/blob/62bf4e439e287e55c843245b49f8d34b1ad024ee/packages/zod/src/v4/core/to-json-schema.ts#L521-L526

https://github.com/colinhacks/zod/blob/62bf4e439e287e55c843245b49f8d34b1ad024ee/packages/zod/src/v4/core/to-json-schema.ts#L617-L628

It'd be also great if https://zod.dev/json-schema was updated to reflect the caveat of dynamic catch.

Happy to help with a PR if the above makes sense.

---

In the meantime, I [patched](https://pnpm.io/cli/patch) the logic locally like this:

```diff
} catch {
-   throw new Error("Dynamic catch values are not supported in JSON Schema");
+   // Assume that catch does not change the type of the schema
+   // Context: https://github.com/colinhacks/zod/issues/5273
+   break;
}
```

It's more like `unrepresentable: "passthrough"` (which does not exist yet) rather than `unrepresentable: "any"`.

dokson:
Closes #5273.

## Summary

`z.toJSONSchema()` crashes on any schema that uses a dynamic `.catch((ctx) => …)`, even when the caller has opted into lossy conversion via `unrepresentable: "any"`. This makes the option unusable for any tree containing a context-dependent catch — which is the exact scenario the option was added to handle.

`catchProcessor` derives a static `default` by invoking `def.catchValue(undefined)`. When the callback inspects its `ctx` argument it throws, and the processor unconditionally re-throws `"Dynamic catch values are not supported in JSON Schema"`.

## Fix

Gate the re-throw on `ctx.unrepresentable === "throw"`. Under `"any"`, skip the `default` and fall through with the inner schema — which has already been populated by the preceding `process(def.innerType, …)` call. This is the same passthrough-on-the-representable-side shape that `pipe(transform, …)` already produces today under `"any"`, so no new flag value is introduced.

```ts
const schema = z.string().catch((ctx) => `${ctx.issues.length} issues`);

z.toJSONSchema(schema);                              // throws (unchanged)
z.toJSONSchema(schema, { unrepresentable: "any" });  // { type: "string" }
```

## Behaviour matrix

| Schema | `unrepresentable` | Result | Status |
|---|---|---|---|
| `z.string().catch("x")` | `"throw"` (default) | `{ type: "string", default: "x" }` | unchanged |
| `z.string().catch((ctx) => …)` | `"throw"` (default) | throws | unchanged |
| `z.string().catch((ctx) => …)` | `"any"` | `{ type: "string" }` | new |
| `z.bigint().catch((ctx) => …)` | `"any"` | `{}` | inner already unrepresentable |
| `z.optional(z.string().catch(dynFn))` | `"any"` | `{ type: "string" }` (with optional) | new, transitive |

Default behaviour is fully unchanged; the new path is opt-in via the existing flag.

## On `"passthrough"` vs `"any"`

The issue author notes their patch is "more like `unrepresentable: 'passthrough'` (which does not exist yet) rather than `unrepresentable: 'any'`." Agreed semantically, but adding a new union variant would need to be applied coherently across every unrepresentable processor and is a separate API decision. The current PR reuses the existing `"any"` flag — the same pattern that `pipe(transform, …)` already relies on for de-facto passthrough behaviour — and avoids growing the surface area. Happy to follow up with a `"passthrough"` proposal if preferred.

## Changes

- `packages/zod/src/v4/core/json-schema-processors.ts` — gate the re-throw on `unrepresentable === "throw"`.
- `packages/zod/src/v4/classic/tests/to-json-schema.test.ts` — inline-snapshot test for the `"any"` path next to the existing static-vs-dynamic catch tests.
- `packages/docs/content/json-schema.mdx` — new "Dynamic `.catch()` values" sub-section under `unrepresentable`, documenting the three cases (the docs caveat the issue explicitly asked for).

## Test plan

- [x] `pnpm vitest run to-json-schema` — 334/334 passing locally.
- [x] Full suite on CI.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -325,6 +325,12 @@ describe("toJSONSchema", () => {
     // Dynamic catch values
     const dynamicCatchSchema = z.string().catch((ctx) => `${ctx.issues.length}`);
     expect(() => z.toJSONSchema(dynamicCatchSchema)).toThrow("Dynamic catch values are not supported in JSON Schema");
+    expect(z.toJSONSchema(dynamicCatchSchema, { unrepresentable: "any" })).toMatchInlineSnapshot(`
+      {
+        "$schema": "https://json-schema.org/draft/2020-12/schema",
+        "type": "string",
+      }
+    `);
   });
 
   test("string formats", () => {
--- a/packages/zod/src/v4/core/json-schema-processors.ts
+++ b/packages/zod/src/v4/core/json-schema-processors.ts
@@ -518,7 +518,10 @@ export const catchProcessor: Processor<schemas.$ZodCatch> = (schema, ctx, json, 
   try {
     catchValue = def.catchValue(undefined as any);
   } catch {
-    throw new Error("Dynamic catch values are not supported in JSON Schema");
+    if (ctx.unrepresentable === "throw") {
+      throw new Error("Dynamic catch values are not supported in JSON Schema");
+    }
+    return;
   }
   json.default = catchValue;
 };


colinhacks:
@pullfrog drop the doc changes then merge

PR #5926, opened 2026-05-01
fix(v4): break circular import between classic schemas and iso (#5275)

### LINKED ISSUE #5275: Circular imports between `schemas` and `iso`
Issue is a little related to https://github.com/colinhacks/zod/issues/5095, https://github.com/colinhacks/zod/issues/227

When I look into `schemas`, it imports from `iso`, which itself imports from `schemas` again. According to #227 this shouldn't be the case anymore, but I still see this on the latest version of zod.

https://github.com/colinhacks/zod/blob/62bf4e439e287e55c843245b49f8d34b1ad024ee/packages/zod/src/v4/classic/schemas.ts#L5

https://github.com/colinhacks/zod/blob/62bf4e439e287e55c843245b49f8d34b1ad024ee/packages/zod/src/v4/classic/iso.ts#L2

This caused issues in our build stack, although I'm not sure what the ultimate culprit is (likely outdated rollup dependency?), but in general it seems like a good idea to avoid circular dependencies.

If you're happy with that, I can have a look to resolve these circular dependencies, maybe add a test or some configuration to avoid circular dependencies in the future. What do you think?

dokson:
Closes #5275.

## Summary

`packages/zod/src/v4/classic/schemas.ts` imported `./iso.js` to forward the deprecated `ZodString.{datetime,date,time,duration}` proxies; `./iso.js` imported back from `./schemas.js` to extend `ZodStringFormat`. Rollup (and any bundler that reports cycles) flagged this two-way dependency, which is what kachkaev opened the issue about and what other consumers (instantsearch.js etc.) hit downstream.

## Fix

Move the `ZodISODateTime / ZodISODate / ZodISOTime / ZodISODuration` class definitions into `schemas.ts` next to their `ZodStringFormat` base, and have the deprecated `ZodString` proxies call the underlying `core._iso*` builders directly. `iso.ts` becomes a thin namespace facade that re-exports the classes from `schemas.ts` and exposes the `z.iso.{datetime,date,time,duration}` builder functions.

Public API is unchanged. All of `z.iso.datetime()`, `z.iso.ZodISODateTime`, top-level `ZodISODateTime`, and `z.string().datetime()` keep their existing shape and behaviour. The dependency between `schemas.ts` and `iso.ts` is now strictly one-way (`iso → schemas`), so the cycle is gone in both directions.

`packages/zod/src/v4/mini/iso.ts` was already cycle-free (`mini/schemas.ts` doesn't import `mini/iso.ts`), so nothing to change there.

## Regression tests

The issue author asked for "a test or some configuration to avoid circular dependencies in the future". This PR adds two complementary checks:

1. **`packages/zod/src/v4/classic/tests/no-circular-imports.test.ts`** — fast static analysis. Walks the `v4/classic` source tree, parses relative `import`/`export … from "./…"` statements, builds the import graph and runs DFS to detect cycles. ~50 lines, zero dependencies, runs in milliseconds. Catches any future cycle anywhere in `v4/classic`. Verified locally: when `import * as iso from "./iso.js"` is reintroduced into `schemas.ts`, the test fails with the exact cycle path.

2. **`packages/treeshake/tests/no-circular-imports.test.ts`** — bundles three zod entries (`zod-full.ts`, `zod-string.ts`, `zod-mini-full.ts`) via rollup's programmatic API (concurrently) and asserts that no `CIRCULAR_DEPENDENCY` warning involves zod source files. Mirrors the exact rollup pass that originally surfaced the issue, so it catches cross-tree cycles (e.g. `classic ↔ core ↔ mini`) that the static check above doesn't.

### Note on the rollup-based test (test 2)

This test runs rollup three times for real — it adds **~35s** to `pnpm test` end-to-end (rollup is CPU-bound on TS parsing, so `test.concurrent` only buys back ~5s vs sequential). The static-analysis test (test 1) already covers the regression scope of #5275 and runs in milliseconds.

Happy to drop test 2 if the cost isn't worth it for the project's CI budget — it's contained to a single new file plus a `vitest.config.ts` in the previously-untested `packages/treeshake` package, both trivial to remove. Or it can be moved to a separate non-default suite if you'd prefer to keep the bundler-level guarantee but off the hot path.

## Changes

- `packages/zod/src/v4/classic/schemas.ts` — drop the `import * as iso` line; inline the `ZodISO*` class definitions; deprecated string proxies call `core._iso*` directly.
- `packages/zod/src/v4/classic/iso.ts` — reduced to ~20 lines; re-exports classes from `schemas.ts` and defines the four `z.iso.*` builder functions.
- `packages/zod/src/v4/classic/tests/no-circular-imports.test.ts` — static-analysis cycle detector (test 1).
- `packages/treeshake/tests/no-circular-imports.test.ts` + `packages/treeshake/vitest.config.ts` — rollup-based cycle detector (test 2; see note above).

## Test plan

- [x] `pnpm vitest run iso string to-json-schema` — 612/612 passing locally.
- [x] `pnpm vitest run --project zod` — 3785/3786 passing (the one failure is the unrelated `recheck-jar` Windows env issue from #5919, not from this PR).
- [x] Verified test 1 actually fails when the cycle is reintroduced.
- [x] Full suite on CI.

### DIFF
--- a/packages/treeshake/tests/no-circular-imports.test.ts
+++ b/packages/treeshake/tests/no-circular-imports.test.ts
@@ -1,0 +1,40 @@
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+import commonjs from "@rollup/plugin-commonjs";
+import resolvePlugin from "@rollup/plugin-node-resolve";
+import typescript from "@rollup/plugin-typescript";
+import { rollup } from "rollup";
+import { expect, test } from "vitest";
+
+const HERE = dirname(fileURLToPath(import.meta.url));
+const PKG_ROOT = resolve(HERE, "..");
+const ZOD_SRC = resolve(PKG_ROOT, "../zod/src");
+
+const ENTRIES = ["zod-full.ts", "zod-string.ts", "zod-mini-full.ts"];
+
+async function circularWarningsFor(entry: string): Promise<string[]> {
+  const warnings: string[] = [];
+  const zodSrc = ZOD_SRC.replace(/\\/g, "/");
+  await rollup({
+    input: resolve(PKG_ROOT, entry),
+    plugins: [
+      resolvePlugin({ exportConditions: ["@zod/source", "default"] }),
+      commonjs(),
+      typescript({ tsconfig: false }),
+    ],
+    onwarn(warning) {
+      if (warning.code !== "CIRCULAR_DEPENDENCY") return;
+      const ids = (warning.ids ?? []).map((id) => id.replace(/\\/g, "/"));
+      if (ids.some((id) => id.includes(zodSrc))) warnings.push(ids.join(" -> "));
+    },
+  });
+  return warnings;
+}
+
+test.concurrent.each(ENTRIES)(
+  "rollup reports no circular deps in zod source for %s (regression: #5275)",
+  async (entry) => {
+    const warnings = await circularWarningsFor(entry);
+    expect(warnings).toEqual([]);
+  }
+);
--- a/packages/treeshake/vitest.config.ts
+++ b/packages/treeshake/vitest.config.ts
@@ -1,0 +1,11 @@
+import { defineProject } from "vitest/config";
+
+export default defineProject({
+  resolve: {
+    conditions: ["@zod/source", "default"],
+  },
+  test: {
+    include: ["tests/**/*.test.ts"],
+    testTimeout: 60000,
+  },
+});
--- a/packages/treeshake/vitest.config.ts
+++ b/packages/treeshake/vitest.config.ts
@@ -6,6 +6,6 @@ export default defineProject({
   },
   test: {
     include: ["tests/**/*.test.ts"],
-    testTimeout: 60000,
+    testTimeout: 180000,
   },
 });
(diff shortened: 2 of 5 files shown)


dokson:
@colinhacks — flagging the second test for your call.

The PR ships two regression tests for #5275:

1. `packages/zod/src/v4/classic/tests/no-circular-imports.test.ts` — fast static analysis (DFS on the import graph). Runs in milliseconds, catches any future cycle within `v4/classic`. **This is the one I'd keep.**
2. `packages/treeshake/tests/no-circular-imports.test.ts` — bundles three zod entries via rollup and asserts no `CIRCULAR_DEPENDENCY` warnings. Mirrors what originally surfaced the issue, but it's slow: ~35s on Windows locally, **timed out at 60s on Linux CI** (just bumped to 180s in d55d2df3 to keep the PR green, but the timing still feels uncomfortably long for every `pnpm test`).

Happy to drop test (2) if you'd prefer to keep the suite tight — test (1) covers the regression scope, and rollup itself will surface any future cycle as a build warning anyway. Removing it is one file plus the new `packages/treeshake/vitest.config.ts`. Just say the word and I'll push the cleanup.

colinhacks:
Thanks!

PR #5927, opened 2026-05-01
fix(v4): preserve indexed-access correlation + add `outputStrict`/`inputStrict` (#5154)

### LINKED ISSUE #5154: Zod 4 regression: parsing with schema of indexed access type
I’m using indexed access types to select between several Zod schemas of different output types depending on a key. Using TypeScript 5.9.2, this type checks successfully with Zod 3.25.76 but not with Zod 4.0.0 and later:

```ts
import * as z from "zod";

const schemas = { a: z.string(), b: z.number(), c: z.boolean() };
type Schemas = typeof schemas;
function p<K extends keyof Schemas>(k: K, v: unknown): z.output<Schemas[K]> {
  return schemas[k].parse(v);
}
```

With Zod 3, since `schemas[k]` has type `Schemas[K]`, TypeScript correctly accepts `schemas[k].parse(v)` at type `z.output<Schemas[K]>`. But with Zod 4, TypeScript weakens its type to `string | number | boolean`, where the correlation with `K` has been lost, leading to this error:

```
test.ts:6:3 - error TS2322: Type 'string | number | boolean' is not assignable to type 'output<{ a: ZodString; b: ZodNumber; c: ZodBoolean; }[K]>'.
  Type 'string' is not assignable to type 'output<{ a: ZodString; b: ZodNumber; c: ZodBoolean; }[K]>'.

6   return schemas[k].parse(v);
    ~~~~~~


Found 1 error in test.ts:6
```

dokson:
Closes #5154.

## Summary

In v3, `z.output<Schemas[K]>` (where `K extends keyof Schemas`) preserved the correlation between `K` and the schema's output type, so `schemas[k].parse(v)` type-checked. v4 broke this: TypeScript now reports

> Type 'string | number | boolean' is not assignable to type 'output<{ a: ZodString; b: ZodNumber; c: ZodBoolean; }[K]>'

on the user's reproducer.

The breakage has two sources:

1. The lenient conditional
   ```ts
   export type output<T> = T extends { _zod: { output: any } } ? T["_zod"]["output"] : unknown;
   ```
   loses correlation when `T = Schemas[K]` is a deferred indexed-access type. TypeScript eagerly resolves the conditional and returns the wide union of all possible outputs.

2. `parse(): core.output<this>` (and the other parse/encode/decode methods) inherits the same problem on the return-type side.

## Why not just go back to v3's strict constraint

Replacing the conditional with `output<T extends { _zod: { output: any } }> = T["_zod"]["output"]` does fix both, but it cascades into ~370 type errors across 15 files. v4 deliberately splits `_$ZodTypeInternals` (the base — no `output`/`input` fields) from `$ZodTypeInternals<O, I>` (with them), and uses `SomeType = { _zod: _$ZodTypeInternals }` as the constraint for many internal generics. Tightening `output<T>` would erase that distinction and force a deep architectural change. This PR explicitly preserves that design.

### 1. Direct indexed access in method return types

Replace `core.output<this>` / `core.input<this>` with direct `this["_zod"]["output"]` / `this["_zod"]["input"]` in `parse`, `safeParse`, `parseAsync`, `safeParseAsync`, `spa`, `decode`, `encode`, `decodeAsync`, `encodeAsync`, `safeDecode`, `safeEncode`, `check`, `with`, `register`, etc. — 33 sites in `classic/schemas.ts` and 6 in `mini/schemas.ts`.

`this` is always shaped as `{ _zod: { output: ... } }` inside these interface declarations, so the conditional in `core.output<this>` was never needed there; direct access works and preserves correlation through `Schemas[K]`.

### 2. Add opt-in strict variants

```ts
export type outputStrict<T extends { _zod: { output: any } }> = T["_zod"]["output"];
export type inputStrict<T extends { _zod: { input: any } }> = T["_zod"]["input"];
```

Users hitting the indexed-access pattern swap `z.output<...>` for `z.outputStrict<...>`:

```ts
function p<K extends keyof Schemas>(k: K, v: unknown): z.outputStrict<Schemas[K]> {
  return schemas[k].parse(v);
}
// ✅ p("a", "x") is `string`, p("b", 1) is `number`, p("c", true) is `boolean`.
```

The lenient `output<T>` / `input<T>` are completely unchanged, so the existing call sites that rely on the `unknown` fallback (and on `SomeType` not exposing `output`/`input`) keep compiling. No public API removal.

## Why this is the right shape

I considered six type-level alternatives (strict constraint, `infer` rename, distributive `T extends any ? …`, intersection `(T & {…})["_zod"]["output"]`, non-distributive `[T] extends […]`, `infer U extends …`). Strict was the only one that fixed the correlation case AND kept constrained-generic patterns working — but cost ~370 cascading errors. The other five each traded one regression for another. The two-part fix here is the smallest change that resolves the reported issue without compromising the v4 design.

## Changes

- `packages/zod/src/v4/core/core.ts` — add `outputStrict<T>` / `inputStrict<T>`. `output` / `input` unchanged.
- `packages/zod/src/v4/classic/schemas.ts` — 33 method return-types switched to direct indexed access on `this`.
- `packages/zod/src/v4/mini/schemas.ts` — 6 sites likewise.
- `packages/zod/src/v4/classic/external.ts` and `packages/zod/src/v4/mini/external.ts` — re-export the new strict types.
- `packages/zod/src/v4/classic/tests/indexed-access-output.test.ts` — regression test for the issue's exact pattern using `outputStrict`.

## Test plan

- [x] `pnpm vitest run --project zod` — 3787/3787 logical pass, 0 type errors. (One failure on Windows from the unrelated `recheck-jar` env issue, irrelevant to types.)
- [x] Regression test added covering the indexed-access pattern.
- [x] Full suite on CI.

### DIFF
--- a/packages/zod/src/v4/classic/external.ts
+++ b/packages/zod/src/v4/classic/external.ts
@@ -10,7 +10,7 @@ import { config } from "../core/index.js";
 import en from "../locales/en.js";
 config(en());
 
-export type { infer, output, input } from "../core/index.js";
+export type { infer, output, input, outputStrict, inputStrict } from "../core/index.js";
 export type { JSONType } from "../core/util.js";
 export {
   globalRegistry,
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -96,8 +96,8 @@ export interface ZodType<
   toJSONSchema(params?: core.ToJSONSchemaParams): core.ZodStandardJSONSchemaPayload<this>;
 
   // base methods
-  check(...checks: (core.CheckFn<core.output<this>> | core.$ZodCheck<core.output<this>>)[]): this;
-  with(...checks: (core.CheckFn<core.output<this>> | core.$ZodCheck<core.output<this>>)[]): this;
+  check(...checks: (core.CheckFn<this["_zod"]["output"]> | core.$ZodCheck<this["_zod"]["output"]>)[]): this;
+  with(...checks: (core.CheckFn<this["_zod"]["output"]> | core.$ZodCheck<this["_zod"]["output"]>)[]): this;
   clone(def?: Internals["def"], params?: { parent: boolean }): this;
   register<R extends core.$ZodRegistry>(
     registry: R,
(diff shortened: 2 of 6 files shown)


colinhacks:
As noted there are legitimate reasons for the distributive behavior in `z.output<>`. APIs like these new "strict" variants are too complicated to explain or document so I'd rather avoid adding them. And the code in the original issue still wouldn't work without code changes so they might we well use direct property access themselves to fix their specific issue.

PR #5928, opened 2026-05-01
feat(v4): add z.deepPartial and runtime z.input / z.output

### LINKED ISSUE #2854: What should deepPartial be replaced with?
It looks like `deepPartial` was deprecated in [3.21.0](https://github.com/colinhacks/zod/releases/tag/v3.21.0). What should be used instead? 

Also just to note that the documentation at https://zod.dev/?id=deeppartial doesn't make any mention of `deepPartial` being deprecated.

### LINKED ISSUE #5224: Codecs: cross-container impedance mismatch issue
I tried implementing a codec for transforming a record<S, T> into a map<K, T>, where S is a subtype of string, K is arbitrary, and T may be either a codec or a type. This would be useful to define arbitrary KV containers decoded from json inputs.

What I could get so far:

```typescript
const noCodecValue = z.object({
  a: bigint(),
});
const yesCodecKey = z.codec(z.string(), z.enum(["a", "b"]), { encode: (input) => input, decode: (input) => {
  if (input === "a") {
    return "a";
  }
  if (input === "b") {
    return "b";
  }
  return z.NEVER;
}});

// basically, we want
// input: record<string, { a: bigint | number | string | BN }>
// output: map<"a" | "b", { a: bigint }>

// what we get:
const yn = z.codec(
  z.record(yesCodecKey._zod.def.in, noCodecValue),
  z.map(
    yesCodecKey._zod.def.out, 
    z.any()  // <-- it should somehow reference the output of noCodecValue
    // putting noCodecValue here will get us z.input<typeof noCodecValue>, which is wrong
  ),
  {
    decode: (input) => new Map(Object.entries(input).map(([k, v]) => [
      yesCodecKey.decode(k), 
      v  // <-- output is already there, decoding is done implicitly, since z.object is not a codec
    ])),
    encode: (input) => Object.fromEntries(
      Array.from(input.entries()).map(([k, v]) => [
        yesCodecKey.encode(k), 
        noCodecValue.encode(v)  // <-- this expects to get the output of noCodecValue, however, we'd like that to be the input
      ]
    )),
  }
);
```

bigint implementation for reference:

```typescript
const bigint = () => {
  return z.codec(
    z.union([
      z.string().regex(z.regexes.integer),
      z.int(),
      z.bigint(),
      z.instanceof(BN),
    ]),
    z.bigint(),
    {
      decode: (input, ctx) => {
        if (typeof input === 'bigint') {
          return input;
        }
        try {
          if (typeof input === "string" && input === "") {
            // special case: for some reason BigInt constructor accepts empty string and returns a 0n
            throw new Error("empty string is not a valid integer");
          }
          if (input instanceof BN) {
            return BigInt(input.toString());
          }
          return BigInt(input);
        } catch (err: any) {
          ctx.issues.push({
            code: "invalid_format",
            format: "json",
            input: input.toString(),
            message: err.message ?? "not a valid integer",
          });
          return z.NEVER;
        }
      },
      encode: (input) => input.toString(),
    }
  );
};
```

There seems to be no plausible way to do it currently.

dokson:
Closes #2854.
Closes #5224.

## Background

`.deepPartial()` was deprecated in v3.21.0 and removed in v4. The reason from #2106:

> The `.deepPartial` API has this same problem [as `discriminatedUnion`]. I'm deprecating it for the same reason.

That problem: a recursive `instanceof` switch over schema types that

1. cannot see user-defined schema subclasses,
2. accumulates edge cases for every new schema type, and
3. silently breaks on advanced shapes (transforms, branded, discriminated unions, lazy/recursive, etc.).

Issue #2854 has been the symptom: 60+ commenters asking what to use instead. Community libraries exist (`zod-deep-partial`, `@traversable/zod`) but the most popular one drops about half the advanced types and stack-overflows on recursive schemas.

## Approach

This PR adds a v4-native implementation that addresses the design concerns of #2106 head-on rather than reproducing the original brittle pattern.

- **Dispatches on `schema._zod.def.type`, not `instanceof`.** Custom schemas with an unknown `def.type` fall through to identity instead of being silently mishandled. The visitor enforces exhaustiveness on the known `def.type` set, so any missed case is a compile error.
- **Cycle-safe** via a `Map<schema, RESOLVING|result>` cache. Lazy schemas unfold lazily, so recursive shapes terminate; shared sub-schemas are visited once. Non-`lazy` cycles — v4's getter-based recursive objects — are broken with a `z.lazy` placeholder that resolves through the cache at parse time.
- **Bottom-up rewrite as the primitive.** One internal traversal in `core/` backs all three public helpers; `deepPartial` is a handful of lines on top of it. `deepStrict`, `deepReadonly`, etc. become handlers rather than new switches.
- **`deepPartial` returns a structural type.** The inferred return is a structural `ZodObject` whose properties are the original properties wrapped in `ZodOptional` — rather than a generic `ZodType<DeepPartial<Out>>`. Keeps `.shape`, `.extend`, `.pick`, etc. usable on the result.
- **`z.input` / `z.output` runtime projections.** Companion helpers that descend through pipes to the input or output side of the composition, sharing the same visitor infrastructure.
- **Covers the full v4 def vocabulary**: object, array, tuple (+rest), record, map, set, union (incl. discriminated union), intersection, optional, nullable, default, prefault, nonoptional, catch, readonly, promise, success, pipe, function, lazy. Leaves (primitives, enum, literal, transform, custom, file, etc.) returned untouched.

The traversal lives in `core/` so both `classic` and `mini` reuse it, but it is not exported from any public barrel; each variant has its own thin `deepPartial` wired to its own `partial`.

### Discriminated union note

Making the discriminator field optional collapses the fast-path lookup (every option ends up with `undefined` as a possible discriminator value). To keep parsing correct, `deepPartial` degrades a discriminated union into a plain `union` over the already-partialed options. Validation semantics are preserved (try-each), the only loss is the discriminator fast-path. This is documented in a test.

## Attribution

The traversal pattern (bottom-up rewrite + `RESOLVING` sentinel) is adapted from @jaens's v3 [gist](https://gist.github.com/jaens/7e15ae1984bb338c86eb5e452dee3010) (Apache-2.0). The v4 implementation here is rewritten: `def.type` dispatch instead of `instanceof`, schema construction via `core.clone` instead of `new ZodFoo(...)`, and full v4 def coverage. Attribution noted at the top of the visitor file.

## Example

```ts
const Node: z.ZodType = z.object({
  name: z.string(),
  children: z.array(z.lazy(() => Node)).optional(),
});

z.deepPartial(Node).parse({});                 // ok
z.deepPartial(Node).parse({ children: [{}] }); // ok
```

## Why the traversal is internal

The brittle thing #2106 identifies is *each* derived helper growing its own `instanceof` switch. A single dispatch primitive fixes that whether or not it is exported: `deepPartial`, `z.input`, and `z.output` are each a handful of lines over one shared traversal, and a future `deepStrict` / `deepReadonly` is a handler rather than another switch.

It is deliberately not exported, because the traversal's contract is not settled enough to freeze as public API:

- It returns `$ZodType`, not the schema type it was handed. A handler can replace any node with a schema of a different type, so echoing the input type back is a claim the traversal cannot keep — every caller has to declare what its own rewrite produces, which is what `DeepPartial<T>` is for.
- Anything non-trivial has to carry its own state beside the traversal rather than threading an accumulator through it.

Keeping it in `core/` and out of the public barrels leaves room to replace it with something better without a deprecation cycle. User-defined `def.type`s still pass through untouched, and the `kind satisfies never` check still turns a missed built-in into a compile error.

## Changes

- `packages/zod/src/v4/core/visit.ts` — the traversal primitive. Internal; not re-exported from `core/index.ts` or either flavor barrel.
- `packages/zod/src/v4/classic/in-out.ts` and `mini/in-out.ts` — `z.input` / `z.output` runtime projections.
- `packages/zod/src/v4/classic/deep-partial.ts` — `deepPartial` for classic (object → `.partial()`, discriminated union → plain union, structural return type).
- `packages/zod/src/v4/mini/deep-partial.ts` — same for mini, wired to `mini.partial`.
- `packages/docs/content/api.mdx` — brief coverage for `z.deepPartial` and `z.input` / `z.output`.
- `packages/zod/src/v4/classic/external.ts` and `mini/external.ts` — re-export `deepPartial`, `input`, `output`.
- `packages/zod/src/v4/classic/tests/deep-partial.test.ts` — 39 tests covering: shallow + 3-level nested objects, arrays, tuples + rest, unions, discriminated union (degradation), intersection, record / map / set, promise, optional / nullable, wrapper composition (default + readonly + optional), recursive (lazy) schemas (cycle termination), primitive leaf preservation, transform passthrough, pipe in/out recursion, traversal identity, targeted def.type rewrites, shared sub-schema visit-once, unknown def.type passthrough, catchall traversal, non-lazy getter cycles, already-resolved lazies, structural-return-type assertions, and `keyof`-on-output preservation.

## Test plan

- [x] `pnpm vitest run --project zod` — full suite green, 0 type errors.
- [x] Cycle case verified via the recursive `Node` schema test.
- [x] Structural-return-type assertions cover `.shape` / `.extend` / `.pick` on `deepPartial(...)` result.
- [x] Full suite on CI.


dokson:
@colinhacks — flagging this for your attention since the deprecation was your call. From [#2106](https://github.com/colinhacks/zod/issues/2106):

> The `.deepPartial` API has this same problem. I'm deprecating it for the same reason.

That problem (recursive `instanceof` switch, no extensibility for user-defined schema types, edge cases on advanced shapes) is what this PR specifically tries to address rather than reintroduce. The traversal dispatches on `def.type` so user-defined types pass through cleanly, the cycle handling is explicit and tested, and the `mapOnSchema` primitive means future helpers (`deepStrict`, `deepReadonly`) don't each grow their own switch.

If the design direction has shifted and the API is still considered an anti-pattern regardless of implementation, happy to close as wontfix — wanted to give you an explicit chance to weigh in before that. Either way the core `mapOnSchema` primitive may stand on its own merits.

dokson:
Extended scope: this also closes #5224 with `z.in()` / `z.out()` recursive pipe projections.

@colinhacks — you proposed exactly this API for #5224:

> I've been playing with the idea of `z.in()` and `z.out()` utilities that recursively traverse the schema and give you the "input" or "output" side of all pipes/codecs/etc.

`mapOnSchema` makes the implementation ~5 lines per direction. Codec extends pipe (`$ZodCodecDef extends $ZodPipeDef`), so a single `def.type === "pipe"` case covers both. Tests include the cross-container `record→map` scenario from the original issue. This is the strongest argument for keeping `mapOnSchema` exported as a primitive — `deepPartial`, `z.in`, `z.out`, and any future `deepStrict`/`deepReadonly` are all one-liners on top of it.

dokson:
@colinhacks — honestly moved by how far this PR has come.
Thank you for the time, the care, and for taking the original proposal somewhere better than I could have on my own.

tsibelman:
Thank you working on this, I very exited to move to v4 but need this ability. Thanks again for working on this.

shaunstoll:
@colinhacks any chance we could merge this PR please?

Issue #5494, opened 2025-12-02
Union type generates anyOf instead oneOf


de-don:
```
export const example = z.union([
  z.string().min(10),
  z.string().min(5)
]);
```

will generate schema like this in `zod@4.1.13`

```
"exampleField": {
  "anyOf":[
    {"type":"string","minLength":10},
    {"type":"string","minLength":5}
  ]
}
```

but in zod v3 it generates:
```
"exampleField": {
  "oneOf":[
    {"type":"string","minLength":10},
    {"type":"string","minLength":5}
  ]
}
```

I think it should use `oneOf` instead of `anyOf`

de-don:
I just recreated this from another account, and closed previous issue, sorry. This is a right one

de-don:
@dosubot, how should I proceed with the case, when I need union from string and int?

enzoferey:
@dosubot On `zod@3.25.76` when using `zod/v4` using `discriminatedUnion` outputs `anyOf` instead of `oneOf` as mentioned here. Why?

colinhacks:
Discriminated union -> `oneOf`
Union -> `anyOf`

This is sound and mirrors the validation behavior of these keywords in JSON Schema. (With discriminated unions, the input can only match one of the union elements so it's implicitly an exclusive union)

de-don:
@colinhacks is it possible somehow to use `discriminatedUnion` with simple types like `string, number, boolean`?
In our case, for example, we have 2 types of string like `uuid_id` and `numeric_id`, and we need union `uuid_id | numeric_id`. 

Now I fixed this via "patching" the generated swagger and replacing `anyOf -> oneOf` because swagger codegeneration works really bad with `anyOf` :(

Issue #5493, opened 2025-12-02
Union type generates anyOf instead oneOf


ddontsov-b2broker:
```
export const example = z.union([
  z.string().min(10),
  z.string().min(5)],
);
```

will generate schema like this in `zod@4.1.13`

```
"exampleField": {
  "anyOf":[
    {"type":"string","minLength":10},
    {"type":"string","minLength":5}
  ]
}
```

but in zod v3 it generates:
```
"exampleField": {
  "oneOf":[
    {"type":"string","minLength":10},
    {"type":"string","minLength":5}
  ]
}
```

I think it should use `oneOf` instead of `anyOf`

PR #5930, opened 2026-05-01
feat(v4): add `union` option to `z.toJSONSchema` for `oneOf` opt-in (#5807, #5493, #5494)

### LINKED ISSUE #5807: v4: toJSONSchema anyOf incompatibility with Gemini API Function Calling
# Zod v4 `anyOf` Incompatibility with Gemini API Function Calling

## Summary

Zod v4's `z.union()` generates `anyOf` in JSON Schema (as per spec), but **Google Gemini API's Function Calling does not support `anyOf`**, causing tool calls to fail silently or return validation errors.

## Environment

- **Zod Version**: 4.3.6
- **LLM Provider**: Google Vertex AI (Gemini API)
- **Model**: gemini-2.5-flash-lite, gemini-2.5-flash, gemini-2.5-pro
- **Framework**: Mastra v1.16.0

### Background: Zod v4 Behavior (Working as Intended)

As documented in https://github.com/colinhacks/zod/issues/4089:
- `z.union()` → `anyOf` (correct per JSON Schema spec)
- `z.discriminatedUnion()` → `oneOf` (correct per JSON Schema spec)

This was fixed in Zod v4.1.13 and is working correctly.

### The Issue: Gemini API Limitation

Google Gemini API's Function Calling **does not support `anyOf`** in tool parameter schemas. When a tool parameter uses `z.union()`, the generated schema contains `anyOf`, which causes:

1. **Silent failures**: LLM ignores the union constraint
2. **Validation errors**: LLM outputs data that doesn't match any branch of the union
3. **Tool call failures**: Request rejected by API with validation errors

### Test Case

```typescript
import { z } from "zod";

// This schema causes issues with Gemini API
const sptScheduleIndexSchema = z.union([
  z.object({ am: z.number().int() }),
  z.object({ pm: z.number().int() }),
]);

const toolInputSchema = z.object({
  dayIndex: z.number().int().min(0),
  sptScheduleIndex: sptScheduleIndexSchema,
  sptName: z.string(),
});

// Generated JSON Schema
console.log(JSON.stringify(z.toJSONSchema(toolInputSchema), null, 2));
```

### Output (Problematic for Gemini)

```json
{
  "type": "object",
  "properties": {
    "dayIndex": {
      "type": "integer",
      "minimum": 0
    },
    "sptScheduleIndex": {
      "anyOf": [
        {
          "type": "object",
          "properties": {
            "am": {"type": "integer"}
          },
          "required": ["am"],
          "additionalProperties": false
        },
        {
          "type": "object",
          "properties": {
            "pm": {"type": "integer"}
          },
          "required": ["pm"],
          "additionalProperties": false
        }
      ]
    },
    "sptName": {
      "type": "string"
    }
  },
  "required": ["dayIndex", "sptScheduleIndex", "sptName"]
}
```

### Error Observed

When Gemini API receives this schema:
- LLM may not respect the union structure
- Tool calls fail with validation errors
- Request may be rejected: `"Unable to submit request because at least one contents field is required"`

## Workaround

Replace complex union types with simple flat schemas for LLM consumption, then transform internally:

```typescript
// LLM-friendly schema (avoids anyOf)
const llmSptScheduleIndexInputSchema = z.object({
  time: z.enum(["am", "pm"]),
  index: z.number().int().min(0),
});

const toolInputSchema = z.object({
  dayIndex: z.number().int().min(0),
  sptScheduleIndex: llmSptScheduleIndexInputSchema,
  sptName: z.string(),
});

// Transform function for internal use
function convertToSptScheduleIndex(input: {
  time: "am" | "pm";
  index: number;
}): { am: number } | { pm: number } {
  return input.time === "am" ? { am: input.index } : { pm: input.index };
}
```

### Generated Schema (Works with Gemini)

```json
{
  "type": "object",
  "properties": {
    "sptScheduleIndex": {
      "type": "object",
      "properties": {
        "time": {
          "type": "string",
          "enum": ["am", "pm"]
        },
        "index": {
          "type": "integer",
          "minimum": 0
        }
      },
      "required": ["time", "index"],
      "additionalProperties": false
    }
  }
}
```

## Impact

This affects:
- ✅ **Structured Output**: Works (uses `discriminatedUnion` → `oneOf`)
- ❌ **Tool/Function Calling**: Fails when using `z.union()` → `anyOf`

## Questions for Zod Team

1. **Should this be documented?**
   - Add a note in Zod v4 docs about LLM provider compatibility
   - Mention that some providers (Gemini) don't support `anyOf`

2. **Is there a flag to control anyOf/oneOf generation?**
   - Allow users to opt into `oneOf` for `z.union()` for compatibility

3. **Should we recommend workarounds?**
   - Document the "flat schema + transform" pattern for LLM compatibility

## Related Issues

- https://github.com/colinhacks/zod/issues/4089 - Original `union` → `anyOf` fix (resolved in v4.1.13)

## Additional Context

- Gemini API Documentation: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini
- JSON Schema spec: https://json-schema.org/understanding-json-schema/reference/combining
- OpenAPI spec on anyOf/oneOf: https://swagger.io/docs/specification/v3_0/data-models/oneof-anyof-allof-not/

## Verification

Tested with:
- Zod v4.3.6
- Gemini models: gemini-2.5-flash-lite, gemini-2.5-flash, gemini-2.5-pro
- Confirmed `discriminatedUnion` (oneOf) works, `union` (anyOf) fails

dokson:
Closes #5807. Also addresses #5493 and #5494 which surface the same need from a different consumer.

## The problem

Multiple downstream consumers of `z.toJSONSchema()` reject the `anyOf` keyword that v4 emits for inclusive unions:

- **Google Gemini Function Calling** (#5807): schemas with `anyOf` cause silent tool-call failures or validation errors.
- **Swagger / OpenAPI codegen** (#5494): generated client bindings break on `anyOf` shapes.
- **Older language-binding generators** (#5493): same.

The default is sound, and stays put. As you noted in https://github.com/colinhacks/zod/issues/5494#issuecomment-3654057725:

> Discriminated union -> `oneOf`
> Union -> `anyOf`
>
> This is sound and mirrors the validation behavior of these keywords in JSON Schema.

`z.union(...)` is inclusive — multiple branches can match the same value — so `anyOf` matches the spec semantics. This PR doesn't change that.

## What this adds

A new opt-in parameter on `z.toJSONSchema()`:

```ts
z.toJSONSchema(schema, { union: "oneOf" });
```

When set, inclusive unions emit `oneOf` instead of `anyOf`. Default is `"anyOf"`.

Discriminated unions and `z.xor()` are unaffected — they're exclusive by construction and keep emitting `oneOf` regardless.

## Why a flag and not just `override`

Users hit by this today work around it by walking the produced schema after the fact:

```ts
z.toJSONSchema(schema, {
  override: (ctx) => {
    if (ctx.zodSchema._zod.def.type === "union" && ctx.jsonSchema.anyOf) {
      ctx.jsonSchema.oneOf = ctx.jsonSchema.anyOf;
      delete ctx.jsonSchema.anyOf;
    }
  },
});
```

It works but it's brittle: it rewrites the path keys after they've been computed (the `path` arg in any user `override` already references `anyOf` indices), it has to be duplicated in every project that hits the issue, and it's not discoverable from the docs. A flag is one parameter, doesn't cost anything when unused, and the path is correct from the start.

## Implementation

- `JSONSchemaGeneratorParams` and `ToJSONSchemaContext` get a `union: "anyOf" | "oneOf"` field.
- `initializeContext` defaults it to `"anyOf"`.
- `unionProcessor` picks the keyword once: `oneOf` if exclusive *or* opted in, else `anyOf`. The path argument and the output key are both written under the chosen keyword.

```ts
const keyword: "oneOf" | "anyOf" = isExclusive || ctx.union === "oneOf" ? "oneOf" : "anyOf";
const options = def.options.map((x, i) =>
  process(x, ctx as any, { ...params, path: [...params.path, keyword, i] })
);
json[keyword] = options;
```

## Test coverage

5 tests in `tests/to-json-schema-union.test.ts`:

- Default emits `anyOf` for inclusive union.
- `union: "oneOf"` emits `oneOf` for inclusive union.
- Explicit `union: "anyOf"` matches the default.
- Discriminated unions stay on `oneOf` regardless of the flag.
- Nested unions inside objects are also rewritten.

`pnpm vitest run to-json-schema*` — 344/344 passing locally with 0 type errors. No existing snapshots changed (default unchanged).

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema-union.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema-union.test.ts
@@ -1,0 +1,44 @@
+import { describe, expect, test } from "vitest";
+import * as z from "zod";
+
+describe("z.toJSONSchema union option", () => {
+  test("default: inclusive union → anyOf", () => {
+    const schema = z.union([z.string(), z.number()]);
+    const json = z.toJSONSchema(schema);
+    expect(json.anyOf).toBeDefined();
+    expect(json.oneOf).toBeUndefined();
+  });
+
+  test('union: "oneOf" → inclusive union emits oneOf (regression #5807, #5493, #5494)', () => {
+    const schema = z.union([z.string(), z.number()]);
+    const json = z.toJSONSchema(schema, { union: "oneOf" });
+    expect(json.oneOf).toBeDefined();
+    expect(json.anyOf).toBeUndefined();
+  });
+
+  test('union: "anyOf" (explicit) matches default', () => {
+    const schema = z.union([z.string(), z.number()]);
+    expect(z.toJSONSchema(schema, { union: "anyOf" })).toEqual(z.toJSONSchema(schema));
+  });
+
+  test("discriminated unions always use oneOf, regardless of union option", () => {
+    const schema = z.discriminatedUnion("kind", [
+      z.object({ kind: z.literal("a"), a: z.string() }),
+      z.object({ kind: z.literal("b"), b: z.number() }),
+    ]);
+    expect(z.toJSONSchema(schema).oneOf).toBeDefined();
+    expect(z.toJSONSchema(schema, { union: "oneOf" }).oneOf).toBeDefined();
+    expect(z.toJSONSchema(schema, { union: "anyOf" }).oneOf).toBeDefined();
+  });
+
+  test("nested unions inside object are also rewritten", () => {
+    const schema = z.object({
+      x: z.union([z.string(), z.number()]),
+      y: z.union([z.boolean(), z.null()]),
+    });
+    const json = z.toJSONSchema(schema, { union: "oneOf" });
+    expect((json.properties as any).x.oneOf).toBeDefined();
+    expect((json.properties as any).y.oneOf).toBeDefined();
+    expect((json.properties as any).x.anyOf).toBeUndefined();
+  });
+});
--- a/packages/zod/src/v4/core/json-schema-processors.ts
+++ b/packages/zod/src/v4/core/json-schema-processors.ts
@@ -335,20 +335,18 @@ export const objectProcessor: Processor<schemas.$ZodObject> = (schema, ctx, _jso
 
 export const unionProcessor: Processor<schemas.$ZodUnion> = (schema, ctx, json, params) => {
   const def = schema._zod.def as schemas.$ZodUnionDef;
-  // Exclusive unions (inclusive === false) use oneOf (exactly one match) instead of anyOf (one or more matches)
-  // This includes both z.xor() and discriminated unions
+  // Exclusive unions (z.xor / discriminated union) always emit oneOf.
+  // Inclusive unions default to anyOf, but the caller can opt into oneOf
+  // via `union: "oneOf"` for consumers that don't accept anyOf.
   const isExclusive = def.inclusive === false;
+  const keyword: "oneOf" | "anyOf" = isExclusive || ctx.union === "oneOf" ? "oneOf" : "anyOf";
   const options = def.options.map((x, i) =>
     process(x, ctx as any, {
       ...params,
-      path: [...params.path, isExclusive ? "oneOf" : "anyOf", i],
+      path: [...params.path, keyword, i],
     })
   );
-  if (isExclusive) {
-    json.oneOf = options;
-  } else {
-    json.anyOf = options;
-  }
+  json[keyword] = options;
 };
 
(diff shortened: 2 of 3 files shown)


colinhacks:
Very allergic to additional params on this function. I'm considering a plugin API that would give Zod the ability to encapsulate some of these common overrides and provides them to users via a consumable API.

PR #2601, opened 2023-07-21
feat: added latitude and longitude in number

hudson-bruno:
This pull request solves https://github.com/colinhacks/zod/issues/2600 adding [Latitude](https://en.wikipedia.org/wiki/Latitude#:~:text=Latitude%20is%20given%20as%20an%20angle%20that%20ranges%20from%20%E2%80%9390%C2%B0%20at%20the%20south%20pole%20to%2090%C2%B0%20at%20the%20north%20pole%2C%20with%200%C2%B0%20at%20the%20Equator.) and [Longitude](https://en.wikipedia.org/wiki/Longitude#:~:text=Longitude%20is%20given%20as%20an%20angular%20measurement%20ranging%20from%200%C2%B0%20at%20the%20Prime%20Meridian%20to%20%2B180%C2%B0%20eastward%20and%20%E2%88%92180%C2%B0%20westward) validation in zod number

Issue #4629, opened 2025-06-05
GeoJSON support


imide:
pretty much what the title says. geojson support is helpful for GIS/weather projects.

there is an [external package with support](https://github.com/infra-blocks/ts-zod-utils/tree/master/src/geojson) however it hasn't been updated and I believe it's better for it to be 1st party.

colinhacks:
Zod tries to mirror the type system of JavaScript/TypeScript. Stuff like this is interesting but out of scope for the core library. Perhaps in the future there could be a `@zod/geojson` with stuff like this but it's probably better as a community library. Consider forking/vendoring that package.

PR #5932, opened 2026-05-01
feat(v4): add `z.latitude()` / `z.longitude()` as bundle-cheap sugar (#2600)

### LINKED ISSUE #2600: [feat] `z.number().latitude()` and `z.number().longitude()`
Latitude and longitude are used in a variety of applications, it feels like this simple validation is missing, currently is possible to validate chaining min and max but it would be better if this validation was abstracted away.

```ts
const latitude = z.number().min(-90).max(90)
const longitude = z.number().min(-180).max(180)

// Proposed
const latitude = z.number().latitude()
const longitude = z.number().longitude()
```

dokson:
Closes #2600. Builds on the implementation closed in #2601 by addressing the specific bundle-size objection that closed it.

## Why this PR exists

The original feature request (#2600, July 2023) asks for `z.number().latitude()` / `.longitude()` as ergonomic sugar over `min(-90).max(90)` / `min(-180).max(180)`. The implementation PR (#2601) was closed by @colinhacks in April 2024:

> "Doesn't seem like there's much demand for this. Zod needs to be careful about bundle size, so I think this is a little too niche, but thank you for the PR."

Two objections: **bundle size** and **demand**. Demand is hard to refute — but bundle size is the actionable one, and #2601 was unnecessarily expensive for what it shipped.

This PR is the same public API at a fraction of the cost.

## Bundle-size argument

#2601 went the maximalist route: 2 new `ZodIssueCode` enum values, 2 issue interface types, 2 case branches in the number parse switch, 2 builder methods, and 2 locale entries × 49 languages (= 98 translation entries to maintain forever). All shipped to every consumer regardless of whether they call `.latitude()`.

`latitude` and `longitude` are pure numeric bounds — there is no information `not_latitude` carries that `too_big`/`too_small` doesn't already say more clearly. So this PR ships them as **thin sugar**:

```ts
// classic
export function latitude(params?: string | core.$ZodNumberParams): ZodNumber {
  return number(params).gte(-90).lte(90);
}
```

| Metric                          | #2601    | this PR  |
|---------------------------------|----------|----------|
| Source bytes added              | ~5,554   | **~1,431** (~74% less) |
| New `ZodIssueCode` enum values  | 2        | **0**    |
| New issue interface types       | 2        | **0**    |
| Parse-switch branches           | 2        | **0**    |
| Locale entries to maintain      | 98       | **0**    |
| Tree-shakeable when unused      | ❌       | **✅**   |

The trade-off: the error message is `"Number must be greater than or equal to -90"` instead of `"Number must be a valid latitude"`. I'd argue this is **more** informative for debugging — the user immediately sees the bound that failed and doesn't have to look up what valid latitude means — but I recognize it's a stylistic call. A future enhancement could pass a custom message through to the underlying `lte` check; for now I left the existing localized error chain untouched.

## Why this is the cheapest correct approach

I considered three alternatives, summarized so future readers can see the trade-offs:

1. **Full schema (#2601 shape)**: best UX (custom error messages), worst bundle. Rejected by @colinhacks already.
2. **String-form regex (e.g. [this StackOverflow answer](https://stackoverflow.com/a/31408260))**: solves a different problem — paired `"40.71,-74.00"` strings — and would require its own factory tree. ~6-8× the cost of this PR for separate primitives. Out of scope for #2600 which is explicitly about `z.number()`.
3. **Sugar over existing checks** (this PR): zero new infrastructure, public API matches the issue's proposal verbatim, error messages localize via the existing `too_big`/`too_small` chain.

(3) wins on every dimension except the cosmetic one (generic error message), which the PR description above argues is actually a feature.

## Why no v3 support

#2601 added v3 + v4. @colinhacks rejected the whole package, so re-proposing v3 would just re-trigger the same objection. v4 is where the new `z.number().gte(...)` machinery lives that makes the sugar cheap — bundle parity in v3 would need the duplicated regex/checks shape from #2601. Easy follow-up if maintainer wants it; not the load-bearing argument here.

## Why no string-form `z.string().latLng(...)`

Different feature, different scope. This PR sticks to the literal #2600 ask. If string-form coordinate validation is also wanted, it's a clean separate PR.

## Public API

```ts
z.latitude();              // number in [-90, 90]
z.longitude();             // number in [-180, 180]
z.number().latitude();     // chainable, same constraint
z.number().longitude();
z.number().int().latitude(); // composes with other number checks
```

Available in both `zod/v4` (classic) and `zod/v4-mini` (mini). Mini is top-level only — no chainable, matching mini's functional pattern.

## Tests

10 cases across `classic/tests/latitude-longitude.test.ts` and `mini/tests/latitude-longitude.test.ts`:

- Bounds inclusivity (`±90`, `±180` accepted; `±90.0001`, `±180.0001` rejected).
- Non-number rejection.
- Chainable parity (`z.number().latitude()` ≡ `z.latitude()`).
- Composition with other number checks (`.int().latitude()`).
- Output type inference resolves to `number`.
- **Issue-code regression test**: failing `latitude(91)` produces `too_big`, not `not_latitude` — confirms no new code was added to the public issue surface.

`pnpm vitest run latitude-longitude` — 20/20 passing locally with 0 type errors.

## Pinging

@colinhacks — this is built specifically to clear the bundle-size objection from #2601. If demand still feels too low to warrant any new builders, that's a fair call and I'll close. But if the original objection was primarily about cost, this version costs ~74% less.

@hudson-bruno — credit for the original implementation in #2601; this PR keeps the same API surface you proposed (top-level + chainable).

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -673,6 +673,10 @@ z.number().nonnegative();
 z.number().negative(); 
 z.number().nonpositive(); 
 z.number().multipleOf(5);              // alias .step(5)
+z.number().latitude();                 // alias .gte(-90).lte(90)
+z.number().longitude();                // alias .gte(-180).lte(180)
+z.latitude();                          // alias z.number().latitude()
+z.longitude();                         // alias z.number().longitude()
 ```
 </Tab>
 <Tab value="Zod Mini">
@@ -686,6 +690,10 @@ z.number().check(z.nonnegative());
 z.number().check(z.negative()); 
 z.number().check(z.nonpositive()); 
 z.number().check(z.multipleOf(5));     // alias .step(5)
+// no .latitude()/.longitude() chainable in mini
+// (functional only — see below)
+z.latitude();                          // number in [-90, 90]
+z.longitude();                         // number in [-180, 180]
 ```
 </Tab>
 </Tabs>
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -984,6 +984,11 @@ export interface _ZodNumber<Internals extends core.$ZodNumberInternals = core.$Z
   /** @deprecated In v4 and later, z.number() does not allow infinite values by default. This is a no-op. */
   finite(params?: unknown): this;
 
+  /** Constrain to a valid latitude (`[-90, 90]`). Sugar over `.gte(-90).lte(90)`. */
+  latitude(params?: string | core.$ZodCheckLessThanParams): this;
+  /** Constrain to a valid longitude (`[-180, 180]`). Sugar over `.gte(-180).lte(180)`. */
+  longitude(params?: string | core.$ZodCheckLessThanParams): this;
+
   minValue: number | null;
   maxValue: number | null;
   /** @deprecated Check the `format` property instead.  */
@@ -1048,6 +1053,12 @@ export const ZodNumber: core.$constructor<ZodNumber> = /*@__PURE__*/ core.$const
     finite() {
       return this;
     },
+    latitude(params) {
+      return this.check(checks.gte(-90)).check(checks.lte(90, params));
+    },
+    longitude(params) {
+      return this.check(checks.gte(-180)).check(checks.lte(180, params));
+    },
   });
 
   const bag = inst._zod.bag;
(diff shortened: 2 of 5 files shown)


colinhacks:
Closing this out, and it's the feature I'm declining rather than the implementation — this is clean work, and the tree-shaking argument holds up.

`z.number().gte(-90).lte(90)` is the whole thing, and where it stops being trivial — whether the antimeridian is inclusive, whether a longitude means anything at the poles — is where core shouldn't be picking a convention on your behalf. Geospatial modeling stays out of scope for the core library; that's the same call as #4629.

Reasoning is written up on #2600. Thanks for looking into this 👍

PR #5934, opened 2026-05-02
Fix enum template literal pattern escaping

gwagjiug:
Numeric enum values used inside `z.templateLiteral()` were converted to strings for the generated regex pattern, but only string enum values were escaped.

That meant `z.templateLiteral([z.enum({ A: 1.2 })])` generated a pattern like `^(1.2)$`, which also accepted `"1x2"`.

This escapes all enum value strings before building the pattern, matching the behavior of `z.literal(1.2)`, and adds regression coverage.


gwagjiug:
@colinhacks Could you review this?

PR #5935, opened 2026-05-02
feat(string): support array of prefixes in startsWith()

### LINKED ISSUE #3683: Feature: support multiple parameters in `.startsWith()`
Currently it quite verbose to define a string that can start with multiple different prefixes:

```js
const schema = z
  .string()
  .startsWith('foo_')
  .or(z.string().startsWith('bar_'))
  .or(z.string().startsWith('baz_'));
```

it would be great if `.startsWith()` would support an array of strings like
```js
const schema = z
  .string()
  .startsWith(['foo_', 'bar_', 'baz_']);
```

maybe I am also missing an alternative solution for the problem.

uchkunr:
Closes #3683.

## What's changed

`.startsWith()` now accepts an array of strings in addition to a single string, allowing multiple allowed prefixes without chaining:

```ts
// before — verbose
z.string().startsWith("foo_").or(z.string().startsWith("bar_")).or(z.string().startsWith("baz_"))

// after
z.string().startsWith(["foo_", "bar_", "baz_"])
```

Single-string usage is unchanged (fully backward compatible).

## Behaviour

```ts
const schema = z.string().startsWith(["foo_", "bar_"]);

schema.parse("foo_hello");  // ok
schema.parse("bar_world");  // ok
schema.parse("baz_nope");   // throws: Invalid string: must start with "foo_" or "bar_"
```

JSON Schema output reflects the union as a pattern:

```json
{ "type": "string", "pattern": "^(foo_|bar_).*" }
```

Single-string JSON Schema output is unchanged (`^foo_.*`).

## Implementation

- `core/errors.ts` — `$ZodIssueStringStartsWith.prefix: string | string[]`
- `core/checks.ts` — normalise to `prefixes` array; build `^(a|b).*` pattern for arrays, keep `^a.*` for single string; guard empty array with never-matching regex (`(?!)`) so runtime and JSON Schema stay consistent
- `core/api.ts` — `_startsWith(prefix: string | string[], ...)`
- `classic/schemas.ts` — `.startsWith(value: string | string[], ...)` method signature
- `locales/*.ts` — all 49 locales updated; `en` uses `"a" or "b"` phrasing, others use `"a", "b"` (language-neutral)

## Tests

Three new test cases in `classic/tests/string.test.ts`:

- array of prefixes — valid and invalid inputs
- error issue carries the original `prefix` array
- English error message format

All 3813 existing tests pass.

---
Note on locale translations: Error messages in all locale files except en and uz were translated using Claude Opus 4.7. Native speaker review and corrections are welcome.


colinhacks:
Going to pass on this. `.startsWith()` taking an array reads fine on its own, but `.endsWith()` and `.includes()` wouldn't follow, so the string API would be inconsistent the day it ships — and it widens `issue.prefix` from `string` to `string | string[]` for everyone with a custom error map.

`z.string().regex(/^(foo_|bar_|baz_)/)` covers it — keep the group, since `/^foo_|bar_/` binds the anchor to the first branch only. Clean work, and the escaping and JSON Schema handling were both right. Thanks for the PR 👍

Issue #5937, opened 2026-05-03
Upgrading from 4.3.6 to 4.4.x causes issues with preprocess + catch


terrymun:
I am using zod to parse payload we received from URLSearchParams, which requires transforming comma-separated string values into an array of string (which can potentially be empty/undefined). In v4.3.6 and below, the following schema works:

```ts
z.object({
  area: z.preprocess(v => v ? v.toString().split(',') : [], z.array(z.string())).catch([])
})
```

And the payload is as follow:

```ts
{}
```

[v4.3.6 on Zod Playground](https://zod-playground.vercel.app?appdata=N4IgzgxgFgpgtgQxALhALwHQHsBGArGCAFwApgAdAOwAJqEAnGBZazAB0Y6whjDBIBu1ALwA%2BakID8EjESwBlIvQCWlAOYkAlBjBsANstIByADRHN1FgG0AuidYYG9BAE8SmMEtUbNvjBAQiaBJbTSoAX00QExABBD0AV14UKxBgcJA7WJh6MGUsShQQABYMAGYMADZokGUwAC0sABMAWVVlFAAzeLAYcKA), which works without any issue ✅ 

However, after upgrading to 4.4.x (4.4.0, 4.4.1, and 4.4.2), I get the following error:

```text
invalid_type - Invalid input: expected nonoptional, received undefined
```

All these versions are showing the same error:

- [v4.4.0 on Zod Playground](https://zod-playground.vercel.app?appdata=N4IgzgxgFgpgtgQxALhALwHQHsBGArGCAFwApgAdAOwAJqEAnGBZazAB0Y6whjDBIBu1ALwA%2BakID8EjESwBlIvQCWlAOYkAlBjBsANstIByADRHN1FgG0AuidYYG9BAE8SmMEtUbNvjBAQiaBJbTSoAX00QExABBD0AV14UKxBgcJA7WJh6MGUsShQQABYMUoAGaJBlMAAtLAATAFlVZRQAM3iwGHCgA)
[...]

PR #5938, opened 2026-05-03
feat(v4): clearer error for missing required object properties

dokson:
Refs #5917 (and the [follow-up comment](https://github.com/colinhacks/zod/issues/5917#issuecomment-4364505232)). Pure DX — does **not** change validation behavior.

## Problem

When a required key is absent from an object input, Zod emits the same issue as an explicit `z.nonoptional(...)` rejecting `undefined`:

```
Invalid input: expected nonoptional, received undefined
```

Since v4.4.0 made `z.undefined()` properties required, this message has been recurring confusion (e.g. [vercel/workflow#1901](https://github.com/vercel/workflow/issues/1901)). Users read it as a Zod bug — "I never sent `undefined`, the key isn't there" — and don't immediately see that the fix is `.optional()` on their schema.

The two cases have *different* root causes but share one message:

| Case | Site | Real meaning |
|---|---|---|
| Required key absent in `z.object` | `core/schemas.ts:1764`, `:2050` | property must be present |
| `z.nonoptional(...)` receiving `undefined` | `core/schemas.ts:3760` | value was sent and rejected |

## Fix

Distinguish the two by tagging the missing-key issue with `received: "missing"`, and render that case as `Invalid input: missing required property` in the English locale. The explicit-`undefined` case keeps its current message — there, "received undefined" is accurate.

```diff
 // core/schemas.ts — both missing-key emit sites
 final.issues.push({
   code: "invalid_type",
   expected: "nonoptional",
   input: undefined,
   path: [key],
+  received: "missing",
 });
```

```diff
 // locales/en.ts
 case "invalid_type": {
+  if (issue.received === "missing") {
+    return `Invalid input: missing required property`;
+  }
   const expected = TypeDictionary[issue.expected] ?? issue.expected;
   const receivedType = util.parsedType(issue.input);
   const received = TypeDictionary[receivedType] ?? receivedType;
   return `Invalid input: expected ${expected}, received ${received}`;
 }
```

`$ZodIssueInvalidType.received` already existed informally (used by primitive parsers to mark `"NaN"`, `"Infinity"`, `"Invalid Date"`); this PR formalizes it on the type and adds `"missing"`.

## Effect on the Vercel Workflow repro

```ts
const Run = z.discriminatedUnion("status", [
  z.object({
    status: z.enum(["pending", "running"]),
    output: z.undefined(),
    error: z.undefined(),
    completedAt: z.undefined(),
  }),
  // ...
]);
Run.parse({ status: "running" });
```

Before:
```
run.output: Invalid input: expected nonoptional, received undefined
run.error: Invalid input: expected nonoptional, received undefined
run.completedAt: Invalid input: expected nonoptional, received undefined
```

After:
```
run.output: Invalid input: missing required property
run.error: Invalid input: missing required property
run.completedAt: Invalid input: missing required property
```

The actual fix for that case still belongs to vercel/workflow (those fields need `.optional()` per the v4.4.0 release notes), but at least the error now points the user there instead of looking like a Zod bug.

## Backward compatibility

- `code` stays `"invalid_type"`, `expected` stays `"nonoptional"` — consumers matching by code/expected keep working.
- `received` is optional and additive on `$ZodIssueInvalidType`.
- Other locales fall back to the previous message until updated — no breakage there either.

## Tests

`packages/zod/src/v4/classic/tests/missing-required-property.test.ts`:
1. Missing key under JIT → new message + `received: "missing"`.
2. Missing key under jitless → identical issue (covers both emit sites).
3. Explicit `z.string().optional().nonoptional()` rejecting `undefined` → unchanged message, no `received` field (non-regression).
4. Discriminated-union repro from Vercel Workflow → all three missing fields get the new message.

Updated existing snapshots that asserted the old string for missing-key cases:
- `optional.test.ts` ("object absent keys require optin optional")
- `partial.test.ts` ("catch/prefault/default")

`nonoptional.test.ts` snapshots are deliberately untouched: they go through `handleNonOptionalResult` (explicit-`undefined` path), not the missing-key path.

Local results:
```
Test Files  341 passed
     Tests  3809 passed
Type Errors no errors

pnpm --filter zod build → ✅
```

(One flaky `redos checker` test fails locally on Windows, same flake mentioned in #5923 / fixed upstream by #5919 — unrelated.)

## Note about CI

`Test with TypeScript latest` will likely show red for the same `baseUrl` deprecation that affects every PR opened since `e58ea4d`; addressed by #5921.


dokson:
@colinhacks just rebased on top of main to resolve the conflicts from #5937/#5941.

I think this PR still has merit even after those landed: #5941 made `catch`/`default`/`prefault`/`preprocess` tolerant of absent keys, so they no longer reach the error path at all. That actually narrows the scope of this PR's `received: "missing"` to the cases where an error *should* be produced — i.e. genuinely required fields without a fallback (e.g. `z.string()` or `z.undefined()` on an absent key). The message is now strictly reserved for true missing-required cases, which makes it semantically more precise than before.

Happy to adjust if you'd prefer a different framing.

colinhacks:
Going to close this out. Changing the default message here is a much bigger breaking change than it looks — all 53 locales inherit it, and anyone asserting on error output gets a diff, for a message that's accurate today and only terse.

If you want this wording, a global error map gets you there:

```ts
z.config({
  customError: (iss) => {
    if (iss.code === "invalid_type" && iss.expected === "nonoptional") {
      return "Missing required property";
    }
  },
});
```

PR #5945, opened 2026-05-04
fix(v4): cidrv6 JSON schema pattern matches runtime

### LINKED ISSUE #5944: cidrv6 JSON schema emits an incomplete pattern that rejects valid IPv6 CIDRs
`z.string().cidrv6()` correctly validates IPv6 CIDRs at runtime using `new URL()`, but the `pattern` emitted by `z.toJSONSchema()` rejects many valid inputs.

**Affected version:** 4.4.3 (current main)

## What happens

```ts
import { z } from "zod";

// Runtime validation — correct
z.string().cidrv6().safeParse("2001:db8::/32").success;    // true
z.string().cidrv6().safeParse("2001:db8:1::/48").success;  // true

// JSON schema — incorrect pattern
z.toJSONSchema(z.string().cidrv6());
// {
//   "$schema": "...",
//   "type": "string",
//   "format": "cidrv6",
//   "pattern": "^(([0-9a-fA-F]{1,4}:){7}...|([0-9a-fA-F]{1,4})?::([0-9a-fA-F]{1,4}:?){0,6})\\/..."
// }
```

The exported `pattern` — `regexes.cidrv6` in `src/v4/core/regexes.ts` — only allows **0 or 1 groups before `::`** via `([0-9a-fA-F]{1,4})?::`. Valid CIDRs with 2+ groups before `::` are not matched:

| Input | Zod runtime | JSON schema `pattern` |
|-------|-------------|----------------------|
| `2001:db8::/32` | ✅ valid | ❌ no match |
| `2001:db8:1::/48` | ✅ valid | ❌ no match |
| `2001:0db8:85a3::/64` | ✅ valid | ❌ no match |
| `fe80::/10` | ✅ valid | ✅ matches |
| `::/0` | ✅ valid | ✅ matches |

`2001:db8::/32` is the canonical documentation prefix (RFC 3849) and appears in virtually every IPv6 tutorial and RFC example.

## Why it matters

The `pattern` field is consumed by OpenAPI validators, API gateways, and frontend validation libraries. Any system that applies the Zod-generated JSON schema directly will reject addresses that Zod itself accepts, silently breaking interoperability.

## Root cause

`$ZodCIDRv6` correctly overrides `inst._zod.check` to use `new URL()`, but `def.pattern` is still set to the incomplete `regexes.cidrv6` before calling `$ZodStringFormat.init()`. `$ZodCheckStringFormat.init()` then adds that pattern to `bag.patterns`, which `stringProcessor` in `json-schema-processors.ts` emits as the `pattern` field.

The runtime check and the schema metadata diverge at `src/v4/core/schemas.ts:885`:

```ts
def.pattern ??= regexes.cidrv6; // not used for validation
```

The comment acknowledges the disconnect but the pattern still leaks into the JSON schema output.

## Possible fixes

1. **Fix `regexes.cidrv6`** to correctly cover all compressed forms (similar to how `regexes.ipv6` enumerates eight alternatives).
2. **Don't set `def.pattern`** for `$ZodCIDRv6` at all, so no `pattern` is emitted in the JSON schema. Validators that need a regex would have to rely on `format: "cidrv6"`.

dokson:
Fixes #5944.

## Problem

`z.string().cidrv6()` validates at runtime via `new URL(...)`, but `z.toJSONSchema(...)` emits a `pattern` from `regexes.cidrv6` that rejects most real-world IPv6 CIDRs:

| Input | Runtime | Old `pattern` |
|---|---|---|
| `2001:db8::/32` | ✅ | ❌ |
| `2001:db8:1::/48` | ✅ | ❌ |
| `2001:0db8:85a3::/64` | ✅ | ❌ |
| `fe80::/10` | ✅ | ✅ |
| `::/0` | ✅ | ✅ |

The body `([0-9a-fA-F]{1,4})?::([0-9a-fA-F]{1,4}:?){0,6}` only allowed 0 or 1 group before `::`, so the canonical RFC 3849 documentation prefix `2001:db8::/32` failed the emitted pattern. OpenAPI validators, API gateways, and any consumer that applies the JSON-schema `pattern` directly would reject inputs that Zod itself accepts.

## Fix

Replace the body of `regexes.cidrv6` with the same eight-alternative enumeration already used by `regexes.ipv6` (which covers all compressed forms correctly), then append the existing `/<prefix 0–128>` segment. Runtime validation already used `new URL(...)` and is unaffected.

## Why fix the regex instead of dropping `pattern`

The issue suggests an alternative: stop emitting `pattern` and let consumers rely on `format: "cidrv6"`. Going with option 1 because:

- `format: "cidrv6"` is not a standard JSON Schema format (the spec defines `ipv4`/`ipv6` but not the CIDR variants). Most generic validators — Ajv without a custom format, OpenAPI tooling, frontend libraries — don't recognize it, so dropping `pattern` would leave them with no validation at all. That's exactly the consumer set the issue calls out.
- Every other format in `regexes.ts` (including `cidrv4`) emits a `pattern`; removing it only for `cidrv6` would be asymmetric.
- It would silently regress existing consumers that today rely on the (buggy-but-working-for-some-inputs) pattern — replacing a partial check with no check at all.

## Tests

`packages/zod/src/v4/classic/tests/string.test.ts` — extended the existing `CIDR v6 validation` test with the failing inputs from the issue, plus an assertion that the `pattern` emitted by `z.toJSONSchema` accepts the same set (the runtime/JSON-schema divergence is the bug).


tomohiro86:
Thanks for picking this up so quickly!

colinhacks:
Thanks!

PR #5948, opened 2026-05-04
feat(v4): parseMaybeAsync / safeParseMaybeAsync (JIT-preserving, follow-up to #5924)

dokson:
## Summary

Re-opens #5924 (closed for a 5x sync regression) with an implementation that preserves the `ZodObject` JIT fastpass on sync schemas and bounds the cost on async ones via a per-schema cache. Refs #5379.

## Approach

1. **Sync first.** Build `syncCtx` with `async: false` and run the schema. If it completes without throwing `$ZodAsyncError` and without returning a Promise, finalize and return inline. The fastpass engages exactly as it does for `safeParse`.
2. **Async fallback on `$ZodAsyncError`.** Re-run the schema with `async: true` and return the Promise.
3. **`WeakSet` cache of async-observed schemas.** After a schema throws `$ZodAsyncError` (or returns a Promise on the defensive path), subsequent calls skip the sync attempt entirely. Eliminates per-call throw cost on async-capable schemas.
4. **Hot path is flattened.** No higher-order `finalize` factory. The sync-first / WeakSet block is intentionally inlined in both factories — extracting a shared `_runMaybeAsync` helper that returns `{ ctx, result }` regresses the sync hot path ~12x because V8 stops inlining and the per-call object allocation dominates. Documented inline.

## Tradeoffs

- **Side-effect doubling on the async fallback.** When the sync attempt throws, any sync `transform` / `refine` / check that ran before the first async step runs again on the async re-run. JSDoc on both methods documents this and points users at `parseAsync` for non-idempotent steps.
- **WeakSet is module-level global state.** Keys are schema instances (GC-eligible). Consumers that construct fresh schemas per request never benefit from the cache and pay the try-sync-first cost on every call. Long-lived schemas (the common case) amortize the throw to a single first call.

## Benchmarks

`packages/bench/object-maybe-async.ts` — three scenarios:

| Scenario | Comparison | Ratio |
|---|---|---|
| Flat sync object (3 fields) | `safeParse` vs `safeParseMaybeAsync` | ~1.0–1.3x (within noise) |
| Nested object + array + discriminated union (sync) | `safeParse` vs `safeParseMaybeAsync` | ~1.0–1.2x (sometimes maybe-async wins) |
| Mostly-sync schema with 1 async refine | `safeParseAsync` vs `safeParseMaybeAsync` | **~1.4x** (down from ~14.5x without the WeakSet cache) |

The ~1.4x on the async case is the steady-state cost: after the first call the schema is routed straight to the async path.

## Files

- `packages/zod/src/v4/core/parse.ts` — `_parseMaybeAsync` / `_safeParseMaybeAsync` (try-sync-then-async + `_asyncObserved` WeakSet + flat sync hot path).
- `packages/zod/src/v4/classic/parse.ts`, `classic/schemas.ts` — public API on `ZodType` with side-effect-doubling caveat in JSDoc.
- `packages/zod/src/v4/mini/parse.ts`, `mini/schemas.ts` — same on `ZodMiniType`.
- `packages/zod/src/v4/classic/tests/parse-maybe-async.test.ts`, `mini/tests/...` — sync return / Promise return / error propagation / StandardSchema integration.
- `packages/bench/object-maybe-async.ts` — three-scenario benchmark.

`~standard.validate` was deliberately **not** rewired through `safeParseMaybeAsync` in this PR — the existing `safeParse` + `try/catch` + `safeParseAsync` path stays intact so this PR is purely additive (new public API + new core functions). The internal `validate` refactor is a separable change that can land on its own merits if `parseMaybeAsync` is declined.

## Test plan

- [x] `pnpm vitest run parse-maybe-async` — 24/24 passing
- [x] `pnpm --filter zod build` (strict tsc clean)
- [x] Bench: sync schemas at parity with `safeParse`; async-refine schemas at ~1.4x of `safeParseAsync`

## Discussion

This is the "try it" alternative @colinhacks named on #5379 as the only remaining shape. With sync-first execution + the WeakSet cache, the regression that closed #5924 is gone, the async-path overhead is bounded to ~1.4x steady-state, and the side-effect-doubling caveat is documented in the public JSDoc. Pinged on #5379 for a fresh read; happy to close if the design call still goes the other way.

### DIFF
--- a/packages/bench/object-maybe-async.ts
+++ b/packages/bench/object-maybe-async.ts
@@ -1,0 +1,83 @@
+import * as z from "../zod/src/v4/index.js";
+import { makeData } from "./benchUtil.js";
+import { metabench } from "./metabench.js";
+
+// 1. small flat object — purely sync, exercises ZodObject fastpass directly.
+{
+  const schema = z.object({
+    string: z.string(),
+    boolean: z.boolean(),
+    number: z.number(),
+  });
+  const DATA = makeData(1000, () => ({
+    number: Math.random(),
+    string: `${Math.random()}`,
+    boolean: Math.random() > 0.5,
+  }));
+  const bench = metabench("flat sync object — safeParse vs safeParseMaybeAsync", {
+    safeParse() {
+      for (const _ of DATA) schema.safeParse(_);
+    },
+    safeParseMaybeAsync() {
+      for (const _ of DATA) schema.safeParseMaybeAsync(_);
+    },
+  });
+  await bench.run();
+}
+
+// 2. nested + discriminated union — sync but non-trivial; proves the
+// fastpass engages end-to-end down the schema tree.
+{
+  const inner = z.object({ tag: z.string(), score: z.number() });
+  const schema = z.object({
+    id: z.string(),
+    nested: z.object({ a: z.string(), b: z.number(), c: z.boolean() }),
+    items: z.array(inner),
+    kind: z.discriminatedUnion("type", [
+      z.object({ type: z.literal("a"), x: z.string() }),
+      z.object({ type: z.literal("b"), y: z.number() }),
+    ]),
+  });
+  const DATA = makeData(500, () => ({
+    id: `${Math.random()}`,
+    nested: { a: "x", b: 1, c: true },
+    items: [
+      { tag: "p", score: 1 },
+      { tag: "q", score: 2 },
+    ],
+    kind: Math.random() > 0.5 ? { type: "a", x: "ok" } : { type: "b", y: 1 },
+  }));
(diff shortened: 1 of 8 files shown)


dokson:
@colinhacks could you take a look when you have a moment? This is the follow-up to #5924 — the regression you flagged was caused by `ctx.async = true` upfront bypassing the `ZodObject` JIT fastpass (gated on `ctx?.async === false` strict). Running sync first and falling back to async only on `$ZodAsyncError` keeps the fastpass engaged; benchmarks now show parity with `safeParse` on sync object schemas.

Would value your read on whether the try-sync-then-async tradeoff (one short-circuited sync pass on async schemas, bailing at the first async check) is acceptable, or if you'd prefer a different shape entirely.

PR #5951, opened 2026-05-04
perf(v4): JIT fastpass for ZodTuple (~1.7-2.6x)

dokson:
## Summary

Extends the `$ZodObject` JIT fastpass pattern to `$ZodTuple`. Tuple length, per-item optout flags, and the presence of `def.rest` are all known at construction, so they're baked into a single compiled function via `Doc` + `new Function`. Same gating shape as `$ZodObject` (`jit && fastEnabled && ctx.async === false && !ctx.jitless`), interpreter captured as `superParse` for fall-through. No public API change.

## Why ZodTuple first

Closest structural match to `$ZodObject`'s static-key shape: items count is known at construction, indices are positional, length validation bounds are constants. Lowest risk for porting the codegen pattern; smallest diff that demonstrates the extension is sound. Array/union/record have less direct correspondence and warrant separate PRs.

## Benchmark

`packages/bench/tuple-fastpass.ts` (mitata, 1000 iter):

| Scenario | jitless | JIT | speedup |
|---|---|---|---|
| 3 items | ~93 µs | ~55 µs | ~1.7x |
| 8 items | ~277 µs | ~108 µs | ~2.6x |
| 2 items + rest | ~189 µs | ~84 µs | ~2.25x |

Speedup grows with tuple width — consistent with the unrolling pattern (each item position becomes a static `input[N]` access + per-index issue prefixing instead of an `Array(items.length)` + post-processing pass).

## Compatibility

- **`globalConfig.jitless: true`** — falls through to the existing interpreter, unchanged.
- **CSP-restricted environments** — `util.allowsEval.value` already gates fastEnabled at construction; no `new Function` call is attempted when blocked.
- **Async items** — fastpass is gated on `ctx.async === false`; async path uses `superParse`.
- **Optout tail truncation** and **explicit-undefined preservation** — replicated in the codegen body, regression-tested.

## Tests

11 new tests in `packages/zod/src/v4/classic/tests/tuple-fastpass.test.ts` build the same schema twice (JIT default and `globalConfig.jitless = true`) and assert byte-identical parse output and issue paths across both:

- Happy path (3 required items)
- Invalid item at index 1 — issue path correctness
- Nested tuple — path stays correct through both layers
- `too_small` / `too_big` invariants
- Rest element with extras (validated and invalid)
- Rest with no extras
- Optout tail — absent input truncates result
- Explicit `undefined` inside input is preserved (not truncated)
- Non-array input — `invalid_type`
- Empty tuple
- Async item — falls through to non-fastpass path

Full suite passes: 3833/3833.

## Files

- `packages/zod/src/v4/core/schemas.ts` — `$ZodTuple` constructor adds `generateFastpass`, fastpass dispatch.
- `packages/zod/src/v4/classic/tests/tuple-fastpass.test.ts` — 11 JIT-vs-interpreter regression tests.
- `packages/bench/tuple-fastpass.ts` — three-scenario benchmark.

## Test plan

- [x] `pnpm vitest run` — 3833/3833
- [x] `pnpm --filter zod build` — strict tsc clean
- [x] Bench: 1.7-2.6x over interpreter path

## Discussion

If the direction is welcome, follow-ups could extend the same pattern to `$ZodUnion` (early-return chain with hoisted option runs — moderate complexity) and `$ZodRecord` (similar to array but with key validation). Skipping `$ZodArray` since the per-element call already dominates and unrolling isn't applicable.


colinhacks:
Closing this out. `z.compile()` in #6085 gets to the same place from the other direction — it compiles the whole schema graph ahead of time, and its global mode wraps every newly-constructed schema, so a per-type fastpass for tuples ends up redundant against it and leaves two codegen paths to keep in step for one container.

PR #5952, opened 2026-05-05
fix: use separate type declarations for ESM and CJS exports

### LINKED ISSUE #5686: Zod incorrectly points to CommonJS declaration when used with ESM

lihan3238:
## Summary

- Fixes #5686: The `exports` field in `packages/zod/package.json` used a single flat `types` condition pointing to `.d.cts` (CJS declarations) for all entry points
- This caused TypeScript to resolve ESM imports to CJS declaration files, resulting in broken type inference (e.g. `z.z.core.$strip` instead of `z.core.$strip`) in ESM projects using `module: "nodenext"`
- This change nests the `types` condition under `import` and `require` sub-conditions so that ESM consumers resolve to `.d.ts` and CJS consumers resolve to `.d.cts`

## Changes

Restructured all export entries in `packages/zod/package.json` from:

```json
".": {
  "types": "./index.d.cts",
  "import": "./index.js",
  "require": "./index.cjs"
}
```

To:

```json
".": {
  "import": {
    "types": "./index.d.ts",
    "default": "./index.js"
  },
  "require": {
    "types": "./index.d.cts",
    "default": "./index.cjs"
  }
}
```

Applied to all entry points: `.`, `./mini`, `./locales`, `./v3`, `./v4`, `./v4-mini`, `./v4/mini`, `./v4/core`, `./v4/locales`, `./v4/locales/*`.

## Test plan

- [ ] Verify `pnpm build` generates both `.d.ts` and `.d.cts` files
- [ ] Verify ESM project with `module: "nodenext"` resolves to `.d.ts`
- [ ] Verify CJS project resolves to `.d.cts`
- [ ] Verify type inference works correctly (`z.object({}).shape` not `z.z.core.$strip`)

### DIFF
--- a/packages/zod/package.json
+++ b/packages/zod/package.json
@@ -53,63 +53,113 @@
     "./package.json": "./package.json",
     ".": {
       "@zod/source": "./src/index.ts",
-      "types": "./index.d.cts",
-      "import": "./index.js",
-      "require": "./index.cjs"
+      "import": {
+        "types": "./index.d.ts",
+        "default": "./index.js"
+      },
+      "require": {
+        "types": "./index.d.cts",
+        "default": "./index.cjs"
+      }
     },
     "./mini": {
       "@zod/source": "./src/mini/index.ts",
-      "types": "./mini/index.d.cts",
-      "import": "./mini/index.js",
-      "require": "./mini/index.cjs"
+      "import": {
+        "types": "./mini/index.d.ts",
+        "default": "./mini/index.js"
+      },
+      "require": {
+        "types": "./mini/index.d.cts",
+        "default": "./mini/index.cjs"
+      }
     },
     "./locales": {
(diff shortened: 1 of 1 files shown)


lihan3238:
Closing this PR as the exports are regenerated by zshy on every build. Will fix this in zshy instead so the correct nested types structure is generated automatically.

PR #5954, opened 2026-05-05
fix: prevent tree-shaking of locale initialization

### LINKED ISSUE #5953: "sideEffects": false in zod/v4/package.json causes esbuild to tree-shake locale initialization since version 4.4.1
# Bug
Since zod 4.4.1, all validation error messages in esbuild-bundled (ESM) applications fall back to the generic ["Invalid input"](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) instead of descriptive locale messages like 'Invalid option: expected one of "km"|"mi"'.

# Root Cause
zod/v4/package.json was updated in 4.4.1 to include "sideEffects": false. This causes esbuild to tree-shake the [config(en())](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) call in zod/v4/classic/external.js — the call that registers the English locale.

Without the locale registered, all errors fall through to the hardcoded fallback in zod/v4/core/util.js:

`message = config.localeError?.(iss) ?? "Invalid input"`

# Reproduction
Minimal self-contained test (requires esbuild and zod@>=4.4.1):
```typescript
import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
import { tmpdir } from "node:os";
import { join } from "node:path";

/**
 * Self-contained reproduction test for zod locale tree-shaking bug.
 *
 * Since zod 4.4.1, `zod/v4/package.json` declares `"sideEffects": false`.
 * This causes esbuild to tree-shake the `config(en())` call in
 * `zod/v4/classic/external.js` that initializes the English locale.
 *
 * Without the locale, all zod validation error messages fall back to
 * the generic "Invalid input" instead of descriptive messages like:
 *   'Invalid option: expected one of "km"|"mi"'
 *
 * The fallback happens here in zod's core/util.js:
 *   message = config.localeError?.(iss) ?? "Invalid input"
 */
describe("zod v4 locale tree-shaking bug (esbuild reproduction)", () => {
    let tmpDir: string;

    beforeAll(() => {
        tmpDir = mkdtempSync(join(tmpdir(), "zod-locale-bug-"));
        // Install zod 4.4.1 (the version that introduced sideEffects: false)
        writeFileSync(join(tmpDir, "package.json"), JSON.stringify({ type: "module" }));
        execSync("npm install zod@4.4.1 --silent", { cwd: tmpDir, stdio: "pipe" });
        // With 4.3.6 it works
        // execSync("npm install zod@4.3.6 --silent", { cwd: tmpDir, stdio: "pipe" });
    });

    afterAll(() => {
        rmSync(tmpDir, { recursive: true, force: true });
    });

    const bundleAndRun = async (sourceCode: string): Promise<string> => {
        const inputFile = join(tmpDir, "input.mjs");
        const outputFile = join(tmpDir, "output.mjs");

        writeFileSync(inputFile, sourceCode);

        await build({
            entryPoints: [inputFile],
            bundle: true,
            format: "esm",
            platform: "node",
            outfile: outputFile,
        });

        const result = execSync(`node ${outputFile}`, {
            stdio: "pipe",
            encoding: "utf-8",
        });

        return result.trim();
    };

    it("enum validation error message is 'Invalid input' instead of descriptive message due to locale tree-shaking", async () => {
        const source = `
            import * as z from "zod/v4";
            const schema = z.enum(["km", "mi"]);
            const result = schema.safeParse("feet");
            if (!result.success) {
                console.log(JSON.stringify({
                    message: result.error.issues[0].message,
                    code: result.error.issues[0].code,
                }));
            }
        `;

        const output = JSON.parse(await bundleAndRun(source));

        // BUG: esbuild tree-shakes the locale, so we get generic "Invalid input"
        // instead of the descriptive 'Invalid option: expected one of "km"|"mi"'
        expect(output.code).toBe("invalid_value");
        // expect(output.message).toBe("Invalid input");
        // This is what the message SHOULD be:
        expect(output.message).toBe('Invalid option: expected one of "km"|"mi"');
    });

    it("workaround: explicit locale import preserves descriptive messages", async () => {
        const source = `
            import { config } from "zod/v4/core";
            import en from "zod/v4/locales/en.js";
            config(en());

            import * as z from "zod/v4";
            const schema = z.enum(["km", "mi"]);
            const result = schema.safeParse("feet");
            if (!result.success) {
                console.log(JSON.stringify({
                    message: result.error.issues[0].message,
                    code: result.error.issues[0].code,
                }));
            }
        `;

        const output = JSON.parse(await bundleAndRun(source));

        // With explicit locale import, the descriptive message works
        expect(output.code).toBe("invalid_value");
        expect(output.message).toBe('Invalid option: expected one of "km"|"mi"');
    });
});
```

# Expected behaviour
Invalid option: expected one of "km"|"mi"
# Actual Behavior
Invalid input

# Workaround
Explicitly import and configure the locale:
```typescript
import { config } from "zod/v4/core";
import en from "zod/v4/locales/en.js";
config(en());
```

Disclaimer: The issue was created using help of AI but I was able to confirming that reverting back to zod version 4.3.6 fixed the issue. When I tried writing a unit test without bundling I had always the english expected message locally, the self contained test which was created by the AI made it clear that it is indeed a combination of changed zod behaviour when using esbuild.

lihan3238:
## Summary

- Replaced `"sideEffects": false` with an explicit array listing `v4/classic/external.js` (and its CJS/MJS variants) as files with side effects
- This prevents esbuild and similar bundlers from tree-shaking the `config(en())` call that registers the English locale in `v4/classic/external.js`

## Problem

Since Zod 4.4.1, `"sideEffects": false` in `package.json` causes esbuild to tree-shake the locale initialization (`config(en())`), so all validation errors fall back to generic "Invalid input" instead of descriptive locale messages like `'Invalid option: expected one of "km"|"mi"'`.

## Fix

The `sideEffects` field now uses an array to explicitly mark only the files that contain module-level side effects (the locale registration), preserving tree-shaking benefits for all other files.

Closes #5953

### DIFF
--- a/packages/zod/package.json
+++ b/packages/zod/package.json
@@ -10,7 +10,11 @@
   "llmsFull": "https://zod.dev/llms-full.txt",
   "mcpServer": "https://mcp.inkeep.com/zod/mcp",
   "funding": "https://github.com/sponsors/colinhacks",
-  "sideEffects": false,
+  "sideEffects": [
+    "v4/classic/external.js",
+    "v4/classic/external.cjs",
+    "v4/classic/external.mjs"
+  ],
   "files": [
     "src",
     "**/*.js",


lihan3238:
Closing this PR. After further investigation, the fix doesn't fully resolve the issue — `write-stub-package-jsons.ts` writes `"sideEffects": false` into every subpath stub, overriding the main package.json. As noted by @pullfrog, the fix needs to target the stub generation script, not the top-level package.json.

PR #5955, opened 2026-05-05
fix: correct ESM/CJS declaration file resolution

### LINKED ISSUE #5686: Zod incorrectly points to CommonJS declaration when used with ESM

lihan3238:
## Summary

- Updated `exports` field in `package.json` to use nested `import`/`require` conditions with distinct type declarations
- ESM (`import`) now resolves to `.d.ts` type declarations; CJS (`require`) resolves to `.d.cts` type declarations
- Updated top-level `types` field from `./index.d.cts` to `./index.d.ts`

## Problem

Both CJS and ESM entry points resolved to the same `.d.cts` declaration file. When TypeScript compiles ESM projects with `module: "nodenext"`, the `.cts` extension causes TypeScript to assume CommonJS module semantics, producing incorrect type paths like `z.z.core.$strip` instead of `z.core.$strip`.

## Fix

Each export entry now uses nested conditional exports:

```json
{
  "import": {
    "types": "./index.d.ts",
    "default": "./index.js"
  },
  "require": {
    "types": "./index.d.cts",
    "default": "./index.cjs"
  }
}
```

This follows the TypeScript documentation recommendation for dual CJS/ESM packages.

Closes #5686

### DIFF
--- a/packages/zod/package.json
+++ b/packages/zod/package.json
@@ -29,7 +29,7 @@
     "inference"
   ],
   "main": "./index.cjs",
-  "types": "./index.d.cts",
+  "types": "./index.d.ts",
   "module": "./index.js",
   "zshy": {
     "exports": {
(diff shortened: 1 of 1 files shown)


lihan3238:
Closing this PR. The underlying zshy PR (colinhacks/zshy#70) was closed by the maintainer as "behaving as intended". Since Zod's exports are auto-generated by zshy, editing package.json directly would be overwritten on the next build. The correct approach would need to come from zshy first.

PR #5963, opened 2026-05-06
fix(v4): allow underscores in domain regex hostnames

### LINKED ISSUE #5960: `z.httpUrl()` doesn't allow underscores in hostname

```ts
const schema = z.object({
    url: z.httpUrl(),
});

// { success: true }

// { success: false }
```

dokson:
Closes #5960

The `domain` regex used by `z.httpUrl()` (via `hostname: regexes.domain`) rejected underscores in DNS labels, so URLs like `https://foo_bar.example.com` failed validation. RFC 3986 §2.3 puts `_` in `unreserved`, and WHATWG URL parses `https://foo_bar.example.com` cleanly — so this was a real validator gap, not a strictness choice.

This PR allows `_` inside non-TLD labels of `regexes.domain`, keeping the existing constraints that labels start and end with an alphanumeric and that the TLD is letters-only.

`regexes.hostname` is intentionally **not** changed: it follows the strict RFC 1035 LDH rule (used by `z.hostname()`), and an existing test asserts that `hostname.parse("example_com")` throws. The two regexes serve different purposes — URL host validation vs. strict DNS hostname — so the divergence is by design and now documented inline.

## Changes
- `regexes.domain`: `[a-zA-Z0-9-]` → `[a-zA-Z0-9_-]` in the label body.
- Comment in `regexes.ts` documenting why `hostname` and `domain` differ on `_`.
- Added `httpurl` test cases for underscore subdomains.

### DIFF
--- a/packages/zod/src/v4/classic/tests/string.test.ts
+++ b/packages/zod/src/v4/classic/tests/string.test.ts
@@ -441,6 +441,9 @@ test("httpurl", () => {
   // subdomains
   httpUrl.parse("https://sub.example.com");
   httpUrl.parse("http://sub.example.com");
+  // underscores in labels (uncommon but valid in DNS hostnames)
+  httpUrl.parse("https://foo_bar.example.com");
+  httpUrl.parse("https://exa_mple.com");
   // paths
   httpUrl.parse("https://example.com/path/to/resource");
   httpUrl.parse("http://example.com/path/to/resource");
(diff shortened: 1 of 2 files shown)


dokson:
- **Framing**: agreed, fixed in the description — the real justification is RFC 3986 / WHATWG URL host rules, not RFC 1035 / DKIM-ACME (which would also need leading-`_` support, out of scope here).
- **`hostname` vs `domain` divergence**: kept intentional. `regexes.hostname` is consumed by `z.hostname()` and there's an existing test (`hostname.parse("example_com")` expected to throw) that pins it to strict RFC 1035 LDH. Loosening it would silently broaden `z.hostname()` semantics for everyone. Added an inline comment in `regexes.ts` so the next reader sees why the two diverge.

colinhacks:
Closing in favor of #6130, which lands the identical regex change with better tests — it pins the boundary case (`http://_asdf.com` still rejected) and updates the copy of the regex inlined in the test file. Thanks for looking into this 👍

colinhacks:
Following up: #6130 is closed too, and so is #5960 — we're not allowing underscores in hostnames at all. RFC 952/1123 restrict host labels to letters, digits and hyphens, and no registry will issue a domain containing one. So this was declined on the merits, not passed over for another PR. Thanks again for looking into it 👍

PR #5978, opened 2026-05-11
test(v4): cover object union property inference

ShikShot:
## Summary

- Adds Zod 4 regression coverage for the object-property union inference case reported in #2654.
- Checks that the original `z.string().or(z.array(z.string()))` object property infers the same output type as the standalone union schema.
- Adds adjacent coverage for the array-or-record and enum-or-record cases reported in the issue thread.

## Validation

- `corepack pnpm vitest run packages/zod/src/v4/classic/tests/object.test.ts`
- Pre-push full suite: `339` test files, `3813` tests, no type errors

Refs #2654
/claim #2654


colinhacks:

PR #5985, opened 2026-05-12
fix: preserve date origin when number timestamp passed to .min()/.max()

### LINKED ISSUE #5980: Passing a number to ZodDate min() and max() functions causes the issue origin to become "number" (zod@4.4.3)
Probably a bug:

```ts
import { z } from 'zod'

z.config({
  customError: (issue) => {
    console.log(issue)
  },
})

// {
//   "origin": "date",
//   "code": "too_small",
//   "inclusive": true,
//   "inst": {},
//   "continue": true
// }

// {
//   "success": true,
// }



// {
//   "origin": "number",
//   "code": "too_small",
//   "inclusive": true,
//   "inst": {},
//   "continue": true
// }

// {
//   "success": true,
// }
```

```
// {
//   "origin": "string",
//   "code": "too_small",
//   "minimum": 5,
//   "inclusive": true,
//   "input": "x",
//   "inst": {},
//   "continue": true
// }
```

o19k:
Fixes #5980.

`z.date().min(timestamp)` and `z.date().max(timestamp)` produce `origin: "number"` in the validation issue when the argument is a number. The origin should always be `"date"` since the schema is a `ZodDate`.

Root cause: `$ZodCheckGreaterThan` / `$ZodCheckLessThan` derive `origin` from `typeof def.value` via `numericOriginMap`. When a `Date` is passed, `typeof` is `"object"` → `"date"`. When a number timestamp is passed, `typeof` is `"number"` → `"number"`.

Fix: normalize the value to a `Date` in `ZodDate.min` / `ZodDate.max` before passing to `checks.gte` / `checks.lte`. This also brings the runtime behavior in line with `$ZodDateInternals.bag`, which already types `minimum` and `maximum` as `Date`.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1365,8 +1365,8 @@ export const ZodDate: core.$constructor<ZodDate> = /*@__PURE__*/ core.$construct
   ZodType.init(inst, def);
   inst._zod.processJSONSchema = (ctx, json, params) => processors.dateProcessor(inst, ctx, json, params);
 
-  inst.min = (value, params) => inst.check(checks.gte(value, params));
-  inst.max = (value, params) => inst.check(checks.lte(value, params));
+  inst.min = (value, params) => inst.check(checks.gte(value instanceof Date ? value : new Date(value), params));
+  inst.max = (value, params) => inst.check(checks.lte(value instanceof Date ? value : new Date(value), params));
 
   const c = inst._zod.bag;
   inst.minDate = c.minimum ? new Date(c.minimum) : null;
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing in favor of #6129, which fixes this in `$ZodCheckGreaterThan` / `$ZodCheckLessThan` rather than in `ZodDate`. Normalizing the bound works for classic, but `ZodMiniDate` has no `.min()` / `.max()` — mini users write `z.date().check(z.gte(timestamp))`, hit the same check, and still get `origin: "number"`. This also flips `_zod.bag.minimum` from a number to a `Date` at runtime, which I'd rather not fold into a bug fix.

Thanks for digging into this 👍

PR #5986, opened 2026-05-12
feat(proposal): mask

holoflash:
## Background

I was trying to find a way to dynamically replace sensitive fields from a server response during schema parsing and ended up building a small utility in a Typescript repo I work on that uses `.meta({ mask })` to tag fields, and then a `withAutoMasking()` wrapper that walks the schema and replaces tagged values on the fly. It supports static values, deterministic array picks (hashed by entity id), and function-based masks. It handles nested objects, arrays, unions, optionals, pipes, etc. The main benefit of this was being able to use production data with some sensitive values replaced dynamically in contexts where it shouldn't be exposed and not have to fiddle with mock data.

I'm really happy I was able to leverage zod in this way and it's made my life a little easier so I figured I'd try to implement it natively in Zod, where it can benefit from proper type safety and a cleaner API.
Consider this a mere suggestion and a wish. No hard feelings if there's no interest in adding such functionality at this point.

## What this PR adds

A `.mask()` method to mark fields for PII replacement and `parseAndMask`/`safeParseAndMask` (+ async variants) that validate and redact in one step:

```ts
const User = z.object({
  id: z.string(),
  name: z.string().mask("REDACTED"),
  age: z.number().mask(0),
  role: z.string().mask(["viewer", "editor"]),
  email: z.string().mask((id) => `${id}@example.com`),
});

User.parseAndMask({
  id: "usr_1",
  name: "Alice",
  age: 28,
  email: "alice@real.com",
  role: "admin",
});
// => { id: "usr_1", name: "REDACTED", age: 0, role: "viewer", email: "usr_1@example.com" }
```

### Mask types

`.mask()` is typed against the schema output, so `z.string().mask(123)` won't compile.

- A value: always used as the replacement
- An array: one is picked deterministically per entity
- A function: receives the entity's id and returns the replacement

### Deterministic masking

Array and function masks need a stable identifier so the same record always produces the same fake data. When masking an object, this identifier is resolved automatically:

1. If the object has a string `id` field, its value is used
2. Otherwise, a hash of the object's masked field values is used as a fallback

### Schema support

The walker handles nested objects, arrays, optionals, nullables, defaults, readonly, unions, discriminated unions, pipes, and lazy schemas. Null/undefined always pass through unmasked.

### Files changed

- `core/mask.ts` (new): recursive schema walker, ~100 LOC
- `core/api.ts`: `_mask` check factory, stores config in `_zod.bag`, no-op at parse time
- `core/parse.ts`: `parseAndMask` / `safeParseAndMask` + async variants
- `classic/schemas.ts`: `.mask()` method on `ZodType` via `_installLazyMethods`
- `classic/parse.ts`: re-exports for classic API
- `core/index.ts`: exports
- `classic/tests/mask.test.ts` (new): 38 tests
- `api.mdx`: docs section

_Disclaimer: The code in this PR was written with a lot of agentic assistance based on my original implementation. Consider this as a conversation starter. Sorry, for skipping the issue. Felt like this context may sell the idea better._

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -3256,7 +3256,7 @@ schema.parseAndMask({ id: "usr_1", name: "Other Name" });
 
 
 <Callout>
-  The deterministic pick algorithm for array masks is stable within a major version. Test fixtures built against `mask([...])` output will not break across patch or minor releases.
+  The deterministic pick algorithm for array masks is stable within a major version — exact picks for known seeds are pinned by fixture tests. Test fixtures built against `mask([...])` output will not break across patch or minor releases.
 </Callout>
 
 ### `.parseAndMask()` / `.safeParseAndMask()`
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -3172,3 +3172,113 @@ z.parse(schema, null); // => null
 ```
 </Tab>
 </Tabs>
+
+## Masking
+
+Use `.mask()` to mark schema fields for PII masking. Then use `.parseAndMask()` instead of `.parse()` to validate data *and* replace masked fields in a single step.
+
+<Tabs groupId="lib" items={["Zod", "Zod Mini"]}>
+<Tab value="Zod">
+```ts
+const User = z.object({
+  id: z.string(),
+  name: z.string().mask("REDACTED"),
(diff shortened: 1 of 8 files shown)


holoflash:
Moving this to an external plugin as suggested.
https://github.com/holoflash/zod-redact

PR #5987, opened 2026-05-13
Add regression tests for object union inference

treydoe1:
/claim #2654

Adds regression coverage for the reported object property inference case where a schema created with `.or()` is placed inside `z.object(...)`.

This covers both `zod/v3` and `zod/v4`, asserting that the inferred object property type matches the standalone schema inference for `string | string[]`.


colinhacks:

PR #3352, opened 2024-03-22
Improve `addQuestionMarks`, fix #2184

colinhacks:
Okay, this is a long time coming. Getting the inferred types to work as expected in generic contexts is quite annoying (see `generics.test.ts` for examples). Some changes I made to that effect seem to interact unhappily with the very specific case described in #2184 - a `ZodObject` schema, with a key corresponding to a union, being cast to `ZodType`.

The change now makes all my tests happy, including assignability in generic contexts and the newly added XOR test.

I'm going to ship this in the next minor since I haven't found a scenario where this breaks or even changes the inferred type signature in non-generic contexts. In generic contexts, it seems to strictly improve assignability characteristics without a big performance hit in the compiler.

cc @chrishoermann

PR #5994, opened 2026-05-14
fix: Schema in object being inferred differently (and weirdly) (#2654)

IbrahimLaeeq:
Fixes #2654.

/claim #2654

Fix verified against the repository's own test suite.

---
*Verified against the repository's own test suite before submission.*


colinhacks:
Closing this out. The diff adds a test but changes no source, and that test already passes on `main` — the inference bug in #2654 was fixed in 3.23.0 by #3352, which replaced the `Pick<Required<T>, R> & Partial<T>` shape that caused it. Nothing left to fix here. Thanks for looking into it 👍

PR #6003, opened 2026-05-16
Add regression test for object union inference

haocyan0723-code:
This adds a focused regression test for the object-property union inference case from #2654. The current v4 type inference already produces the same type for a standalone union schema and the same union inside an object property, so this keeps that behavior covered.\n\nValidation:\n- Ran a standalone TypeScript check against latest zod with the original repro shape.\n\n/claim #2654


haocyan0723-code:
Closing this after review: current main already has equivalent coverage for this inference case, so this PR would not add useful signal for maintainers. Thanks for the quick review.

PR #6010, opened 2026-05-20
test(v3): pin union-in-object inference (issue #2654)

pmjnt:
## Summary

Add v3 regression test locking the inferred type for a union nested in z.object so the original #2654 bug can't silently return.

## Closes

Fixes Algora bounty: https://github.com/colinhacks/zod/issues/2654

## Disclosure

This PR was AI-assisted (Claude Code) and reviewed by a human before submission.


colinhacks:

PR #6012, opened 2026-05-20
chore: drop dead code in core and fix a stale deprecation message

Mohammad-Faiz-Cloud-Engineer:
Three cleanups from a source read. Merged main in and cut this down to what still applies — the `isTransforming` catch fix landed in #6409 with a regression test, and the duplicate `inst.guid` went away when `classic/schemas.ts` moved to an object literal.

- `core/config.ts` was orphaned. Nothing imported it, and the live `config`/`globalConfig` pair in `core.ts` carries the `globalThis` singleton that keeps config shared across the CJS and ESM builds. It was never compiled into the build either.
- `_stringFormat` spread `normalizeParams` into the same def twice. This drops the leading spread, so params keep overriding the internal keys the way they do in `_email`, `_uuid` and `_url`.
- The `ZodFirstPartySchemaTypes` deprecation pointed at `$$ZodFirstPartyTypes`, which exists nowhere in the source. It now names `$ZodTypes`.


colinhacks:
Merged main in and cut this down to the three fixes that still apply.

The `isTransforming` catch fix landed in #6409 back in August with a regression test, and the duplicate `inst.guid` went away when that file moved to an object literal. Dropped the `$ZodReadonly` change as well — the optional chaining there isn't an inconsistency, `$ZodLazy` has the same thing and both came from #5094.

Kept the deprecation typo, the orphaned config module, and the `_stringFormat` spread, though that last one now drops the leading spread rather than the trailing one so params keep overriding the internal keys like they do everywhere else in that file. Merging.

Thanks for digging through the source 👍

PR #6013, opened 2026-05-21
fix(v4): preserve discriminatedUnion inference for recursive getter options

### LINKED ISSUE #5991: discriminatedUnion breaks TS inference for mutually-recursive getter-based schemas; plain union works
## Versions

- zod: `4.3.6`
- typescript: `5.9.3`
- node: `22.22.1`

## Summary

The v4 getter pattern correctly handles mutual recursion when variants are combined with `z.union`, but **fails** with `z.discriminatedUnion` — TS emits `ts(7022)` / `ts(2615)` and the resulting type resolves to `any`.

The User/Post recursive example from the docs works because there's no union. Once a union is introduced over variants that recurse back through it via getter, only `z.union` produces a usable inferred type; `z.discriminatedUnion` does not.

### ✅ Works — plain `z.union`

```ts
import { z } from 'zod';

const variantA = z.object({
    kind: z.literal('a'),
    get child() { return tree; }
});

const variantB = z.object({
    kind: z.literal('b'),
    get sibling() { return tree; }
});

const tree = z.union([variantA, variantB]);

type Tree = z.input<typeof tree>;
//   ^? { kind: 'a'; child: ... | { kind: 'b'; sibling: ... } }
//      | { kind: 'b'; sibling: { kind: 'a'; child: ... } | ... }
```

Type is fully typed and recursive. ✓

### ❌ Broken — `z.discriminatedUnion`

```ts
import { z } from 'zod';

const variantA = z.object({
    kind: z.literal('a'),
    get child() { return tree; }
});

const variantB = z.object({
    kind: z.literal('b'),
    get sibling() { return tree; }
});

const tree = z.discriminatedUnion('kind', [variantA, variantB]);

type Tree = z.input<typeof tree>;
//   ^? any
```

Diagnostics:

- `ts(7022)`: `'tree' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.`
- `ts(2615)`: `Type of property 'child' circularly references itself in mapped type ...`
- `ts(2615)`: `Type of property 'sibling' circularly references itself in mapped type ...`

Adding explicit getter return type annotations (`get child(): typeof tree`) doesn't help — TS still reports `ts(2502)` (`'child' is referenced directly or indirectly in its own type annotation.`).

## Expected behavior

`z.discriminatedUnion` should infer the same recursive shape as `z.union` does.

## Likely cause

`ZodDiscriminatedUnion`'s type computation eagerly maps over variants to build the discriminator → branch lookup, requiring every variant's shape to be fully resolved before the union's type. Plain `ZodUnion` is lazier — it just unions the variants' inferred types, which TS handles through normal lazy expansion.

## Workaround

For now, use `z.union` instead of `z.discriminatedUnion` for recursive schemas, accepting the runtime parse-perf loss (no discriminator short-circuit), or fall back to v3's `z.lazy` + `z.ZodType<X>` cast pattern with a hand-written recursive type.

ATOM00blue:
## Summary
`z.discriminatedUnion` collapsed TypeScript inference to `any` (emitting ts(7022)/ts(2615)) for mutually-recursive, getter-based object schemas that recurse back through the union, while the equivalent `z.union` inferred correctly.

## Root cause
The option constraint `$ZodTypeDiscriminable` was based on `$ZodType` and required the discriminator on each option's resolved `input` (`{ [K in Disc]?: unknown }`). Both force the option's inferred input to be evaluated while the call's argument constraint is checked. For schemas whose input references the union itself this is circular, so TS reports ts(2615)/ts(7022) and falls back to `any`. `z.union` does not constrain option input in argument position, so it stays lazy and infers fine.

## Fix
`$ZodTypeDiscriminable<Disc>` is now based on `SomeType` and, for object options, checks the discriminator against `def.shape` rather than the resolved input. Non-object options (pipes/transforms, which have no `shape`) still validate via `input`, preserving prior behavior. The three factory bodies cast `options` exactly like the existing `union`/`xor` factories.

Preserved guarantees (covered by existing + new type tests): an object option missing the discriminator is still a type error; a pipe option whose object lacks the discriminator is still a type error; an empty options list is still rejected; non-recursive and nested discriminated unions infer unchanged.

## Test plan
- New type + runtime tests for the recursive getter case (classic + mini); both fail without the fix
- `pnpm test` — 339 files / 3815 tests pass, no type errors
- `tsc -p tsconfig.test.json --noEmit` clean

Closes #5991

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1692,7 +1692,7 @@ export function discriminatedUnion<
   // const [options, params] = args;
   return new ZodDiscriminatedUnion({
     type: "union",
-    options,
+    options: options as any as core.$ZodType[],
     discriminator,
     ...util.normalizeParams(params),
   }) as any;
--- a/packages/zod/src/v4/classic/tests/recursive-types.test.ts
+++ b/packages/zod/src/v4/classic/tests/recursive-types.test.ts
@@ -629,3 +629,51 @@ test("recursive type with `id` meta", () => {
     },
   });
 });
+
+test("mutual recursion through discriminatedUnion getter", () => {
+  // Mutually-recursive getter variants that recurse back through the union.
+  // Previously z.discriminatedUnion broke inference here (ts7022/ts2615 -> any),
+  // while the equivalent z.union worked. Both should now infer the same type.
+  // Children are optional so the recursion has a terminating runtime value.
+  const variantA = z.object({
+    kind: z.literal("a"),
+    get child() {
+      return tree.optional();
+    },
+  });
+
+  const variantB = z.object({
+    kind: z.literal("b"),
+    get sibling() {
+      return tree.optional();
+    },
+  });
+
+  const tree = z.discriminatedUnion("kind", [variantA, variantB]);
+  type Tree = z.input<typeof tree>;
+
+  type _Tree =
+    | {
+        kind: "a";
+        child?: _Tree | undefined;
+      }
+    | {
+        kind: "b";
+        sibling?: _Tree | undefined;
+      };
+
+  // The inferred type must be the recursive union shape, not `any`.
+  expectTypeOf<Tree>().toEqualTypeOf<_Tree>();
+  expectTypeOf<Tree>().not.toBeAny();
+
+  // discriminatedUnion and the union equivalent must infer the same type.
+  const treeUnion = z.union([variantA, variantB]);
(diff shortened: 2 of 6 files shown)


colinhacks:
Closing — this landed in #6422. Your diagnosis is what the fix was built on, and both of your tests shipped with it; the final version drops the compile-time discriminator check entirely and moves the guarantee to construction time, which also covers recursive pipe options. Thanks for digging into this one 👍

Issue #5377, opened 2025-10-23
Reparse against the output of a Schema (codecs)


Sparticuz:
It would be helpful to provide a way to re-parse the output of a schema, specifically when working with Codecs.

I pull data from my database and encode it for use in my (react-hook-) form. The user can then edit data as needed in the form, then submits the form, which gets validated using the zodResolver. If successful, the data is passed back to the server, where it's parsed again and committed to the database.

The biggest issue with working with html forms has been the difficulty with Numbers, so I wanted a codec that would encode the number to a string, then decode it back to an int (or float in places). I got encode working great, it's passing strings to my client. However, when the form submits, it submits the parsed data, which is a number. (I don't think it's that great of an idea to submit the non-parsed data). When I try to verify the data that was sent from the client, because it's already been parsed, the field is a number already and is failing with 'expected string'.

I would like to request a way to parse against the output of the codec.

```ts
const schema = z.object({
  num: z.codec(z.string().regex(z.regexes.integer), z.int(), {
    decode: (str) => Number.parseInt(str, 10),
    encode: (num) => num.toString(),
  })
});
```
should be able to reparse and pass with the data satisfying the second parameter of z.codec.

The workarounds that I see is `const reparseSchema = schema.omit({ num: true }).extend({ num: z.int() });`
However, this requires me to have the schema defined twice.

My request:

```ts
await schema.reparseAsync({ num:42 })
```

Related to #5081 but that was closed before codecs were a thing

Sparticuz:
I believe that all the discussion on re-parsing schema's has been pre-codec and I think that codec's might change the conversation surrounding this. A reparse function should be able to grab the correct output functions from the original schema without the need for any manual schema management.

akomm:
I think it might not be obvious, but you have two schemas anyway. What you try not to do is to "not" inline them into one, so avoid creating two definitions, even though you create them anyway. By doing that you couple some representation of a serialization (through form) to your "natural" data/model. What that means is, that you needlessly always parse one schema, fail then parse another. And its still not enough, because you would have to somehow be able to transform back and forth or carry input/output data. The point is, what you try to avoid can't be avoided, because its natural. You can only make it worse by trying to fuse it in one place, which will start obstacle course of edge-cases zod would have to implement and handle and which would be inefficient in performance.

[...]

PR #6017, opened 2026-05-22
feat(v4): add toOutputSchema()

### LINKED ISSUE #6016: Proposal: `toOutputSchema()` — get a plain schema for the output side of a composed codec schema
## Problem

Schemas composed from codecs have distinct input and output types. At the type level, `z.input<>` and `z.output<>` give you both sides. But at the value level, there's no way to get a **plain schema object** that validates the output shape — one where `.parse()` and `.encode()` are identical (no transformation).

My use case involves [nestjs-zod](https://github.com/BenLorantfy/nestjs-zod) and response validation with `ZodSerializerDto`. nestjs-zod can operate in two modes: by default it calls `.parse()` on the response, or with `codec: true` it calls `.encode()`. Neither works for my case.

Here's the setup. I have a codec-based schema describing data that arrives from an external service in encoded form:

```ts
const isoDate = z.codec(z.iso.datetime(), z.date(), {
  decode: (s) => new Date(s),
  encode: (d) => d.toISOString(),
});

const UserSchema = z.object({
  name: z.string(),
  createdAt: isoDate,
  updatedAt: isoDate.optional(),
});
```

In my NestJS controller, I fetch data from the external service, decode it, and then want to return the decoded form as the response. I want response validation to confirm the object matches the output shape before sending it:

```ts
class UserResponseDto extends createZodDto(UserSchema) {}
// or: createZodDto(UserSchema, { codec: true })

@Controller('users')
export class UserController {
  @Get(':id')
  @ZodSerializerDto(UserResponseDto)
  async getUser(@Param('id') id: string) {
    const raw = await this.externalService.getUser(id);  // returns encoded form
    const user = UserSchema.decode(raw);                  // decode to output form

    // At this point, `user` is { name: string, createdAt: Date, ... }
    // I want to validate and return it as-is.
    return user;
  }
}
```

The problem is that `ZodSerializerDto` will either:
- Call `.parse(user)` (default) — fails because `.parse()` expects the input types (ISO string, not `Date`)
- Call `.encode(user)` (`codec: true`) — transforms the data back to the input representation (Date → ISO string), which isn't what I want — I want to return the decoded object

What I actually need is a plain schema for the output shape — one I can pass to `ZodSerializerDto` that will validate `{ name: string, createdAt: Date, ... }` without transforming it.

For a single codec this is trivial — just use the output schema directly. But when codecs are spread across a larger composed schema, manual reconstruction duplicates the shape and defeats the purpose of composition:

```ts
const UserSchema = z.object({
  name: z.string(),
  createdAt: isoDate,
  updatedAt: isoDate.optional(),
  metadata: jsonMetadata,
  scores: z.array(z.object({
    category: z.string(),
    value: z.codec(z.string(), z.number(), {
      decode: (s) => Number(s),
      encode: (n) => String(n),
    }),
  })),
});

// ❌ No way to get a plain schema for the output shape of UserSchema
// without manually rebuilding the entire structure
```

`toJSONSchema(schema, { io: "output" })` already walks the tree resolving codecs to their output schemas, but produces JSON Schema — not a zod schema.

This is closely related to #5377 (and draft PR #5704), which describe the same underlying problem — needing to validate already-decoded data against the output side of a composed codec schema. The difference is that instead of a `reparse()` method as proposed there, I think a `toOutputSchema()` function makes more sense: it gives you a schema object you can pass to any framework, not just a one-off parse call.

## Proposed solution

A standalone `toOutputSchema(schema)` function that recursively walks a composed schema and resolves every pipe/codec to its output schema, returning a new plain zod schema (no codec transformations):

```ts
import { toOutputSchema } from "zod/v4";

const UserOutputSchema = toOutputSchema(UserSchema);
// A plain schema where:
//   createdAt expects Date (not ISO string)
//   scores[].value expects number (not string)
//   all non-codec fields are unchanged
//   .parse() and .encode() are identical — no transformations

class UserResponseDto extends createZodDto(UserOutputSchema) {}
```

The function recursively walks the schema tree, resolving each codec to its output schema. Non-codec parts of the tree are left unchanged — if there are no codecs in the subtree, the same schema instance is returned.

## Questions

Would a standalone `toOutputSchema()` function (following the pattern of `toJSONSchema()`) make sense as a built-in feature?

ensconced:
## Summary

- Adds a standalone `toOutputSchema(schema)` function that recursively walks a composed zod schema and resolves every pipe/codec to its `.out` schema, returning a plain zod schema for the output (decoded) shape
- Exported from core, classic, and mini layers (standalone function only — follows the `toJSONSchema` pattern)
- Also adds `.outputSchema()` instance method on `ZodType` in the classic layer (matching `inst.toJSONSchema`)

Closes #6016. Also addresses the use case in #5377 and draft PR #5704.

## Test plan

- [x] New test file `packages/zod/src/v4/classic/tests/output-schema.test.ts` with 10 test cases:
  - Leaf schemas return same instance
  - Simple codec resolved to `.out` schema
  - Object with codec fields validates decoded shape
  - Array of objects with codecs
  - Discriminated union with codec fields
  - Optional/nullable codec fields
  - Identity optimisation (no codecs → same instance)
  - Deeply nested: codec inside optional inside array inside object
  - `z.output<>` type inference on result
  - Standalone `z.toOutputSchema()` matches instance `.outputSchema()`
- [x] All existing classic tests pass (160 files, 2204 tests, 0 type errors)
- [x] All existing mini tests pass (30 files, 340 tests, 0 type errors)

### DIFF
--- a/packages/zod/src/v4/classic/external.ts
+++ b/packages/zod/src/v4/classic/external.ts
@@ -31,6 +31,7 @@ export {
   NEVER,
 } from "../core/index.js";
 export { toJSONSchema } from "../core/json-schema-processors.js";
+export { toOutputSchema } from "../core/to-output-schema.js";
 export { fromJSONSchema } from "./from-json-schema.js";
 
 export * as locales from "../locales/index.js";
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -3,6 +3,7 @@ import { util } from "../core/index.js";
 import * as processors from "../core/json-schema-processors.js";
 import type { StandardSchemaWithJSONProps } from "../core/standard-schema.js";
 import { createStandardJSONSchemaMethod, createToJSONSchemaMethod } from "../core/to-json-schema.js";
+import { toOutputSchema } from "../core/to-output-schema.js";
 
 import * as checks from "./checks.js";
 import * as iso from "./iso.js";
@@ -147,6 +148,11 @@ export interface ZodType<
     params?: core.ParseContext<core.$ZodIssue>
   ): Promise<parse.ZodSafeParseResult<core.output<this>>>;
 
+  // output schema
+  /** Returns a new schema that validates the *output* (decoded) shape,
+   *  with all pipes/codecs resolved to their `.out` schemas. */
+  outputSchema(): this;
+
   // refinements
   refine<Ch extends (arg: core.output<this>) => unknown | Promise<unknown>>(
     check: Ch,
@@ -245,6 +251,7 @@ export const ZodType: core.$constructor<ZodType> = /*@__PURE__*/ core.$construct
   inst.safeDecode = (data, params) => parse.safeDecode(inst, data, params);
   inst.safeEncodeAsync = async (data, params) => parse.safeEncodeAsync(inst, data, params);
   inst.safeDecodeAsync = async (data, params) => parse.safeDecodeAsync(inst, data, params);
+  inst.outputSchema = () => toOutputSchema(inst) as any;
 
   // All builder methods are placed on the internal prototype as lazy-bind
   // getters. On first access per-instance, a bound thunk is allocated and
--- a/packages/zod/src/v4/classic/tests/output-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/output-schema.test.ts
@@ -1,0 +1,135 @@
+import { expect, expectTypeOf, test } from "vitest";
+import * as z from "zod/v4";
+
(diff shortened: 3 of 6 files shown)


colinhacks:
Closing this out — `z.output()` in #5928 covers it. Same projection, and it returns `ZodType<output<T>, output<T>>`, so `.parse()` and `.encode()` agree on the result, which is what #6016 asked for.

Your review caught something real though: checks attached to a pipe were getting dropped by the projection. That applied to `z.output()` as well, and it's fixed now. Thanks for the writeup 👍

PR #6020, opened 2026-05-22
fix: json schema tuple rest compliance

mneetika:
### Summary
Fixes JSON Schema to Zod v4 translation compliance regarding positional tuples with trailing elements. 

By default, both Draft-7 and Draft-2020-12 specifications allow open-ended arrays when positional items are specified, unless explicitly locked down by `additionalItems: false` or `items: false`. The parser previously defaulted to generating a strict Zod tuple that rejected trailing elements when these configuration properties were omitted or evaluated to booleans.

### Root Cause
In `packages/zod/src/v4/classic/from-json-schema.ts`, the logic transforming `prefixItems` (Draft-2020-12) and `items` arrays (Draft-7) expected a nested schema object to evaluate `.rest()`. It did not account for:
1. Implicitly allowing any trailing elements when configuration attributes are completely missing (`undefined`).
2. Expressly handling JSON Schema boolean parameters (`true` / `false`) for array continuation rules.

### Changes
* Introduced `getTupleRest` helper to evaluate boolean schema constraints and map `undefined`/`true` conditions to `.rest(z.any())`.
* Unified the rest parameter assignment across Draft-7 (`additionalItems`) and Draft-2020-12 (`items`) compilation branches.
* Appended 5 regression test suites inside `from-json-schema.test.ts` to cover boolean constraints and default open-ended schemas.

#### 1. Input Schema (Draft-2020-12 Open Tuple)

```JSON
{
  "$schema": "[https://json-schema.org/draft/2020-12/schema](https://json-schema.org/draft/2020-12/schema)",
  "type": "array",
  "prefixItems": [{ "type": "string" }]
}
```

##### Before This PR
Generated a rigid tuple matching only the explicit prefix length. Passing trailing elements threw a validation error.

```TypeScript
// Parsed Output
z.tuple([z.string()]); 

schema.parse(["hello", 42]); // ❌ Throws ZodError (unexpected element)
```

##### After This PR
Correctly attaches an open-ended fallback rest layer to match specification defaults

```TypeScript
// Parsed Output
z.tuple([z.string()]).rest(z.any()); 

schema.parse(["hello", 42]); // ✅ Returns ["hello", 42]
```

#### 2. Input Schema (Draft-2020-12 Closed Tuple)

```JSON
{
  "$schema": "[https://json-schema.org/draft/2020-12/schema](https://json-schema.org/draft/2020-12/schema)",
  "type": "array",
  "prefixItems": [{ "type": "string" }],
  "items": false
}
```

#### After This PR
Correctly interprets false to suppress .rest() injection, maintaining strict length boundaries.

```TypeScript
// Parsed Output
z.tuple([z.string()]); 

schema.parse(["hello"]);     // ✅ Returns ["hello"]
schema.parse(["hello", 42]); // ❌ Throws ZodError
```

---

### ⚠️ Behavioral Note for Existing Users
This PR introduces a behavioral change for existing `fromJSONSchema` invocations where a tuple schema relies on implicit defaults for trailing elements (e.g., `{ prefixItems: [{ type: "string" }] }`). 

* **Previous Behavior:** Generated a closed tuple that strictly rejected trailing elements.
* **New Behavior:** Generates an open tuple (`.rest(z.any())`) that accepts trailing elements, fully conforming to Draft-7 and Draft-2020-12 specifications. 

Because `fromJSONSchema` is an experimental feature, this alignment does not break stable API surfaces but is surfaced here for changelog tracking.

### DIFF
--- a/packages/docs/content/json-schema.mdx
+++ b/packages/docs/content/json-schema.mdx
@@ -18,6 +18,10 @@ import { Accordion, Accordions } from 'fumadocs-ui/components/accordion';
   **Experimental** — The `z.fromJSONSchema()` function is experimental and is not considered part of Zod's stable API. It is likely to undergo implementation changes in future releases.
 </Callout>
 
+<Callout type="info">
+  **Note on Tuples:** Per Draft-07 and Draft-2020-12 specifications, positional tuples generated via `prefixItems` or `items[]` arrays are open-ended by default and will append `.rest(z.any())`. To enforce a strict, closed tuple boundary, you must explicitly declare `items: false` or `additionalItems: false` within your source JSON Schema.
+</Callout>
+
 Zod provides `z.fromJSONSchema()` to convert a JSON Schema into a Zod schema.
 
 ```ts
--- a/packages/docs/content/json-schema.mdx
+++ b/packages/docs/content/json-schema.mdx
@@ -18,10 +18,6 @@ import { Accordion, Accordions } from 'fumadocs-ui/components/accordion';
   **Experimental** — The `z.fromJSONSchema()` function is experimental and is not considered part of Zod's stable API. It is likely to undergo implementation changes in future releases.
 </Callout>
 
-<Callout type="info">
-  **Note on Tuples:** Per Draft-07 and Draft-2020-12 specifications, positional tuples generated via `prefixItems` or `items[]` arrays are open-ended by default and will append `.rest(z.any())`. To enforce a strict, closed tuple boundary, you must explicitly declare `items: false` or `additionalItems: false` within your source JSON Schema.
-</Callout>
-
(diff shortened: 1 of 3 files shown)


colinhacks:
Rebased this onto #6201, which just landed and makes positional elements optional up to `minItems` — the two changes had to be resolved together, since both rewrite the same `rest` derivation. Also pushed a commit on top: it pins the boolean `items: true` / `additionalItems: true` cases, which `getTupleRest` handles but nothing covered, and drops the docs callout, since spec-default behavior doesn't need one.

`getTupleRest` is the right shape — collapsing `false` / `undefined` / `true` / schema-object into one place is clearer than the nested ternaries this replaces. Merging.

PR #6021, opened 2026-05-23
fix(v4): detect async results when Promise is patched

### LINKED ISSUE #6019: `superRefine(async ...)` silently drops `ctx.addIssue` when `globalThis.Promise` is patched (Zone.js / Angular NgZone / `@opentelemetry/context-zone`)
## Human Notes

## Description

## Minimum reproduction

```js
// repro.mjs
import { z } from "zod";

console.log(
);

const schema = z
    ctx.addIssue({
      code: "custom",
      message: "Always fails",
      path: ["email"],
    });
  });

```

```json
// package.json
{
  "name": "repro",
  "private": true,
  "type": "module",
  "dependencies": {
    "zod": "4.4.3",
    "zone.js": "0.16.2"
  }
}
```

### Expected output

```
result.success: false
```

### Actual output

```
result.success: true
```

## Root cause

```js
}
    await _;
  });
}
```

## Suggested fix

```js

// …
}
    await _;
    // …
  });
}
```

## Affected environments

puneetdixit200:
Fixes #6019.

This replaces v4 async detection with a shared thenable check so async refinements, transforms, and nested parser results are still awaited when libraries such as Zone.js replace `globalThis.Promise`. I added a focused regression test that patches `globalThis.Promise` and verifies `superRefine(async ...)` issues are returned instead of being dropped.

Validated with the focused regression test, the full `packages/zod/src/v4` Vitest tree, `pnpm lint:check`, `pnpm build`, `git diff --check`, and the full pre-push `pnpm test` suite.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -2096,7 +2096,7 @@ export const ZodTransform: core.$constructor<ZodTransform> = /*@__PURE__*/ core.
       };
 
       const output = def.transform(payload.value, payload);
-      if (output instanceof Promise) {
+      if (util.isPromise(output)) {
         return output.then((output) => {
           payload.value = output;
           payload.fallback = true;
--- a/packages/zod/src/v4/classic/tests/refine.test.ts
+++ b/packages/zod/src/v4/classic/tests/refine.test.ts
@@ -105,6 +105,47 @@ describe("async refinements", () => {
       })
     ).rejects.toThrow();
   });
+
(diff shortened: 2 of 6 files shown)


puneetdixit200:
Pushed b3b0b37c with a test-only follow-up covering patched-Promise async handling across transform, pipe, array, tuple, record, union, and intersection paths. Verification: focused refine.test Vitest run, Biome check for the touched file, git diff --check, no instanceof Promise matches under packages/zod/src/v4, and the full pre-push test suite passed.

PR #6022, opened 2026-05-23
fix(from-json-schema): drop redundant inclusive bound for draft-04 exclusive ranges

francisjohnjohnston-web:
## What

`fromJSONSchema` converts JSON Schema draft-04 boolean exclusive bounds correctly in *behaviour* but
emits a redundant, contradictory second check. For `{ type: "number", minimum: 5, exclusiveMinimum: true }`
the converter applied **both** `.min(5)` (inclusive, `>= 5`) **and** `.gt(5)` (exclusive, `> 5`), leaving
the resulting schema with two number-range checks where one is dominated by the other. Same for
`maximum` / `exclusiveMaximum: true`.

This change skips the inclusive `.min()` / `.max()` when the sibling exclusive flag is the boolean `true`,
so only the exclusive bound is emitted.

## Why

In draft-04, when `exclusiveMinimum` is the boolean `true`, the sibling `minimum` *becomes* the exclusive
bound — it is not an independent inclusive constraint. Emitting `.min()` as well produces a schema whose
internal check list carries a duplicate, weaker bound. Validation output is unaffected (the exclusive
check binds), so the duplication is invisible at the `.parse()` level and only shows up when inspecting
`schema._zod.def.checks` or round-tripping the schema — which is why it is easy to miss in review.

## Safety / blast radius

- Internal converter behaviour only; no public API, signature, or exported-type change.
- Behaviour-preserving for all valid inputs (the exclusive bound was already the binding constraint).
- The **draft-07 numeric form** (`{ minimum: 3, exclusiveMinimum: 5 }`, where the two are legitimately
  distinct independent bounds) is deliberately untouched — the guard only triggers on `=== true`.

## Testing

- Added `draft-04 boolean exclusive bounds do not emit a redundant inclusive bound` to
  `from-json-schema.test.ts`, asserting both the validation behaviour (boundaries rejected, interior
  accepted) and the structure (exactly `[greater_than 5 exclusive, less_than 10 exclusive]`).
- Full file suite green from repo root: **146/146 tests pass, 0 type errors** (`vitest run`).

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -336,10 +336,12 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
       let numberSchema: ZodNumber = type === "integer" ? z.number().int() : z.number();
 
       // Apply constraints
-      if (typeof schema.minimum === "number") {
+      // In draft-04, `exclusiveMinimum: true` makes the sibling `minimum` exclusive, so the
+      // inclusive `.min()` must be skipped to avoid emitting a contradictory second bound.
+      if (typeof schema.minimum === "number" && schema.exclusiveMinimum !== true) {
         numberSchema = numberSchema.min(schema.minimum);
       }
-      if (typeof schema.maximum === "number") {
+      if (typeof schema.maximum === "number" && schema.exclusiveMaximum !== true) {
         numberSchema = numberSchema.max(schema.maximum);
       }
       if (typeof schema.exclusiveMinimum === "number") {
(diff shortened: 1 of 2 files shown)


colinhacks:
Merging this. Pushed two commits on top.

The first pins the issue count. The check-list assertion covers the internal shape, but the thing a user actually sees is that an out-of-range value reported two issues instead of one — `expect(schema.safeParse(4).error!.issues).toHaveLength(1)` is what regresses if this is ever undone. Also dropped the `as any`: `JSONSchema` already types `exclusiveMinimum` as `number | boolean`, so the draft-04 literal type-checks unaided.

The second unstacks the comment, since `check:comments` landed after you cut this branch and fails on the pair.

One correction worth recording, because the squash message says otherwise than the description here: the duplication is not invisible at the `.parse()` level, and it does not show up in round-tripping. It is the reverse. `toJSONSchema` output is byte-identical with and without this fix on all four targets — `exclusiveMinimum >= (minimum ?? -Infinity)` in the processor makes the redundant bound invisible to the exporter — so no JSON Schema validator ever sees a difference. Zod's own parse output is the only place it surfaces. The bounds also aren't contradictory: `>=5` and `>5` are both true of every value that trips them, and the weaker one never fires alone.

Good investigation — the guard itself is right, and `!== true` is exactly the correct test to leave the draft-06+ numeric form alone. Thanks for looking into it 👍

PR #6023, opened 2026-05-24
fix(from-json-schema): allow extra items when additionalItems/items is `true`

francisjohnjohnston-web:
## What

`fromJSONSchema` converts a draft-07 array tuple (`items` as an array) with
`additionalItems: true`, and a draft-2020-12 tuple (`prefixItems`) with `items: true`,
into a **fixed-length** `z.tuple([...])` that rejects any extra items.

```ts
  $schema: "http://json-schema.org/draft-07/schema#",
  type: "array",
  items: [{ type: "string" }, { type: "number" }],
  additionalItems: true,
});
schema.parse(["hello", 42, "extra"]); // ❌ throws — but additionalItems:true should allow it
```

## Why it's wrong

In both drafts a boolean `true` means "additional items of any type are allowed".
The converter already handles the `false` case (no rest — correct) and the
schema-object case (typed `.rest(...)` — correct), but a boolean `true` falls through
to `rest = undefined`, producing the *same* fixed-length tuple as `false`.

## Why a reviewer might miss it

The existing tests only cover `additionalItems: false`, and the `false`/`true` paths
look almost identical — the missing branch is invisible unless you specifically
exercise the `true` case. Every other `additionalItems` value behaves correctly, so
it never surfaces in normal use.

## Fix

Map a boolean `true` to `.rest(z.unknown())` so extra items of any type are accepted.
The `false` (no rest) and schema-object (typed rest) cases are unchanged. Blast radius
is the tuple `rest` derivation only — no public API or signature change.

## Tests

Added regression tests for draft-07 `additionalItems: true` and draft-2020-12
`items: true` (both assert mixed-type extras are accepted). Full `from-json-schema`
suite: **148/148 pass, 0 type errors**.

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -468,9 +468,11 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
         // Tuple with prefixItems (draft-2020-12)
         const tupleItems = prefixItems.map((item) => convertSchema(item as JSONSchema.JSONSchema, ctx));
         const rest =
-          items && typeof items === "object" && !Array.isArray(items)
-            ? convertSchema(items as JSONSchema.JSONSchema, ctx)
-            : undefined;
+          items === true
+            ? z.unknown()
+            : items && typeof items === "object" && !Array.isArray(items)
+              ? convertSchema(items as JSONSchema.JSONSchema, ctx)
+              : undefined;
         if (rest) {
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing in favor of #6020, which just merged. Correct diagnosis — a boolean `true` did fall through to the same closed tuple as `false` — but #6020 fixes the same derivation through one `getTupleRest` helper that also handles the absent case, where `items`/`additionalItems` is omitted entirely and the tail should be open. Your two test cases went in with it.

Issue #6027, opened 2026-05-25
Bug: toJSONSchema `$ref` pointer not encoded when meta id contains `/` or `~`


dinwwwh:
## Description

According to [RFC 6901 (JSON Pointer)](https://datatracker.ietf.org/doc/html/rfc6901), reference tokens inside a JSON Pointer must escape two characters:

- `~` → `~0`
- `/` → `~1`

When `z.toJSONSchema()` builds the `$ref` URI fragment from a schema's `.meta({ id })`, it uses the raw id without applying these escapes. The `$defs` key itself is a plain JSON object key and does not need escaping — but the `$ref` pointer path does, and it currently doesn't get it.

## Steps to reproduce

```typescript
import { z } from 'zod';

const User = z.object({ name: z.string() })
  .meta({ id: 'Shared/User~' });

console.log(z.toJSONSchema(z.object({ User })));
```

## Actual output

```json
{
  "properties": {
    "User": {
      "$ref": "#/$defs/Shared/User~"
    }
  },
  "$defs": {
    "Shared/User~": {}
  }
}
```

The `$ref` pointer token `Shared/User~` is not escaped — `/` should be `~1` and `~` should be `~0`.

## Expected output

```json
{
  "properties": {
    "User": {
      "$ref": "#/$defs/Shared~1User~0"
    }
  },
  "$defs": {
    "Shared/User~": {}
  }
}
```

The `$defs` key stays as the original unescaped id. Only the `$ref` pointer path needs encoding.

me-saurabhkohli:
Confirmed on `main` (`bbc68f9`). Root cause is in `packages/zod/src/v4/core/to-json-schema.ts` at the line:

```typescript
return { defId, ref: defUriPrefix + defId };
```

The `$ref` URI fragment is a [JSON Pointer (RFC 6901)](https://datatracker.ietf.org/doc/html/rfc6901), which requires two escapes inside reference tokens:
- `~` → `~0`
- `/` → `~1`

The `$defs` object **key** is a plain JSON key and should remain unescaped — only the pointer path inside `$ref` needs encoding.

Fix is straightforward:
```typescript
const encodedDefId = defId.replace(/~/g, "~0").replace(/\//g, "~1");
return { defId, ref: defUriPrefix + encodedDefId };
```

The same issue exists in the `external` code path. I've opened a PR with the fix and 3 regression tests covering `/`, `~`, and both characters combined.

PR #6028, opened 2026-05-25
fix(v4): encode / and ~ in $ref JSON Pointer paths (RFC 6901)

mixelburg:
Fixes #6027

When a schema has a `.meta({ id })` containing `/` or `~`, the generated `$ref` URI fragment was using the raw id, producing an invalid JSON Pointer.

Per [RFC 6901](https://datatracker.ietf.org/doc/html/rfc6901), reference tokens in a JSON Pointer must escape:
- `~` → `~0`
- `/` → `~1`

The `$defs` key (a plain JSON object key) correctly keeps the raw id. Only the `$ref` pointer path needs the encoding.

**Before:**
```json
{ "$ref": "#/$defs/Shared/User~" }
```

**After:**
```json
{ "$ref": "#/$defs/Shared~1User~0" }
```

Applied the same fix to both the self-contained path and the external registry path.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2212,6 +2212,15 @@ test("id is observable in override callback", () => {
   expect(seenIds).toContain("Inner");
 });
 
+test("$ref pointer encodes / and ~ in meta id (RFC 6901)", () => {
+  // ids with slashes or tildes need RFC 6901 encoding in the $ref pointer path
+  // but the $defs key should remain the raw id
+  const User = z.object({ name: z.string() }).meta({ id: "Shared/User~" });
+  const result = z.toJSONSchema(z.object({ user: User }));
+  expect((result as any).$defs?.["Shared/User~"]).toBeDefined();
+  expect((result.properties as any)?.user?.$ref).toBe("#/$defs/Shared~1User~0");
+});
+
 test("describe with id on wrapper", () => {
   // Test that $ref propagation works when processor sets a different ref (readonly -> innerType)
   // but parent was extracted due to having an id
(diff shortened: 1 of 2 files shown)


wahajahmed010 (commented):
## Review: encode / and ~ in $ref JSON Pointer paths (RFC 6901)

This is a well-scoped fix addressing a real spec violation. The 8-line change is correct and the test captures the right behavior. A few observations:

**Strengths:**
- Escape order is correct: `~` before `/`, so a literal `/` encoded as `~1` won't be double-encoded by the `~` → `~0` pass on a subsequent iteration.
- Both code paths (external registry and self-contained) are covered — no missed spots.
- The test correctly asserts the `$defs` key keeps the raw id while the `$ref` uses the encoded form.

**Potential edge cases to consider:**

1. **Double-encoding on re-processing** — If `toJSONSchema` is called on a schema that already has a pre-encoded `~0` or `~1` in its meta id, the replace will encode them again (producing `~00` or `~10`). For example, a meta id of `My~0Key` would produce a `$ref` of `#/$defs/My~000Key`. This is an edge case that relies on users intentionally putting encoded sequences in their ids, but worth documenting or protecting against if the function might be called multiple times on the same schema.

2. **No test for `~` or `/` at the start/end of the id** — The test only covers `Shared/User~` (slash in the middle, tilde at the end). A test with a leading `~` or trailing `/` would provide more confidence — e.g., `meta({ id: "~root" })`.

3. **No negative test** — A test that an id WITHOUT special characters passes through unchanged would help prevent regression (though admittedly this is covered implicitly by other tests).

The fix is spec-correct and the implementation is minimal. Good work on catching a conformance gap that many implementations miss.

**Verdict:** Ready to merge after considering the double-encode edge case (low risk).

mixelburg:
Closing due to inactivity — 6 weeks with no review or feedback. Happy to reopen if there's interest.

PR #6033, opened 2026-05-28
Fix: escape JSON Pointer tokens in $ref per RFC 6901

wahajahmed010:
## Summary

Fixes #6027

When a schema's `meta({ id })` contains `/` or `~` characters, `z.toJSONSchema()` builds the `$ref` URI fragment using the raw id without applying RFC 6901 escaping. Per RFC 6901, reference tokens in a JSON Pointer must escape `~` as `~0` and `/` as `~1`.

## Changes

- Added `escapePointerToken` helper inside `extractDefs` that applies RFC 6901 encoding (`~` -> `~0`, `/` -> `~1`)
- Applied escaping to the `$ref` path in both the self-contained and external schema URI branches of `makeURI`
- The `$defs` dictionary key remains unescaped (it is a plain JSON object key, not a pointer token)
- Added a regression test that verifies the `$ref` is correctly encoded while the `$defs` key stays as the original id

## Before

```json
{
  "$ref": "#/$defs/Shared/User~"
}
```

## After

```json
{
  "$ref": "#/$defs/Shared~1User~0"
}
```

The `$defs` key remains `"Shared/User~"` (unescaped) as expected.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -3115,6 +3115,18 @@ test("cycle detection - mutual recursion", () => {
   `);
 });
 
+test("RFC 6901: $ref pointer tokens escape ~ and / in meta id", () => {
+  // When a schema's meta id contains / or ~, the $defs key stays unescaped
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #6034, opened 2026-05-28
fix(url): reject IPv6 link-local addresses parsed as fake schemes

### LINKED ISSUE #6031: Bug: `z.string().url()` incorrectly validates IPv6 link-local addresses ("fe80:0000:0000:0000:0000:0000:0000:0001") as valid URLs

### Environment
- **Zod Version**: 3.25.76
- **TypeScript Version**: 5.x

### Test Results

```javascript
import { z } from "zod";

```

xxiaoxiong:
## Description

The `z.string().url()` validator incorrectly accepts bare IPv6 link-local addresses like `fe80::1` as valid URLs.

**Root cause:** The `URL` constructor parses `fe80::1` by treating `fe80` as a URL scheme and `::1` as the path. This produces a `URL` object with `origin === null` and an empty `hostname`, bypassing the validation.

**Fix:** After successful `new URL()` parsing, check whether the resulting URL has a null origin and no hostname. If so, only accept it if the protocol is a known opaque scheme (`file:`, `blob:`, `data:`, `mailto:`). All other cases are rejected.

## Test cases

Invalid (previously incorrectly accepted):
- `fe80::1`
- `fe80::abcd:1234`

Still valid:
- `http://google.com`
- `https://[::1]`
- `file:///C:/path`
- `mailto:test@test.com`

Fixes #6031

### DIFF
--- a/packages/zod/src/v3/tests/string.test.ts
+++ b/packages/zod/src/v3/tests/string.test.ts
@@ -300,6 +300,11 @@ test("url validations", () => {
   expect(() => url.parse("asdf")).toThrow();
   expect(() => url.parse("https:/")).toThrow();
   expect(() => url.parse("asdfj@lkjsdf.com")).toThrow();
+  // IPv6 addresses without proper scheme should be rejected
+  expect(() => url.parse("fe80::1")).toThrow();
+  expect(() => url.parse("fe80::abcd:1234")).toThrow();
+  // Known opaque schemes like mailto: should still work
+  url.parse("mailto:test@test.com");
 });
 
 test("url error overrides", () => {
--- a/packages/zod/src/v3/types.ts
+++ b/packages/zod/src/v3/types.ts
@@ -877,7 +877,17 @@ export class ZodString extends ZodType<string, ZodStringDef, string> {
       } else if (check.kind === "url") {
         try {
           // @ts-ignore
-          new URL(input.data);
+          const url = new URL(input.data);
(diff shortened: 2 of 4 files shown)


xxiaoxiong:
Thank you for the thorough review! I have updated the fix based on your feedback:

1. **More targeted heuristic**: Replaced the hardcoded allowlist with a regex check that detects when the URL parser has mistaken an IPv6 hex prefix for a scheme. The regex `/^[0-9a-f]{1,4}:/i` tests the parsed URL protocol. This catches bare IPv6 addresses like "fe80::1" without rejecting opaque-scheme URLs like "javascript:", "tel:", "about:", etc.

2. **v4 fix included**: Added the same check to `packages/zod/src/v4/core/schemas.ts` (the `new URL(trimmed)` call) with matching logic and error reporting.

This version no longer uses an allowlist, so any legitimate opaque-scheme URL will pass through unaffected.

xxiaoxiong:
Good catch on the `c:` false positive! Fixed with:

1. Added `url.pathname.includes(":")` guard — IPv6 addresses mis-parsed as schemes (like `fe80::1`) have colons in their pathname (`:1`), while opaque schemes like `c:`, `a:2`, `mailto:test@test.com` have empty pathnames without colons.

2. Added IPv6 test cases to v4 `string.test.ts` — matching the v3 test coverage: 2 IPv6 rejection cases + 1 `mailto:` success case.

PR #6035, opened 2026-05-28
fix: enforce RFC 1035 length limits in regexes.domain

### LINKED ISSUE #5965: regexes.domain is missing RFC 1035 length limits
## Context

```ts
```

```ts
```

## Question

## Note

emmayusufu:
Closes #5965.

## Problem

`regexes.domain` (used by `z.httpUrl()` for the host check) has no length anchors:

```ts
export const domain: RegExp = /^([a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,}$/;
```

`regexes.hostname` (used by `z.hostname()`) already enforces the RFC 1035 limits:

```ts
```

So `z.httpUrl()` accepts hosts that exceed the 253-char total or have a TLD longer than 63 chars, while `z.hostname()` rejects them. The two validators disagree on what a valid DNS name is.

## Change

```ts
export const domain: RegExp = /^(?=.{1,253}$)([a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,63}$/;
```

- `(?=.{1,253}$)` caps the total length at 253 chars
- `[a-zA-Z]{2,63}` caps the TLD label at 63 chars

The per-label cap for non-TLD labels was already there via `[a-zA-Z0-9-]{0,61}`. The `domain` pattern never allows a trailing dot, so I dropped the `\.?` from the lookahead that `hostname` carries.

## A note on intent

The issue raised this as a question: RFC 3986 / WHATWG URL don't themselves impose DNS length limits, so the looseness in `z.httpUrl()` might be deliberate. I went with closing the gap because `z.hostname()` and `z.httpUrl()` validating the same host differently is surprising. If you'd rather keep `httpUrl` permissive, happy to close this.

## Tests

Added cases to the existing `httpurl` test for the total-length cap, the TLD cap, and a host at the 253-char boundary (which still parses). Full `string.test.ts` suite passes (94 tests).

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -381,7 +381,7 @@ schema.parse("http://example.com"); // ❌
   This restricts the protocol to `http`/`https` and ensures the hostname is a valid domain name with the `z.regexes.domain` regular expression:
 
   ```ts
-  /^([a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,}$/
+  /^(?=.{1,253}$)([a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,63}$/
   ```
 </Callout>
 
--- a/packages/zod/src/v4/classic/tests/string.test.ts
+++ b/packages/zod/src/v4/classic/tests/string.test.ts
@@ -464,6 +464,16 @@ test("httpurl", () => {
   ).toThrow();
   expect(() => httpUrl.parse("http://asdf.c")).toThrow();
   expect(() => httpUrl.parse("mailto:asdf@lckj.com")).toThrow();
+  // total host length over the RFC 1035 limit of 253 chars (built from valid-length labels)
+  const longHost = `${`${"a".repeat(50)}.`.repeat(5)}com`; // 5 * 51 + 3 = 258 chars
+  expect(longHost.length).toBeGreaterThan(253);
+  expect(() => httpUrl.parse(`http://${longHost}`)).toThrow();
+  // TLD over the 63-char label limit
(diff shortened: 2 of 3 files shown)


wahajahmed010 (commented):
## Review: RFC 1035 length limits in `regexes.domain`

Clean, minimal fix. One-line regex change with good test coverage. Observations:

### LGTM

- **Adds `(?=.{1,253}$)` lookahead** — correctly bounds total hostname length per RFC 1035 without affecting the structural match
- **Caps TLD at 63 chars** (`{2,63}` instead of `{2,}`) — matches RFC 1035 label length limit
- **Tests cover three key scenarios**: over-length host (258 chars), over-length TLD (64 chars), and boundary case (251 chars)

### Minor: trailing dot behavior

The `domain` regex rejects trailing dots, while `hostname` allows them via `\.?$`. Consistent with existing behavior — domain is intentionally stricter.

### Summary

Simple, correct, well-tested. Ready to merge.

colinhacks:
Pushed a commit on top. The docs page and the `httpurl` test each embed a literal copy of `z.regexes.domain`, and the new length limits made both stale — those are now in sync. Also moved the boundary fixture to exactly 253 characters; it was 251, so the comment was wrong and neither side of the limit was pinned.

Checked the regex for backtracking before landing this, since a length assertion in front of a repeated label group is the classic hazard. It's clean: `recheck` proves the label body linear and the lookahead safe on its own, and the lookahead means the body never sees more than 253 characters. A differential fuzz over 400k hostnames found no behavior change beyond the two intended limits.

Merging once CI is green.

emmayusufu:
Thanks for landing this and for the fixes on top. I had missed that the docs and the httpurl test both carried their own copy of the regex. Also good to see the backtracking check come back clean, that lookahead in front of the repeated labels was the part I was least sure about.

PR #6037, opened 2026-05-29
fix(to-json-schema): apply RFC 6901 encoding to $ref pointer path

me-saurabhkohli:
## Summary

When a schema's `.meta({ id })` contains `/` or `~`, the `$ref` pointer generated by `z.toJSONSchema()` is invalid per [RFC 6901](https://datatracker.ietf.org/doc/html/rfc6901).

**Before** (broken):
```json
{ "$ref": "#/$defs/Shared/User~" }
```

**After** (correct):
```json
{ "$ref": "#/$defs/Shared~1User~0" }
```

## Root cause

In `to-json-schema.ts`, the `$ref` path is built by concatenating the raw `defId` string:

```typescript
return { defId, ref: defUriPrefix + defId };  // ← missing RFC 6901 encoding
```

RFC 6901 §3 requires that inside a JSON Pointer reference token:
- `~` is encoded as `~0`
- `/` is encoded as `~1`

The `$defs` object **key** is a plain JSON key (not a pointer) and must stay unencoded so validators can look up the definition by its original id.

## Fix

```typescript
const encodedDefId = defId.replace(/~/g, "~0").replace(/\//g, "~1");
return { defId, ref: defUriPrefix + encodedDefId };
```

The same fix is applied to the `external` code path.

## Tests

Three regression tests added to `to-json-schema.test.ts`:
1. meta id containing `/` → `$ref` uses `~1`, `$defs` key stays raw
2. meta id containing `~` → `$ref` uses `~0`, `$defs` key stays raw
3. meta id containing both `/` and `~` → correct combined encoding

All 3817 existing tests continue to pass.

Fixes #6027

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -3129,3 +3129,32 @@ test("recursive lazy with describe does not stack overflow", () => {
   expect(result).toBeDefined();
   expect(result.$defs).toBeDefined();
 });
+
+// RFC 6901: $ref pointer tokens must escape '~' → '~0' and '/' → '~1'
+// but the $defs key itself is a plain JSON object key and stays unescaped.
+// See https://github.com/colinhacks/zod/issues/6027
+test("toJSONSchema encodes '/' in meta id as '~1' in $ref pointer (RFC 6901)", () => {
+  const User = z.object({ name: z.string() }).meta({ id: "Shared/User" });
+  const result = z.toJSONSchema(z.object({ User }));
+
+  // $defs key uses the raw id (plain object key — no escaping needed)
+  expect(result.$defs).toHaveProperty("Shared/User");
+  // $ref pointer encodes '/' as '~1'
+  expect((result as any).properties.User.$ref).toBe("#/$defs/Shared~1User");
+});
+
+test("toJSONSchema encodes '~' in meta id as '~0' in $ref pointer (RFC 6901)", () => {
+  const Model = z.object({ value: z.number() }).meta({ id: "My~Model" });
+  const result = z.toJSONSchema(z.object({ Model }));
+
+  expect(result.$defs).toHaveProperty("My~Model");
+  expect((result as any).properties.Model.$ref).toBe("#/$defs/My~0Model");
+});
+
+test("toJSONSchema encodes both '~' and '/' in meta id in $ref pointer (RFC 6901)", () => {
+  const Schema = z.object({ x: z.string() }).meta({ id: "Shared/User~" });
+  const result = z.toJSONSchema(z.object({ Schema }));
+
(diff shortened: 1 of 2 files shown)


me-saurabhkohli:
Hi @colinhacks , just a gentle ping on this one when you get a chance!

The fix is minimal (9 lines changed in `to-json-schema.ts`) and both automated reviewers (Copilot + Pullfrog) gave it a green signal. Happy to address any feedback or rebase if needed.

wahajahmed010 (commented):
**Clean fix — nice work!** 🎉

The encoding order is correct: replace `~` → `~0` first, then `/` → `~1`. This avoids double-encoding a `/` that appears inside `~0` (which would create an invalid round-trip).

I verified the round-trip manually for edge cases:
- `id = "a/b~c"` → encodes to `a~1b~0c` → decodes correctly back (per RFC 6901 §4: `~1`→`/` first, then `~0`→`~`).
- `id = "My~1Model"` → encodes to `My~01Model` → decodes back to `My~1Model`. ✅

One minor thing worth noting: this only fixes `toJSONSchema` output. If there are other code paths that construct `$ref` pointers from `schema.id` or `defId` manually (outside `extractDefs`), they would have the same bug. But that is out of scope for this PR.

Tests are thorough — covering `/`, `~`, and combined cases. This is a textbook-quality first PR. Welcome to the Zod community! 🦌

me-saurabhkohli:
Thanks @wahajahmed010!! Really appreciate the thorough validation and the edge-case walkthrough 🙏.

On the out-of-scope note: @wahajahmed010 / @colinhacks, happy to extend this PR to cover any other places that manually construct $ref pointers from `schema.id` / `defId` outside `extractDefs` if you'd rather land it all together. 
Or, if you'd prefer to keep this one focused, just let me know if you guys will open an issue or I can open a separate one with the details. Happy to help either way, just let me know! Thanks!

me-saurabhkohli:
Hi @wahajahmed010 / @colinhacks , please review whenever you get a chance and let me know if anything is required on this PR. Thanks and appreciate your help!!

me-saurabhkohli:
Hi @wahajahmed010 / @colinhacks , please let me know if you have any observation on this PR and would be happy to address. Thanks and appreciate your help!!

colinhacks:
Closing in favor of #6144. Same fix and the same correct escape order, but it puts the escaping behind a single `encodeJSONPointerSegment` helper instead of repeating the replace chain at both call sites, and it still merges clean — this one now conflicts with `main`. Thanks for looking into it 👍

me-saurabhkohli:
> Closing in favor of #6144. Same fix and the same correct escape order, but it puts the escaping behind a single `encodeJSONPointerSegment` helper instead of repeating the replace chain at both call sites, and it still merges clean — this one now conflicts with `main`. Thanks for looking into it 👍

Thanks for the update @colinhacks.

PR #6041, opened 2026-05-29
Locales: added Slovak (sk) language

belicam:
Added error map locale support for Slovak language 🇸🇰


wahajahmed010 (commented):
Great work on the Slovak locale, @belicam! 🇸🇰 The structure follows the existing pattern perfectly.

**Review notes:**

1. **Coverage**: All error map cases handled — matches existing locales. ✅

2. **Format dictionary**: All 22 format types covered. ✅

3. **Sizing**: "znakov" (characters), "bajtov" (bytes), "prvkov" (elements) — correct genitive plural. Verb "mať" matches Slovak syntax. ✅

4. **Grammar**: All translations use correct Slovak grammar and gender agreement ("Neplatný", "očakávané"/"obdržané", "príliš veľká/príliš malá"). ✅

5. **Integration**: Import/export in `index.ts` and docs listing are correct. ✅

Minor note: `template_literal` = "vstup" is slightly less descriptive than some locales use, but it is ultimately a preference call for the native speaker and consistent with the general approach.

Tip: Could add a locale test, though no existing locale has one either — not blocking.

Thanks for contributing to Zod localization! 🎉

PR #6042, opened 2026-05-30
fix(core): Throw error if enum schema has no values

RobinTail:
`z.literal([])` throws since 4.0.9: see https://github.com/colinhacks/zod/commit/4e7a3ef180f6a5525d9021638e9df20b3ca50456#diff-f84405ad925758ccb224a20fe7ca8c6ee375d7faa88431059e008077550f8cd9R2785

but `z.enum([])` does not.

Since the signatures of both methods allow empty arrays, I'm suggesting to implement the same runtime validation for consistency.

### DIFF
--- a/packages/zod/src/v4/classic/tests/enum.test.ts
+++ b/packages/zod/src/v4/classic/tests/enum.test.ts
@@ -155,9 +155,7 @@ test("exclude", () => {
   expect(UnhealthyEnum.safeParse("Salad").success).toEqual(false);
   expectTypeOf<z.infer<typeof UnhealthyEnum>>().toEqualTypeOf<"Pasta" | "Pizza" | "Tacos" | "Burgers">();
 
-  const EmptyFoodEnum = FoodEnum.exclude(foods);
-  expectTypeOf<typeof EmptyFoodEnum>().toEqualTypeOf<z.ZodEnum<{}>>();
-  expectTypeOf<z.infer<typeof EmptyFoodEnum>>().toEqualTypeOf<never>();
+  expect(() => FoodEnum.exclude(foods)).toThrow(/no valid values/);
 });
 
 test("error map inheritance", () => {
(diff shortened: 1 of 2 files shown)


wahajahmed010 (commented):
## Review: Throw error if enum schema has no values

This is a good consistency fix bringing `z.enum([])` in line with the existing `z.literal([])` behavior. The 4-line change is clean and the runtime check is placed correctly.

**Substantive observations:**

1. **Missing direct test for `z.enum([])`** — The test only covers the `FoodEnum.exclude(foods)` path that produces an empty enum through exclusion. There is no test for calling `z.enum([])` directly. While the constructor check covers both paths, a test for the direct case would:
   - Document the expected behavior explicitly for anyone reading the tests
   - Guard against future refactors that might bypass the constructor path
   - Mirror how the `z.literal([])` test suite covers the direct case

2. **Error message specificity** — The message says "Cannot create enum schema with no valid values." When triggered by `exclude()`, a user seeing this error has no indication it came from an `exclude()` call vs. a direct empty call. Consider whether distinguishing these in the message would aid debugging (e.g., referencing the method that triggered the creation). That said, this is a minor DX polish concern.

3. **`$ZodType.init(inst, def)` runs before the throw** — The init call runs before the empty check. If `init()` has side effects or allocates resources, those happen even on the throw path. This mirrors the `z.literal([])` pattern so it's consistent, but worth noting.

The logic is sound, CI is green, and the change is consistent with existing patterns. Nice work.

colinhacks:
Going to decline this one. An empty enum is reachable from inside the library — `z.object({}).keyof()` produces one, and so does excluding every member of an enum — so throwing in the constructor turns working code into a crash, and the empty enum already rejects every input anyway. The same guard is safe on `z.literal()` because nothing inside the library builds one.

Thanks for looking into it 👍

Issue #5993, opened 2026-05-14
z.never() within union not behaving as it used to since 4.4.0


didinele:
A user filed a bug for discord.js where a previously functional input no longer worked.

Boiling down the issue to its most minimal form (past all the nested union and intersection types), the difference boiled down to this:

<img width="1496" height="141" alt="Image" src="https://github.com/user-attachments/assets/5f192544-2d14-4356-a054-e15c01196f1f" />

<img width="1507" height="126" alt="Image" src="https://github.com/user-attachments/assets/58502340-97bc-4551-adcd-0d39adb9bebb" />

The issue still occurs as of 4.4.3:

<img width="1495" height="136" alt="Image" src="https://github.com/user-attachments/assets/ccc5907a-983f-4689-b6bd-7af566c56693" />

For now, I decided to fix it by replacing the predicate with 
```ts
choices: z.array(z.any()).length(0).optional()
```

See https://github.com/discordjs/discord.js/pull/11532

MerlijnW70:
I've been investigating the root cause of #5993 and wanted to share my findings to help clarify the scope of this regression.

## Root Cause Analysis

The `nonoptional` error (triggered in `handlePropertyResult`) occurs because `z.undefined()`, `z.void()`, `z.any()`, and `z.unknown()` do not set the `optin === "optional"` flag. As a result, when parsing an absent key against these types, `handlePropertyResult` incorrectly treats them as required.

## Observations

- **Scope:** This is a regression since 4.4.0 affecting both runtime behavior and type inference (e.g., `z.object({ a: z.any() })` now infers as `{ a: any }` instead of `{ a?: any }`).
- **Complexity:** Because this affects fundamental types used in discriminated unions, tuples, and general object schema optionality, a 'surgical' fix via the `optin` flag risks side effects on existing snapshots and type inference patterns.

## Conclusion

While this is clearly a regression, it touches upon the broader optionality semantics currently evolving in the v4 codebase. Given the potential impact on type-snapshots and schema consistency, I believe a definitive fix requires a decision on the intended optionality model for these types.

I hope this diagnostic analysis helps in triaging the issue!

colinhacks:
Intentional, same root cause as #5957. `z.undefined()` was marked optional on the input side, which propagated through your union and let `choices` be absent — but the inferred type has always required it, so the runtime was accepting input its own types rejected. 4.4.0 dropped that flag and it went out as a bug fix in a minor.

[...]

PR #6044, opened 2026-05-31
fix: union containing z.undefined() should be optional-in for object keys

sapirbaruch:
Ran into this via the discord.js bug report in #5993 — after digging in I traced it to a change in 4.4.0 (commit b6066b3e, PR #5661) that stripped `optin`/`optout` off `$ZodUndefined`.

**The problem:**

`z.object` decides whether a missing key is valid by checking `el._zod.optin === "optional"`. Before 4.4, `z.undefined()` set that flag, so a union like:

```ts
z.union([z.never(), z.array(z.never()), z.undefined()])
```

was treated as optional-in and a missing key would silently coerce to `undefined` (which the union accepts). After the change, neither `z.undefined()` nor the union propagate that flag, so the object bails early with `"expected nonoptional"` before the schema even runs.

**The fix:**

In `$ZodUnion`'s lazy `optin`/`optout` computation, also check whether any option's `_zod.values` set includes `undefined`. `z.undefined()` has `values = new Set([undefined])`, so a union with a `z.undefined()` member now correctly inherits the optional-in/out behavior.

```ts
// before
def.options.some((o) => o._zod.optin === "optional")

// after
def.options.some((o) => o._zod.optin === "optional" || o._zod.values?.has(undefined))
```

This is intentionally conservative: the new check only fires when a member's static value set contains `undefined`, which is exactly when the union will accept a missing key.

**Repro:**

```ts
const schema = z.object({
  autocomplete: z.literal(true),
  choices: z.union([z.never(), z.never().array(), z.undefined()]),
});

schema.parse({ autocomplete: true });        // 4.3.x: ok — 4.4.x: throws nonoptional
schema.parse({ autocomplete: true, choices: undefined });  // ok in both
```

The updated snapshot in `optional.test.ts` (line ~144) reflects the corrected behavior: only the `z.undefined()` standalone field still requires an explicit key; a union containing `z.undefined()` does not.

Refs #5993.

### DIFF
--- a/packages/zod/src/v4/classic/tests/never-union-object.test.ts
+++ b/packages/zod/src/v4/classic/tests/never-union-object.test.ts
@@ -1,0 +1,42 @@
+import { expect, test } from "vitest";
+import * as z from "zod/v4";
+
+// Regression test for https://github.com/colinhacks/zod/issues/5993
+
+const choicesSchema = z.union([z.never(), z.array(z.never()), z.undefined()]);
+
+// The exact discord.js scenario: choices field in an object
+const optionSchema = z.object({
+  name: z.string(),
+  choices: choicesSchema,
+});
+
+test("z.never() union in object: accepts choices: undefined", () => {
+  expect(optionSchema.safeParse({ name: "opt", choices: undefined }).success).toBe(true);
+});
+
+test("z.never() union in object: accepts choices: []", () => {
+  expect(optionSchema.safeParse({ name: "opt", choices: [] }).success).toBe(true);
+});
+
+test("z.never() union in object: rejects choices: [{name:'x',value:'y'}]", () => {
+  // This is the broken case reported in #5993
+  const result = optionSchema.safeParse({ name: "opt", choices: [{ name: "x", value: "y" }] });
+  expect(result.success).toBe(false);
+});
+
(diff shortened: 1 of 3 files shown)


wahajahmed010 (commented):
Great catch and clean fix, @sapirbaruch! The root cause analysis is spot-on — PR #5661 cleaned up optin/optout from $ZodUndefined, but the union propagation needed to catch undefined through the values set instead. The fix is conservative and correct.

**What I like:**
- Clear reproduction path from discord.js downstream
- The optional chaining on _zod.values handles schemas that do not define a values set (e.g., ZodString, ZodNumber) gracefully
- Test coverage is thorough — covers all 4 states (explicit undefined, empty array, populated array, missing key)

**One thought for completeness:**
The tests cover a single-depth union in an object. Worth also adding a nested union test: a z.union inside another z.union with z.undefined(). The lazy chain should propagate optin correctly, but locking it in with a test would prevent regression. Not a blocker.

Excellent first contribution! 🦌

colinhacks:
Closing this — the 4.4.0 behavior is intentional. `z.undefined()` was marked optional on the input side, which propagated through unions and let a required key go missing; the inferred type never agreed, so that was an unintentional soundness hole. 4.4.0 dropped the flag, as a bug fix in a minor.

Append `.optional()` to the union if you want the key to be omittable. Thanks for the repro 👍

PR #6051, opened 2026-06-03
fix: escape JSON Pointer tokens in toJSONSchema $ref (RFC 6901)

godfengliang:
Fixes colinhacks/zod#6027

## Problem
When a schema id contains `/` or `~` characters, the `$ref` URI fragment was not being escaped per RFC 6901 (JSON Pointer):
- `~` should become `~0`
- `/` should become `~1`

For example, with id `Shared/User~`, the `$ref` was:
  `#/$defs/Shared/User~` (incorrect)

Now correctly produces:
  `#/$defs/Shared~1User~0`

The `$defs` object key remains unescaped (as it's a plain JSON object key, not a JSON Pointer).

## Fix
Apply `split("~").join("~0").split("/").join("~1")` to the defId in both `$ref` construction paths in `extractDefs`:
1. The shared schema path (line ~263)
2. The self-contained schema path (line ~274)

## Testing
Verified with three test cases:
- `id: "Shared/User~"` -> `$ref: "#/$defs/Shared~1User~0"` ✓
- `id: "MySchema"` -> `$ref: "#/$defs/MySchema"` ✓ (no change for normal ids)
- `id: "My~Schema"` -> `$ref: "#/$defs/My~0Schema"` ✓

### DIFF
--- a/packages/zod/src/v4/core/to-json-schema.ts
+++ b/packages/zod/src/v4/core/to-json-schema.ts
@@ -260,7 +260,9 @@ export function extractDefs<T extends schemas.$ZodType>(
       // otherwise, add to __shared
       const id: string = entry[1].defId ?? (entry[1].schema.id as string) ?? `schema${ctx.counter++}`;
       entry[1].defId = id; // set defId so it will be reused if needed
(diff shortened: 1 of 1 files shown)


colinhacks:

PR #6052, opened 2026-06-03
fix: narrow format type on ZodInt, ZodFloat32, ZodFloat64, ZodInt32, …

### LINKED ISSUE #6045: Distinguish different number formats in type-level
Currently the situation at the type-level is following:

```typescript
export interface ZodInt extends ZodNumberFormat {}
export interface ZodFloat32 extends ZodNumberFormat {}
export interface ZodFloat64 extends ZodNumberFormat {}
export interface ZodInt32 extends ZodNumberFormat {}
export interface ZodUInt32 extends ZodNumberFormat {}
export declare function int(...): ZodInt;
export declare function float32(...): ZodFloat32;
export declare function float64(...): ZodFloat64;
export declare function int32(...): ZodInt32;
export declare function uint32(...): ZodUInt32;
```

Also same with ZodBigIntFormat, they do have string-literal types but no ability to narrow, since both return same type:

```typescript
export declare function int64(...): ZodBigIntFormat;
export declare function uint64(...): ZodBigIntFormat;
```

This makes it impossible to use these to for instance derive Postgres types which are required to exist typed at type-level too (for Kysely etc).

## Suggestion

What I would like them to be is something like:

```typescript
export interface ZodInt extends ZodNumberFormat {
  _zod: {
    def: {
      format: "safeint"
    }
  }
}
export interface ZodFloat32 extends ZodNumberFormat {
  _zod: {
    def: {
      format: "float32"
    }
  }
}
export interface ZodFloat64 extends ZodNumberFormat {
  _zod: {
    def: {
      format: "float64"
    }
  }
}
export interface ZodInt32 extends ZodNumberFormat {
  _zod: {
    def: {
      format: "int32"
    }
  }
}
export interface ZodUInt32 extends ZodNumberFormat {
  _zod: {
    def: {
      format: "uint32"
    }
  }
}
```

This would allow me to use that format (which already exists in runtime) as differentiator.

Thanks.

abhishek-chaudhary2003:
Fixes #6045

Added generic <Format> parameter to $ZodCheckNumberFormatDef, 
$ZodNumberFormatInternals, $ZodNumberFormat in core, and 
ZodNumberFormat/ZodMiniNumberFormat in classic/mini packages.

This allows ZodInt, ZodFloat32, ZodFloat64, ZodInt32, ZodUInt32 
to be distinguished at the type level:

// before
ZodInt extends ZodNumberFormat {}  // no format info

// after  
ZodInt extends ZodNumberFormat<"safeint">  // TypeScript knows the format

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1,5 +1,5 @@
 import * as core from "../core/index.js";
-import { util } from "../core/index.js";
+import { type $ZodNumberFormats, util } from "../core/index.js";
 import * as processors from "../core/json-schema-processors.js";
 import type { StandardSchemaWithJSONProps } from "../core/standard-schema.js";
 import { createStandardJSONSchemaMethod, createToJSONSchemaMethod } from "../core/to-json-schema.js";
@@ -1130,8 +1130,8 @@ export function number(params?: string | core.$ZodNumberParams): ZodNumber {
 }
 
 // ZodNumberFormat
-export interface ZodNumberFormat extends ZodNumber {
-  _zod: core.$ZodNumberFormatInternals;
+export interface ZodNumberFormat<Format extends $ZodNumberFormats = $ZodNumberFormats> extends ZodNumber {
+  _zod: core.$ZodNumberFormatInternals<Format>;
 }
 export const ZodNumberFormat: core.$constructor<ZodNumberFormat> = /*@__PURE__*/ core.$constructor(
   "ZodNumberFormat",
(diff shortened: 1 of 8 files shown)


Ciantic:
I think this is roughly correct approach, it would also need to make ZodBigIntFormats generic too, it has same problem. However, why would the test refer to postgres, that is just my use case.

colinhacks:
Pushed three commits on top: rebased onto main, reverted the play.ts changes, and added type tests for the pinned format literal and the conditional-type dispatch this enables.

Two things differ from the original approach. Every format now has a name in both entry points — `ZodInt64` and `ZodUInt64` in classic, plus the mini aliases — since classic named the five number formats and nothing for the bigints. And the `| undefined` on the bigint check def comes off in its own commit rather than as a side effect of this one: nothing can construct that state, and the runtime was already asserting past it.

Merging. Thanks for putting this together 👍

PR #6053, opened 2026-06-03
fix: preserve explicit superRefine issue input

frastefanini:
`superRefine` currently uses nullish assignment when backfilling an issue's `input`, so an issue author who explicitly passes `input: null` or `input: undefined` has that value replaced with the parsed payload value when `reportInput` is enabled.

This switches the defaulting check to property presence in both `superRefine` paths, so omitted `input` still reports the parsed value while explicit nullish values are preserved. It also adds regression coverage for both `.superRefine(...)` and top-level `z.superRefine(...)` through `.check(...)`.

Validated with `pnpm test`, `pnpm lint:check`, and `pnpm format:check`.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -2088,7 +2088,7 @@ export const ZodTransform: core.$constructor<ZodTransform> = /*@__PURE__*/ core.
 
           if (_issue.fatal) _issue.continue = false;
           _issue.code ??= "custom";
-          _issue.input ??= payload.value;
+          if (!("input" in _issue)) _issue.input = payload.value;
           _issue.inst ??= inst;
           // _issue.continue ??= true;
           payload.issues.push(util.issue(_issue));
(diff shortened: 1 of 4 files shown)


wahajahmed010 (commented):
Nice catch, @frastefanini! The `??=` nullish assignment was silently swallowing explicit `null`/`undefined` inputs on issues — subtle but real bug.

**Review:**

1. **Root cause analysis**: Correct. `_issue.input ??= payload.value` replaces *any* nullish value with the parsed payload. Since `reportInput` docs say it "backfills" omitted input, explicit `null`/`undefined` should be preserved. ✅

2. **Fix**: `if (!("input" in _issue))` is the right approach — it checks property presence rather than value truthiness. Found in both `ZodTransform` (classic schemas.ts) and `_superRefine` (core/api.ts). ✅

3. **Tests**: Good coverage — tests both `.superRefine()` and `.check().superRefine()` paths, with three cases per test (default, explicit null, explicit undefined). ✅

4. **Edge cases considered**:
   - `issue.code` fallback is left as `??=` — correct, since `code` should never intentionally be null/undefined ✅
   - `reportInput: true` gating is covered (test uses it) ✅
   - The default case (omitted `input`) still gets the parsed value via the presence check ✅

5. **One question**: What happens when `reportInput` is false/not set? Looking at the code, `input` defaulting is gated internally — this change should not affect that path since the nullish coalescing in the issue construction itself is separate. Could be worth a test to confirm, but not a blocker.

Clean, well-scoped fix for a real bug. Welcome to open source! 🎉

PR #6054, opened 2026-06-04
test: cover all v4 locale error maps

zckjzfy:
## Summary
- add a shared smoke test that exercises every exported v4 locale error map
- cover common issue branches including type, enum value, size, string format, object keys, union, map, and set errors
- raises zod package line coverage from 73.94% to 90.36% locally

## Test
- pnpm exec vitest run --project zod packages/zod/src/v4/core/tests/locales/all-locales.test.ts --reporter=dot
- pnpm exec vitest run --project zod --coverage --reporter=dot
- git push hook: pnpm test (341 files / 3864 tests)

### DIFF
--- a/packages/zod/src/v4/core/tests/locales/all-locales.test.ts
+++ b/packages/zod/src/v4/core/tests/locales/all-locales.test.ts
@@ -1,0 +1,43 @@
+import { expect, test } from "vitest";
+import * as z from "zod/v4";
+
+type LocaleFactory = () => { localeError: z.core.$ZodErrorMap };
+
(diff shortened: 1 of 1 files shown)


colinhacks:
Going to close this. Every locale always returns a message — a missing dictionary entry still produces one, just with the English key embedded in it — so asserting the message is a non-empty string passes on the exact breakage that has actually shipped here more than once. The coverage jump is the tell rather than the win: it executes every locale file without checking anything about what they emit, and we already run a parity check against `en` that catches the real failure mode.

Thanks for digging into this, though 👍

PR #6064, opened 2026-06-05
fix(toJSONSchema): preserve default value through pipe/transform in input mode

### LINKED ISSUE #6049: v4: `toJSONSchema` drops `default` when the defaulted schema contains a transform/pipe
### Summary

### Zod version

`zod@4.4.3`

### Minimal reproduction

```ts
import { z } from "zod";

```

### Expected

```json
```

### Actual

### Notes

  ```ts
  z.toJSONSchema(
    { io: "input" },
  );
  ```

VihAMBR:
## Problem

`z.string().transform(s => s).default("hello")` passed to `z.toJSONSchema({ io: "input" })` silently drops the `default` keyword:

```ts
z.toJSONSchema(z.string().transform(s => s).default("hello"), { io: "input" })
// emits { type: "string" }  ← default: "hello" is lost

// without the transform it works correctly:
z.toJSONSchema(z.string().default("hello"), { io: "input" })
```

Fixes #6049.

## Root cause

`process()` in `to-json-schema.ts` runs `isTransforming(schema)` after every processor and unconditionally deletes `result.schema.default` when the check returns `true`. For a `ZodDefault` whose inner type is a `ZodPipe` containing a transform, `isTransforming` returns `true` — so the `default` value that `defaultProcessor` just set on the schema is immediately wiped.

The deletion was intended for cases where `default` is inherited from a *pipe* (output-side default), not for `ZodDefault`, which always declares an explicit input-side default.

## Fix

Skip the `delete result.schema.default` step when `def.type === "default"`. The `ZodDefault` processor always sets `default` intentionally; removing it here is wrong regardless of whether the inner type transforms.

```ts
// to-json-schema.ts
  // ZodDefault explicitly declares the input-side default — preserve it even through transforms
  if (def.type !== "default") {
    delete result.schema.default;
  }
}
```

## Test changes

- Updated existing snapshot for `a.default(1234)` (type-changing transform): with this fix, `default: 1234` now appears in the input schema. The `default` keyword in JSON Schema is advisory and does not need to validate against the schema type, so this is spec-compliant and more informative for tooling.
- Added a new test reproducing the exact case from #6049.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2628,6 +2628,35 @@ test("defaults/prefaults", () => {
   expect(z.toJSONSchema(c, { io: "input" })).toMatchInlineSnapshot(`
     {
       "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": 1234,
+      "type": "string",
+    }
+  `);
+});
+
+test("default preserved through pipe/transform (issue #6049)", () => {
+  // z.string().transform(s => s).default("hello") should emit default: "hello" in io: "input" mode
+  const schema = z
+    .string()
+    .transform((s) => s)
+    .default("hello");
+  expect(z.toJSONSchema(schema, { io: "input" })).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": "hello",
+      "type": "string",
+    }
+  `);
+  // With a proper output terminus, default appears in output mode too
+  const schemaWithPipe = z
+    .string()
+    .transform((s) => s.toUpperCase())
+    .pipe(z.string())
+    .default("hello");
+  expect(z.toJSONSchema(schemaWithPipe, { io: "input" })).toMatchInlineSnapshot(`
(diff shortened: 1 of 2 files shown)


wahajahmed010 (commented):
A clean, well-documented fix. The root-cause analysis is excellent — `isTransforming` runs after every processor and the unconditional `delete result.schema.default` was colliding with `defaultProcessor`.

**Logic:** Correct. Checking `def.type !== "default"` is the right discriminant — only ZodDefault intentionally sets `result.schema.default` on the input side. One subtle edge case: if a future processor also sets `default` on input for a non-ZodDefault schema, this guard would not cover it. A snapshot approach ("was it set before?") would be more robust, but that is future-proofing, not a current bug.

**Tests:** Snapshot update and new `#6049` repro test look correct. The pipe-through-transform test covering both sides is thorough.

**Edge case:** The static `def.type` check assumes ZodDefault def.type is always `"default"`. From reading the Zod source — confirmed, it is. Safe.

Nice work — minimal, correct, well-explained.

VihAMBR:
Reopening from my main account (VihaanAgarwal).

VihAMBR:
Reopened as #6105.

PR #6065, opened 2026-06-06
feat: add .exactPartial() to ZodObject

andersk:
`.exactPartial()` mirrors `.partial()` but wraps fields in `ZodExactOptional` instead of `ZodOptional`, so missing keys are accepted while explicit `undefined` is rejected.

`.exactPartial()` therefore matches TypeScript’s builtin [`Partial<>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype) utility type, whereas `.partial()` only matches `Partial<>` when [`exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes) is off.  Also, the difference may be important at runtime, e.g., `{k: 1, ...{}}.k !== {k: 1, ...{k: undefined}}.k`.

- Fixes #5983.


wahajahmed010 (commented):
**Review:** Clean implementation. Using `ZodExactOptional` with `util.partial()` is exactly right.

**One issue before merge:**
The error message in the refinement case says `.partial()` but users calling `.exactPartial()` will see the wrong method name. The shared `util.partial()` function likely hardcodes the `.partial()` string — should be generalized or pass the caller name through.

Tests are solid: type inference, wrapper checks, absent vs undefined, masked variant, refinement rejection — both classic and mini APIs covered. Nice work.

andersk:
@wahajahmed010 I improved the error message by passing the caller name.

colinhacks:
Merging this. Pushed three commits on top: docs for `exactOptional()` and `.exactPartial()` in the API reference, a test pinning `.exactPartial()` against TypeScript's `Partial<>` — that equivalence is the point of the method and wasn't covered — and a change passing the caller name to `util.partial()` positionally rather than as an options object.

`exactOptional()` had never been documented at all, which is most of why this was hard to find in the first place. Also merged `main` in so CI ran against what actually lands.

PR #6067, opened 2026-06-07
fix(toJSONSchema): preserve default value through transform in input mode

### LINKED ISSUE #6049: v4: `toJSONSchema` drops `default` when the defaulted schema contains a transform/pipe

chatman-media:
## Summary

When a `ZodDefault` schema wraps a schema that contains a `.transform()` or `.pipe()`, `z.toJSONSchema(..., { io: "input" })` was incorrectly dropping the `default` keyword.

```ts
z.toJSONSchema(withoutTransform, { io: "input" });

z.toJSONSchema(withTransform, { io: "input" });
// { type: "string" }  ✗ — default: "hello" is lost
```

The same bug affects nested transforms inside a default:

```ts
z.toJSONSchema(
  { io: "input" },
);
// default: false is dropped
```

## Root cause

In `to-json-schema.ts`, the `process()` function calls `isTransforming(schema)` after each processor runs, and when it returns `true` it unconditionally deletes `result.schema.default`:

```ts
  delete result.schema.default;  // ← too aggressive
}
```

`isTransforming` recurses into the inner type of a `ZodDefault`, so it returns `true` whenever the wrapped schema contains any transform — even if the transform is an identity function. This wipes the `default` that `defaultProcessor` just set.

The deletion was intended for `ZodPipe`/`ZodTransform` nodes where a default comes from the output side of a pipe. For `ZodDefault`, the user has explicitly declared the default value, and it belongs in the JSON Schema output regardless of what the inner type does.

## Fix

Skip the `delete result.schema.default` step when `def.type === "default"`:

```ts
  if (def.type !== "default") {
    delete result.schema.default;
  }
}
```

## Test plan

- Added a new test `"default is preserved through identity transform in input mode"` covering the exact cases from #6049 (simple pass-through transform and nested transform inside a union default).
- Updated the existing `"defaults/prefaults"` snapshot for `c = a.default(1234)` to include `default: 1234` in io:input mode. The JSON Schema `default` keyword is purely advisory and not required to be valid against the schema type, so this is spec-compliant.
- All 3813 existing tests continue to pass.

Closes #6049

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2628,7 +2628,49 @@ test("defaults/prefaults", () => {
   expect(z.toJSONSchema(c, { io: "input" })).toMatchInlineSnapshot(`
     {
       "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": 1234,
       "type": "string",
+    }
+  `);
+});
+
+test("default is preserved through identity transform in input mode", () => {
+  // z.string().transform(s => s) is a pass-through — the input type is still string
+  // and the default "hello" is a valid string, so it should appear in io:"input"
+  const withTransform = z
+    .string()
+    .transform((s) => s)
+    .default("hello");
+  expect(z.toJSONSchema(withTransform, { io: "input" })).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": "hello",
+      "type": "string",
+    }
+  `);
+
+  // nested: union containing an object with a transform inside a default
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing in favor of #6374.

Exempting `ZodDefault` from the check reintroduces the bug #4557 fixed — this PR rewrites the existing `defaults/prefaults` snapshot so `z.string().transform(v => v.length).pipe(z.number()).default(1234)` emits `default: 1234` on a `type: "string"` input schema. `.default()` is typed against the output type, so it isn't generally valid input; #6374 keeps it only when it actually is, which fixes both cases in the issue without giving that back.

PR #6072, opened 2026-06-08
fix: surface @deprecated on re-exported compat aliases

### LINKED ISSUE #6038: @deprecated JSDocs not propagated
Several exported members are deprecated with JSDocs but those deprecations are not communicated properly. IDEs do not show deprecation messages in tooltips and do not display the deprecated member with a strikethrough.

For example, `z.ZodTypeAny` is deprecated in v4, but shows normally in VSCode and Cursor. Following the definition takes us directly to the definition of `z.ZodType`. It is only if we happen to look at v4/classic/compat.ts do we see
```ts
export type {
  /** @deprecated Use z.ZodType (without generics) instead. */
  ZodType as ZodTypeAny,
  /** @deprecated Use `z.ZodType` */
  ZodType as ZodSchema,
  /** @deprecated Use `z.ZodType` */
  ZodType as Schema,
};
```

Members that are re-exported in this way lose their `@deprecated` tag and implementers only see the deprecated member as an alias for the member they link to.

I have a simple PR ready, but the contributing guide suggests I make an issue first to get feedback/approval.

MahinAnowar:
Closes #6038.

`@deprecated` written on an `export { X as Y }` / `export type { X as Y }` specifier is dropped once the alias is re-exported again through the package entry, so none of the Zod 3 compat aliases actually rendered as deprecated in editors — `ZodTypeAny`, `ZodSchema`, `Schema`, `TypeOf`, `Infer`, `ZodFirstPartySchemaTypes` (compat.ts), `ZodFlattenedError`, `ZodFormattedError`, `ZodErrorMap` (errors.ts), and `_max` / `_min` (core/api.ts).

This redeclares each as a standalone aliased declaration with the JSDoc attached directly — the same form `inferFlattenedErrors`, `inferFormattedError`, and `BRAND` already use in compat.ts — so the tag survives re-export.

Verified against the built `.d.ts` via the language service: a fixture importing these through the package entry reports **0** deprecation diagnostics (6385) before the change and **9** after (the two `_` value aliases carry the tag on their emitted `const` declarations as well).

### DIFF
--- a/packages/zod/src/v4/classic/compat.ts
+++ b/packages/zod/src/v4/classic/compat.ts
@@ -3,14 +3,12 @@
 import * as core from "../core/index.js";
 import type { ZodType } from "./schemas.js";
 
-export type {
-  /** @deprecated Use `z.output<T>` instead. */
-  output as TypeOf,
-  /** @deprecated Use `z.output<T>` instead. */
-  output as Infer,
-  /** @deprecated Use `z.core.$$ZodFirstPartyTypes` instead */
-  $ZodTypes as ZodFirstPartySchemaTypes,
-} from "../core/index.js";
+/** @deprecated Use `z.output<T>` instead. */
+export type TypeOf<T> = core.output<T>;
+/** @deprecated Use `z.output<T>` instead. */
+export type Infer<T> = core.output<T>;
+/** @deprecated Use `z.core.$$ZodFirstPartyTypes` instead */
+export type ZodFirstPartySchemaTypes = core.$ZodTypes;
 
 /** @deprecated Use the raw string literal codes instead, e.g. "invalid_type". */
 export const ZodIssueCode = {
@@ -54,14 +52,12 @@ export function getErrorMap(): core.$ZodErrorMap<core.$ZodIssue> | undefined {
   return core.config().customError;
 }
 
-export type {
-  /** @deprecated Use z.ZodType (without generics) instead. */
-  ZodType as ZodTypeAny,
-  /** @deprecated Use `z.ZodType` */
-  ZodType as ZodSchema,
-  /** @deprecated Use `z.ZodType` */
-  ZodType as Schema,
-};
+/** @deprecated Use z.ZodType (without generics) instead. */
+export type ZodTypeAny = ZodType;
+/** @deprecated Use `z.ZodType` */
+export type ZodSchema = ZodType;
+/** @deprecated Use `z.ZodType` */
+export type Schema = ZodType;
 
 /** Included for Zod 3 compatibility */
 export type ZodRawShape = core.$ZodShape;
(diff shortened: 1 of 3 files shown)


wahajahmed010 (commented):
This is a well-scoped fix for a real DX issue. Verified the problem — JSDoc `@deprecated` tags on `export type { X as Y }` re-exports don\u2019t survive the TypeScript declaration emitter in many cases. The approach (redeclare aliases as standalone type aliases with the JSDoc attached directly) is the standard workaround.

**Logic:** Correct. The change from `export type { X as Y }` to `export type Y = X` preserves semantic equivalence for consumers while letting the JSDoc tag survive.

**One thing to double-check:** `_max` and `_min` changed from `export { ... as ... }` (re-exports of existing function declarations) to `export const _max = _lte`. This changes the runtime semantics slightly — they are now new const bindings rather than alias re-exports. In practice this is fine (same reference for `const`), but verify that tools/transformers that tree-shake based on export origin don\u2019t misbehave. Not a blocker, just worth noting.

**Tests:** No tests added. For a pure type-declaration change this makes sense, but adding a quick `.d.ts` snapshot or a language-service check would prevent regressions.

**Verified count:** 8 aliases fixed (TypeOf, Infer, ZodFirstPartySchemaTypes, ZodTypeAny, ZodSchema, Schema, ZodFlattenedError, ZodFormattedError, ZodErrorMap) + 2 `_max`/`_min` value aliases. That matches the \u201c9 after\u201d claim in the description.

Clean fix, good documentation.

MahinAnowar:
Thanks for taking a look!

On `_max`/`_min`: agreed it's a non-issue — `export const _max = _lte` just aliases an `@__NO_SIDE_EFFECTS__`-annotated function, so bundlers tree-shake it the same as the previous `export { _lte as _max }`; the circular-deps check and both TS test jobs stay green.

On tests: a `@deprecated` symbol doesn't produce a compile error, so there's no clean runtime/type-assertion hook. I verified it at the declaration level via the language service instead — a fixture importing these through the package entry reports 0 deprecation diagnostics (6385) on `main` and 9 after this change. Happy to add a committed `.d.ts`/language-service check if the maintainers want one, though it'd need a build step the current suite doesn't do.

colinhacks:
Pushed two changes on top of this.

`ZodTypeAny`, `ZodSchema` and `Schema` were re-exports of `ZodType`, so they inherited its three defaulted type parameters. Redeclaring them as bare aliases surfaces the tag but collapses them to zero arity, which breaks `z.ZodSchema<string>` with TS2315 — the Zod 3 idiom this compat layer exists to serve. I restated the parameters and forwarded them.

I also reverted the `core/api.ts` hunk. `_max` and `_min` are value re-exports, and TypeScript already reports their deprecation through the alias — only type-only export specifiers drop the tag, so that change fixed nothing and traded a hoisted function re-export for a TDZ-bound const.

Verified against the built `.d.ts` with the language service: all nine type aliases now report deprecated where none did before, and `_max`/`_min` still do. Merging once CI is green. Also closed #6265, which covered three of the same aliases.

PR #6081, opened 2026-06-11
fix(treeifyError): guard against Object.prototype inherited keys in path segments

### LINKED ISSUE #6070: treeifyError throws TypeError when an issue path contains a key inherited from Object.prototype (e.g. toString)
`treeifyError` throws a `TypeError` whenever a validation issue's `path` contains a segment whose name matches a property inherited from `Object.prototype` (for example `toString`, `valueOf`, `constructor`, or `hasOwnProperty`).

Minimal reproduction:

```
import zod, { treeifyError } from 'zod';

const schema = zod.object({ toString: zod.string() });

if (!result.success) {
    treeifyError(result.error); // throws TypeError
}
```

The parse correctly fails: `{}.toString` resolves to the inherited `Object.prototype.toString` function, so the `toString` field is reported as `invalid_type (expected string, received function)` with `path: ["toString"]`. That part is expected. The bug is that `treeifyError` then throws while building the tree:

```
node_modules/zod/v4/core/errors.js:113
                        curr.errors.push(mapper(issue));
                                    ^
    at processError (node_modules/zod/v4/core/errors.js:113:37)
    at treeifyError (node_modules/zod/v4/core/errors.js:120:5)
```

Root cause:

`treeifyError` builds the tree using a plain object as a dictionary:

```
curr.properties ?? (curr.properties = {});
(_a = curr.properties)[el] ?? (_a[el] = { errors: [] });
curr = curr.properties[el];
```

When the path segment `el` is an `Object.prototype` key, `curr.properties[el]` returns the inherited member (e.g. the `toString` function) instead of `undefined`. The `??` short-circuits, so the node is never created, and `curr` is assigned the inherited function which has no `errors` array. The following `curr.errors.push(...)` then throws.
A null-prototype object (`Object.create(null)`) or a `Map` for `properties`, or an `Object.hasOwn` guard, would fix it.

Environment:
zod: 4.4.3
Node.js: v24.15.0

1795771535y-cell:
Closes #6070

When a path segment (e.g. `toString`, `valueOf`, `constructor`) matches an `Object.prototype` property, `curr.properties[el]` returns the inherited member instead of `undefined`. The `??=` short-circuits, the node is never created, and `curr` is assigned the inherited function (which has no `errors` array).

Fix: add an `Object.hasOwn` guard before accessing `curr.properties[el]` to ensure inherited prototype members don't shadow real property lookups.

### DIFF
--- a/packages/zod/src/v4/core/errors.ts
+++ b/packages/zod/src/v4/core/errors.ts
@@ -370,7 +370,10 @@ export function treeifyError<T, U>(error: $ZodError<T>, mapper = (issue: $ZodIss
           const terminal = i === fullpath.length - 1;
           if (typeof el === "string") {
             curr.properties ??= {};
-            curr.properties[el] ??= { errors: [] };
+            // Guard against prototype-polluted keys (toString, valueOf, etc.)
+            if (!Object.hasOwn(curr.properties, el)) {
+              curr.properties[el] = { errors: [] };
+            }
             curr = curr.properties[el];
           } else {
             curr.items ??= [];
--- a/packages/zod/src/v4/core/errors.ts
+++ b/packages/zod/src/v4/core/errors.ts
@@ -371,7 +371,7 @@ export function treeifyError<T, U>(error: $ZodError<T>, mapper = (issue: $ZodIss
           if (typeof el === "string") {
             curr.properties ??= {};
             // Guard against prototype-polluted keys (toString, valueOf, etc.)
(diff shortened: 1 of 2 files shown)


1795771535y-cell:
@CodeRabbit Thank you for the suggestion! I've added a test case in \packages/zod/src/v4/core/tests/prototype-keys.test.ts\ that verifies \	reeifyError\ correctly handles Object.prototype inherited keys like \	oString\.

PR #6082, opened 2026-06-11
fix(json-schema): encode `$ref` pointer per RFC 6901

tsushanth:
Fixes #6027.

The `$ref` URI fragment built in `makeURI` interpolated the meta `id` directly into the pointer path. RFC 6901 §3 requires reference tokens to escape `~` as `~0` and `/` as `~1`; without that, a meta `id` containing either character produces a `$ref` that consumers can't dereference back to the `$defs` key.

The encoding applies to the pointer path only — the `$defs` key itself is a plain JSON object key and stays raw — and the replacement order is pinned (`~` first) so a `/` turning into `~1` is not re-encoded as `~01`.

Before:

```jsonc
{
  "properties": { "User": { "$ref": "#/$defs/Shared/User~" } },
  "$defs": { "Shared/User~": {} }
}
```

After:

```jsonc
{
  "properties": { "User": { "$ref": "#/$defs/Shared~1User~0" } },
  "$defs": { "Shared/User~": {} }
}
```

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2212,6 +2212,28 @@ test("id is observable in override callback", () => {
   expect(seenIds).toContain("Inner");
 });
 
+test("$ref encodes `/` and `~` per RFC 6901 (issue #6027)", () => {
+  // RFC 6901 §3 requires `~` and `/` to be escaped inside reference tokens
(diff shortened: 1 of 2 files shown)


tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

PR #6084, opened 2026-06-12
fix(toJSONSchema): RFC 6901 encode ref pointer tokens when id contains / or ~

1795771535y-cell:
Closes #6027

Per RFC 6901, JSON Pointer reference tokens must escape ~ as ~0 and / as ~1.

Previously, toJSONSchema() used the raw schema id (from .meta()) directly in ref paths without encoding these characters. This broke JSON Schema consumption when the id contained / or ~.

Example:
```ts
const User = z.object({ name: z.string() }).meta({ id: 'Shared/User~' });
z.toJSONSchema(z.object({ User }));
```

Before: ref: `#/defs/Shared/User~` (invalid pointer)
After: ref: `#/defs/Shared~1User~0` (valid RFC 6901 pointer)

The defs key remains as the original unencoded id.


1795771535y-cell:
@Pullfrog Thank you for the suggestion! I've added a test case in \packages/zod/src/v4/classic/tests/to-json-schema.test.ts\ that verifies RFC 6901 encoding works correctly when the schema id contains \/\ and \~\ characters.

PR #6085, opened 2026-06-12
feat(v4): z.compile — ahead-of-time schema compilation

colinhacks:
Ahead-of-time compilation for v4 schemas. Two entry points, intentionally non-overlapping:

- `z.compile(schema)` — returns a cloned schema whose parse path runs an AOT-compiled fast path. The clone is a normal `ZodType` (`.parse`, `.safeParse`, composition, Standard Schema all work); the original is untouched. Derivations of the clone don't inherit the fast path — compile the final schema.
- `import "zod/compile"` — global mode. Installs a post-processor that wraps every newly-constructed schema with a one-shot lazy compile shim (compile on first parse), mirroring the existing `generateFastpass` pattern so builder-chain intermediates never pay compile cost.

The core design decision is the failure model: the compiled fast path is a happy-path validator that returns the parsed output or an `INVALID` sentinel. On `INVALID`, the wrapper re-runs the original interpreter to produce the canonical `ZodError`. That gives 100% error parity by construction — there is no second error-path codegen to maintain (the main reason this is preferable to arktype-style dual `Allows`/`Apply` compilation). User `.refine`/`.transform` callbacks run at most twice on invalid input, matching the existing Standard Schema sync-then-async bound.

Anything the compiler can't model exactly throws `ZodCompileUnsupportedError` at codegen time — no silently-dead fast paths. Containers island unsupported children; unions and discriminated unions deliberately don't island, since a falsely-rejecting compiled branch would corrupt match semantics. Forward direction only, no async, and global mode respects `config().jitless` so `import "zod/compile"` is inert under CSP.

## Performance

Median **2.4x** across a 55-schema benchmark (`packages/bench/compile-matrix.ts`), scaling with how much work the schema does per parse — what compilation removes is per-node dispatch and allocation, not the checks:

| | |
| --- | --- |
| Biggest wins | `z.array(z.string())` x100 **13.9x**, `z.array(z.object())` x50 **9.3x**, 20-key object **8.9x**, object union **8.4x**, `.pipe()` **7.7x** |
| Typical | nested object 4.5x, tuple 4.8x, discriminated union 4.3x, intersection 4.3x, strict object 3.0x, flat 5-key object 2.5x |
| No win | bare `z.string()` — one `typeof` has no dispatch to remove. Leaves nested in a container are inlined into the parent and still benefit. |
| Forced fallbacks | recursive schemas, `z.xor` — 1.00x, the bypass checks cost nothing measurable |

Against arktype 2.1.19, matched on contract:

| case | ark, returns input | ark, rejects unknown | ark, allocates | `z.object()` | `z.strictObject()` |
| --- | --- | --- | --- | --- | --- |
| simple 2-key object | 130.0M | 13.0M | 1.2M | **64.3M** | **44.6M** |
| nested object (moltar) | 19.9M | 6.0M | 174k | **30.5M** | **17.6M** |

Both rejecting undeclared keys, zod is 2.9x faster on moltar and 3.4x on the simple object. Arktype's fast path validates in place — it neither allocates nor strips, a weaker contract than anything zod offers — and even against that, `z.object()` wins on moltar (30.5M vs 19.9M), losing only the flat two-key case. Arktype can build fresh output via `.onDeepUndeclaredKey("delete")`, but that costs it ~20x its own fast path, so that column says more about an unoptimized mode on their side than about validation speed.

Benchmarking this needed four separate corrections before the numbers meant anything: results must escape (a discarded one lets V8 delete the parse outright — `z.string()` measured 625M ops/sec), inputs must arrive through an array load (a constant makes the call loop-invariant and V8 hoists interpreter code far more readily than a `new Function` closure), how many schemas share the process is the single largest lever on the result, and runtime/compiled must be interleaved because absolute throughput drifts tens of percent between runs. `wiki/compile.md` documents all four; earlier figures in `wiki/compile-plan.md` predate them and shouldn't be quoted.

## Cost

Three axes, measured against `origin/main` on one machine.

**Bundle**, esbuild `--bundle --minify` + gzip under `--conditions=@zod/source`:

| fixture | main | PR | delta |
| --- | --- | --- | --- |
| `zod-mini-boolean` | 2789 | 2814 | +25 (+0.9%) |
| `zod-mini-string` | 3115 | 3136 | +21 (+0.7%) |
| `zod-mini-object` | 4216 | 4239 | +23 (+0.5%) |

That is the `postProcessor` hook in `$constructor`, which is intrinsic to the design and paid whether or not `zod/compile` is imported. `packages/treeshake/bundle-size.test.ts` now holds these fixtures to a ceiling, so a regression fails CI instead of waiting for a reviewer to measure. An earlier measurement put mini at +64; the difference was `urlCanParse`, a module-scope const whose initializer calls `.bind()` and so could not be tree-shaken. It resolves on first use now. The ~2000-line compiler itself tree-shakes cleanly.

**Construction**, schemas built per second, three runs each:

| | main | PR |
| --- | --- | --- |
| classic `z.string()` | 1.62M / 1.57M / 1.66M | 1.61M / 1.68M / 1.60M |
| classic `z.object` (5 keys) | 121k / 120k / 126k | 127k / 120k / 122k |
| mini `z.string()` | 1.57M / 1.63M / 1.68M | 1.70M / 1.63M / 1.53M |
| mini `z.object` (5 keys) | 144k / 145k / 156k | 132k / 161k / 146k |

Every pair overlaps and the differences fall on both sides, so the hook's construction cost is below this harness's noise floor (~±8%). Mini is listed separately because a fixed cost is a larger fraction there.

**Parse** is the benchmark section above.

## Correctness

The differential harness runs the entire v4 test corpus a second time with global compile enabled (`vitest.compile.config.ts`, wired into the default `pnpm test`), asserting per fixture that the fast path actually produced the value rather than silently falling back. That's what caught every divergence fixed here:

- `z.creditCard()` compiled to its shape regex without the Luhn digit, and `z.record(z.email(), …)` accepted every key — both because a string *format* lives on the def rather than in `checks`. `generateStringCheck` kept a second copy of the format table; it now delegates, and the `if (def.pattern)` catch-all is an allowlist so an unclassified format loses its fast path instead of silently accepting more than the runtime.
- A compiled strict object skipped an undeclared `__proto__` while scanning for unknown keys, bypassing the hardening in #6221.
- A schema whose subtree contains a cycle followed reference cycles in the input until the stack ran out. Those now go to the runtime, whose memoizer terminates them; a compiled node also bails when handed a back edge so a transform on the cycle still raises `$ZodCyclicError`.
- `.default(v).readonly().optional()` dropped the default, and `.default(v)` compiled to `undefined` whenever `defaultValue` was a data property rather than an accessor (which is how `deepPartial` stores it).

The optional path is now keyed on the `optin` ladder from #6419 rather than on the inner's type, which is the structural property it was missing.

Records with a validating key schema also compile now — `z.email()`, `z.string().min(2)`, `z.number()`, template literals and key transforms, in strict and loose mode — which was the largest remaining fallback in the corpus.

Full rationale, scope cuts and open questions are in `wiki/compile.md`.


colinhacks:
@pullfrog review

PR #6086, opened 2026-06-12
fix: reject bare IPv6 addresses in z.string().url()

### LINKED ISSUE #6031: Bug: `z.string().url()` incorrectly validates IPv6 link-local addresses ("fe80:0000:0000:0000:0000:0000:0000:0001") as valid URLs

farhanswitch:
## Fixes

Fixes #6031

## Summary

`z.string().url()` incorrectly accepts certain bare IPv6 addresses such as `fe80::1` as valid URLs.

The root cause is that JavaScript's `new URL()` parser interprets the first IPv6 hex group as a URI scheme. For example, in `fe80::1`, the `fe80:` portion satisfies the URI scheme grammar because it starts with a letter and is followed by alphanumeric characters. As a result, `new URL("fe80::1")` succeeds and produces a URL object with:

```ts
{
  protocol: "fe80:"
}
```

This issue only affects bare IPv6 addresses whose first segment happens to be a valid scheme name (for example `fe80`, `abcd`, etc.). Addresses that begin with `::` or a digit, such as `::1` or `2001:db8::1`, already fail during `new URL()` parsing.

## Solution

After `new URL()` successfully parses the input, this change validates the original input against the existing IPv6 regex.

This ensures that bare IPv6 addresses are rejected while avoiding false positives for short scheme-like strings such as `c:`.

## Behavior Changes

| Input | Before | After |
|---------|---------|---------|
| `fe80::1` | ✅ Accepted | ❌ Rejected |
| `fe80::abcd:1234` | ✅ Accepted | ❌ Rejected |
| `fe80:0000:...:0001` | ✅ Accepted | ❌ Rejected |
| `abcd::1` | ✅ Accepted | ❌ Rejected |
| `::1` | ❌ Rejected | ❌ Rejected |
| `2001:db8::1` | ❌ Rejected | ❌ Rejected |
| `c:` | ✅ Accepted | ✅ Accepted |
| `http://[fe80::1]` | ✅ Accepted | ✅ Accepted |

## Scope

This fix has been applied to:

- `src/v3/types.ts`
- `src/v4/core/schemas.ts`

Corresponding test cases have also been added for both versions.

### DIFF
--- a/packages/zod/src/v3/tests/string.test.ts
+++ b/packages/zod/src/v3/tests/string.test.ts
@@ -300,6 +300,19 @@ test("url validations", () => {
   expect(() => url.parse("asdf")).toThrow();
   expect(() => url.parse("https:/")).toThrow();
   expect(() => url.parse("asdfj@lkjsdf.com")).toThrow();
+
+  // bare IPv6 addresses must not be accepted as URLs
+  expect(() => url.parse("fe80::1")).toThrow();
+  expect(() => url.parse("fe80::abcd:1234")).toThrow();
+  expect(() => url.parse("fe80:0000:0000:0000:0000:0000:0000:0001")).toThrow();
+  expect(() => url.parse("abcd::1")).toThrow();
+  expect(() => url.parse("::1")).toThrow();
+  expect(() => url.parse("2001:db8::1")).toThrow();
+
+  // valid URLs containing IPv6 in brackets must still be accepted
+  url.parse("http://[::1]");
+  url.parse("http://[2001:db8::1]");
+  url.parse("http://[fe80::1]");
 });
 
 test("url error overrides", () => {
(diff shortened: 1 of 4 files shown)


colinhacks:
Closing in favor of #6132, which fixes the same bug with the same oracle `z.ipv6()` uses at runtime — `new URL("http://[...]")` rather than `regexes.ipv6`. That matters: `regexes.ipv6` has no IPv4-tail alternative, so this version still accepts `fe80::ffff:192.168.1.1` and `abcd:0:0:0:0:0:1.2.3.4`, and leaves `z.url()` disagreeing with `z.ipv6()` on them.

Good diagnosis of the scheme-parsing root cause, though — thanks for digging into it 👍

colinhacks:
Following up — I ended up closing #6132 too and closing #6031 as working-as-intended. `z.url()` is documented as deferring to the platform's `new URL()`, which parses `fe80::1` as a scheme, so accepting it is the contract rather than a bug; `z.httpUrl()` is the strict API. Nothing wrong with your read of the mechanism.

PR #6090, opened 2026-06-13
fix: escape JSON Pointer tokens in toJSONSchema $ref paths

Malaybhai11:
Per RFC 6901, JSON Pointer path segments must escape  as  and  as . The  function in  was using the raw schema id without escaping, producing invalid  URIs for ids containing these characters.

Added  helper and applied it to both the self-contained and external registry ref paths.

Fixes #6027

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -3129,3 +3129,183 @@ test("recursive lazy with describe does not stack overflow", () => {
   expect(result).toBeDefined();
   expect(result.$defs).toBeDefined();
 });
+
+test("JSON Pointer escaping - tilde in schema id", () => {
+  const tildeSchema = z.string().meta({ id: "my~tilde" });
+  const result = z.toJSONSchema(
+    z.object({
+      value: tildeSchema,
+    })
+  );
+  expect(result).toMatchInlineSnapshot(`
+    {
+      "$defs": {
+        "my~tilde": {
+          "type": "string",
+        },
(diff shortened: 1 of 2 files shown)


Malaybhai11:
Thanks for the review! I've added test coverage for the JSON Pointer escaping:

- `escapeJsonPointer` behavior: tilde → `~0`, slash → `~1`, combined special chars
- `toJSONSchema` integration: schema IDs with `~` and `/` produce valid `$ref` paths
- Registry and reused-ref modes both covered

All 170 tests pass. Let me know if you want any additional cases.

wahajahmed010 (commented):
### Review: #6090 - Fix: escape JSON Pointer tokens in toJSONSchema $ref paths

**2 files, +186 additions (mostly tests) — clean RFC 6901 compliance fix.**

**What works well:**
- Correctly encodes ~ -> ~0 and / -> ~1 in $ref pointer paths
- Comprehensive test coverage: tilde-only, slash-only, mixed, complex multi-character, reused refs, registry schemas
- Inline snapshots make expected output crystal clear
- Minimal production code change (4 lines inserted)

**Perf nit:**
- escapeJsonPointer is defined inside extractDefs, so a new function closure is created each call. Hoisting to module scope would avoid that allocation.

**Verdict:** Correct fix, good tests. Ready to merge.

Malaybhai11:
@wahajahmed010 Thanks for the thorough review! Good catch on the closure allocation — I'll hoist  to module scope in a follow-up commit. Appreciate the detailed feedback on the test coverage too.

colinhacks:

PR #6094, opened 2026-06-16
fix(6070):treeifyError throws TypeError when an issue path contains a key inherited from Object.prototype (e.g. toString)

zhangsan582:
Fixes `z.treeifyError` path handling when object keys collide with inherited `Object.prototype` properties like `toString` or `constructor`.

  `properties` containers now use null-prototype records, so those path segments are treated as own keys in the generated error tree instead of resolving inherited object
  methods. Added regression coverage for both direct object paths and mixed object/array paths.


zhangsan582:
cancle

PR #6098, opened 2026-06-17
fix(preprocess): infer input type from the preprocessor

ValeraS:
starting from version 4.4.2 the preprocess method does not infer input type from preprocessor function. Documentation says:

By default, the input type of a z.preprocess() schema is unknown, since the preprocessor is expected to handle arbitrary input. To narrow the input type, annotate the preprocessor's parameter directly:

```ts
const trimmed = z.preprocess(
  (val: string | null | undefined) => val?.trim() ?? "",
  z.string()
);
 
type Input = z.input<typeof trimmed>;  // string | null | undefined
type Output = z.output<typeof trimmed>; // string
```

but currently `type Input = z.input<typeof trimmed>` is `unknown`.

This PR fixes it.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -2429,12 +2429,12 @@ export function invertCodec<A extends core.SomeType, B extends core.SomeType>(co
 }
 
 // ZodPreprocess
-export interface ZodPreprocess<B extends core.SomeType = core.$ZodType>
-  extends ZodPipe<core.$ZodTransform, B>,
-    core.$ZodPreprocess<B> {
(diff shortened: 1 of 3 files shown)


colinhacks:
Closing — #5967 landed this in 9d20fdc3.

Worth naming the difference, since it's the non-obvious part: this binds both slots of the inner transform, and typing its output makes `$ZodPipeDef`'s `transform?` field contravariant in the preprocessor's return type. A narrowed `ZodPreprocess` then stops being assignable to the bare `ZodPreprocess` — `(value: number, ...)` isn't assignable to `(value: unknown, ...)`. The merged fix binds the input slot only and pins that invariant with a test.

PR #6103, opened 2026-06-18
fix(url): reject bare IPv6 addresses in z.url()

### LINKED ISSUE #6031: Bug: `z.string().url()` incorrectly validates IPv6 link-local addresses ("fe80:0000:0000:0000:0000:0000:0000:0001") as valid URLs

MerlijnW70:
Fixes #6031.

## Problem
`z.url()` accepts bare IPv6 link-local addresses such as `fe80::1` as valid URLs, while correctly rejecting `::1`, `2001:db8::1`, and `192.168.1.1`:

```js
z.url().safeParse("fe80::1");   // ✅ VALID  — bug (should be invalid)
z.url().safeParse("2001:db8::1"); // ❌ invalid — correct
```

The cause is the WHATWG `URL` constructor: a bare IPv6 address whose first group starts with a hex letter (`fe80:`) is parsed as `scheme:opaque-path`, so `new URL("fe80::1")` succeeds.

## Fix
Before normal URL handling, reject any input that is itself a valid IPv6 literal by probing `new URL("http://[<input>]")` — the same technique Zod's `ipv6()` validator already uses. A cheap pre-check (only hex digits, colons, dots and a zone `%`, and no `/`) keeps schemed URIs and ordinary URLs out of this branch, so the existing behaviour is preserved:

- rejected: `::1`, `2001:db8::1`, `fe80::1`, `fe80::abcd:1234`, full-form `fe80:0000:…:0001`, `dead:beef:…` (×8)
- still valid: `http://[2001:db8::1]`, `https://example.com`, `c:`, `mailto:foo@bar.com`, `face:b00c`

## Tests
Adds a regression test in `url.test.ts` covering both the rejected bare-IPv6 cases and the preserved valid cases. Verified red without the fix, green with it; full suite (3817 tests) passes.

### DIFF
--- a/packages/zod/src/v4/classic/tests/url.test.ts
+++ b/packages/zod/src/v4/classic/tests/url.test.ts
@@ -11,3 +11,25 @@ test("url regex", () => {
     "^example\\.com$"
   );
 });
+
+test("bare IPv6 addresses are not valid URLs (#6031)", () => {
+  const url = z.url();
+  // The URL constructor mis-parses IPv6 addresses whose first group starts
+  // with a hex letter (e.g. "fe80::1") as `scheme:opaque-path`; reject them.
+  for (const addr of [
+    "::1",
+    "2001:db8::1",
+    "fe80::1",
+    "fe80::abcd:1234",
+    "fe80:0000:0000:0000:0000:0000:0000:0001",
+    "dead:beef:dead:beef:dead:beef:dead:beef",
+  ]) {
+    expect(url.safeParse(addr).success, addr).toBe(false);
+  }
+
+  // A bracketed IPv6 host inside a real URL stays valid; so do schemed URIs
+  // whose opaque part happens to look hex-ish.
+  for (const valid of ["http://[2001:db8::1]", "https://example.com", "c:", "mailto:foo@bar.com", "face:b00c"]) {
+    expect(url.safeParse(valid).success, valid).toBe(true);
+  }
+});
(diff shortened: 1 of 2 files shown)


wahajahmed010 (commented):
## Review: colinhacks/zod#6103 — fix(url): reject bare IPv6 addresses

**Logic:** ✅ Solid. The pre-check (`includes(":") && !includes("/") && hex-only`) is a good heuristic — cheap to run, catches the edge case without touching the normal URL path. The `new URL("http://[...]")` probe is the authoritative check.

**Tests:** ✅ Covers both valid (bracketed IPv6, schemed URIs, `c:`, `mailto:`) and invalid (bare IPv6 variants). Good.

**Edge case:** What about IPv6 addresses with a zone ID suffix like `fe80::1%eth0`? The regex `[0-9a-fA-F:.%]+` allows `%`, and `new URL("http://[fe80::1%eth0]")` should work. So this is handled. ✅

**Performance:** The pre-check regex is O(n) on trimmed length. For non-IPv6 URLs (vast majority), it's a single `includes(":")` check that short-circuits. ✅

**Nit:** The `@ts-ignore` on `new URL(...)` is a bit broad — would be cleaner to cast or use a type assertion. Minor.

**Verdict:** Clean fix. Ship it.

MerlijnW70:
glad i could help

PR #6105, opened 2026-06-19
fix(toJSONSchema): preserve default value through pipe/transform in input mode

### LINKED ISSUE #6049: v4: `toJSONSchema` drops `default` when the defaulted schema contains a transform/pipe

VihaanAgarwal:
## Problem

`z.string().transform(s => s).default("hello")` passed to `z.toJSONSchema({ io: "input" })` silently drops the `default` keyword:

```ts
z.toJSONSchema(z.string().transform(s => s).default("hello"), { io: "input" })
// emits { type: "string" }  ← default: "hello" is lost

// without the transform it works correctly:
z.toJSONSchema(z.string().default("hello"), { io: "input" })
```

Fixes #6049.

## Root cause

`process()` in `to-json-schema.ts` runs `isTransforming(schema)` after every processor and unconditionally deletes `result.schema.default` when the check returns `true`. For a `ZodDefault` whose inner type is a `ZodPipe` containing a transform, `isTransforming` returns `true` — so the `default` value that `defaultProcessor` just set on the schema is immediately wiped.

The deletion was intended for cases where `default` is inherited from a *pipe* (output-side default), not for `ZodDefault`, which always declares an explicit input-side default.

## Fix

Skip the `delete result.schema.default` step when `def.type === "default"`. The `ZodDefault` processor always sets `default` intentionally; removing it here is wrong regardless of whether the inner type transforms.

```ts
// to-json-schema.ts
  // ZodDefault explicitly declares the input-side default — preserve it even through transforms
  if (def.type !== "default") {
    delete result.schema.default;
  }
}
```

## Test changes

- Updated existing snapshot for `a.default(1234)` (type-changing transform): with this fix, `default: 1234` now appears in the input schema. The `default` keyword in JSON Schema is advisory and does not need to validate against the schema type, so this is spec-compliant and more informative for tooling.
- Added a new test reproducing the exact case from #6049.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2628,6 +2628,35 @@ test("defaults/prefaults", () => {
   expect(z.toJSONSchema(c, { io: "input" })).toMatchInlineSnapshot(`
     {
       "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": 1234,
+      "type": "string",
+    }
+  `);
+});
+
+test("default preserved through pipe/transform (issue #6049)", () => {
+  // z.string().transform(s => s).default("hello") should emit default: "hello" in io: "input" mode
+  const schema = z
+    .string()
+    .transform((s) => s)
+    .default("hello");
+  expect(z.toJSONSchema(schema, { io: "input" })).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": "hello",
+      "type": "string",
+    }
+  `);
+  // With a proper output terminus, default appears in output mode too
+  const schemaWithPipe = z
+    .string()
+    .transform((s) => s.toUpperCase())
+    .pipe(z.string())
+    .default("hello");
+  expect(z.toJSONSchema(schemaWithPipe, { io: "input" })).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": "hello",
       "type": "string",
     }
   `);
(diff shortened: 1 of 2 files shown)


wahajahmed010 (commented):
## Review: colinhacks/zod#6105 — fix(toJSONSchema): preserve default value through pipe/transform

**Overall:** Small, focused fix. Correctly addresses issue #6049.

### ✅ What I like
- **Surgical change**: One conditional check in to-json-schema.ts — exactly the right scope.
- **Correct logic**: ZodDefault explicitly declares the input-side default, so it should survive the isTransforming branch. The check def.type != "default" is the right discriminator.
- **Good test coverage**: Tests both z.string().transform().default() and the pipe variant. Inline snapshots make expected output explicit.
- **Existing test updated**: The defaults/prefaults test snapshot now includes default: 1234 — confirming no regression.

### ⚠️ Minor concern
- **Edge case: nested transforms**: What about z.string().transform(s => s).transform(s => s).default("hello")? The process function recurses through the chain, so the def.type != "default" check should still work. A test for this would be nice.

### 🔒 Security / Performance
- No security concerns. No performance impact (single type check).

**Verdict:** LGTM. Clean fix, well-tested.

wahajahmed010 (approved):
## Review: zod#6105

**Verdict: ✅ Approve**

Clean fix for a subtle bug. The root cause is well-diagnosed: `isTransforming()` returns true for a `ZodDefault` wrapping a pipe with a transform, causing the `default` that `defaultProcessor` just set to be immediately deleted. The one-line fix (`if (def.type !== "default")`) is minimal and correct.

**What I like:**
- The fix is surgically precise — it only skips the deletion when the schema is explicitly a `ZodDefault`, which always declares an intentional input-side default
- The existing snapshot update (`a.default(1234)` now emits `default: 1234` in input mode) is actually more correct per JSON Schema spec (default is advisory, doesn't need to validate against type)
- Tests reproduce the exact issue from #6049 with both a simple transform and a pipe

**No concerns on:** security, performance, error handling, or breaking changes. The snapshot change is low-risk and strictly more informative for tooling.

VihaanAgarwal:
Going to close this one for now to focus my contributions elsewhere. Happy to reopen if it'd be useful down the line.

PR #6109, opened 2026-06-21
perf: optimize `$ZodObjectJIT` fastpass

TomerAberbach:
Drop the fastpass wrapper frame in `$ZodObjectJIT`, the compiled parser that backs `z.object().parse()` and everything built on it.

`generateFastpass` returned `(payload, ctx) => fn(shape, payload, ctx)`, a forwarding closure that ran on every object parse and that V8 wouldn't inline. It now returns the compiled function directly; `shape` is captured once and passed at the call site, so the forwarding frame disappears from the hot path.

Profiled with a mixed object schema (string/number/boolean/nested/array) over ~30k CPU samples:

| | Before | After |
|---|---|---|
| Wrapper frame self time | 139ms | removed |
| Total | 4.06s | 3.92s (−3.4%) |

### DIFF
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -2086,11 +2086,11 @@ export const $ZodObjectJIT: core.$constructor<$ZodObject> = /*@__PURE__*/ core.$
 
       doc.write(`payload.value = newResult;`);
       doc.write(`return payload;`);
-      const fn = doc.compile();
-      return (payload: any, ctx: any) => fn(shape, payload, ctx);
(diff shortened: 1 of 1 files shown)


colinhacks:
Closing this out. I re-ran it against its own merge base and the 3.4% doesn't reproduce; it lands as a coin flip. On current main it's a regression: `z.object` forwards five bindings now rather than three, and pushing them as arguments costs more than the frame it saves.

Right area, though. Closing those bindings into the compiled parser removes the frame and the arguments both, which is what #6464 does — −10% on a nested object, −14% on a three-key one. Thanks for looking into this 👍

TomerAberbach:
Nice, glad something came of it :)

PR #6111, opened 2026-06-22
Update object refinement extend tests

Hamjoon:
Updates object refinement tests to match extend behavior for non-overlapping keys while preserving coverage for overlapping key rejection.

Tested with:
- pnpm test packages/zod/src/v4/classic/tests/object.test.ts
- pnpm test


Hamjoon:
Closing this PR because it was opened accidentally from an internal research experiment branch. This was not intended as an upstream contribution. Sorry for the noise.

PR #6126, opened 2026-06-25
fix(zod): guard zero divisor in bigint multipleOf check

spokodev:
## Problem

A bigint `multipleOf` check with a zero divisor throws an uncaught `RangeError` out of `safeParse`:

```ts
z.bigint().multipleOf(0n).safeParse(5n);
// RangeError: Division by zero
```

This violates the documented `safeParse` contract, which states you "get back a plain result object containing either the successfully parsed data or a `ZodError`". `parse()` is affected too: it throws a `RangeError` instead of a `ZodError`.

## Asymmetry with the number path

The number path handles a zero divisor gracefully and returns a normal failed result:

```ts
z.number().multipleOf(0).safeParse(5);
// { success: false, error: ZodError }
```

It uses `util.floatSafeRemainder`, where `value % 0` evaluates to `NaN` rather than throwing. The bigint branch has no such guard, so the two numeric types behave differently for the same input shape.

## Root cause

In `packages/zod/src/v4/core/checks.ts`, the `$ZodCheckMultipleOf` bigint branch computes `payload.value % (def.value as bigint)`. In JavaScript, `bigint % 0n` throws `RangeError: Division by zero`, and that throw escapes `safeParse`.

## Fix

Guard the zero divisor before the modulo, mirroring the number path. Zero is treated as a multiple of zero; any other value is a non-multiple. This keeps the throw inside the validation result instead of crashing the parse call.

```ts
typeof payload.value === "bigint"
  ? def.value === BigInt(0)
    ? payload.value === BigInt(0)
    : payload.value % (def.value as bigint) === BigInt(0)
  : util.floatSafeRemainder(payload.value, def.value as number) === 0;
```

## Tests

Added a test in `bigint.test.ts` asserting `safeParse` returns a result object (not a throw), that `0n` is a multiple of `0n`, and that `parse` throws `ZodError`. It fails on the current `main` with `RangeError: Division by zero` and passes with the fix.

Full v4 suite stays green (`218` files, `2762` tests, no type errors), including the existing bigint and number `multipleOf` coverage.

### DIFF
--- a/packages/zod/src/v4/classic/tests/bigint.test.ts
+++ b/packages/zod/src/v4/classic/tests/bigint.test.ts
@@ -45,6 +45,20 @@ test("failing validations", () => {
   expect(() => multipleOfFive.parse(BigInt(13))).toThrow();
 });
 
+test("multipleOf zero divisor returns result object instead of throwing", () => {
+  const schema = z.bigint().multipleOf(BigInt(0));
+
+  // safeParse must return a result object, never throw a RangeError
+  const result = schema.safeParse(BigInt(5));
+  expect(result.success).toEqual(false);
+
+  // zero is a multiple of zero
+  expect(schema.safeParse(BigInt(0)).success).toEqual(true);
+
+  // parse must throw a ZodError, not a RangeError
+  expect(() => schema.parse(BigInt(5))).toThrow(z.ZodError);
+});
+
 test("min max getters", () => {
   expect(z.bigint().min(BigInt(5)).minValue).toEqual(BigInt(5));
   expect(z.bigint().min(BigInt(5)).min(BigInt(10)).minValue).toEqual(BigInt(10));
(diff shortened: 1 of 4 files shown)


colinhacks:
Pushed a commit on top, with two changes.

The guard here only covered the runtime check. The compiler generates `x % 0n` for the same schema, so a zero divisor still escaped `safeParse` as a `RangeError` under `zod/compile` — and under the compile-mode test project, which re-runs these files with global compilation on. There's no expression to emit for a zero divisor, so the compiler now refuses the check the way it already refuses a NaN comparison bound.

I also flipped the semantics: nothing is a multiple of zero, rather than `0n` being a multiple of itself. Treating `0n` as a multiple is the mathematical answer, but `z.number().multipleOf(0)` rejects every input including `0`, so keeping the bigint answer different swaps one number/bigint asymmetry for another.

Rebased onto main as well, since the branch predated the compiler entirely.

colinhacks:
Closing in favor of #6466, which fixes the same bug. This one treats `0n` as a valid multiple of `0n`, where the number branch rejects it, and the runtime guard on its own still leaves the compiled fast path throwing.

Thanks for looking into this!

PR #6129, opened 2026-06-25
fix(checks): date min/max issue origin should reflect input, not bound

### LINKED ISSUE #5980: Passing a number to ZodDate min() and max() functions causes the issue origin to become "number" (zod@4.4.3)

MerlijnW70:
## Problem

The issue `origin` produced by `z.date().min()` / `.max()` changes based on the *type of the bound* passed to `min`/`max`, rather than the schema:

```ts
z.date().min(new Date(2020, 6, 10)).safeParse(tooSmall);       // origin: date  ✅
z.date().min(new Date(2020, 6, 10).getTime()).safeParse(tooSmall); // origin: number ❌ (should be date)
```

## Root cause

In `$ZodCheckLessThan` / `$ZodCheckGreaterThan`, `origin` is derived from `numericOriginMap[typeof def.value]` — the runtime type of the bound. A numeric timestamp bound therefore yields `origin: number`, while a `Date` bound yields `date`.

## Fix

Derive `origin` from the value being validated (`payload.value`) instead of the bound, falling back to the previous bound-derived origin for inputs whose type isn't numeric/bigint/object (e.g. a failed base-type parse). The reported `minimum` / `maximum` still come from the bound and are unchanged.

## Test

Added a case in `date.test.ts` asserting `origin: date` for both `min` and `max` when the bound is a numeric timestamp. Existing date tests (which use a `Date` bound) are unaffected.

Fixes #5980

---
_Restores #6100, which closed automatically when my fork was briefly deleted. The fix commit is unchanged._

### DIFF
--- a/packages/zod/src/v4/classic/tests/date.test.ts
+++ b/packages/zod/src/v4/classic/tests/date.test.ts
@@ -53,6 +53,16 @@ test("date max", () => {
   `);
 });
 
+test("date min/max origin is 'date' for a numeric timestamp bound (#5980)", () => {
+  const minRes = z.date().min(benchmarkDate.getTime()).safeParse(beforeBenchmarkDate);
+  expect(minRes.success).toEqual(false);
+  expect(minRes.error!.issues[0]!).toHaveProperty("origin", "date");
+
+  const maxRes = z.date().max(benchmarkDate.getTime()).safeParse(afterBenchmarkDate);
+  expect(maxRes.success).toEqual(false);
+  expect(maxRes.error!.issues[0]!).toHaveProperty("origin", "date");
+});
+
 test("min max getters", () => {
   expect(minCheck.minDate).toEqual(benchmarkDate);
   expect(minCheck.min(afterBenchmarkDate).minDate).toEqual(afterBenchmarkDate);
--- a/packages/zod/src/v4/classic/tests/date.test.ts
+++ b/packages/zod/src/v4/classic/tests/date.test.ts
@@ -53,14 +53,34 @@ test("date max", () => {
   `);
 });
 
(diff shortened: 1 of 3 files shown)


colinhacks:
This is the one I'm taking for #5980 — closed #5985, #6269 and #6338 in its favor. Fixing it in `$ZodCheckGreaterThan` / `$ZodCheckLessThan` is the right layer: it's what `multiple_of` and the length and size checks in that file already do, and it's the only approach that also covers `zod/v4-mini`, where there's no `.min()` / `.max()` to patch.

One thing to tighten before this lands — the `date.test.ts` case should use `toMatchInlineSnapshot` like the two tests above it, so it pins the message as well as the origin. `Too small: expected number to be >=1667606400000` is the half of this bug users actually see.

colinhacks:
Pushed a follow-up on top. The `date.test.ts` case now uses inline snapshots like the two tests above it, so it pins the message and not just the origin — `Too small: expected number to be >=1667606400000` is the half of this users actually see. Added a mini case too, since `ZodMiniDate` has no `.min()` / `.max()` and only the core fix reaches that path.

Merging once CI comes back.

PR #6130, opened 2026-06-25
fix(regexes): allow underscores inside httpUrl hostname labels

### LINKED ISSUE #5960: `z.httpUrl()` doesn't allow underscores in hostname

MerlijnW70:
## Problem

`z.httpUrl()` rejects URLs whose hostname contains an underscore, even though the WHATWG URL parser accepts them:

```ts
const schema = z.httpUrl();
schema.safeParse(https://example.com/image.jpg);      // ✅ ok
schema.safeParse(https://foo_bar.example.com/img.jpg); // ❌ fails (should pass)
schema.safeParse(https://exa_mple.com/img.jpg);        // ❌ fails (should pass)
```

`new URL(https://foo_bar.example.com)` parses fine and yields hostname `foo_bar.example.com`, so `z.httpUrl()` was rejecting hosts the URL parser had already accepted.

## Root cause

`z.httpUrl()` validates the URL host against `regexes.domain`, whose per-label character class `[a-zA-Z0-9-]` excluded `_`.

## Fix

Add `_` to the inner character class (`[a-zA-Z0-9_-]`) so underscores are permitted **inside** a label, but not as its first or last character — mirroring exactly how hyphens are already handled. Leading/trailing-underscore labels (e.g. `_asdf.com`) remain invalid.

`regexes.domain` is only used by `z.httpUrl()`, so this change has no effect on other formats (e.g. email).

## Test

Extended the `httpurl` test in `string.test.ts`:
- positive: `foo_bar.example.com`, `exa_mple.com`, `julia_artberg.artstation.com`
- negative (boundary): `_asdf.com` still throws.

Fixes #5960

---
_Restores #6101, which closed automatically when my fork was briefly deleted. The fix commit is unchanged._

### DIFF
--- a/packages/zod/src/v4/classic/tests/string.test.ts
+++ b/packages/zod/src/v4/classic/tests/string.test.ts
@@ -430,7 +430,7 @@ test("httpurl", () => {
   const httpUrl = z.url({
     protocol: /^https?$/,
     hostname: z.regexes.domain,
-    // /^([a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,}$/
+    // /^([a-zA-Z0-9](?:[a-zA-Z0-9_-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,}$/
   });
 
   httpUrl.parse("https://example.com");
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing this out — declining. Underscores aren't valid in hostnames: RFC 952/1123 restrict host labels to letters, digits and hyphens, and no registry will issue a domain containing one. The RFC 2181 relaxation people cite covers DNS labels generally — SRV records and friends — not the host component of a URL, and `z.regexes.domain` is a strict domain check on purpose.

If you need looser host rules, `z.url({ protocol: /^https?$/, hostname: /your-own-regex/ })` takes any regex. Thanks for looking into this 👍

PR #6131, opened 2026-06-25
fix(json-schema): RFC 6901 encode $ref pointer for ids containing / or ~

MerlijnW70:
Fixes #6027.

## Problem
When a schema's `.meta({ id })` contains `/` or `~`, `z.toJSONSchema()` built the `$ref` pointer from the raw id, producing an invalid JSON Pointer:

```jsonc
"$ref": "#/$defs/Shared/User~"   // invalid per RFC 6901
```

## Fix
Per [RFC 6901](https://datatracker.ietf.org/doc/html/rfc6901), a JSON Pointer reference token must escape `~` → `~0` and `/` → `~1` (in that order). `makeURI` now encodes the def id when building the `$ref` path, in both the self-contained (`#/$defs/…`) and external (`__shared`) branches.

The `$defs`/`definitions` **object key** stays the raw id — it is a plain JSON key and needs no escaping, exactly as the issue specifies.

```jsonc
"$ref": "#/$defs/Shared~1User~0",
"$defs": { "Shared/User~": { /* … */ } }
```

## Tests
Two regression tests added in `to-json-schema.test.ts` (draft-2020-12 `$defs` and draft-04 `definitions`). Verified red without the fix, green with it. Encoding is a no-op for ordinary ids, so no existing snapshots change.

---
_Restores #6102, which closed automatically when my fork was briefly deleted. The fix commit is unchanged._

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2182,6 +2182,28 @@ test("id is stripped from $defs entries (draft-2020-12)", () => {
   expect((result.$defs?.Inner as any).id).toBeUndefined();
 });
 
+test("$ref pointer is RFC 6901 encoded when id contains / or ~ (#6027)", () => {
+  // Per RFC 6901, a JSON Pointer reference token must escape `~` -> `~0` and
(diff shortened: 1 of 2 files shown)


colinhacks:

MerlijnW70:
@colinhacks np, was a good learning exercise

PR #6132, opened 2026-06-25
fix(url): reject bare IPv6 addresses in z.url()

### LINKED ISSUE #6031: Bug: `z.string().url()` incorrectly validates IPv6 link-local addresses ("fe80:0000:0000:0000:0000:0000:0000:0001") as valid URLs

MerlijnW70:
Fixes #6031.

## Problem
`z.url()` accepts bare IPv6 link-local addresses such as `fe80::1` as valid URLs, while correctly rejecting `::1`, `2001:db8::1`, and `192.168.1.1`:

```js
z.url().safeParse("fe80::1");   // ✅ VALID  — bug (should be invalid)
z.url().safeParse("2001:db8::1"); // ❌ invalid — correct
```

The cause is the WHATWG `URL` constructor: a bare IPv6 address whose first group starts with a hex letter (`fe80:`) is parsed as `scheme:opaque-path`, so `new URL("fe80::1")` succeeds.

## Fix
Before normal URL handling, reject any input that is itself a valid IPv6 literal by probing `new URL("http://[<input>]")` — the same technique Zod's `ipv6()` validator already uses. A cheap pre-check (only hex digits, colons, dots and a zone `%`, and no `/`) keeps schemed URIs and ordinary URLs out of this branch, so the existing behaviour is preserved:

- rejected: `::1`, `2001:db8::1`, `fe80::1`, `fe80::abcd:1234`, full-form `fe80:0000:…:0001`, `dead:beef:…` (×8)
- still valid: `http://[2001:db8::1]`, `https://example.com`, `c:`, `mailto:foo@bar.com`, `face:b00c`

## Tests
Adds a regression test in `url.test.ts` covering both the rejected bare-IPv6 cases and the preserved valid cases. Verified red without the fix, green with it; full suite (3817 tests) passes.

---
_Restores #6103, which closed automatically when my fork was briefly deleted. The fix commit is unchanged._

### DIFF
--- a/packages/zod/src/v4/classic/tests/url.test.ts
+++ b/packages/zod/src/v4/classic/tests/url.test.ts
@@ -11,3 +11,25 @@ test("url regex", () => {
     "^example\\.com$"
   );
 });
+
+test("bare IPv6 addresses are not valid URLs (#6031)", () => {
+  const url = z.url();
+  // The URL constructor mis-parses IPv6 addresses whose first group starts
+  // with a hex letter (e.g. "fe80::1") as `scheme:opaque-path`; reject them.
+  for (const addr of [
+    "::1",
+    "2001:db8::1",
+    "fe80::1",
+    "fe80::abcd:1234",
+    "fe80:0000:0000:0000:0000:0000:0000:0001",
+    "dead:beef:dead:beef:dead:beef:dead:beef",
+  ]) {
+    expect(url.safeParse(addr).success, addr).toBe(false);
+  }
+
+  // A bracketed IPv6 host inside a real URL stays valid; so do schemed URIs
+  // whose opaque part happens to look hex-ish.
+  for (const valid of ["http://[2001:db8::1]", "https://example.com", "c:", "mailto:foo@bar.com", "face:b00c"]) {
+    expect(url.safeParse(valid).success, valid).toBe(true);
+  }
+});
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -484,6 +484,31 @@ export const $ZodURL: core.$constructor<$ZodURL> = /*@__PURE__*/ core.$construct
       // Trim whitespace from input
       const trimmed = payload.value.trim();
 
+      // A bare IPv6 address (e.g. "fe80::1") is not a URL. The URL constructor
+      // mis-parses addresses whose first group starts with a hex letter as
+      // `scheme:opaque-path`, so it would otherwise accept them. Reject any
+      // input that is itself a valid IPv6 literal. The cheap pre-check (only
+      // hex digits, colons, dots and zone `%`, and no `/`) keeps schemed URIs
+      // such as "c:" or "mailto:x" — and ordinary URLs — out of this branch.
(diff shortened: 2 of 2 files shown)


wahajahmed010 (commented):
## Review: fix(url): reject bare IPv6 addresses in z.url()

**Overall:** Clean, well-scoped fix. The approach is correct and the test coverage is thorough.

### ✅ Good
- **Detection strategy** is smart: cheap pre-filter (hex chars + colons, no `/`) avoids false positives for normal URLs, then validates via `new URL("http://[...]")` to confirm it's a real IPv6.
- **Edge cases covered**: `::1`, `fe80::1`, mixed hex, `dead:beef:...`, and the valid cases (`http://[2001:db8::1]`, `c:`, `mailto:`, `face:b00c`).
- **No false positives** on schemed URIs with hex-like opaque parts (`c:`, `face:b00c`).
- **Minimal diff** — 1 test file + 1 source file, no collateral changes.

### ⚠️ Minor
- The `// @ts-ignore` on `new URL(...)` is a bit broad. Consider a more targeted suppression or a type guard. Not a blocker.
- The regex pre-check `trimmed.includes(":") && !trimmed.includes("/") && /^[0-9a-fA-F:.%]+$/.test(trimmed)` could theoretically match a very long string — but in practice URL inputs are short, so this is fine.

### 🔒 Security
- No injection risk. The `new URL()` probe is safe (no network call).
- Prevents a class of URL validation bypass.

**Verdict:** Approve. Ready to merge.

MerlijnW70:
@wahajahmed010  thanks for pointing out the minors

wahajahmed010 (commented):
## Review: colinhacks/zod#6132 — fix(url): reject bare IPv6 addresses in z.url()

**Logic:** The approach is correct and well-reasoned. The WHATWG URL constructor mis-parses bare IPv6 addresses starting with hex letters as scheme:opaque-path, so probing with http://[input] is the right detection strategy. The pre-check (includes ":", no "/", only hex/colon/dot/%) is a good optimization to avoid false positives on normal URLs.

**Tests:** Covers the reported bug case (fe80::1), other bare IPv6 variants, and confirms valid URLs still pass. Good.

**Edge case:** The pre-check regex /^[0-9a-fA-F:.%]+$/ would match an empty string or just ":" — but the includes(":") guard prevents that. However, what about input like "%" or "." alone? The includes(":") check handles it. Fine.

**Performance:** The pre-check is cheap (regex + includes), and the URL probe only runs for the small subset of inputs that look IPv6-ish. No concern.

**Verdict:** Clean, minimal fix. Ship it.

colinhacks:
Closing this out — I think the current behavior is right. `z.url()` is documented as accepting any WHATWG-compatible URL, and it deliberately just defers to the platform's `new URL()`; `new URL("fe80::1")` succeeds, so accepting it is the contract working, not a bug. `mailto:` and friends pass for the same reason.

If you want the strict behavior, `z.httpUrl()` rejects every address in that issue today. Good investigation, and the diagnosis of the scheme parsing was exactly right — thanks for digging into it 👍

Issue #5003, opened 2025-07-29
toJSONSchema .catch() returning as required


samchungy:
Looks like [4.0.6 changed ZodCatch](https://github.com/colinhacks/zod/commit/7dd7484802b1351c8b81d3d523aadd876fcdf73e#diff-f84405ad925758ccb224a20fe7ca8c6ee375d7faa88431059e008077550f8cd9L3332) due to a bug with how it parses, fair enough.

However:

```ts
toJSONSchema(z.object({ a: z.string().catch('foo') }, { io: 'input' }))
```

now renders `a` as `required`.

```ts
{
  '$schema': 'https://json-schema.org/draft/2020-12/schema',
  type: 'object',
  properties: { a: { default: 'a', type: 'string' } },
  required: [ 'a' ]
}
```

colinhacks:
Yeah sorry, didn't mean to pull this out from under you. I had to roll back the optionality on `ZodCatch` due to some pretty serious bugs it introduced that I didn't anticipate: https://github.com/colinhacks/zod/issues/4883

Though after thinking about this a more during the course of resolving that issue, I'm not convinced that caught properties should actually be considered optional. I suppose it's a question of whether the JSON Schema should reflect what you *can* pass in vs what you *should* pass in. In general I think *should* is a better policy, and it seems to align with most people's expectations. I made this change in tandem with the equivalent change to Zod's inferred input type. ZodCatch used to open up the input type to be `T | {}` (basically "whatever"). I undid that, so now it's back to just `T`, because despite the existence of the catch, most users (say, tRPC users) still want their typings to stay strict. 

I know you've got your own test suite to contend with, but I'm pretty happy with this behavior as-is. You may need to hack this with override on your end.

PR #6133, opened 2026-06-25
fix(json-schema): keep preprocess object properties required in input mode

### LINKED ISSUE #5968: Object property with “preprocess” is marked as optional when output JSON using `toJSONSchema [io: 'input']` since v4.4.3
Object property with “preprocess” is marked as optional when output JSON using `toJSONSchema [io: 'input']` since v4.4.3.

## Minimal reproduction
```ts
import * as z from 'zod';

const schema = z.object({
  noPreprocess: z.string(),
  withPreprocess: z.preprocess(v => v, z.string()),
});

console.log(z.toJSONSchema(schema, { io: "input" }));

/*
{
  type: 'object',
  properties: {
    noPreprocess: { type: 'string' },
    withPreprocess: { type: 'string' }
  },
  required: [ 'noPreprocess' ]
}
*/
```

## JSON output using v4.4.2
In v4.4.2, `withPreprocess` is included in `required`.
```
{
  type: 'object',
  properties: {
    noPreprocess: { type: 'string' },
    withPreprocess: { type: 'string' }
  },
  required: [ 'noPreprocess', 'withPreprocess' ]
}
```

MerlijnW70:
Fixes #5968.

## Problem

`z.toJSONSchema(schema, { io: "input" })` marked object properties defined with `z.preprocess(...)` as **not required** — a regression since v4.4.3 (#5941):

```ts
z.toJSONSchema(
  z.object({ a: z.string(), b: z.preprocess((v) => v, z.string()) }),
  { io: "input" },
);
// required: ["a"]   ❌  ("b" dropped)
```

This contradicts runtime parsing — the absent key is rejected:

```ts
z.object({ b: z.preprocess((v) => v, z.string()) }).parse({});
// ✖ ZodError: expected string, received undefined
```

## Cause

`z.preprocess(fn, inner)` builds a `ZodPipe` whose `in` is a transform. A bare transform reports `optin: "optional"` (so the preprocessor can still run on absent keys at parse time), and `ZodPipe` inherits `optin` from its `in`. The object processor's required-keys filter (`optin === undefined`) therefore excluded preprocessed properties.

## Fix

In the JSON Schema object processor only, resolve input-optionality through transform-`in` pipes to the inner `out` schema. A preprocessed property is now required unless its inner schema is genuinely optional — restoring the v4.4.2 behavior and the design of #5929. Runtime `optin` is left untouched.

```ts
// required: ["a", "b"]   ✅
```

## Tests

Regression test added in `to-json-schema.test.ts` covering a required preprocess property and an optional-inner preprocess property. Verified red without the fix, green with it. No existing snapshots change (the encoding is a no-op for non-preprocess schemas).

---
_Restores #6112, which closed automatically when my fork was briefly deleted. The fix commit is unchanged._

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2900,6 +2900,9 @@ test("catch on a transforming schema", () => {
           "type": "string",
         },
       },
+      "required": [
+        "a",
+      ],
       "type": "object",
     }
   `);
@@ -3009,6 +3012,7 @@ test("input type", () => {
       "required": [
         "a",
         "d",
+        "f",
         "g",
       ],
       "type": "object",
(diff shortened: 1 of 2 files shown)


colinhacks:
Pushed a commit on top. The same defect hits `z.catch()`: #5939 changed it in the same release and quietly dropped `f: z.string().catch("hello")` out of the `input type` snapshot. That reverted the position #4941 landed on and I defended in #5003 — an input JSON Schema should describe what you *should* send, not everything the parser happens to tolerate. So I extended your `inputOptin` helper to resolve past catch as well, restored that snapshot line, and updated the object snapshot from #6409 for the same reason.

Nice approach on the helper — resolving structurally means the nested-preprocess and `z.transform(fn).pipe(T)` cases fall out for free. Merging once CI is green.

MerlijnW70:
good to see all my fixes work :)

PR #6134, opened 2026-06-25
fix(json-schema): don't let metadata override Zod-derived keywords

### LINKED ISSUE #5990: `z.toJSONSchema()` allows overriding `type` via metadata
Running 

```ts
z.string().meta({ type: 'number' }).toJSONSchema()
```

returns

```json
{
  "type": "number"
}
```

I would propose two solutions, maybe both together are best:

1. ["All metadata fields get copied into the resulting JSON Schema."](https://zod.dev/json-schema#ztojsonschema) becomes opt-out (opt-in would've been better in my opinion but that would be a breaking change)
2. zod values take precedence and overwrite meta

MerlijnW70:
Fixes #5990.

## Problem

`z.toJSONSchema()` copied a schema's registered metadata over the generated schema with `Object.assign`, letting metadata clobber keywords Zod derives from the schema itself:

```ts
z.toJSONSchema(z.string().meta({ type: "number" }));
// { "$schema": "...", "type": "number" }   ❌  contradicts the schema
```

## Fix

Metadata now fills only keys Zod did **not** already produce. Schema-derived keywords (`type`, `properties`, `format` from validators, …) take precedence, so the example above keeps `type: "string"`.

Metadata is still additive for everything else:

- It contributes keys Zod doesn't emit (`description`, `examples`, custom vocabulary).
- It can still supply a `type` for an otherwise-unrepresentable schema (e.g. `z.unknown().meta({ type: "string" })` → `type: "string"`).

```ts
z.toJSONSchema(z.string().meta({ type: "number" }));
// { "$schema": "...", "type": "string" }   ✅
```

This matches proposal 2 in the issue ("zod values take precedence and overwrite meta"). Note it's a behavior change for anyone relying on metadata to override a derived keyword; the empty-registry escape hatch (`schema.toJSONSchema({ metadata: z.registry() })`) is unaffected.

## Tests

Added a regression test in `to-json-schema.test.ts` covering all three cases (override blocked, additive metadata preserved, `type` still supplied for unrepresentable schemas). No existing snapshots change.

---
_Restores #6113, which closed automatically when my fork was briefly deleted. The fix commit is unchanged._

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -1841,6 +1841,34 @@ describe("toJSONSchema", () => {
   });
 });
 
+test("metadata does not override Zod-derived keywords", () => {
+  // `type` is derived from the schema and must not be clobbered by metadata.
+  expect(z.toJSONSchema(z.string().meta({ type: "number" } as any))).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "type": "string",
+    }
+  `);
+
+  // Metadata still contributes keys Zod did not produce.
+  expect(z.toJSONSchema(z.string().meta({ description: "a name", format: "email" } as any))).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "description": "a name",
+      "format": "email",
+      "type": "string",
+    }
+  `);
+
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing this — the metadata merge is intended. Metadata gets copied into the output verbatim, and that's the only per-schema way to give a schema a form Zod can't derive, like `z.date().meta({ type: "string", format: "date-time" })` under `unrepresentable: "any"`. Making derived keywords win takes that away.

If you want metadata suppressed, `z.toJSONSchema(schema, { metadata: z.registry() })`. To rewrite what Zod derived, use `override`. Thanks for looking into this 👍

PR #5905, opened 2026-04-30
Defer strict object keys in intersections

colinhacks:
## Summary
- Defer strict object `unrecognized_keys` issues while parsing intersection operands so sibling schemas can reconcile keys before checks and transforms run.
- Preserve deferred key metadata through pipes so transformed strict object operands still produce their output before intersection reconciliation.
- Add regressions for strict intersections with checks, superRefine, transforms, nested defaults, and strict/refined object composition.

Fixes #5663
Fixes #4017
Fixes #2200

## Test plan
- pnpm vitest run packages/zod/src/v4/classic/tests/intersection.test.ts
- pnpm vitest run packages/zod/src/v4/classic/tests/object.test.ts packages/zod/src/v4/classic/tests/pipe.test.ts packages/zod/src/v4/classic/tests/transform.test.ts
- pnpm dev --eval '<original issue repro cases>'
- pnpm biome check packages/zod/src/v4/core/schemas.ts packages/zod/src/v4/classic/tests/intersection.test.ts
- git push pre-push hook: pnpm test

PR #6135, opened 2026-06-25
fix(intersection): run checks/refinements/transforms on strict branches

### LINKED ISSUE #5663: intersection (.and(...)) does not run check, superRefine or transform functions (and probably others)

MerlijnW70:
Fixes #5663.

## Problem

In `z.intersection(a, b)`, checks/refinements/transforms attached to a **strict** object branch were silently skipped, and an intersection of strict objects could pass validation that should have failed:

```ts
const schema = z.intersection(
  z.strictObject({ x: z.string() }).superRefine((_d, ctx) =>
    ctx.addIssue({ code: "custom", message: "fail" })
  ),
  z.strictObject({ a: z.string() }),
);

schema.safeParse({ x: "test", a: "hello" }).success;
// → true   ❌  (the superRefine never ran)
```

`z.object()` branches work; only `strictObject`/`.strict()` branches were affected. The same happens for `.check()`, `.refine()`, and `.transform()`.

## Cause

Each branch is parsed against the **full** input, so a strict branch flags the *other* branch's keys as `unrecognized_keys`. `handleIntersectionResults` already reconciles these (only keys unrecognized by *both* sides are reported). But those issues are produced during each branch's own parse and mark the payload **aborted**, so `runChecks` (and the pipe step backing `.transform()`) skip the branch's own checks — before the intersection ever reconciles the keys.

## Fix

Parse intersection branches in a non-aborting mode via an internal `nonAbortingUnrecognizedKeys` context flag. When set:

- `runChecks` treats `unrecognized_keys` issues as non-blocking, so the branch's checks/refinements run.
- the pipe step (`.transform()`) proceeds to its `out` step and carries the `unrecognized_keys` issues forward.

Any other issue still aborts as before, the issues are still recorded and reconciled by `handleIntersectionResults`, and **behavior outside intersections is unchanged** (standalone strict objects, the existing "unrecognized by both sides errors" case, etc.).

## Tests

Added a regression test in `intersection.test.ts` covering `superRefine`, `refine`, and `transform` on strict branches (each verified to run, fail, and merge correctly). Verified red without the fix, green with it. Full v4 suite (2764 tests) passes with no other snapshot changes.

---
_Restores #6114, which closed automatically when my fork was briefly deleted. The fix commit is unchanged._

### DIFF
--- a/packages/zod/src/v4/classic/tests/intersection.test.ts
+++ b/packages/zod/src/v4/classic/tests/intersection.test.ts
@@ -67,6 +67,47 @@ test("object intersection: strict + strict", () => {
   `);
 });
 
+test("strict intersection runs checks/superRefine/transform on each branch (#5663)", () => {
+  // Each strict branch flags the other branch's keys as unrecognized; those
+  // issues are reconciled away, but they must not skip the branch's own checks.
+  let superRefineCalls = 0;
+  const sr = z.intersection(
+    z.strictObject({ x: z.string() }).superRefine((_d, ctx) => {
+      superRefineCalls++;
+      ctx.addIssue({ code: "custom", message: "fail-left" });
+    }),
+    z.strictObject({ a: z.string() }).superRefine((_d, ctx) => {
+      superRefineCalls++;
+      ctx.addIssue({ code: "custom", message: "fail-right" });
+    })
+  );
+  const srResult = sr.safeParse({ x: "test", a: "hello" });
+  expect(superRefineCalls).toBe(2);
+  expect(srResult.success).toBe(false);
+  expect(srResult.error?.issues.map((i) => i.message).sort()).toEqual(["fail-left", "fail-right"]);
+
+  // `.refine()` likewise runs and can fail the parse.
+  const rf = z.intersection(
+    z.strictObject({ x: z.string() }).refine((d) => d.x === "ok", "bad-x"),
+    z.strictObject({ a: z.string() })
+  );
+  expect(rf.safeParse({ x: "no", a: "hello" }).success).toBe(false);
+  expect(rf.safeParse({ x: "ok", a: "hello" }).success).toBe(true);
+
+  // `.transform()` on a strict branch runs and its output is merged.
+  let transformCalls = 0;
+  const tx = z.intersection(
+    z.strictObject({ x: z.string() }).transform((v) => {
+      transformCalls++;
+      return { ...v, x: v.x.toUpperCase() };
+    }),
+    z.strictObject({ a: z.string() })
+  );
+  const txResult = tx.safeParse({ x: "test", a: "hello" });
+  expect(transformCalls).toBe(1);
+  expect(txResult).toMatchObject({ success: true, data: { x: "TEST", a: "hello" } });
+});
+
 test("deep intersection", () => {
   const Animal = z.object({
     properties: z.object({
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -29,6 +29,12 @@ export interface ParseContextInternal<T extends errors.$ZodIssueBase = never> ex
   readonly async?: boolean | undefined;
   readonly direction?: "forward" | "backward";
   readonly skipChecks?: boolean;
+  /** @internal When set, `unrecognized_keys` issues do not abort a schema's
+   * checks. Set while parsing the branches of an intersection: each strict
+   * branch flags the *other* branch's keys as unrecognized, and those issues
+   * are reconciled afterward by `handleIntersectionResults`, so they must not
+   * skip the branch's own checks/transforms. */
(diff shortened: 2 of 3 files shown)


colinhacks:
Closing in favor of #5905, which covers the same ground plus one case this misses: the nested-intersection default from the original report still fails here, since the abort check in `handleIntersectionResults` runs before `mergeValues` and this PR leaves it in place.

The mechanism is also the one #5905 started with — marking `unrecognized_keys` non-aborting and teaching the pipe to forward them — before I replaced it with a `loosen` flag on the parse context that skips those keys at the source. Good investigation. Thanks for looking into this 👍

PR #6136, opened 2026-06-25
fix(errors): report non-finite numbers as "Infinity" in invalid_type message

### LINKED ISSUE #5697: Weird error message after update | 4.3.6

MerlijnW70:
Fixes #5697.

## Problem

Parsing a non-finite number produced a confusing, self-contradictory message:

```ts
z.number().finite().safeParse(Infinity).error;
// {
//   code: "invalid_type",
//   expected: "number",
//   received: "Infinity",                                  // ← correct
//   message: "Invalid input: expected number, received number"  // ❌ "number"
// }
```

The issue's `received` field is already `"Infinity"`, but the rendered message said "received number".

## Cause

The locale message recomputes the received type via `util.parsedType(issue.input)`. That helper special-cased `NaN` (→ `"nan"`, displayed as `"NaN"`) but not non-finite numbers, so `Infinity`/`-Infinity` fell through to `"number"`.

## Fix

`parsedType` now returns `"Infinity"` for non-finite numbers, mirroring the existing `NaN` handling and matching the issue's `received` field:

```ts
z.number().finite().safeParse(Infinity).error;
// message: "Invalid input: expected number, received Infinity"   ✅
```

Because every locale shares `parsedType`, this fixes the message consistently across all of them.

## Tests

Updated the existing `Infinity validation` and `.finite() validation` snapshots in `number.test.ts` — they previously captured the buggy "received number" message (right next to `received: "Infinity"`); they now assert the corrected message. Full v4 suite passes.

---
_Restores #6115, which closed automatically when my fork was briefly deleted. The fix commit is unchanged._

### DIFF
--- a/packages/zod/src/v4/classic/tests/number.test.ts
+++ b/packages/zod/src/v4/classic/tests/number.test.ts
@@ -22,7 +22,7 @@ test("Infinity validation", () => {
         "code": "invalid_type",
         "received": "Infinity",
         "path": [],
-        "message": "Invalid input: expected number, received number"
+        "message": "Invalid input: expected number, received Infinity"
       }
     ]],
       "success": false,
@@ -36,7 +36,7 @@ test("Infinity validation", () => {
         "code": "invalid_type",
         "received": "Infinity",
         "path": [],
-        "message": "Invalid input: expected number, received number"
+        "message": "Invalid input: expected number, received Infinity"
       }
     ]],
       "success": false,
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing in favor of #5906, which handles this in the locale layer instead. Changing `parsedType()` is the approach I closed in #5737 — it makes a type-naming helper return a value, and it still reports `-Infinity` as `Infinity`.

PR #6137, opened 2026-06-25
fix(core): detect promises robustly when globalThis.Promise is patched

### LINKED ISSUE #6019: `superRefine(async ...)` silently drops `ctx.addIssue` when `globalThis.Promise` is patched (Zone.js / Angular NgZone / `@opentelemetry/context-zone`)

MerlijnW70:
Fixes #6019.

## Problem

When `globalThis.Promise` is replaced by a subclass — Zone.js's `ZoneAwarePromise`, Angular NgZone, `@opentelemetry/context-zone`, etc. — zod silently drops issues added inside an `async` `superRefine`/`refine`, and `parseAsync` resolves "successfully" before the async refinement runs:

```ts
import "zone.js"; // patches globalThis.Promise
import { z } from "zod";

const schema = z.object({ email: z.string() }).superRefine(async (_d, ctx) => {
  await fetch(/* ... */);
  ctx.addIssue({ code: "custom", message: "Always fails", path: ["email"] });
});

(await schema.safeParseAsync({ email: "x" })).success; // → true   ❌
```

Reported in production via `@opentelemetry/context-zone`: a duplicate-email check silently let invalid submissions through.

## Cause

A native promise returned from a user `async` function is **not** `instanceof` the patched global `Promise` (a subclass), so zod's `instanceof Promise` checks treated the async result as a synchronous value. The native-ness then propagated up the parse chain — native `.then()` yields native promises — so every downstream `instanceof Promise` check failed too, and `parseAsync` never awaited the refinement.

## Fix

Add `util.isPromise`, which keeps the fast `instanceof Promise` path and falls back to the `[object Promise]` tag. That tag is shared across promise subclasses and realms, while still excluding ordinary thenable objects:

```ts
export function isPromise(x: unknown): x is Promise<unknown> {
  return x instanceof Promise || Object.prototype.toString.call(x) === "[object Promise]";
}
```

Every promise check in the parse pipeline (`runChecks`, transforms, codecs, objects, unions, records, pipes, `parse`/`parseAsync`, …) now uses it. In an unpatched runtime the `instanceof` fast path matches exactly as before, so **behavior is unchanged**; only foreign/subclass promises that previously slipped through are now detected.

## Tests

Added a regression test in `async-refinements.test.ts` that swaps `globalThis.Promise` for a subclass (restored in `finally`) and asserts that an `async` `superRefine` fails the parse and an `async` `transform` resolves correctly. Verified red without the fix, green with it. Full v4 suite (2764 tests) passes.

---
_Restores #6116, which closed automatically when my fork was briefly deleted. The fix commit is unchanged._

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -2096,7 +2096,7 @@ export const ZodTransform: core.$constructor<ZodTransform> = /*@__PURE__*/ core.
       };
 
       const output = def.transform(payload.value, payload);
-      if (output instanceof Promise) {
+      if (util.isPromise(output)) {
         return output.then((output) => {
           payload.value = output;
           payload.fallback = true;
--- a/packages/zod/src/v4/classic/tests/async-refinements.test.ts
+++ b/packages/zod/src/v4/classic/tests/async-refinements.test.ts
@@ -66,3 +66,30 @@ test("async refine that uses value", async () => {
   expect(r2.success).toBe(true);
   expect(r2.data).toEqual("asdf123");
 });
+
+test("async refinements/transforms survive a patched global Promise (#6019)", async () => {
+  // Zone.js / Angular NgZone / @opentelemetry/context-zone replace
+  // globalThis.Promise with a subclass. A native promise returned by a user
+  // `async` refinement is not `instanceof` that subclass, so zod must not rely
+  // on `instanceof Promise` to detect async results.
+  const RealPromise = globalThis.Promise;
+  class ZoneAwarePromise<T> extends RealPromise<T> {}
+  (globalThis as any).Promise = ZoneAwarePromise;
+  try {
+    // Sanity check: this is the condition that triggered the bug.
+    expect((async () => {})() instanceof globalThis.Promise).toBe(false);
+
+    const refine = z.object({ email: z.string() }).superRefine(async (_data, ctx) => {
+      await new RealPromise<void>((r) => setTimeout(r, 1));
+      ctx.addIssue({ code: "custom", message: "always fails", path: ["email"] });
+    });
+    const refineResult = await refine.safeParseAsync({ email: "test@example.com" });
+    expect(refineResult.success).toBe(false);
+    expect(refineResult.error?.issues.map((i) => i.message)).toEqual(["always fails"]);
+
+    const transform = z.string().transform(async (s) => s.toUpperCase());
+    expect(await transform.parseAsync("hi")).toEqual("HI");
+  } finally {
+    (globalThis as any).Promise = RealPromise;
+  }
+});
--- a/packages/zod/src/v4/core/checks.ts
+++ b/packages/zod/src/v4/core/checks.ts
@@ -1119,7 +1119,7 @@ export const $ZodCheckProperty: core.$constructor<$ZodCheckProperty> = /*@__PURE
         {}
       );
 
-      if (result instanceof Promise) {
+      if (util.isPromise(result)) {
         return result.then((result) => handleCheckPropertyResult(result, payload, def.property));
       }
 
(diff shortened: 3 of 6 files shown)


wahajahmed010 (commented):
## Review: colinhacks/zod#6137 — fix(core): detect promises robustly when globalThis.Promise is patched

**Overall:** Clean, focused fix for a real production bug. The approach is correct and well-tested.

### ✅ What I like
- **Root cause is clear**: `instanceof Promise` breaks when Zone.js/Angular/OpenTelemetry patch `globalThis.Promise` with a subclass. Native async functions return native promises, not the subclass.
- **Fix is minimal**: A single `isPromise()` utility using `Object.prototype.toString.call(x) === "[object Promise]"` as a fallback. This is the standard cross-realm promise detection pattern (used by Lodash, Bluebird, etc.).
- **Mechanically correct**: Every `instanceof Promise` check in the codebase (30+ sites across schemas.ts, parse.ts, checks.ts, classic/schemas.ts) is replaced consistently. No missed spots.
- **Test is excellent**: The ZoneAwarePromise subclass + async refine + async transform test covers the exact production scenario. The `try/finally` cleanup is proper.

### 🔍 Potential issues

1. **`isPromise` doesn't handle thenables**: The function name says "isPromise" but it only detects actual Promise objects (via `[object Promise]` tag). Thenables (objects with a `.then` method) are not detected. This is fine for the current use case — zod's own code returns real promises — but worth documenting in the function JSDoc to prevent future misuse.

2. **`[object Promise]` tag is reliable but slow**: `Object.prototype.toString.call()` is ~10x slower than `instanceof`. Since this runs on every schema parse (potentially thousands of times), it's worth noting that the fast path (`instanceof`) is checked first, so the slow path only runs when Promise is patched. The performance impact in the patched-Promise case is negligible since parsing is already async-heavy there.

3. **One missed spot?**: I checked the diff — all `instanceof Promise` occurrences are replaced. The `ZodTransform` in `classic/schemas.ts` (line 2099) and the `handleCodecAResult` in `core/schemas.ts` (lines 4106, 4113) are both covered. Good.

### 🧪 Tests
- One new test case covering: patched Promise + async superRefine + async transform.
- The test verifies both the refine (issue detection) and transform (value transformation) paths.
- Sanity check confirms `(async () => {})() instanceof ZoneAwarePromise` is `false` — proving the bug exists.

### 🔒 Security
- The `[object Promise]` tag check is prototype-safe (uses `Object.prototype.toString` directly, not `obj[Symbol.toStringTag]` which could be overridden). No prototype pollution vector here.

### ⚡ Performance
- Fast path: `instanceof` (same as before) — no regression for 99.9% of users.
- Slow path: `toString` call — only triggered when Promise is patched, which is already a degraded environment.

**Verdict:** Ship it. This is a textbook fix for a real-world compatibility issue. No blockers.

colinhacks:
Closing in favor of #6021. It makes the identical change across the same 43 call sites, but `Object.prototype.toString.call()` runs on every sync parse result — that measures ~13% slower on `z.object().parse`, where the plain thenable check costs ~2%. Thanks for looking into this 👍

PR #6138, opened 2026-06-25
fix(preprocess): infer input type from the preprocessor argument

### LINKED ISSUE #5966: Narrowing the input type of `z.preprocess` no longer works after updating to latest version (v4.4.3)
Take for example, this sample from the docs:

```ts
const trimmed = z.preprocess(
  (val: string | null | undefined) => val?.trim() ?? "",
  z.string()
);
 
type Input = z.input<typeof trimmed>;  // Now gives you `unknown` instead of  `string | null | undefined` as stated in the docs
type Output = z.output<typeof trimmed>; // string

```

I can assert this to be a regression given it still worked as intended in v4.4.0

MerlijnW70:
Fixes #5966.

## Problem

The input type of `z.preprocess` no longer reflects the preprocessor's argument type — a regression since v4.4.3. From the docs example:

```ts
const trimmed = z.preprocess(
  (val: string | null | undefined) => val?.trim() ?? "",
  z.string(),
);

type Input = z.input<typeof trimmed>;  // unknown   ❌  (docs say string | null | undefined)
type Output = z.output<typeof trimmed>; // string
```

## Cause

`ZodPreprocess`/`$ZodPreprocess` were hardcoded to a bare `$ZodTransform`, whose input is `unknown`. Since a pipe's input is its `in` schema's input, the annotated argument type of the preprocessor function was discarded.

## Fix

Add an `In` type parameter (defaulted to `unknown`) to the preprocess types and type the inner transform as `$ZodTransform<unknown, In>`, so the pipe's input reflects the preprocessor argument:

```ts
type Input = z.input<typeof trimmed>;  // string | null | undefined   ✅
```

The parameter defaults to `unknown`, so this is backward compatible — an un-annotated preprocessor (`(v) => v`) still infers `unknown`. Runtime behavior is unchanged (the change is purely type-level).

## Tests

Added a type test in `preprocess-types.test.ts` asserting the annotated-argument case infers the correct input, and kept the existing un-annotated case (`unknown`). Full v4 suite passes with no type errors.

---
_Restores #6117, which closed automatically when my fork was briefly deleted. The fix commit is unchanged._

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -2429,12 +2429,12 @@ export function invertCodec<A extends core.SomeType, B extends core.SomeType>(co
 }
 
 // ZodPreprocess
-export interface ZodPreprocess<B extends core.SomeType = core.$ZodType>
-  extends ZodPipe<core.$ZodTransform, B>,
-    core.$ZodPreprocess<B> {
+export interface ZodPreprocess<B extends core.SomeType = core.$ZodType, In = unknown>
+  extends ZodPipe<core.$ZodTransform<unknown, In>, B>,
+    core.$ZodPreprocess<B, In> {
   "~standard": ZodStandardSchemaWithJSON<this>;
-  _zod: core.$ZodPreprocessInternals<B>;
-  def: core.$ZodPreprocessDef<B>;
+  _zod: core.$ZodPreprocessInternals<B, In>;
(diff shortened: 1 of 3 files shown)


colinhacks:
Closing in favor of #5967, which landed this in 9d20fdc3 — the two diffs are identical apart from the generic's name. #5967 also pins the part that isn't obvious: the narrowing has to ride on the transform's input slot only, since binding its output as well drops assignability to the bare `ZodPreprocess`.

PR #6140, opened 2026-06-25
fix(checks): don't throw from safeParse on bigint multipleOf(0n)

MerlijnW70:
## Summary

`z.bigint().multipleOf(0n).safeParse(10n)` throws `RangeError: Division by zero`
instead of returning a result object. `safeParse` is contractually never supposed
to throw, and the number equivalent already returns `{ success: false }`:

```ts
z.number().multipleOf(0).safeParse(10);   // { success: false }  ✅
z.bigint().multipleOf(0n).safeParse(10n); // throws RangeError    ❌
```

## Fix

The `multiple_of` check's bigint branch computed `value % def.value` directly, and
`value % 0n` throws. The number branch already tolerates a 0 divisor —
`floatSafeRemainder(value, 0)` is `NaN`, which `!== 0`, so the value is reported as
not-a-multiple. Guard the bigint branch so a `0n` divisor is likewise treated as
not-a-multiple, making the two branches consistent and keeping `safeParse`
throw-free.

## Test

Adds a regression test asserting `z.bigint().multipleOf(0n).safeParse(10n)` returns
`success: false`, alongside the matching `z.number().multipleOf(0)` behavior. Fails
on `main`, passes here; full `zod` suite stays green.

### DIFF
--- a/packages/zod/src/v4/classic/tests/bigint.test.ts
+++ b/packages/zod/src/v4/classic/tests/bigint.test.ts
@@ -52,3 +52,13 @@ test("min max getters", () => {
   expect(z.bigint().max(BigInt(5)).maxValue).toEqual(BigInt(5));
   expect(z.bigint().max(BigInt(5)).max(BigInt(1)).maxValue).toEqual(BigInt(1));
 });
+
+test("multipleOf(0n) does not throw from safeParse", () => {
+  // `value % 0n` throws RangeError; safeParse must never throw. Mirror the
+  // number branch, where z.number().multipleOf(0) reports a failure instead.
+  const schema = z.bigint().multipleOf(BigInt(0));
+  const result = schema.safeParse(BigInt(10));
+  expect(result.success).toBe(false);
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing as a duplicate of #6126, filed a couple of hours earlier.

Your read on the semantics is the one I went with, though — nothing is a multiple of zero, matching the number branch — over #6126's original take that `0n` is a multiple of itself. #6126 now carries that, plus a fix for the compiled path, which throws the same `RangeError` and neither PR covered. Thanks for looking into this!

colinhacks:
Carried this forward in #6466. Your fork has maintainer edits turned off so I couldn't push to the branch directly — your commit is preserved there with authorship intact.

The runtime guard was the right call, but it was only half the fix: the compiler emits `x % 0n !== 0n`, so a compiled schema still threw and this test failed under compile mode. #6466 declines a zero divisor at compile time as well. Good catch on the original bug 👍

PR #6141, opened 2026-06-26
fix(v4): ipv6() JSON Schema pattern should accept IPv4-embedded addresses

mahirhir:
## Problem

`z.ipv6()`'s emitted JSON Schema `pattern` rejects RFC 4291 IPv4-embedded IPv6 addresses (e.g. `::ffff:192.168.0.1`, `2001:db8::192.168.0.1`, `64:ff9b::192.168.0.1`) that `z.ipv6().parse()` accepts at runtime. So `z.toJSONSchema(z.ipv6())` produces a `pattern` that fails inputs the schema itself considers valid.

## Why

This is the same mismatch #5945 fixed for `cidrv6`. `$ZodIPv6` is the structural twin of `$ZodCIDRv6` — both pair a URL-parser runtime check with a regex-only JSON Schema pattern. #5945 fixed cidrv6 and left ipv6 untouched. zod's existing tests already assert these v4-embedded forms are valid at runtime, so the JSON Schema pattern was the outlier.

## Fix

Updates `regexes.ipv6` to accept the standard v4-embedded forms, scoped to match the bar #5945 set (not a bit-exact WHATWG-parser mirror).

Verified both ways: a parity probe (18 valid / 18 invalid) and end-to-end `z.toJSONSchema(z.ipv6()).pattern` checks against runtime `parse()`. Existing suite passes; +13/−1, no version bump.


mahirhir:
Closing this one. The regex change makes existing snapshots in template-literal.test.ts and to-json-schema.test.ts mismatch on stable TS, so it isn't correct as it stands and I'd rather pull it than leave a red PR open. If I can land the fix without disturbing those snapshots I'll reopen with that sorted.

PR #6142, opened 2026-06-26
feat(types): implement native .partial method for ZodTuple schemas

HarshalPatel1972:
### Description
This PR introduces a native `.partial()` instance method to the `ZodTuple` schema engine. This allows developers to relax element schemas within a tuple down to optional types while fully preserving the fixed-length tuple positional constraints.

### Technical Context
Previously, `ZodTuple` lacked an independent `.partial()` implementation, requiring developers to rely on global recursive strategies like `.deepPartial()` from a parent object. When trying to make tuple items optional directly, type structures often fell back to generic array mappings, losing critical fixed-length index metadata. 

### Resolution Strategy
* **Native Class Expansion:** Injected the `partial()` signature directly onto the `ZodTuple` prototype inside `packages/zod/src/v3/types.ts`.
* **Positional Index Retention:** Utilized `.map()` over the internal `this._def.items` schema collection to explicitly wrap each positional index node inside a `ZodOptional` wrapper.
* **Variadic Rest Support:** Added handling to ensure that if a variadic rest schema element (`this._def.rest`) exists on the tuple configuration, it is automatically relaxed via `.optional()` as well.

### DIFF
--- a/packages/zod/src/v3/types.ts
+++ b/packages/zod/src/v3/types.ts
@@ -3460,6 +3460,17 @@ export class ZodTuple<
     });
   }
 
+  partial(): ZodTuple<
+    { [K in keyof T]: T[K] extends ZodTypeAny ? ZodOptional<T[K]> : never } & (ZodTupleItems | []),
+    Rest extends ZodTypeAny ? ZodOptional<Rest> : null
+  > {
+    return new ZodTuple({
+      ...this._def,
+      items: this._def.items.map((item: any) => item.optional()) as any,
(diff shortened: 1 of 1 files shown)


colinhacks:
Closing this out. Zod 3 only takes security and correctness fixes at this point, and a new method on `ZodTuple` is neither.

Zod 4 doesn't have tuple `.partial()` either, and I'm not inclined to add it — for a fixed-length tuple, spelling the members out with `z.tuple([a.optional(), b.optional()])` is explicit and doesn't need a dedicated method. Thanks for looking into this, though 👍

PR #6144, opened 2026-06-28
fix: escape JSON Pointer reserved characters in toJSONSchema $ref

MaksZhukov:
`toJSONSchema` built `$ref` pointers by interpolating a schema's raw `meta({ id })` into the path, so an id containing the JSON Pointer reserved characters `/` or `~` produced an invalid reference. Per RFC 6901, tokens in a JSON Pointer must escape `~` as `~0` and `/` as `~1`.

```ts
z.toJSONSchema(z.object({ User: z.object({ name: z.string() }).meta({ id: "Shared/User~" }) }))
// $ref was "#/$defs/Shared/User~"  →  now "#/$defs/Shared~1User~0"
```

The `$defs` key is a plain object key and stays as the original, unescaped id; only the `$ref` pointer path is encoded.

Closes #6027.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2098,6 +2098,15 @@ test("extract schemas with id", () => {
   `);
 });
 
+test("escapes JSON Pointer reserved characters in $ref but not in $defs key", () => {
+  const User = z.object({ name: z.string() }).meta({ id: "Shared/User~" });
+  const result = z.toJSONSchema(z.object({ User }));
+  // the $ref pointer escapes `/` -> `~1` and `~` -> `~0` (RFC 6901),
+  // while the $defs key keeps the original id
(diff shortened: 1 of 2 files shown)


wahajahmed010 (commented):
## Review: colinhacks/zod#6144 — fix: escape JSON Pointer reserved characters in toJSONSchema $ref

**Logic:** Correct per RFC 6901. The encodeJSONPointerSegment function properly escapes ~ before / (order matters — ~0 before ~1). The $defs key correctly stays unescaped since it's a plain object key, not a JSON Pointer path.

**Tests:** One test case covering the exact scenario from the issue (Shared/User~). Would be nice to also test edge cases like ids with only ~, only /, or neither, but the fix is simple enough that the single test covers the regression.

**Security:** No concern — this is purely a serialization fix, no logic changes.

**Performance:** The replace calls run on every def reference, but ids are typically short strings. Negligible.

**Verdict:** Correct, minimal, well-documented. Ship it.

PR #6151, opened 2026-07-03
fix(json-schema): preserve default value on ZodDefault wrapping a transform

### LINKED ISSUE #6049: v4: `toJSONSchema` drops `default` when the defaulted schema contains a transform/pipe

amitmishra11:
## Bug

\`z.toJSONSchema()\` with \`io: "input"\` silently drops the \`default\` keyword whenever the schema passed to \`.default()\` contains a \`.transform()\` or \`.pipe()\`.

Minimal reproduction (from #6049):

\`\`\`ts
import { z } from "zod";

// { "$schema": "...", "default": "hello", "type": "string" }  -- correct

// { "$schema": "...", "type": "string" }  -- "default" is missing
\`\`\`

## Root cause

In \`packages/zod/src/v4/core/to-json-schema.ts\`, the \`process()\` function contains:

\`\`\`ts
  delete result.schema.default;
}
\`\`\`

\`isTransforming()\` recurses into inner types, so it returns \`true\` for a \`ZodDefault\` whose inner type is a transforming pipe. The \`default\` value that \`defaultProcessor\` just set on \`result.schema\` is then deleted, even though it was explicitly provided by the user.

## Fix

Skip the \`delete result.schema.default\` step when \`def.type === "default"\`. The \`ZodDefault\` schema exists specifically to carry a default value; that value must always survive to the output. The guard that deletes \`default\` is correct for \`ZodPipe\` itself (where the default might describe output-side data), but not for the \`ZodDefault\` wrapper.

## Tests

- Updated the existing \`defaults/prefaults\` inline snapshot for the \`c = a.default(1234)\` case: the default is now preserved on the input schema (the value type mismatch, if any, is the caller's responsibility, consistent with how \`prefault\` behaves).
- Added a new test \`default is preserved when inner schema contains a transform (issue #6049)\` that covers the exact reproductions from the issue report.

All 3813 tests pass.

Fixes #6049

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2625,10 +2625,62 @@ test("defaults/prefaults", () => {
       "type": "number",
     }
   `);
+  // The default value (1234) is preserved on the input schema even though it
+  // does not match the input type (string). The default is explicitly set by
+  // the user and is surfaced as-is; it is the user's responsibility to supply
+  // a type-compatible default value.
   expect(z.toJSONSchema(c, { io: "input" })).toMatchInlineSnapshot(`
     {
       "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": 1234,
       "type": "string",
+    }
+  `);
+});
+
+test("default is preserved when inner schema contains a transform (issue #6049)", () => {
+  // Identity transform: default value is a valid string, must appear on input schema
+  const withIdentityTransform = z
+    .string()
+    .transform((s) => s)
+    .default("hello");
+  expect(z.toJSONSchema(withIdentityTransform, { io: "input" })).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": "hello",
+      "type": "string",
+    }
+  `);
+  // Output schema: transform is unrepresentable by default, use unrepresentable:"any"
+  expect(z.toJSONSchema(withIdentityTransform, { unrepresentable: "any" })).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": "hello",
+    }
+  `);
+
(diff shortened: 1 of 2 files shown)


wahajahmed010 (commented):
## Review: colinhacks/zod#6151

**fix(json-schema): preserve default value on ZodDefault wrapping a transform**

### What it does
Fixes #6049: z.toJSONSchema() with io: "input" was dropping the default keyword when the schema passed to .default() contained a .transform() or .pipe(). The fix checks if the schema type is "default" before deleting the default value.

### Strengths
- Minimal, targeted fix - one line change in the core logic
- Comprehensive test coverage: identity transform, union with transform, and the original issue repro
- Good inline comment explaining the rationale
- Handles both input and output schema paths correctly

### Concerns
1. Edge case: transform that changes type - The fix preserves defaults on ZodDefault even when the inner transform changes the type (e.g., z.string().transform(Number).default(123)). The test comment acknowledges this is "user's responsibility" - which is fair, but a note in the docs would help prevent confusion.
2. unrepresentable: "any" output test - The output schema test drops type entirely (since the transform is unrepresentable). This is correct behavior but worth double-checking that downstream consumers handle schemas without type gracefully.

### Verdict
Clean, correct, well-tested. Ship it.

colinhacks:
Closing in favor of #6374.

Exempting `ZodDefault` from the check reintroduces the bug #4557 fixed — this PR rewrites the existing `defaults/prefaults` snapshot so `z.string().transform(v => v.length).pipe(z.number()).default(1234)` emits `default: 1234` on a `type: "string"` input schema. The added comment says a type-compatible default is the user's responsibility, but `.default()` is typed against the *output* type, so users can't supply one — that's what `.prefault()` is for. #6374 keeps the default only when it's valid input, which covers both cases in the issue.

PR #6152, opened 2026-07-03
ci: add CVE Lite dependency audit workflow

sonukapoor:
This adds a CVE Lite dependency audit workflow to the zod repository. CVE Lite CLI is an OWASP Lab Project that scans lockfiles locally for vulnerable dependencies using the OSV advisory database -- no package installation required, no network calls to registry endpoints during the scan.

A scan of the current pnpm lockfile found 25 findings across 1,247 packages, including 2 high-severity direct dependencies (next and vite) that have patched versions available. The full breakdown is 2 high direct, 1 critical transitive, 14 high transitive, 7 medium transitive, and 1 low transitive.

The workflow runs on every push and pull request to main, plus a weekly Monday schedule to catch newly published advisories against unchanged dependencies. When findings are present, it generates a SARIF report and uploads it to GitHub Code Scanning via the codeql-action/upload-sarif step, so vulnerabilities appear in the Security tab with file-level location data. The workflow uses `fail-on: high` so CI blocks on high or critical findings.

All GitHub Actions references are pinned to immutable commit SHA digests rather than mutable version tags, following supply chain security best practices.

More about CVE Lite CLI and how it works: https://owasp.org/cve-lite-cli

### DIFF
--- a/.github/workflows/cve-lite.yml
+++ b/.github/workflows/cve-lite.yml
@@ -1,0 +1,36 @@
+name: CVE Lite dependency audit
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    - cron: '0 6 * * 1'
+
+permissions:
+  contents: read
+  security-events: write
+
+jobs:
+  scan:
+    name: Scan dependencies
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Scan for vulnerabilities
+        uses: OWASP/cve-lite-cli@2eed959b8641042472d2810444393b88d5454e62 # v1
+        with:
+          fail-on: high
+          sarif: 'true'
+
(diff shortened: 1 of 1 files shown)


sonukapoor:
One quick note: CVE Lite CLI v1.26.0 shipped a scheduled fix mode that can automate what the companion fix PR does manually. Adding `fix: true` and `create-pr: true` to this workflow would have it open a single batched PR with the next round of upgrades automatically - OSV-confirmed versions and advisory IDs included. Here's an example of what that looks like: https://github.com/sonukapoor/cve-lite-test/pull/2

Want me to update this PR to include that? It does require enabling "Allow GitHub Actions to create and approve pull requests" in your repo settings first.

colinhacks:
Going to pass on this. zod ships with zero dependencies, so nothing this scan finds is reachable by anyone who installs the package — every finding here is in the dev toolchain, and the two direct highs are the docs site's Next.js and the test runner's Vite. With `fail-on: high` the workflow is red on this PR right now, and it would keep main red indefinitely, since most of what it flags is transitive ReDoS advisories under vitest and next that we can't upgrade past.

Thanks for putting it together, though 👍

PR #6160, opened 2026-07-07
fix(toJSONSchema): escape RFC 6901 special characters in $ref pointers

n0liu:
When `z.toJSONSchema()` builds a `$ref` from a schema’s `.meta({ id })`, it inserts the raw id into the pointer path. Per [RFC 6901](https://datatracker.ietf.org/doc/html/rfc6901), reference tokens in a JSON Pointer must escape `~` as `~0` and `/` as `~1`, so an id containing either character produces an invalid pointer.

For example, `z.string().meta({ id: "Shared/User~" })` emitted `"$ref": "#/$defs/Shared/User~"` instead of `"#/$defs/Shared~1User~0"`.

The `$defs` object key is a plain JSON key and stays raw — only the pointer path is escaped.

resolves #6027

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2173,6 +2173,16 @@ test("describe with id", () => {
   `);
 });
 
+test("$ref pointer escapes RFC 6901 special characters in meta id", () => {
+  const User = z.object({ name: z.string() }).meta({ id: "Shared/User~" });
+  const result = z.toJSONSchema(z.object({ User })) as any;
+
+  // The $defs key is a plain object key and stays raw.
+  expect(result.$defs["Shared/User~"]).toBeDefined();
+  // The $ref pointer escapes `/` as `~1` and `~` as `~0` (RFC 6901).
+  expect(result.properties.User.$ref).toBe("#/$defs/Shared~1User~0");
+});
+
 test("id is stripped from $defs entries (draft-2020-12)", () => {
   // The `id` in `.meta()` is a registration tag — it determines the $defs key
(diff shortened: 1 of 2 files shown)


wahajahmed010 (commented):
## Review: colinhacks/zod#6160 — fix(toJSONSchema): escape RFC 6901 special characters in $ref pointers

**Summary:** Clean, focused fix. The `escapeJSONPointerSegment` function correctly escapes `~` → `~0` and `/` → `~1` per RFC 6901. The test covers the edge case of both characters in a meta id.

**Logic:** ✅ Correct. The escape order matters — `~` first, then `/` — and the implementation gets this right.

**Tests:** ✅ One test covering `Shared/User~` with assertions on both the raw `$defs` key and the escaped `$ref` path.

**Edge case consideration:** The function only escapes in the `$ref` path, not the `$defs` key itself (which stays raw). This is correct per the spec — JSON Pointer escaping only applies to reference tokens, not object keys.

**No issues found.** Good to merge.

wahajahmed010 (commented):
## Review: colinhacks/zod#6160 — toJSONSchema $ref escaping

**Overall:** Clean, well-scoped fix. The `escapeJSONPointerSegment` function correctly implements RFC 6901 escaping (`~` → `~0`, `/` → `~1`).

**Strengths:**
- Minimal diff — only the escaping function and its call sites
- Test covers the exact failure case with `Shared/User~`
- Correctly keeps `$defs` keys raw (they are plain object keys, not pointer paths)

**Nit:** The function is only used in `extractDefs`. If someone adds a `$ref` elsewhere, they might miss the escaping. Consider making it a module-level utility or adding a lint rule.

**Verdict:** Looks good. No issues.

colinhacks:

Issue #6163, opened 2026-07-07
Wrong input type inferred from record with enum keys and default values


dsech:
When defining a Record with enums as keys and values with defaults, then zod is able to complete the missing keys with the default values in the parsed result:
```ts
import z from "zod"

const schema = z.record(z.enum(["UP", "DOWN"]), z.string().default("unknown"))

type Input = z.input<typeof schema>

// @ts-ignore
const input: Input = {
    UP: "somewhere"
}

console.log(z.parse(schema, input))
```
output:
```
{ UP: 'somewhere', DOWN: 'unknown' }
```

The only issue is that the inferred type `Input` is:
```ts
type Input = {
    UP: string | undefined;
    DOWN: string | undefined;
}
```
which requires using `// @ts-ignore` when defining partial inputs, otherwise typescript complains with:
```
error TS2741: Property 'DOWN' is missing in type '{ UP: string; }' but required in type 'Record<"UP" | "DOWN", string | undefined>
```

Assuming it is intended for zod to generate the missing fields in the output, then it would be amazing if the inferred `z.input` type to be Partial. For example:
```ts
type Input = Partial<{
    UP: string | undefined;
    DOWN: string | undefined;
}>
```

colinhacks:
The `Partial<>` you're asking for isn't quite right — `z.record(z.enum([...]), z.string())` really does reject a missing key, so making the input type partial across the board would type-check input the parser refuses.

But there's a real bug underneath it. When the value schema has a default, the key genuinely is omittable, and `z.object()` already infers that correctly. `z.record()` does the same thing now — fixed in #6460. Thanks for the writeup 👍

dsech:
Indeed, the `Partial<>` is expected only when zod can prefill the values because defaults are defined. I didn't realize this issue applies for optionals too, but it makes complete sense.

Thank you for fixing this and, of course, for making zod happen!

PR #6165, opened 2026-07-08
fix(v4): infer partial input for defaulted enum records

weifanglab:
Align record input inference with the existing exhaustive-record runtime behavior when the value schema accepts absent input.

For `z.record(z.enum([...]), z.string().default(...))`, parsing already runs the value schema for every enum key and fills defaults for missing keys. This updates `$InferZodRecordInput` so those missing enum keys are optional in `z.input`, while keeping the output type exhaustive. Adds a type regression test for the defaulted enum-record case.

Refs #6163.

### DIFF
--- a/packages/zod/src/v4/classic/tests/record.test.ts
+++ b/packages/zod/src/v4/classic/tests/record.test.ts
@@ -25,11 +25,17 @@ test("type inference", () => {
   const recordWithTypescriptEnum = z.record(z.enum(Enum), z.string());
   type recordWithTypescriptEnum = z.infer<typeof recordWithTypescriptEnum>;
 
+  const recordWithEnumKeysAndDefault = z.record(z.enum(["UP", "DOWN"]), z.string().default("unknown"));
+  type recordWithEnumKeysAndDefaultInput = z.input<typeof recordWithEnumKeysAndDefault>;
+  type recordWithEnumKeysAndDefaultOutput = z.output<typeof recordWithEnumKeysAndDefault>;
+
   expectTypeOf<booleanRecord>().toEqualTypeOf<Record<string, boolean>>();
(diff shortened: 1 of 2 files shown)


wahajahmed010 (commented):
## Review: zod#6165 — Infer partial input for defaulted enum records

**Logic:** ✅ Correct. The change makes `$InferZodRecordInput` check `Value["_zod"]["optin"]` for `"optional"` — which is what `z.string().default(...)` sets internally. This aligns the input type with the actual runtime behavior where missing keys get filled by defaults.

**Tests:** ✅ A type regression test is added covering both `z.input` and `z.output` for a defaulted enum record. The test uses `expectTypeOf` which is the standard pattern in this codebase.

**Edge case:** The `optin` check only covers the `"optional"` case. If a value schema uses `.default()` on a non-optional type (e.g., `z.string().default("x")`), the `optin` is `"optional"` so this works. But what about `z.record(z.enum([...]), z.string().nullable().default(null))`? The `optin` would still be `"optional"` so it should work — but worth verifying.

**Breaking changes:** None. This only widens the input type (makes keys optional), which is backward-compatible.

**Nit:** The condition `Value["_zod"]["optin"] extends "optional"` is a bit fragile — it depends on an internal property name. If `optin` is ever renamed or restructured, this breaks silently. A more robust approach would be checking `Value extends $ZodDefault` or using a branded type guard. But given the existing patterns in the codebase, this is consistent.

**Verdict:** Clean, minimal fix. Good to merge.

weifanglab:
Thanks for your review!  If there's anything need I to do, please feel free to let me know.  @wahajahmed010

colinhacks:
Closing this out — the fix landed in #6460, which closed #6163.

Key optionality on the input side now follows the value schema itself, so `.default()`, `.prefault()` and `.optional()` all make the key omittable. The input-mode JSON Schema had the same defect independently and was emitting the full `required` list, so that got fixed alongside it. Thanks for looking into this!

PR #6166, opened 2026-07-08
test: add readonly deep inference coverage

lfazioli:
## Summary
- add regression coverage for `.readonly()` on nested structures and collections
- strengthen readonly type inference assertions for nested objects
- document the readonly coverage in the changelog

## Testing
- pnpm vitest run packages/zod/src/v4/classic/tests/readonly.test.ts
- pnpm vitest run packages/zod/src/v4/classic/tests/readonly.deep.test.ts

### DIFF
--- a/packages/docs-v3/CHANGELOG.md
+++ b/packages/docs-v3/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### 3.10
 
+- Added regression coverage for `.readonly()` on nested structures and collections, including deep inference and freezing behavior for parsed values.
+
 - New parser that allows parsing to continue after non-fatal errors have occurred. This allows Zod to surface more errors to the user at once.
 
 ### 3.9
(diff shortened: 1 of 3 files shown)


colinhacks:
Declining this. The `deep inference` block you uncommented references `deepReadonlySchemas_0` and `crazyDeepReadonlySchema`, which have never existed in this repo — it's dead code inherited from v3, and it's inert because `.readonly()` in v4 is shallow, so there's no deep inference to cover. The array you added in its place is the `flat inference` array under a different name, so the new file re-asserts what `readonly.test.ts` already asserts, and the two cases that would have been new are deleted rather than tested.

The changelog edit also needs to come out — `packages/docs-v3/CHANGELOG.md` is a frozen v3 document, and test coverage doesn't go in a changelog either way.

Thanks for looking at this 👍

PR #6167, opened 2026-07-08
feat(v4): add z.formData() for typed FormData parsing

shahrryyar:
## Summary

Adds **`z.formData(shape)`** — a new helper in `zod/v4` (classic) that accepts a native `FormData` instance and returns a fully typed, validated plain object.

## Motivation

HTML `FormData` is the universal currency of web form submissions — used by React Router, Remix, SvelteKit, Next.js server actions, and every full-stack framework. Every value in `FormData` is a string (or `File`), so developers currently write repetitive `z.coerce.number()` / `z.coerce.boolean()` chains by hand. `z.formData()` eliminates that boilerplate and handles the edge cases (unchecked checkboxes, multi-value fields) that are easy to miss.

## API

```ts
const schema = z.formData({
  username: z.string().min(3),
  age:      z.number().int().positive(),
  agree:    z.boolean(),         // checkbox: absent → false, 'on'/'true'/'1' → true
  tags:     z.array(z.string()), // FormData.getAll() — multi-value
  avatar:   z.file().optional(),
});

// In a server action / route handler:
const data = schema.parse(await request.formData());
//    ^? { username: string; age: number; agree: boolean; tags: string[]; avatar?: File }
```

## Coercion rules

| Zod schema                 | Coercion applied                                       |
|----------------------------|--------------------------------------------------------|
| `z.number()`               | `Number(value)`                                        |
| `z.boolean()`              | `'on'/'true'/'1'/'yes'` → `true`; absent key → `false` |
| `z.date()`                 | `new Date(value)`                                      |
| `z.array(element)`         | `fd.getAll(key)` + per-element coercion               |
| `z.file()`                 | `File` object passed through unchanged                |
| `z.string()` / everything else | string as-is                                      |

Absent (non-boolean, non-array) keys are omitted from the coerced object so that `optional` / `default` / `nullable` schemas in the shape work as expected.

## Implementation

- **No new schema class** — implemented as a `preprocess()`-based pipe over `ZodObject`. Three files changed: `form-data.ts` (new), `external.ts` (one export line added), and `tests/form-data.test.ts` (new).
- **Coercion uses `def.type` dispatch** — `_unwrap()` walks wrapper schemas (`optional/nullable/default/catch/pipe/readonly`) to reach the inner type, keeping the implementation compatible with user-defined wrapper types.
- **No new dependencies.**

## Tests

21 new tests covering:
- string, number, boolean, date, `File`, and array field types
- checkbox absent-key → `false` semantic
- multi-value arrays (`getAll()`)
- `optional` / `default` / `nullable` absent-key behaviour
- realistic signup form round-trip
- TypeScript type inference (`expectTypeOf`)
- non-`FormData` input rejection with a clear error message

## CI

✅ All 341 existing test files (3,853 tests) pass  
✅ 0 TypeScript type errors  
✅ biome format + lint clean  
✅ Pre-push hook passes

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         node: ["lts/*"]
-        typescript: ["5.5", "latest"]
+        typescript: ["5.5", "6.0"]
     name: Test with TypeScript ${{ matrix.typescript }} on Node ${{ matrix.node }} (${{ matrix.os }})
     steps:
       - uses: actions/checkout@v4
--- a/packages/zod/src/v4/classic/external.ts
+++ b/packages/zod/src/v4/classic/external.ts
@@ -50,3 +50,6 @@ export type {
   ZodCoercedDate,
 } from "./coerce.js";
 export * as coerce from "./coerce.js";
+
+// formData
+export { formData, type ZodFormData } from "./form-data.js";
--- a/packages/zod/src/v4/classic/form-data.ts
+++ b/packages/zod/src/v4/classic/form-data.ts
@@ -1,0 +1,202 @@
+/**
+ * z.formData — parse HTML FormData into a typed, validated object.
+ *
+ * Every value in a FormData entry is a string (or File).  This helper
+ * bridges that gap by inspecting the target Zod schema for each key and
+ * applying the right coercion before validation:
+ *
+ *   - ZodNumber / ZodCoercedNumber  → Number(value)
+ *   - ZodBoolean / ZodCoercedBoolean→ checkbox semantics: "on"|"true"|"1" → true,
+ *                                      absent key → false
+ *   - ZodDate / ZodCoercedDate      → new Date(value)
+ *   - ZodArray                      → FormData.getAll(key) (multi-value)
+ *   - ZodFile                       → raw File entry from FormData.get(key)
+ *   - ZodString / everything else   → string as-is
+ *
+ * Keys absent from FormData are omitted from the coerced object so that
+ * optional / default / nullable schemas in the shape work as expected.
+ *
+ * Usage:
+ *   const schema = z.formData({
+ *     name:    z.string().min(1),
+ *     age:     z.number().int().positive(),
+ *     agree:   z.boolean(),
+ *     tags:    z.array(z.string()),
+ *     avatar:  z.file().optional(),
+ *   });
+ *
+ *   // In a server action / route handler:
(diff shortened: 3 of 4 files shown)


shahrryyar (commented):
ready to be merged and serve developers

wahajahmed010 (commented):
## Review: colinhacks/zod#6167 — `z.formData()`

**Summary:** Adds a `z.formData(shape)` helper that accepts a native `FormData` instance and returns a typed, validated object. Well-implemented with solid test coverage.

### ✅ What I like
- Clean API surface — mirrors `z.object()` shape syntax, feels native to Zod
- Sensible coercion rules: numbers via `Number()`, booleans via checkbox semantics, arrays via `getAll()`, files pass through
- Absent keys are omitted (not `null`), so `optional`/`default`/`nullable` work naturally
- Unchecked checkboxes → `false` (not missing) — correct HTML form behavior
- `_unwrap` correctly handles nested wrappers (optional, nullable, default, prefault, catch, readonly, pipe)
- 219 lines of tests covering all coercion types, edge cases, and type inference

### ⚠️ Things to consider
1. **`_coerceEntry` casts non-file values to `string`** — if a `File` is passed for a non-file schema (e.g., `z.string()`), it gets cast to `"[object File]"`. This is an edge case (wrong schema for the field), but a type guard or early rejection would be more defensive.

2. **`_unwrap` doesn't handle `effect` type** — if someone wraps a field in `z.effect()`, the unwrap won't reach the inner type. Unlikely in form shapes, but worth a note.

3. **Empty string for number fields returns `undefined`** — this means `z.number()` with an empty `<input>` will fail validation rather than silently becoming `NaN`. This is the *correct* behavior, but worth documenting explicitly since it differs from `z.coerce.number()` which would produce `NaN`.

### Verdict
Well-designed feature. Tests are comprehensive. No blocking issues. Ready to merge.

shahrryyar:
Got it, will work on it and update the same PR with the fixed errors you
suggested

shahrryyar:
> Got it, will work on it and update the same PR with the fixed errors you suggested

done and updated the PR.!!!!!!!!

shahrryyar:
hey arent you going to review it? its ready

shahrryyar (commented):
.

colinhacks:
Closing this out — `FormData` parsing is out of scope for Zod. What an unchecked checkbox, an empty file input, or a repeated key means is a decision each framework makes differently, so it belongs in the layer above Zod rather than in it.

You can build this with `z.codec()` today. Thanks for taking a run at it 👍

Issue #5642, opened 2026-01-16
Partial object schema with default fields has default fields, I was expecting undefined


erwanvivien:
Use any of the following:

```ts
z.object({ a: z.string().default(false) }).partial()

z.object({ a: z.string().default(false) })
```

I was expecting the partial to return an empty record `{}` but it returns `{ a: false }`

https://zod-playground.vercel.app?appdata=N4IgzgxgFgpgtgQxALhAegFQYDoDsCqYMABArgJ7ED2AZsQC6zE1UA2rVA7gJa4DmyPHgBeAOioAjAFYwI9ABTBSyYmLD0ATrz7yAlKIAmMGggCurBSdZFdxAL76ADgg31uCVnqG4xkmXMVlVVF1LX49Q2MzC3krG3tdbwBJYk4EMGIYAA9HWTd%2BBiZnV3dWBipiDRh6Uw1cUnr4R3pKKogqDQNiYDs8DDRvX2k8wIQVNU1tCKMTc0sPeIc8EAAaEAA3D1MYMBQAbRAekABdNfWYDTBuKlwUEAAWUQAGUQBGAHZVkG4wAC0qAwAWV43BQcRgdiAA

Issue #5315, opened 2025-10-06
Partial adds default value (changed from v3 -> v4)


linco95:
We have a patch route in our backend, where we validate our request data with Zod. 
In Zod v3 we could use `OurSchema.partial()` to let the user include the fields they wanted to patch. However while updating to Zod v4 we got failing tests on this route, as Zod now adds the default values for the missing fields in `OurSchema`. This would then overwrite those fields if the user tried to patch one of the other fields on the schema/object.

We haven't seen anything about this specific change in the documentation (migration guide), and we tried consulting the Zod AI but we couldn't find any workaround for this. Is this an intended change? Do you have any recommendation on how we should write the schema instead?

---
I made an example of this behaviour change. Given this example, the `birth_year` would be set to `2025` if our patch request contained `{ name: 'some new name` }`

### Example
Schema
```
z.object({
  name: z.string(),
  birth_year: z.number().default('2025')
}).partial()
```

Value
```
{name: "John"}
```

Results
_v3.25.76_
```
{
    "name": "John"
}
```

_v4.1.11_
```
{
    "name": "John",
    "birth_year": "2025"
}
```

linco95:
Making it required first did indeed help. Not sure if it could've been documented more clearly in the migration guide (or if I just simply missed it), but thanks mr @dosu bot, and Colin for answering this question in the linked discussion

Issue #5235, opened 2025-09-13
[v4] Partial object runs `default` handler on missing properties


Mysak0CZ:
In my codebase a common pattern is something like this:

```ts
/** Describes data saved in database */
const SavedDataSchema = z.object({
    oldKey: z.number(),
    // Property added in a newer version
    // Has default to allow for automatic data migration, while still raising errors if there does exist an incompatible value already
    keyThatGotAddedInNewVersion: z.string().default(""),
});

/** Describes changes to data, used as part of delta update mechanism */
const UpdateSchema = SavedDataSchema.partial();
```

After updating to v4, data started being randomly deleted, which I traced to Zod using `default` even in partial objects:
```ts
// Data comes in as part of delta update
const dataIn = { oldKey: 2 };
const parsed = UpdateSchema.parse(dataIn);
// Results in:
parsed === { oldKey: 2, keyThatGotAddedInNewVersion: "" }
```

Only after finding the cause of this issue, I realized "[defaults applied within optional fields](https://zod.dev/v4/changelog?id=defaults-applied-within-optional-fields)" applies to `partial` objects as well (which wasn't at all clear from how it is written).
This, however, makes me wonder what the proper solution for the above problem is.

Also I find it rather confusing that inner type has higher priority than what is written later (especially in code that passes schemas around as arguments, not always knowing which schema will be used) - I would expect that the wrapped type used later will override the current behavior, instead of possibly doing nothing.

colinhacks:
Zod has to make a call on this behavior, and there's no approach that makes sense to everybody. The fact that `.partial()` clobbered defaults was a huge usability issue in Zod 3—consider searching to find the related issues there. 

You can get the behavior you're looking for by making it required first:

```ts
/** Describes data saved in database */
const SavedDataSchema = z.object({
  oldKey: z.number(),
  // Property added in a newer version
  // Has default to allow for automatic data migration, while still raising errors if there does exist an incompatible value already
  keyThatGotAddedInNewVersion: z.string().default(""),
});

/** Describes changes to data, used as part of delta update mechanism */
const A = SavedDataSchema.partial();
console.log(A.parse({})); // => { keyThatGotAddedInNewVersion: "" }

const B = SavedDataSchema.required().partial();
console.log(B.parse({})); // => {}
```

Issue #4799, opened 2025-07-02
zod v4 .partial erases defaults


spence-s:
This previously worked with zod v3 and is an undocumented in the migration guides from what I could tell

```typescript
import {z} from 'zod/v4';

const objWithDefaults = z.object({
  a: z.string().default('defaultA'),
  b: z.string().default('defaultB'),
  c: z.string().default('defaultC'),
});

const objPartialWithOneRequired = objWithDefaults.partial().required({a: true});

const test = objPartialWithOneRequired.parse({}); // => errors and does not use the default value for "a"
```

However, if you remove the `partial` call, `required` then uses the default 

```ts
import {z} from 'zod/v4';

const objWithDefaults = z.object({
  a: z.string().default('defaultA'),
  b: z.string().default('defaultB'),
  c: z.string().default('defaultC'),
});

const objPartialWithOneRequired = objWithDefaults.required({a: true});

const test = objPartialWithOneRequired.parse({}); // => works defaults are preserved { a: 'defaultA', ....}
```

This broke a ton of stuff for me where I was using the pattern in the first example. Seems like a bug to me, partial should not strip defaults.

colinhacks:
Good call. Found a clean way to fix this, landed in 3.25.69

PR #6172, opened 2026-07-09
fix(v4): stop broadening optional/default/prefault fields with partial

### LINKED ISSUE #6171: `.partial()` produces inaccurate inferred types for optional and defaulted fields
`.partial()` wraps every field in `ZodOptional`, even when a field already accepts absent input. Runtime behavior is correct, but the inferred types become inaccurate.

### Already-optional fields are double-wrapped

```ts
const schema = z.object({ a: z.string().optional() }).partial();
type A = typeof schema.shape.a;
// ZodOptional<ZodOptional<ZodString>>
```

Repeated `.partial()` calls keep nesting `ZodOptional` unnecessarily.

### Defaulted fields get broadened to `undefined`

```ts
const schema = z.object({ foo: z.string().default("foo") }).partial();
type Foo = z.output<typeof schema>;
// { foo?: string | undefined }
```

But `foo` is never `undefined` after parsing:

```ts
schema.parse({});
// { foo: "foo" }
```

The inferred output should reflect that.

### This is a type-only issue

I'm not proposing any runtime changes. Defaults applying through `.partial()` are intentional (#5235, #5315, #5642, #4799), and `.required().partial()` remains the escape hatch for disabling them. This is only about making the inferred types match the values `.partial()` produces.

Environment

- Zod version: 4.4.3

acontreras89:
Fixes #6171.

While looking into the incorrect types produced by `.partial()`, we found this logic already implemented but commented out.

Simply enabling it doesn't quite work: it keys on `_zod.optin`, which also changes the behavior of `catch`/`transform`/`preprocess`. Their wrapping `ZodOptional` is still semantically meaningful, so skipping it changes `partial().parse(...)`.

This instead keys on `def.type` (`optional`, `default`, `prefault`), fixing the two typing issues without changing runtime behavior:

- already-optional fields no longer become `ZodOptional<ZodOptional<...>>`;
- `default`/`prefault` fields no longer infer `T | undefined` when parsing always produces a value.

This may not be the final direction—especially with the optionality work in flight—but it seemed like a useful starting point. If the intent is to solve this through `optin`/`fallback` instead, we're happy to reshape or fold this into that work. Refs #5235, #5315, #5642, #4799.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1437,6 +1437,18 @@ export type SafeExtendShape<Base extends core.$ZodShape, Ext extends core.$ZodLo
     : Ext[K];
 };
 
+// `.partial()` leaves a field untouched when it already accepts absent input
+// *and* produces a deliberate value for it (`optional`, `default`, `prefault`);
+// otherwise it wraps in `ZodOptional`. Keyed on `def.type` rather than `optin`
+// on purpose: `catch`/`transform`/`preprocess` also report `optin: "optional"`
+// at runtime, but their static optin defers to the inner type, and the wrapping
+// `ZodOptional` does real work for them (clobbering their fallback on absent
+// input). Matching on `def.type` keeps the inferred type and the runtime shape
+// in sync for every schema.
+type PartialField<T extends core.SomeType> = T extends { _zod: { def: { type: "optional" | "default" | "prefault" } } }
+  ? T
+  : ZodOptional<T>;
+
 export interface ZodObject<
   /** @ts-ignore Cast variance */
   out Shape extends core.$ZodShape = core.$ZodLooseShape,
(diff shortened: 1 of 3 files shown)


colinhacks:
This is the right approach. Keying on `def.type` rather than `optin` is what keeps `catch` and `preprocess` wrapped, and I verified `.partial().parse()` output is identical to `main` across optional/default/prefault/catch/preprocess/transform — so this really is inference-only, which is what the issue asked for. Closed #6173 and #6264 in favor of it.

One thing before it lands: `util.partial` is shared with zod-mini, so `m.partial(obj).shape.def` is now `ZodMiniDefault` at runtime while `mini/schemas.ts` still types it as `ZodMiniOptional<ZodMiniDefault<...>>`. Mirror `PartialField` into the two `partial` overloads there and the divergence goes away. Needs a rebase on `main` as well — the prototype refactor in #6318 landed since, though it doesn't touch `partial` itself.

colinhacks:
Closing this. Keying the object type signature on specific schema types is the part I don't want — an allowlist of `optional`/`default`/`prefault` is fragile, and every new wrapper type becomes one more entry somebody has to remember to add in two places at once.

The current behavior is intended. `.partial()` wraps every field, and defaults still apply through the wrapper. Thanks for digging into this, though 👍

PR #6179, opened 2026-07-13
fix: use exact flag in English locale too_small/too_big messages + fix .partial() double-wrapping

deepshekhardas:
Fixes two bugs in Zod v4:

## Bug 1: `exact` size messages show wrong wording
When using `.size()` with `exact: true`, the English locale error messages incorrectly show `>=` / `<=` instead of `exactly`.

**Before:** `"Too small: expected array to have >=2 items"`
**After:** `"Too small: expected array to have exactly 2 items"`

## Bug 2: `.partial()` double-wraps optional/default/prefault fields
When calling `.partial()` on an object that already has optional/default/prefault fields, those fields were being unnecessarily wrapped in `ZodOptional`, causing nested wrappers and broader output types.

Changes:
- Added `PartialField<T>` helper type that checks if a field's def.type is already `optional | default | prefault` and leaves it untouched
- Updated both `partial()` and `partial(mask)` signatures in classic and mini layers
- Changed `$ZodCatch.optin` from `T['_zod']['optin']` to `'optional'` for correct type matching
- Added `partialSkipTypes` set in `core/util.ts` and uncommented the skip logic

Closes #6177
Closes #6172
Closes #6173

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1437,6 +1437,18 @@ export type SafeExtendShape<Base extends core.$ZodShape, Ext extends core.$ZodLo
     : Ext[K];
 };
 
+// `.partial()` leaves a field untouched when it already accepts absent input
+// *and* produces a deliberate value for it (`optional`, `default`, `prefault`);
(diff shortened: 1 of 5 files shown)


deepshekhardas:
Superseded by #6250, #6242 — cleaner branches with targeted fixes

PR #6180, opened 2026-07-13
fix(toJSONSchema): escape RFC 6901 special characters in \ pointers

deepshekhardas:
Fixes JSON Pointer escaping per RFC 6901 for meta id containing / or ~ characters. Closes #6027.


deepshekhardas:
Superseded by #6257 — cleaner single-purpose fix

PR #6181, opened 2026-07-13
fix(formatError): prevent throwing on Object.prototype keys in path

deepshekhardas:
Fixes formatError throwing a TypeError when path segments match Object.prototype members (toString, constructor, valueOf). Related to #5265.

### DIFF
--- a/packages/zod/src/v4/classic/tests/error.test.ts
+++ b/packages/zod/src/v4/classic/tests/error.test.ts
@@ -709,3 +709,19 @@ test("error serialization", () => {
     `);
   }
 });
+
+test("formatError does not throw on Object.prototype keys in path", () => {
+  const schema = z.object({
+    toString: z.string(),
+    constructor: z.string(),
+    valueOf: z.string(),
+  });
+  const result = schema.safeParse({});
+  expect(result.success).toEqual(false);
+  if (!result.success) {
+    const formatted = z.formatError(result.error) as any;
+    expect(formatted.toString._errors.length).toBeGreaterThan(0);
+    expect(formatted.constructor._errors.length).toBeGreaterThan(0);
+    expect(formatted.valueOf._errors.length).toBeGreaterThan(0);
+  }
+});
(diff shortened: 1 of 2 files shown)


anp0429:
**The fix resolves the crash broadly.** 11 distinct shapes throw `TypeError: Cannot read properties of undefined (reading 'push')` at `errors.ts:318` on `main`  top-level `toString`/`hasOwnProperty`, nested and non-terminal positions, `__defineGetter__`, paths through `invalid_element`/`invalid_key` recursion, and repeat-format calls (one even hits `Cannot assign to read only property 'name'`, because the fabricated node *is* the inherited function). All 11 pass on this branch. The committed test covers one shape; the recursion and non-terminal paths are distinct code paths , happy to contribute them.

**One residual case this fix doesn't cover: a `"__proto__"` path element.** The new `curr[el] = { _errors: [] }` doesn't create an own key for `__proto__` - bracket-assignment sets the object's *prototype* instead:

```js
const curr = { _errors: [] };
if (!Object.prototype.hasOwnProperty.call(curr, "__proto__")) {
  curr["__proto__"] = { _errors: [] };   // the fix's node-creation line
}
Object.prototype.hasOwnProperty.call(curr, "__proto__"); // false
Object.keys(curr);                                        // ["_errors"] — node gone
Object.getPrototypeOf(curr) === Object.prototype;         // false - prototype replaced
```

So errors on a `__proto__` path land on the prototype chain: absent from `Object.keys` / serialization, and the formatted object's prototype gets swapped. Using `Object.defineProperty(curr, el, { value: { _errors: [] }, writable: true, enumerable: true, configurable: true })` for node creation treats `__proto__` as an ordinary key and resolves it. Glad to push that plus a test as a follow-up.

wahajahmed010 (approved):
## Review: colinhacks/zod#6181

**Logic**: ✅ The old code used `curr[el] = curr[el] || { _errors: [] }` which silently fails when `el` is an Object.prototype key like `toString`, `constructor`, or `valueOf` — those exist on the prototype and are truthy, so the `||` short-circuits and never creates the `_errors` array. The fix correctly uses `hasOwnProperty.call` to check actual own-properties.

**Structure**: ✅ The refactor also simplifies the logic — the `terminal` check is now independent of the initialization block, which is cleaner.

**Tests**: ✅ Good test coverage with the three prototype keys that would have caused the bug.

**Security**: ✅ This is a correctness/security fix — prevents silent data loss when schema paths collide with Object.prototype keys.

**No issues found.** Clean fix, well tested.

deepshekhardas:
Superseded by #6243 — cleaner branch with focused changes

PR #6182, opened 2026-07-13
feat: allow structured ZodIssueMessage values for deferred error message rendering

romankoho:
## Motivation

Today, a `ZodIssueMessage` can only be a string. This means that validation messages must be fully resolved when the schema is created.

This becomes problematic in applications that support multiple languages. The translation function (`t`) is typically only available in the UI layer, while schemas are often defined as reusable constants or shared across multiple environments.

As a result, developers have to either:

- recreate the schema whenever the active language changes,
- inject the translation function into schema factories, or
- keep schemas inside UI components instead of defining them once.

In React, for example, this often leads to patterns like:

```ts
export function createFormSchema(t) {
  return z.object({
    firstName: z.string({
      message: t("some.translationKey", { someKey: "abc" }),
    }),
  });
}
```

```tsx
const { t, i18n } = useTranslation();

const schema = useMemo(
  () => createFormSchema(t),
  [i18n.language]
);
```

The validation logic itself is independent of the current language, yet the schema must be recreated whenever the language changes because the error messages have already been translated.

## Proposed solution

Allow `ZodIssueMessage` to also accept a structured object:

```ts
type ZodIssueMessage =
  | string
  | {
      key: string;
      values?: Record<string, unknown>;
    };
```

Schemas can then remain language-agnostic:

```ts
export const formSchema = z.object({
  firstName: z.string({
    message: {
      key: "some.translationKey",
      values: {
        someKey: "abc",
      },
    },
  }),
});
```

The rendering layer can translate the message when it is actually displayed:

```tsx
const issue = error.message;

if (typeof issue === "string") {
  return issue;
}

return t(issue.key, issue.values);
```

Libraries such as `@hookform/error-message` (or custom error renderers) can extract the translation key and interpolation values and perform localization at render time.

## Benefits

- Separates validation logic from presentation concerns.
- Schemas can be defined once and reused regardless of the active language.
- Eliminates the need to recreate schemas on language changes.
- Enables richer integrations with i18n libraries without imposing any particular translation solution.
- **Fully backwards compatible.** This change is purely additive: existing code that provides a string message continues to work unchanged, while applications that prefer deferred localization can opt into the new object format.

### DIFF
--- a/packages/zod/src/v4/classic/tests/error-utils.test.ts
+++ b/packages/zod/src/v4/classic/tests/error-utils.test.ts
@@ -79,7 +79,7 @@ test(".flatten()", () => {
 });
 
 test("custom .flatten()", () => {
-  type ErrorType = { message: string; code: number };
+  type ErrorType = { message: string | { key: string; values?: object }; code: number };
   const flattened = parsed.error!.flatten((iss) => ({ message: iss.message, code: 1234 }));
   expectTypeOf(flattened).toMatchTypeOf<{
     formErrors: ErrorType[];
@@ -159,7 +159,7 @@ test(".format()", () => {
 });
 
 test("custom .format()", () => {
-  type ErrorType = { message: string; code: number };
+  type ErrorType = { message: string | { key: string; values?: object }; code: number };
   const formatted = parsed.error!.format((iss) => ({ message: iss.message, code: 1234 }));
   expectTypeOf(formatted).toMatchTypeOf<{
     _errors: ErrorType[];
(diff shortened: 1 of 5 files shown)


colinhacks:
Going to decline this one. `message` is the rendered string by design — the structured payload is the issue itself, which already carries `code` plus every interpolation value the message was built from (`minimum`, `origin`, `format`, and so on), so a render-time translator is `t(issue.code, issue)` over `error.issues`.

Worth noting the schema recreation in your example isn't necessary either: the `error` param is a function called at parse time, not when the schema is constructed, so a module-level schema picks up whatever language `t` is on at the moment you call `.parse()`. And `.refine(fn, { params: { ... } })` already attaches arbitrary structured data to an issue if you want a key and values of your own.

Thanks for the detailed writeup though 👍

PR #6183, opened 2026-07-13
fix(v4): preserve input-valid default on transforming schema in toJSONSchema (#6049)

### LINKED ISSUE #6049: v4: `toJSONSchema` drops `default` when the defaulted schema contains a transform/pipe

NmaaAlhawary:
`z.toJSONSchema(schema, { io: "input" })` dropped the `default` keyword whenever the defaulted schema contained a transform:

```ts
z.string().transform((s) => s).default("hello")
// input mode → { type: "string" }   ← default "hello" lost
```

The strip dates to #4557, which broadened the input-mode deletion from `def.type === "pipe"` to any `isTransforming(schema)`. That correctly removes an output-typed default that can't fit the input type — e.g. `z.string().transform((v) => v.length).pipe(z.number()).default(1234)`, where `1234` doesn't belong on a `string` input — but it also removed defaults that are valid input.

A `.default()` value now survives input extraction when its JSON type matches the schema's resolved input type, so `"hello"` is kept while the `1234`-on-`string` case still strips.

Closes #6049

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         node: ["lts/*"]
-        typescript: ["5.5", "latest"]
+        typescript: ["5.5", "5"]
     name: Test with TypeScript ${{ matrix.typescript }} on Node ${{ matrix.node }} (${{ matrix.os }})
     steps:
       - uses: actions/checkout@v4
@@ -28,8 +28,8 @@ jobs:
           node-version: ${{ matrix.node }}
       - uses: pnpm/action-setup@v4
       - run: pnpm install
-      - run: pnpm add typescript@${{ matrix.typescript }} -w
       - run: pnpm build
+      - run: pnpm add typescript@${{ matrix.typescript }} -w
       - run: pnpm test
       - run: pnpm run --filter @zod/resolution test:all
       - run: pnpm run --filter @zod/integration test:all
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2674,6 +2674,40 @@ test("keep input-valid default on transforming schema (#6049)", () => {
   `);
 });
 
+test("strip type-mismatched default on transforming schema (#4557)", () => {
+  // default is output-typed (number) but the input type is string → still stripped in input mode
+  const a = z
+    .string()
+    .transform((v) => v.length)
+    .pipe(z.number())
+    .default(1234);
+  expect(z.toJSONSchema(a, { io: "input" })).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "type": "string",
+    }
+  `);
+  expect(z.toJSONSchema(a, { io: "output" })).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": 1234,
+      "type": "number",
+    }
+  `);
+
+  // same polarity without an explicit .pipe()
+  const b = z
+    .string()
+    .transform((s) => s.length)
+    .default(42);
+  expect(z.toJSONSchema(b, { io: "input" })).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "type": "string",
+    }
+  `);
+});
+
 test("falsy prefaults (false, 0, empty string)", () => {
   // boolean prefault false
   const a = z.boolean().prefault(false);
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2633,6 +2633,47 @@ test("defaults/prefaults", () => {
   `);
 });
 
+test("keep input-valid default on transforming schema (#6049)", () => {
+  // default value matches the input type (string) → preserved in input mode
+  const a = z
+    .string()
+    .transform((s) => s)
+    .default("hello");
+  expect(z.toJSONSchema(a, { io: "input" })).toMatchInlineSnapshot(`
+    {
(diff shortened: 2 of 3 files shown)


NmaaAlhawary:
The failing checks are pre-existing, not related to this change — every PR since 2026-07-08 fails the same way because `typescript@latest` now resolves to 7.0.2 and zshy crashes during `pnpm build` before any tests run. Root-cause analysis in #6185.

The full suite passes locally on this branch with the repo's pinned toolchain (339 files, 3813 tests, no type errors).

wahajahmed010 (commented):
Solid fix. The `defaultMatchesInputType` JSON-type check is the right granularity for a default-time prefilter (cheaper than a real validator), and `resolveInputType` correctly walks past ref chains via a guard set to avoid cycles.

One nit: commit 2/2 is an unrelated CI workflow change (TS 5 cap + build-before-install order, refs #6185) bundled in the same PR. Please split it into a separate PR so revert granularity is clean and the toJSONSchema fix isn't held up by CI churn. Also, since the CI change makes `typescript: [5.5, 5]` a hard requirement to even build, a brief note in the PR description about the temporary TS 7 deferral would help reviewers.

wahajahmed010 (commented):
Nice fix and well-scoped — `keepDefault` cleanly threads the input-type check into the existing transform-stripping branch. The follow-up commit moving `pnpm build` ahead of the matrix TS swap is also a sensible CI hardening.

A few things worth tightening before merge:

1. **Missing negative test.** The new `keep input-valid default on transforming schema (#6049)` test only covers the keep case. The motivation hinges on `#4557` continuing to strip *type-mismatched* defaults on transforms — but there's no assertion that `z.string().transform(s => s.length).default(42)` still drops the default in `io: "input"` mode. Without that test, a future refactor that flips the polarity of `keepDefault` would still go green. Please add an inline-snapshot test that mirrors the one in the PR description for the `1234` case (or any clear `default: number` on a `string` transform).

2. **`as any` cast in `defaultMatchesInputType`.** The `declared.includes(t as any)` looks like a friction point with `JSONSchema.BaseSchema["type"]`. Worth tightening — either by narrowing `type` to a typed string union up front, or by replacing the `includes` with a typed `Set<string>` lookup. Low risk today, but a future change to the JSON Schema type representation could quietly break the comparison.

3. **`prefault` symmetry.** `def.type === "default"` is checked here, but `prefault` is also a way to set a default-typed value (and there's a separate `result.schema._prefault` handling just below in this function). Is the intent that `prefault` follows the same rule, or is `prefault` always emitted on the input side? If the former, the gate is incomplete; if the latter, a one-line comment explaining the asymmetry would help future readers.

4. **Cycle guard.** `resolveInputType` correctly guards against cycles via the `Set`, but it silently returns `undefined` if the chain doesn't terminate at a typed `seen.schema.type`. That collapses to `false` in `defaultMatchesInputType`, which strips the default — which is the safe default, but worth a one-line comment noting that an unresolved input type is treated as "not safe to keep."

Approving on the substantive fix; the test coverage gap is the main blocker for me.

NmaaAlhawary:
Thanks for the careful read — all four points addressed in 1ea15523.

**1. Missing negative test.** Agreed this was the real gap. Added `strip type-mismatched default on transforming schema (#4557)`, covering both the piped case from the description (`z.string().transform(v => v.length).pipe(z.number()).default(1234)`) and the bare one (`z.string().transform(s => s.length).default(42)`), each asserting the default is absent in `io: "input"`. The piped case also asserts `default: 1234` still shows up in `io: "output"`, so the test pins the polarity in both directions rather than just the absence.

I sanity-checked that it actually bites: inverting `keepDefault` to `!defaultMatchesInputType(...)` locally fails 3 tests — the new one, the #6049 keep test, and the pre-existing `defaults/prefaults`. So a refactor that flips the polarity can't go green.

**2. `as any` cast.** Replaced with a `Set<string>` lookup, which drops the cast entirely:

```ts
const expected: string[] = jsonType === "number" ? ["number", "integer"] : [jsonType];
const declared = new Set<string>(Array.isArray(type) ? type : [type]);
return expected.some((t) => declared.has(t));
```

I kept the `Array.isArray` branch even though `JSONSchema.BaseSchema["type"]` is a single keyword today — JSON Schema permits an array of type keywords, and an `override` hook or metadata can produce one at runtime. Comment added noting that.

**3. `prefault` symmetry.** The latter — `prefault` is always emitted on the input side, deliberately. It's input-typed by definition (substituted *before* parsing), so unlike `default` it can't carry an output-typed value that fails the input schema, and the check would never have anything to reject. The ordering already encodes this: the strip block runs at the top and the `_prefault` → `default` assignment happens right after it, so `prefault` is never a candidate for stripping. Added a comment at the gate spelling that out.

**4. Cycle guard.** Right, `undefined` is the intended outcome for every non-terminating path — cyclic, unprocessed, or a chain that legitimately never reaches a typed schema (a union, `z.any()`). It collapses to `false` and strips, which is the conservative choice and matches pre-#6049 behavior, so the fix can only ever *narrow* what gets stripped, never widen it. Documented on `resolveInputType`.

Full suite locally: 339 files, 3815 tests, no type errors.

On splitting the CI commit from your earlier review — happy to pull `ca6054fd` into its own PR if you'd still prefer that. I left it here for now since checks are green on this branch and separating it would put this PR back on red until the other one lands, but say the word and I'll split it.

colinhacks:
Closing in favor of #6374.

Checking the default against the input type is the right approach, but resolving only the `type` keyword misses the union case in the issue — `z.union([z.boolean(), z.object({ a: z.string().transform(x => x) })]).default(false)` still drops its default, since a union never sets `type`. It also keeps defaults that fail the schema's own constraints: `z.string().min(5).transform(s => s).default("hi")` emits `default: "hi"` alongside `minLength: 5`. #6374 checks validity by parsing, so both hold. The `.github/workflows/test.yml` change is also unrelated and would need to come out.

Good investigation. Thanks for looking into this 👍

PR #6184, opened 2026-07-13
fix(v4): use exact flag in English locale messages

### LINKED ISSUE #6176: v4: `.length()` / `.size()` error messages ignore the `exact` flag
## Zod version

v4.4.3

## Minimal reproduction

```ts
import { z } from "zod";

z.string().length(4).parse("abc");
```

## Expected

Since `.length(4)` represents an exact constraint, the error message should indicate that the expected length is **exactly 4**, as it did in Zod v3.

## Actual

The error message uses range wording (e.g. `>=4` / `<=4`), even though the generated issue already contains `exact: true`.

## Additional context

While investigating this, I found that:

* `exact: true` is assigned for fixed-length validations in `core/checks.ts`.
* `exact` is part of the `too_small` / `too_big` issue types.
* None of the v4 locale renderers currently use `issue.exact` when formatting messages.
* The v3 English locale did use `issue.exact` to produce messages such as "exactly 4".
* I couldn't find an existing issue or PR covering this behavior.

Before opening a PR, I'd like to confirm whether this is an unintended regression. If so, I'd be happy to submit a focused PR updating the English locale together with the relevant tests.

deepshekhardas:
Updates the v4 English locale to honor the exact flag on too_small/too_big issues. Closes #6176.

### DIFF
--- a/packages/zod/src/v4/locales/en.ts
+++ b/packages/zod/src/v4/locales/en.ts
@@ -71,15 +71,26 @@ const error: () => errors.$ZodErrorMap = () => {
         if (issue.values.length === 1) return `Invalid input: expected ${util.stringifyPrimitive(issue.values[0])}`;
         return `Invalid option: expected one of ${util.joinValues(issue.values, "|")}`;
       case "too_big": {
-        const adj = issue.inclusive ? "<=" : "<";
         const sizing = getSizing(issue.origin);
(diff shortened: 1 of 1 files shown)


deepshekhardas:
Superseded by #6242 — cleaner locale-only fix

PR #6186, opened 2026-07-13
ci: build with pinned TypeScript before installing matrix version

### LINKED ISSUE #6185: CI: `pnpm build` fails on every PR since TypeScript 7.0.2 became `latest` (zshy crash)
### Summary

### Mechanism

```yaml
- run: pnpm build
```

```
```

```sh
```

### Possible directions

NmaaAlhawary:
CI has failed on every PR since 2026-07-08, when TypeScript 7.0.2 became npm's `latest`. Root cause in #6185: the workflow installs the matrix TypeScript before `pnpm build`, and `zshy@0.7.2` (peer dep `typescript >5.5.0`) crashes on TS 7's compiler API:

```
Build failed: TypeError: Cannot read properties of undefined (reading 'readFile')
```

This moves `pnpm build` above `pnpm add typescript@${{ matrix.typescript }} -w`, so the package builds with the lockfile's pinned TypeScript while the matrix version still applies to the test steps — which is what the matrix exists to exercise.

Verified locally on `main`:

- `pnpm build` with the pinned TypeScript: passes
- `pnpm vitest run` after `pnpm add typescript@latest -w` (7.0.2): 339 files, 3813 tests, no type errors
- `@zod/resolution` / `@zod/integration` pin their own `typescript@^5.8.2` and are unaffected by the root swap

The "TypeScript 5.5" job in recent runs was cancelled by matrix fail-fast rather than failing on its own, so this one change should restore both legs.

Closes #6185

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,8 +28,10 @@ jobs:
           node-version: ${{ matrix.node }}
       - uses: pnpm/action-setup@v4
       - run: pnpm install
-      - run: pnpm add typescript@${{ matrix.typescript }} -w
+      # Build with the lockfile's pinned TypeScript — zshy doesn't support TS 7 (see #6185).
+      # The matrix TypeScript applies to the test steps below, which is what the matrix is for.
       - run: pnpm build
+      - run: pnpm add typescript@${{ matrix.typescript }} -w
       - run: pnpm test
       - run: pnpm run --filter @zod/resolution test:all
       - run: pnpm run --filter @zod/integration test:all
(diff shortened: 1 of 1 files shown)


NmaaAlhawary:
First run confirmed the diagnosis one level deeper: `pnpm build` and `pnpm test` both passed under TS 7 (3811 tests, no type errors), but `@zod/resolution test:all` then crashed with the same zshy `readFile` error — its harness (`test-resolution.ts`) invokes zshy internally, and zshy peer-resolves the *root* TypeScript, which the matrix step had swapped to 7.0.2.

Pushed 988120c8: the resolution/integration steps now also run before the matrix swap. Both packages pin their own `typescript@^5.8.2`, so they were never meant to vary with the matrix — after this, only `pnpm test` exercises the matrix TypeScript.

colinhacks:
Closing — this landed in #6345, which makes the same reordering and adds a TypeScript 6 leg plus a guard for the resolution suite. Thanks for tracking it down.

Issue #1933, opened 2023-01-25
Access to value in `.catch` that failed to parse.


andrejpavlovic:
I'd like to be able to access the value that failed parsing in `.catch` and do something with it. For example:

```typescript
const schema = z
  .discriminatedUnion("type", [
    z.object({ type: z.literal("type1") }),
    z.object({ type: z.literal("unknown"), value: z.unknown() })
  ])
  .catch((val: unknown) => ({ type: "unknown" as const, value: val }));

const result = schema.parse({ type: "some-other-type" });
// result would look like
// {type: "unknown", original: { type: "some-other-type" }}
```

Then in other places I can maybe log what the original value was, or try to do some more parsing in the `.catch` callback itself

rotu:
I think this is fixed?

The above code gives me this value for `result`:
```js
{
  type: 'unknown',
  value: { error: [Getter], input: { type: 'some-other-type' } }
}
```

andrejpavlovic:
I think since my post the first parameter in `catch` callback is the error. It still doesn't provide the actual value.

kachkaev:
Hmmm looks like the input is passed indeed:

https://github.com/colinhacks/zod/blob/c4458caadc24f25063b6a5d0dbbbc9e1d6d60735/src/types.ts#L470-L473

`git bame` refers to https://github.com/colinhacks/zod/commit/ac0135e00df75abd57a93a4816fe0fdaa31e94e8

andrejpavlovic:
Ah, that makes sense. Issue closed!

andrejpavlovic:
@rotu So this should work:

```typescript
const schema = z
  .discriminatedUnion("type", [
    z.object({ type: z.literal("type1") }),
    z.object({ type: z.literal("unknown"), value: z.unknown() })
  ])
  .catch(result => ({ type: "unknown" as const, value: result.input }));
```

PR #6192, opened 2026-07-14
Preserve the original input in catch callbacks after coercion

zelinewang:
`ZodCatch` currently builds its callback context after the inner schema has
parsed the payload. If that schema coerces or transforms the value before a
later check fails, both `ctx.value` and the deprecated `ctx.input` expose the
intermediate value instead of the input passed to `.parse()`.

This breaks the original-input access added for `.catch()` in #1933. It also
conflicts with the current API documentation, which describes `ctx.value` as
the input value. For example, a failed `z.coerce.number()` refinement receives
`5` instead of `"5"`, so a dynamic fallback cannot distinguish a string input
from a numeric input.

This change captures the input before running the inner schema and restores it
only in the catch callback context. The parsed result and validation issue
inputs are unchanged. The regression test covers both synchronous checks and
asynchronous refinements after coercion.


wahajahmed010 (approved):
## Review: colinhacks/zod#6192

**Logic**: ✅ The bug was that `ctx.input` in catch callbacks showed the *coerced* value instead of the original input. The fix captures `payload.value` before the inner type runs (which may coerce it), then passes that original value as both `value` and `input` to the catch callback.

**Edge cases**: ✅ The fix is applied in both the sync and async paths. The `input` variable is captured once at the top of the forward-direction block, so it correctly holds the pre-coercion value.

**Tests**: ✅ Good test covering both sync `.parse()` and async `.parseAsync()` paths, verifying that `ctx.value` and `ctx.input` both equal the original string input after coercion.

**Breaking changes**: None — this fixes a bug where the documented contract (`ctx.input` = original input) was violated.

**No issues found.** Clean, well-tested behavioral fix.

PR #6197, opened 2026-07-16
test: add regression coverage for union-in-object inference (#2654)

ShreyasUday:
Closes #2654

Adds focused regression tests pinning the union in object inference fix for every edge case reported in the issue thread. Both v3 compat and v4 classic are covered to prevent silent regressions.

**v3** (`packages/zod/src/v3/tests/issue-2654.test.ts`):
- `string | string[]` object property (original repro by @ericallam)
- `string[] | Record<string, string>` union property (@szulcus)
- `enum | Record<string, enum>` union property (@AlexGalays)
- mixed required + optional keys with union

**v4** (`packages/zod/src/v4/classic/tests/issue-2654.test.ts`):
- all of the above, plus:
- union with `.nullable()` inside object
- union inside nested `z.object()`

Full suite: 3831 tests pass, no type errors.

/claim #2654

### DIFF
--- a/packages/zod/src/v3/tests/issue-2654.test.ts
+++ b/packages/zod/src/v3/tests/issue-2654.test.ts
@@ -1,0 +1,88 @@
+// @ts-ignore TS6133
+import { expect, test } from "vitest";
+import * as z from "zod/v3";
+import { util } from "../helpers/util.js";
+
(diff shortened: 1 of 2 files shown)


colinhacks:

PR #6207, opened 2026-07-19
fix: error messages for exact length checks should say 'exactly N'

### LINKED ISSUE #6176: v4: `.length()` / `.size()` error messages ignore the `exact` flag

GowsiSM:
**Scope note:** This PR only updates the English locale (`en.ts`). Other locales will be updated in a separate follow-up PR to keep this focused and reviewable.

When using `.length()` or `.size()` on strings, arrays, or other sizable types, the error messages incorrectly show range language (`>=` or `<=`) even though the validation requires an exact value.

The issue object correctly sets the `exact: true` flag, but the English locale message formatter in `en.ts` ignores it and always
renders range-based messages.

This fix:
- Adds a check for the `exact` flag in `too_big` and `too_small` cases
- Returns "exactly N" when `exact` is true
- Keeps range messages (`>=N`, `<=N`) for `min()` and `max()`
- Updates snapshots in `array.test.ts` and `validations.test.ts`
- Adds a new test file verifying both exact and range messages

Fixes #6176


wahajahmed010:
**Review: fix exact-length error messages**

Thanks for tackling this — the bug is real and the fix is correct in scope. Two concerns before this can land:

**1. Other locales are not updated (scope concern)**

`en.ts` is one of ~50 locale files. The `exact` flag is set on the issue regardless of locale, so users on `de.ts`, `fr.ts`, `ja.ts`, etc. will still see the incorrect `>=N` / `<=N` messages for `.length()` and `.size()`. Either:

- Update all 50 locales (large but thorough), OR
- Land this PR as an en-only fix and file a follow-up issue explicitly listing the other locales

I'd lean toward the second — a partial fix that ships in one locale is still better than nothing, and forcing the same change across translations in one PR will be a nightmare to review and rebase.

**2. `as any` cast is unnecessary**

The `exact` field is already declared on the official interface in `packages/zod/src/v4/core/errors.ts`:
```ts
export interface $ZodIssueTooBig<Input> extends $ZodIssueBase {
  readonly code: "too_big";
  ...
  readonly exact?: boolean;
}
```

So `(issue as any).exact` should just be `issue.exact`. Removing the cast surfaces any type issues and is the right pattern for the codebase.

**Minor — `sizing.unit ?? "elements"`**

You added the nullish-coalescing fallback in the `too_big` exact branch but not in the `too_small` exact branch (it matches the original, which also lacked it). Worth deciding: either both branches should be defensive, or neither. Consistency is fine here, but make it a conscious choice — easy to miss.

**Verdict:** Logic is correct, tests cover the new behavior including regression of `min()`/`max()`. Once the `as any` is dropped and the scope (en-only vs all) is clarified, this is good to merge.

GowsiSM:
Thanks for the detailed review and helpful suggestions!

I've addressed the points you raised:

- Removed the unnecessary `as any` cast and switched to `issue.exact`.
- Added the defensive `?? "elements"` fallback for consistency.
- Clarified in the PR description that this is intentionally an English-only change, with other locales to be handled in a follow-up PR.
- Added the missing `size()` exact-length test and fixed the formatting/import ordering issues.

Regarding the failing **"Test with TypeScript latest"** check, I don't think it's related to this PR. I was able to reproduce the same `TypeError: Cannot read properties of undefined (reading 'readFile')` on a clean checkout of `main` with `typescript@latest` (currently TS 7.0.2), which looks like a `zshy` compatibility issue. That also explains why the TypeScript 5.5 job was cancelled due to the matrix's fail-fast behavior.

The full test suite passes locally under TypeScript 5.5, including the new tests.

colinhacks:
Closing in favor of #6177, which makes the same `en.ts` change with fuller coverage — both the `too_small` and `too_big` sides for `set.size()` and `map.size()`, which this one doesn't test.

This PR also carries a `test.sh` harness with `cd /app` in it, a random-suffixed test filename, and unrelated `.gitignore` and `attw.test.ts` edits, none of which should land here. Thanks for looking into it 👍

PR #6215, opened 2026-07-21
fix: encode / and ~ in toJSONSchema \ pointer (issue #6027)

deepshekhardas:
When a schema's \id\ (set via \.meta({ id })\) contains \/\ or \~\ characters, the generated \\\ pointer in the JSON Schema output was not escaped per RFC 6901, producing invalid pointers.

This adds a JSON Pointer escape step (\~\ → \~0\, \/\ → \~1\) when building \\\ URIs in \makeURI\, while keeping the \\\ key unescaped (JSON object keys don't need it).


deepshekhardas:
Superseded by #6257 — cleaner single-commit fix

PR #6216, opened 2026-07-21
fix: use 'exactly' in locale messages when issue.exact is true (issue #6176)

deepshekhardas:
﻿When .length() or .size() is used (which sets exact: true on the issue), the error message should say "exactly N" rather than ">=N" / "<=N".

Changes:
- locales/en.ts: too_big and too_small handlers now check issue.exact and produce "exactly N" messages when set
- tests/locales_en.test.ts: new test covering exact and non-exact messages

z.string().length(4).safeParse("abc") now produces "Too small: expected string to have exactly 4 characters" instead of "Too small: expected string to have >=4 characters".

### DIFF
--- a/packages/zod/src/v4/classic/tests/locales_en.test.ts
+++ b/packages/zod/src/v4/classic/tests/locales_en.test.ts
@@ -1,0 +1,24 @@
+import { expect, test } from "vitest";
+import * as z from "zod/v4";
+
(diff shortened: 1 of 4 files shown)


deepshekhardas:
Superseded by #6242 — cleaner locale-only fix

Issue #6154, opened 2026-07-04
z.iso.datetime() rejects extended-year ISO strings produced by Date#toISOString() (e.g. year 10000+)


kkawaharanet:
## Description

`z.iso.datetime()` rejects ISO 8601 datetime strings using the expanded year representation (leading `+`/`-` sign with a 6-digit year).

## Reproduction

```ts
import { z } from "zod";

const schema = z.iso.datetime();

const date1 = new Date("9999-12-31T00:00:00.000Z").toISOString();
console.log(date1); // "9999-12-31T00:00:00.000Z"
console.log(schema.safeParse(date1)); // { success: true, data: '9999-12-31T00:00:00.000Z' }

const date2 = new Date("+010000-01-01T00:00:00.000Z").toISOString();
console.log(date2); // "+010000-01-01T00:00:00.000Z"
console.log(schema.safeParse(date2));
// { success: false, error: ZodError: [{ code: "invalid_format", format: "datetime", message: "Invalid ISO datetime", ... }] }
```

## Expected behavior

Both `safeParse` calls should succeed, since both inputs are valid outputs of `Date#toISOString()`.

Per the [docs](https://zod.dev/api?id=iso-datetimes#iso-datetimes):

> The `z.iso.datetime()` method enforces ISO 8601;

ISO 8601 permits an expanded year representation (leading `+`/`-` sign with a 6-digit year) for years outside 0000–9999.

Rejecting a valid ISO 8601 expanded-year string is a mismatch between the documented behavior and the implementation.

## Actual behavior

The year-10000 case fails with `Invalid ISO datetime`.

## Additional context

Zod version: 4.4.3

colinhacks:
Closing as intended behavior. The `z.iso.datetime()` regex accepts a strict subset of ISO 8601 — it already rejects basic format (`20200101T061500Z`), week dates (`2020-W01-1`), ordinal dates (`2020-001`), comma decimal separators and `24:00`, all of which the spec permits just as much as expanded years. The narrowness is the point, so "ISO 8601 allows it" isn't on its own a reason to widen the regex.

[...]

PR #6217, opened 2026-07-21
fix: accept extended-year ISO strings in datetime regex (issue #6154)

deepshekhardas:
﻿Date#toISOString() produces extended-year ISO strings (e.g. "+010000-01-01T00:00:00.000Z") for years >= 10000 or <= -000000. The zod datetime regex only accepted 4-digit years, causing false validation failures for valid ISO 8601 strings.

Changes:
- regexes.ts: dateSource now accepts [±]\d{6} extended-year format as a third alternative branch (with basic month/day validation)
- tests/datetime.test.ts: test for extended year parsing
- Updated to-json-schema inline snapshots reflecting the changed regex


deepshekhardas:
Superseded by #6256 — cleaner regex fix

PR #6218, opened 2026-07-21
fix: move pnpm build before matrix TypeScript swap in CI (issue #6185)

deepshekhardas:
﻿The `pnpm build` step was running after `pnpm add typescript@latest`, which caused zshy to crash on TypeScript 7.x (which ships the Go-native compiler without the JS compiler API).

Moving `pnpm build` before the matrix TypeScript swap ensures the build uses the repo's pinned TypeScript. The matrix swap continues to apply to the test and typecheck phase (which is the intended purpose of the matrix).


deepshekhardas:
Superseded by #6253 — cleaner CI fix

PR #6222, opened 2026-07-22
fix(object): report __proto__ as unrecognized key in strict mode (#6220)

### LINKED ISSUE #6220: .strict() silently accepts an own __proto__ input key instead of reporting unrecognized_keys
**Not a vulnerability report** — no prototype pollution occurs; the existing defense works. This is a question about `.strict()`'s reporting semantics that falls out of where that defense sits.

### What happens

`z.object(...).strict()` (and `z.strictObject`) reports every unknown input key as `unrecognized_keys` — except an own `__proto__` key, which is silently accepted and silently dropped from the parse output. So strict mode flags every unknown key except the one an attacker would actually send, and downstream code that inspects `parsed.data` never learns the key was there.

### Repro (zod 4.4.3, fresh install, Node 24)

```js
import { z } from 'zod';

const strict = z.object({ name: z.string() }).strict();

// JSON.parse creates an OWN "__proto__" property — the wire reality for any JSON body.
const wire = JSON.parse('{"name":"a","__proto__":{"admin":true}}');
console.log(Object.getOwnPropertyNames(wire)); // [ 'name', '__proto__' ]

const res = strict.safeParse(wire);
console.log(res.success);                                          // true   (expected: false)
console.log(Object.getOwnPropertyNames(res.data));                 // [ 'name' ]  — dropped
console.log(Object.getPrototypeOf(res.data) === Object.prototype); // true — defense intact 👍

// Control: any other unknown key is reported as documented.
console.log(strict.safeParse(JSON.parse('{"name":"a","other":1}')).success); // false, unrecognized_keys
```

### Where it comes from

The pollution guard in `handleCatchall` skips `__proto__` **before** the `unrecognized.push(key)` branch:

https://github.com/colinhacks/zod/blob/912f0f51b0ced654d0069741e7160834dca742ee/packages/zod/src/v4/core/schemas.ts#L1870-L1878

```ts
for (const key in input) {
  // skip __proto__ so it can't replace the result prototype via the
  // assignment setter on the plain {} we build into
  if (key === "__proto__") continue;   // ← also exempts it from the strict check below
  if (keySet.has(key)) continue;
  if (t === "never") {
    unrecognized.push(key);
    continue;
  }
```

The comment justifies not *copying* the key into the result (clearly right, please keep it). It doesn't say the key should also be exempt from strict's *reporting* — that looks like a side effect of where the `continue` sits. `z.record` has the same silent skip (schemas.ts#L2957 (same permalink base)).

### Why it matters

For anyone using `.strict()` as a trust-boundary contract ("a body carrying anything unexpected is refused loudly, never stripped silently"), this is the one input where the contract silently doesn't hold — and because the key is also absent from the output, a downstream check can't catch it either. We hit this in an e-signature API where a rejected-extra-keys guarantee is load-bearing; our unit tests proved the downstream authority rejects a smuggled `__proto__`, while the deployed pipeline never showed it that key at all. (Worked around with an own-key check beside the parse.)

### Suggested fix, if unintended

Report before skipping — keeps the copy-defense byte-for-byte identical:

```ts
if (key === "__proto__") {
  if (t === "never") unrecognized.push(key);
  continue;
}
```

Happy to send a PR with tests if you'd take this. And if the current behavior is intentional, a line in the `.strict()` docs naming the exemption would let integrators know to pin own-keys themselves.

gonzoblasco:
Fixes #6220

When using `.strict()` (or `z.strictObject`), an own `__proto__` key in the input was silently accepted and dropped from the output without being reported as `unrecognized_keys`. This meant strict mode flagged every unknown key except the one an attacker would actually send, and downstream code inspecting the parsed data never learned the key was there.

The prototype pollution defense (skipping `__proto__` in the copy loop) is preserved — the fix only moves the `unrecognized` reporting before the `continue`, so the key is still not copied into the result but is now correctly surfaced in the error issues.

### Changes

- **`schemas.ts`**: In `handleCatchall`, report `__proto__` as unrecognized before skipping the copy
- **`object.test.ts`**: Updated the existing test to expect `__proto__` to be reported as unrecognized, and verified the error contains the key

### Testing

- 3811 tests pass (339 test files)
- TypeScript: clean

### DIFF
--- a/packages/zod/src/v4/classic/tests/object.test.ts
+++ b/packages/zod/src/v4/classic/tests/object.test.ts
@@ -709,9 +709,15 @@ describe("__proto__ in object catchall paths", () => {
     }
   });
 
-  test("strict does not surface __proto__ as unrecognized", () => {
+  test("strict surfaces __proto__ as unrecognized", () => {
     const schema = z.object({ name: z.string() }).strict();
     const result = schema.safeParse(protoInput());
-    expect(result.success).toBe(true);
+    expect(result.success).toBe(false);
+    if (!result.success) {
+      expect(result.error.issues[0].code).toBe("unrecognized_keys");
+      if (result.error.issues[0].code === "unrecognized_keys") {
+        expect(result.error.issues[0].keys).toContain("__proto__");
+      }
+    }
   });
 });
--- a/packages/zod/src/v4/classic/tests/object.test.ts
+++ b/packages/zod/src/v4/classic/tests/object.test.ts
@@ -720,4 +720,25 @@ describe("__proto__ in object catchall paths", () => {
       }
     }
   });
+
+  test("strict does not report inherited __proto__ from prototype chain", () => {
+    // for...in iterates enumerable inherited keys. __proto__ is an accessor
+    // on Object.prototype, not an enumerable key — so it's never visited.
+    // Properties inside the __proto__ object (e.g. isAdmin) ARE iterated
+    // and reported as unrecognized. This test documents that boundary.
+    const input = Object.create({ __proto__: { isAdmin: true } });
+    input.name = "alice";
+    const schema = z.object({ name: z.string() }).strict();
+    const result = schema.safeParse(input);
+    expect(result.success).toBe(false);
+    if (!result.success) {
+      expect(result.error.issues[0].code).toBe("unrecognized_keys");
+      if (result.error.issues[0].code === "unrecognized_keys") {
+        // isAdmin is inherited via the prototype chain and iterated by for...in
+        expect(result.error.issues[0].keys).toContain("isAdmin");
+        // __proto__ itself is not an enumerable key, so it's not reported
+        expect(result.error.issues[0].keys).not.toContain("__proto__");
+      }
+    }
+  });
 });
--- a/packages/zod/src/v4/core/errors.ts
+++ b/packages/zod/src/v4/core/errors.ts
@@ -82,6 +82,12 @@ export interface $ZodIssueNotMultipleOf<Input extends number | bigint = number |
 
 export interface $ZodIssueUnrecognizedKeys extends $ZodIssueBase {
   readonly code: "unrecognized_keys";
+  /**
+   * The unrecognized key names. May include "__proto__" as a literal string.
+   * Consumers should treat these as raw strings — avoid passing them through
+   * Object.assign, spread, or JSON.parse without sanitization, as "__proto__"
+   * can re-introduce prototype pollution if re-interpreted as an accessor.
+   */
   readonly keys: string[];
   readonly input?: Record<string, unknown>;
 }
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -1870,7 +1870,10 @@ function handleCatchall(
   for (const key in input) {
     // skip __proto__ so it can't replace the result prototype via the
     // assignment setter on the plain {} we build into
-    if (key === "__proto__") continue;
+    if (key === "__proto__") {
+      if (t === "never") unrecognized.push(key);
+      continue;
+    }
     if (keySet.has(key)) continue;
     if (t === "never") {
       unrecognized.push(key);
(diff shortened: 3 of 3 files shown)


gonzoblasco:
The CI failure on "Test with TypeScript latest" is a pre-existing issue with `zshy` (zod's build tool) crashing on the latest TypeScript — `TypeError: Cannot read properties of undefined (reading 'readFile')`. The same error occurs on the base branch. The other CI jobs (TS 5.5, lint, circular deps) all pass cleanly. No changes needed on this PR's side.

wahajahmed010:
### Review of #6222 — `fix(object): report __proto__ as unrecognized key in strict mode`

The fix is well-scoped and correct. Moving the `unrecognized.push(key)` above the `continue` cleanly resolves the reporting asymmetry without touching the prototype-pollution guard (the key is still never copied into the result), and the test rename + new assertions lock the contract in.

**A few things worth confirming before merge:**

1. **Strict-only? — look at the surrounding `t` value.** The patch only adds the `unrecognized.push` call when `t === "never"`, which corresponds to `.strict()` / `z.strictObject`. Worth a quick mental check: are there other "catchall modes" where an own `__proto__` arriving through `JSON.parse` should also be reported (e.g. `.passthrough`, `.strip`)? If yes, the `t === "never"` guard is the right scope, but a one-line comment explaining *why* `__proto__` is special-cased to strict mode only (vs. silently dropped in passthrough/strip by design) would help future readers — the current PR description justifies the strict fix but doesn't address the other modes, which is where the next maintainer will look first when triaging similar issues.

2. **Symmetry with `Object.hasOwn`.** The loop is `for (const key in input)`, which iterates enumerable string keys including inherited ones (modulo the `__proto__` skip). For JSON-parsed inputs this is fine because `__proto__` is own, but the same `unrecognized.push` would also fire for `__proto__` inherited via the input's prototype chain. Probably harmless (the input didn't contain it), but worth a sentence in the PR description or a test that explicitly covers `Object.create({ __proto__: {...} })` if you want the contract to be precisely "own keys only."

3. **Error key naming.** `unrecognized.keys` will now contain the literal string `"__proto__"`. Any downstream consumer that stringifies this and re-parses it (or that filters it through `Object.assign` / spread) could re-introduce exactly the problem the guard is preventing. Consider documenting the format in the issue type's docstring, or — stronger — adding a `JSON.stringify`-style safe repr in the error payload. Out of scope for this fix, but worth filing a follow-up issue.

4. **The unrelated CI failure** (`zshy` vs. TS latest) is correctly diagnosed as pre-existing on the base branch. No action here.

Logic, error handling, and security posture all look right. Approving the direction; the only change I'd want before merge is the comment in point 1 to make the strict-only scope explicit.

— automated review via the daily code-review pipeline

gonzoblasco:
Thanks for the thorough review - all three points addressed.

**Point 1 (strict-only scope):** Added a comment in `handleCatchall` explaining why `__proto__` is only reported in strict mode: passthrough/strip/catchall silently drop it by design since the key is never copied into the result regardless of mode, so there is nothing to report.

**Point 2 (inherited `__proto__` via prototype chain):** Added a test using `Object.create({ __proto__: { isAdmin: true } })`. The `for...in` loop iterates enumerable inherited keys, so `isAdmin` (inside the `__proto__` object) is reported as unrecognized. But `__proto__` itself is an accessor on `Object.prototype`, not an enumerable key, so it is never visited by `for...in` and never reported. The test asserts both sides of that boundary.

**Point 3 (error key naming / `"__proto__"` in `unrecognized.keys`):** Added a JSDoc comment on the `$ZodIssueUnrecognizedKeys.keys` field warning consumers that the array may contain the literal string `"__proto__"` and should not be passed through `Object.assign`, spread, or `JSON.parse` without sanitization. Did not change the runtime value — `stringifyPrimitive` already wraps strings in double quotes for error messages, and changing the API surface would be a breaking change for consumers comparing against `"__proto__"` directly.

**CI failure:** The `zshy` build error and the `attw` test failure are both pre-existing on the base branch - unrelated to this PR.

colinhacks:
Going with #6221 for this one — it lands the same behavior with a smaller diff, and its test asserts the resulting issue code and keys.

I left the `errors.ts` JSDoc out. The warning is aimed at consumers who re-interpret issue `keys` through a spread or `Object.assign`, which is a caller-side concern rather than something the issue type should carry.

Worth noting for anyone reading later: since #6354 a declared `__proto__` shape key round-trips as a real field, so the check has to sit after the `keySet` lookup or `.strict()` starts rejecting input the schema explicitly allows. Folded that into #6221.

PR #6223, opened 2026-07-23
fix: accept exact decimal multiples in multipleOf

spokodev:
## What

`z.number().multipleOf(0.07)` rejects some exact decimal multiples:

```ts
z.number().multipleOf(0.07).safeParse(2.03).success // false  — but 2.03 === 29 * 0.07
z.number().multipleOf(0.07).safeParse(4.06).success // false  — 58 * 0.07
z.number().multipleOf(0.07).safeParse(1.96).success // true   — 28 * 0.07
z.number().multipleOf(0.07).safeParse(2.10).success // true   — 30 * 0.07
```

`floatSafeRemainder` compares `val / step` to its rounded value with a one-ULP tolerance (`Number.EPSILON * Math.max(Math.abs(ratio), 1)`). But `2.03 / 0.07 === 28.999999999999993`, ~1.1 ULP away from 29: `val` and `step` are each already rounded to the nearest double, and the division rounds a third time, so the quotient's error can exceed a single ULP. Neighbouring multiples happen to land within one ULP, so the rejection is inconsistent (and it's a regression vs Zod 3, which accepts `2.03`).

## Fix

Widen the tolerance to a few ULPs. To pick a safe factor I swept ~2M exact decimal multiples (steps with up to 4 decimal places):

- the quotient error for a genuine multiple never exceeds **~1.17 ULP**;
- the nearest genuine **non**-multiple that must stay rejected sits **~7×10⁴ ULP** away (e.g. `step 0.1` vs `6.1000000001`).

Four ULPs removes every false rejection with a ~18000× margin before any false acceptance becomes possible. The existing small-step / scientific-notation regression tests (`multipleOf(1e-7)` rejecting `2.5e-7`, `multipleOf(1e-10)` accepting `1e-9`, etc.) still pass.

## Tests

Added `.multipleOf() accepts exact decimal multiples` asserting `2.03`, `4.06`, `8.54` pass and the non-multiple `2.04` is still rejected. It fails on `main` and passes with the change; the number test file is green (`vitest run … number.test.ts`).


colinhacks:
Good investigation. Merged main in and joined the tolerance comment onto one line — this repo doesn't hard-wrap prose comments. Also corrected the ULP figure: the quotient sits 2 ULP below 29, and 1.1 was its ratio to the old tolerance rather than a ULP count. Merging once CI is green.

PR #6224, opened 2026-07-23
fix(core): treat any/unknown/coerced-boolean object properties as optional-in

### LINKED ISSUE #5957: v4.4 now errors on z.coerce.boolean parsing undefined
In earlier versions of Zod, the following code behaved differently.

```typescript
const test = z.object( { foo: z.coerce.boolean() } )
const result = test.parse({}) // before = { foo: false },  after = ZodError "Invalid input: expected nonoptional, received undefined"
```

I might understand this being intentional (although our code relied on the old behavior when parsing process.env to provide fallbacks to `false`). But I didn't see it actually documented in the release notes.

### LINKED ISSUE #5997: Undocumented (?) change to `z.any()` object property behavior in v4.4.0
Since [v4.4.0](https://github.com/colinhacks/zod/releases/tag/v4.4.0), object properties defined as `z.any()` are now treated as required (i.e. the object will fail to parse if the property is unset). The release notes say this is now the expected behavior for `z.undefined()`, but don't mention `z.any()`, so I assume this is unintended?

Reproduction: [v4.3.6](https://zod-playground.vercel.app?appdata=N4IgzgxgFgpgtgQxALhAejQAgMIHsB2AZgJYDmArgE4yYAusmANrhAozYbpZjJZV5jgwwYBKWGYAFLgAOtYgTYBKADr4MmAF4A6CARKlJO5q3ZhtxWpKWr8avfjC1MkWIkwBeLdtwAjAFYwEFbAaph0wrQAgvgAngAK-LS8tLHI3ghx1moAvrZq1LRU%2BC7Q8EgANCAAbmzkwigA2iChJeGY%2BAhC6SogAFK4UHYguSAAulXVvGAK%2BCggACzaAMzaAGwgVcRgAFq4ACYAssT4xCiEbGAwOUA), [v4.4.0](https://zod-playground.vercel.app?appdata=N4IgzgxgFgpgtgQxALhAejQAgMIHsB2AZgJYDmArgE4yYAusmANrhAozYbpZjJZV5jgwwYBKWGYAFLgAOtYgTYBKADr4MmAF4A6CARKlJO5q3ZhtxWpKWr8avfjC1MkWIkwBeLdtwAjAFYwEFbAaph0wrQAgvgAngAK-LS8tLHI3ghx1moAvrZq1LRU%2BC7Q8EgANCAAbmzkwigA2iChJeGY%2BAhC6SogAFK4UHYguSAAulXVvGAK%2BCggACzaSwAMIFXEYABauAAmALLE%2BMQohGxgMDlAA)

mabdullah465:
# fix(core): treat any/unknown/coerced-boolean object properties as optional-in

Fixes #5997
Fixes #5957

## Root cause

`handlePropertyResult` in `packages/zod/src/v4/core/schemas.ts` decides whether
a missing object key is an error by checking `el._zod.optin === "optional"`:

```ts
if (!isPresent && !isOptionalIn) {
  if (!result.issues.length) {
    final.issues.push({
      code: "invalid_type",
      expected: "nonoptional",
      input: undefined,
      path: [key],
    });
  }
  return;
}
```

`$ZodAny`, `$ZodUnknown`, and `$ZodBoolean` (even with `coerce: true`) never
set `optin = "optional"`, so an absent key on these properties gets rejected
as a missing required property — even though these schemas are semantically
supposed to tolerate an absent/undefined input:

- `z.any()` / `z.unknown()` accept `undefined` unconditionally.
- `z.coerce.boolean()` coerces via `Boolean(payload.value)`, and
  `Boolean(undefined) === false` — so it should coerce an absent key to
  `false`, not throw.

This is the shared root cause behind two separately-reported regressions:

- #5997 — `z.any()` object properties became required in v4.4.0 (undocumented)
- #5957 — `z.coerce.boolean()` rejects an absent key instead of coercing it to `false`

## Fix

Set `inst._zod.optin = "optional"`:
- Unconditionally in `$ZodAny` and `$ZodUnknown`.
- In `$ZodBoolean`, only when `def.coerce` is `true` — a plain, non-coercing
  `z.boolean()` correctly remains required (covered by a new test).

## Testing

- Added test cases to `anyunknown.test.ts` and `coerce.test.ts` covering both
  reported regressions plus the "non-coerced boolean still required" guard.
- Ran the full package test suite (`pnpm vitest run --project zod`): **336
  test files, 3808 tests, 0 failures, 0 type errors** — no regressions
  elsewhere in the suite.

## Disclosure

I used Claude Code (Anthropic) to help trace the root cause through the
source, prepare this patch, and write/run the verification above. I reviewed
the change and test results myself before opening this PR.

### DIFF
--- a/packages/zod/src/v4/classic/tests/anyunknown.test.ts
+++ b/packages/zod/src/v4/classic/tests/anyunknown.test.ts
@@ -24,3 +24,17 @@ test("check never inference", () => {
   expect(() => t1.parse("asdf")).toThrow();
   expect(() => t1.parse(null)).toThrow();
 });
+
+test("any object property tolerates an absent key (#5997)", () => {
+  const schema = z.object({ foo: z.any() });
+  const result = schema.safeParse({});
+  expect(result.success).toBe(true);
+  expect(result.data).toEqual({});
+});
+
+test("unknown object property tolerates an absent key", () => {
+  const schema = z.object({ foo: z.unknown() });
+  const result = schema.safeParse({});
+  expect(result.success).toBe(true);
+  expect(result.data).toEqual({});
+});
--- a/packages/zod/src/v4/classic/tests/coerce.test.ts
+++ b/packages/zod/src/v4/classic/tests/coerce.test.ts
@@ -75,6 +75,19 @@ test("boolean coercion", () => {
   expect(schema.parse(new Date(1670139203496))).toEqual(true);
 });
 
+test("coerced boolean object property tolerates an absent key (#5957)", () => {
+  const schema = z.object({ foo: z.coerce.boolean() });
+  const result = schema.safeParse({});
+  expect(result.success).toBe(true);
+  expect(result.data).toEqual({ foo: false });
+});
+
+test("non-coerced boolean object property is still required", () => {
+  const schema = z.object({ foo: z.boolean() });
+  const result = schema.safeParse({});
+  expect(result.success).toBe(false);
+});
+
 test("bigint coercion", () => {
   const schema = z.coerce.bigint();
   expect(schema.parse("5")).toEqual(BigInt(5));
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -1204,6 +1204,11 @@ export const $ZodBoolean: core.$constructor<$ZodBoolean> = /*@__PURE__*/ core.$c
     $ZodType.init(inst, def);
     inst._zod.pattern = regexes.boolean;
 
+    // A coercing boolean accepts undefined (Boolean(undefined) === false),
+    // so an absent object key should be coerced rather than rejected as a
+    // missing required property.
+    if (def.coerce) inst._zod.optin = "optional";
+
     inst._zod.parse = (payload, _ctx) => {
       if (def.coerce)
         try {
@@ -1445,6 +1450,9 @@ export interface $ZodAny extends $ZodType {
 export const $ZodAny: core.$constructor<$ZodAny> = /*@__PURE__*/ core.$constructor("$ZodAny", (inst, def) => {
   $ZodType.init(inst, def);
 
+  // `any` accepts undefined as a valid value, so an absent object key
+  // should not be treated as a missing required property either.
+  inst._zod.optin = "optional";
   inst._zod.parse = (payload) => payload;
 });
 
@@ -1474,6 +1482,9 @@ export const $ZodUnknown: core.$constructor<$ZodUnknown> = /*@__PURE__*/ core.$c
   (inst, def) => {
     $ZodType.init(inst, def);
 
+    // Same as `any`: unknown accepts undefined, so an absent object key
+    // should not be treated as a missing required property.
+    inst._zod.optin = "optional";
     inst._zod.parse = (payload) => payload;
   }
 );


wahajahmed010:
### Review of #6224 — `fix(core): treat any/unknown/coerced-boolean object properties as optional-in`

Nice root-cause analysis — pinning the regression to a single property (`inst._zod.optin`) instead of a per-call branch in `handlePropertyResult` is the right shape. The fix is minimal, the test for "non-coerced `z.boolean()` still required" is the exact guard I would have asked for, and bundling two issues with a shared root cause into one PR is reasonable here.

**Substantive points:**

1. **`$ZodUnknown` test gap.** You added a `safeParse({})` test for `z.unknown()` — good. But the most common failure mode for `unknown` properties historically has been `null` vs. `undefined` (e.g. `safeParse({foo: null})` for `{foo: z.unknown()}`). Since the new `optin = "optional"` flag now short-circuits the missing-key path entirely, please add a one-line test asserting that an explicit `null` value still produces an `invalid_type` issue, so we don't silently regress null handling in the process.

2. **`$ZodAny` value-passthrough is unchanged, but worth a regression note.** `z.any()`'s `parse` is literally `(payload) => payload`, so it returns `undefined` for an absent key. With the new `optin = "optional"`, the caller now gets `result.data = {}` (no `foo` key) instead of an error. That's clearly the intended behavior per #5997, but the behavior contract is now subtly different from v3 in a way that may surprise consumers who do `result.data.foo ?? "default"`. Probably fine, but worth a one-line note in the changelog/release notes under "Minor behavior change" so the v4 → v4.4.x migration story stays honest.

3. **Coerced booleans and the `nan` / `0` / `""` edge case.** `Boolean(undefined) === false`, so this fix is correct for the absent case. But the same coercion path also makes `z.coerce.boolean()` accept the string `"false"` as `true` (since non-empty strings are truthy in JS), which is a long-standing footgun unrelated to your change. Don't fix it here — just flagging because the diff is touching exactly this surface and a maintainer might want to bundle a follow-up.

4. **Disclosure transparency.** Thanks for the explicit Claude Code disclosure and the "I reviewed the change and test results myself" line. The change is small enough to audit by eye, and the diff matches your description, so I don't have anything to add there — but the practice itself is appreciated. Other contributors should follow this pattern.

5. **Type concern — `inst._zod.optin` write site.** You're mutating an internal property directly in the constructor for `$ZodAny` / `$ZodUnknown` / `$ZodBoolean` (coerce). If `$ZodType.init` ever grows logic that derives `optin` from other constructor state (e.g. default values, catchall, `optional()` chaining), the order-of-operations matters: your writes happen *after* `init`, so a future `init` enhancement that initializes `optin = "required"` would silently overwrite your `"optional"`. Cheap insurance: assign unconditionally *inside* `init` if you control it, or add a comment that the write must come after `init`. Not blocking, but the kind of thing that bites in two release cycles.

No blockers from my side. The fix and tests are correct, the bundling is justified, and the disclosure is appreciated. Approving with the v3-vs-v4.4.x changelog note as the one ask.

— automated review via the daily code-review pipeline

colinhacks:
Closing this — the current behavior is intentional. `z.any()` and `z.unknown()` infer as required keys, and requiring the key at parse time is what makes the parser agree with the type it advertises. Same for `z.coerce.boolean()`.

`z.any().optional()` and `z.coerce.boolean().default(false)` are the ways to opt into an omittable key. Thanks for digging into this!

PR #6225, opened 2026-07-24
docs: add type-safe generics feature highlight to README

deepshekhardas:
Adds a section highlighting the new type-safe generics support in Zod v4 README, making it easier for users to discover this feature.


colinhacks:
This isn't the recommended pattern actually https://zod.dev/library-authors#how-to-accept-user-defined-schemas

PR #6228, opened 2026-07-24
added Hindi Locale

vedanshshetti:
# Added Hindi locale to Zod 4 Locales
- the new locale file is under `packages/zod/src/v4/locales/hi.ts`.
- it contains exactly the same contents as the other locale files.
- it uses correct **`ISO 639-1` compliant** naming standards.

> Side Note: I really love the file structure, it's all so simple, i've seen lots of other repos which are just... bad.


deepshekhardas:
Thanks for the Hindi locale! One thing — the locale won't be accessible until it's also exported from locales/index.ts. I've opened #6229 with the missing export added.

```diff
+ export { default as hi } from './hi.js';
```

PR #6229, opened 2026-07-25
feat: add Hindi locale (hi.ts) for Zod v4 error messages

deepshekhardas:
# Added Hindi locale to Zod 4 Locales

- the new locale file is under \packages/zod/src/v4/locales/hi.ts\.
- it contains exactly the same contents as the other locale files.
- it uses correct **ISO 639-1 compliant** naming standards.
- also registers the locale export in \locales/index.ts\.


colinhacks:
Closing this one out — #6315 landed Hindi alongside Kannada in 0175a043, so `hi.ts` is on `main` now.

No fault in this implementation; it came down to #6315 covering two locales in one merge, and both PRs needing the same registration work either way. Thanks for putting it together 👍

PR #6232, opened 2026-07-25
fix: accept exact decimal multiples in multipleOf

deepshekhardas:
Floating-point division accumulates rounding error from representing both val and step as floats. A single-ULP tolerance rejected exact decimal multiples; widening to 4 ULPs covers the error range while still rejecting genuine non-multiples.


deepshekhardas:
Superseded by original PR #6223 — the fix was already applied by the author.

PR #6234, opened 2026-07-26
fix: object & record intersection incorrectly validates keys owned by the object side

### LINKED ISSUE #2573: Intersection between `z.object` and `z.record` incorrectly merges validators

devthedevil:
## Summary

When intersecting an object schema with a record schema (`z.object({...}).and(z.record(...))`), the record side of the intersection was validated against the *entire* input, including keys that structurally belong to the object's own shape. If the record's `valueType` didn't happen to match the shape of those object-owned fields, this produced a spurious validation failure even though the object side already validated those keys correctly.

Closes #2573

## Root cause

`$ZodIntersection`'s `parse` runs `left` and `right` independently against the full input, then reconciles the two results in `handleIntersectionResults`. That reconciliation only special-cases `unrecognized_keys` issues (keys neither side wants) by requiring *both* sides to reject a key before it's reported. A record schema, however, doesn't reject an object-owned key as "unrecognized" — it tries to validate it against `valueType` and fails with an ordinary `invalid_type` issue, which isn't covered by that reconciliation and leaks straight through.

## Change

Before running each side of the intersection, filter out the sibling object schema's own shape keys from the input handed to the record side, so the record never attempts to validate keys that structurally belong to the object:

```ts
inst._zod.parse = (payload, ctx) => {
  const input = payload.value;
  const leftInput = getIntersectionSideInput(def.left, def.right, input);
  const rightInput = getIntersectionSideInput(def.right, def.left, input);
  const left = def.left._zod.run({ value: leftInput, issues: [] }, ctx);
  const right = def.right._zod.run({ value: rightInput, issues: [] }, ctx);
  ...

function getIntersectionSideInput(thisSchema, otherSchema, input) {
  if (thisSchema.def.type === "record" && otherSchema.def.type === "object" && util.isPlainObject(input)) {
    const shapeKeys = new Set(Object.keys(otherSchema._zod.def.shape));
    const filtered = {};
    for (const key of Object.keys(input)) if (!shapeKeys.has(key)) filtered[key] = input[key];
    return filtered;
  }
  return input;
}
```

This only changes behavior for object+record intersections specifically (checked in both orderings). Object+object, record+record, array, and all other intersection combinations are untouched, since the filter only activates when one side is `"record"` and the sibling is `"object"`.

## Testing

Added a regression test in `intersection.test.ts` covering: the reported bug (both `object.and(record)` and `record.and(object)` orderings), that genuinely-invalid extra keys under the record's `valueType` still correctly fail, and that the object side's own validation errors still surface correctly.

Ran the full existing suite locally to check for regressions — all passing, no changes needed elsewhere:

```
✓ intersection.test.ts (12 tests, including the new one)
✓ record.test.ts (24 tests)
✓ object.test.ts (59 tests)

95/95 passing
```

### DIFF
--- a/packages/zod/src/v4/classic/tests/intersection.test.ts
+++ b/packages/zod/src/v4/classic/tests/intersection.test.ts
@@ -204,7 +204,9 @@ test("object and record intersection: record should not validate keys owned by t
   // the object side. If the record's valueType didn't match the object's
   // field shapes, this produced a spurious validation failure even though
   // the object side already validated those keys correctly.
-  const Config = z.object({ name: z.object({ first: z.string() }) }).and(z.record(z.string(), z.object({ sub: z.string() })));
+  const Config = z
+    .object({ name: z.object({ first: z.string() }) })
+    .and(z.record(z.string(), z.object({ sub: z.string() })));
 
   const valid = { name: { first: "Ada" }, extra: { sub: "value" } };
   expect(Config.parse(valid)).toEqual(valid);
--- a/packages/zod/src/v4/classic/tests/intersection.test.ts
+++ b/packages/zod/src/v4/classic/tests/intersection.test.ts
@@ -190,7 +190,7 @@ test("invalid deep merge of object and array combination", async () => {
     })
   );
 
-  const students = [{ name: "John", surname": "Doe" }];
+  const students = [{ name: "John", surname: "Doe" }];
 
   expect(() => Registry.parse({ students })).toThrowErrorMatchingInlineSnapshot(
     `[Error: Unmergable intersection. Error path: ["students",0,"name"]]`
--- a/packages/zod/src/v4/classic/tests/intersection.test.ts
+++ b/packages/zod/src/v4/classic/tests/intersection.test.ts
@@ -190,9 +190,37 @@ test("invalid deep merge of object and array combination", async () => {
     })
   );
 
-  const students = [{ name: "John", surname: "Doe" }];
+  const students = [{ name: "John", surname": "Doe" }];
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing in favor of #6412. Good diagnosis — the reconciliation only understood `unrecognized_keys` and a record reports `invalid_key`, which is exactly right.

I went at it from the other end: the record was classifying the issue off `def.partial` instead of off the key schema, so fixing the classification lets the existing reconciliation handle it with no new filtering. That also covers a record intersected with another record, which the special case here doesn't reach.

Thanks for digging into this one 👍

PR #6235, opened 2026-07-26
fix: add offset option to z.iso.time() for timezone-aware time strings

### LINKED ISSUE #6204: Add support for RFC3339 full-time in z.iso.time()
## Problem

`z.iso.time()` validates RFC3339 `partial-time`, explicitly rejecting timezone offsets.

However, OpenAPI defines

```yaml
type: string
format: time
```

as RFC3339 `full-time`. As a result, valid OpenAPI values like

```text
10:15:30Z
10:15:30+02:00
```

cannot be validated using the built-in ISO validators.

## Use case

I'm using [Hey API](https://heyapi.dev/) to generate Zod schemas from an OpenAPI specification.

For most OpenAPI formats there is a corresponding Zod validator, but `format: time` currently has no equivalent because `z.iso.time()` implements `partial-time` instead of `full-time`.

This makes it impossible to generate a correct validator for an OpenAPI schema like:

```yaml
type: string
format: time
```

without resorting to a custom regex or `.refine()`.

## Possible solutions

Some possible APIs could be:

```ts
z.iso.time({ offset: true })
```

(similar to `z.iso.datetime({ offset: true })`)

or

```ts
z.iso.fullTime()

```

## Conclusion

Since OpenAPI and Zod are commonly used together, it would be great if OpenAPI's standard format: time could be mapped directly to a built-in Zod validator without requiring custom validation logic.

Related to #5542

okxint:
`z.iso.datetime({ offset: true })` accepts strings like `10:15:30Z` and `10:15:30+02:00`, but `z.iso.time()` has no equivalent option. This is a gap: OpenAPI 3.x specifies `format: time` as RFC 3339 `full-time`, which requires a timezone offset. There's currently no way to validate those strings with `z.iso.time`.

**Changes:**

`packages/zod/src/v4/core/regexes.ts` — add `offset?: boolean` to `time()`. When true, appends `(?:Z|[+-](?:[01]\d|2[0-3]):[0-5]\d)` to the pattern (same offset regex used by `datetime()`).

`packages/zod/src/v4/core/schemas.ts` — add `offset?: boolean` to `$ZodISOTimeDef`. `$ZodISOTimeParams` is derived from this via `StringFormatParams`, so the option is automatically available at the call site with no API file changes needed.

`packages/zod/src/v4/classic/tests/index.test.ts` — extend the existing `z.iso.time` test block with offset cases.

Fixes #6204

### DIFF
--- a/packages/zod/src/v4/classic/tests/index.test.ts
+++ b/packages/zod/src/v4/classic/tests/index.test.ts
@@ -150,6 +150,14 @@ test("z.iso.time", () => {
   expect(z.safeParse(c, d1).success).toEqual(true);
   expect(z.safeParse(c, d2).success).toEqual(true);
   expect(z.safeParse(c, d3).success).toEqual(false);
+
+  // offset: true allows timezone suffixes (Z, +HH:MM, -HH:MM)
+  const withOffset = z.iso.time({ offset: true });
+  expect(z.safeParse(withOffset, "10:15:30Z").success).toEqual(true);
+  expect(z.safeParse(withOffset, "10:15:30+02:00").success).toEqual(true);
+  expect(z.safeParse(withOffset, "10:15:30-05:30").success).toEqual(true);
+  expect(z.safeParse(withOffset, "10:15:30").success).toEqual(false); // no offset — rejected when offset required
+  expect(z.safeParse(withOffset, "bad data").success).toEqual(false);
 });
 
 test("z.iso.duration", () => {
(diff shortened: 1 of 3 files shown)


colinhacks:
Closing — the JSON Schema half of #6204 was fixed separately, and I'm not adding an offset option to `z.iso.time()`. The format validators are deliberately narrower than the specs they're named after, and `z.string().regex()` is the escape hatch when you need a wider format. Thanks for taking a run at it 👍

PR #6238, opened 2026-07-27
fix(toJSONSchema): serialize bigint defaults without throwing

### LINKED ISSUE #6209: BigInt with default value fails to run toJSONSchema even before the override function is run
## Summary

When exporting JSON Schema from a schema that contains a `bigint` default value, Zod throws before the `override` callback is invoked.

This makes it impossible to normalize `bigint` output (for example to `{ type: "integer", format: "int64" }`) inside `override` for schemas that also use `.default(0n)`.

## Minimal reproduction

```ts
import { z } from 'zod'

const schema = z.object({
  startAt: z.coerce.bigint().optional().default(0n)
})

const jsonSchema = schema.toJSONSchema({
  io: 'input',
  unrepresentable: 'any',
  override: ({ zodSchema, jsonSchema }) => {
      if (zodSchema._zod.def.type === 'bigint') {
        jsonSchema.type = 'integer'
        jsonSchema.format = 'int64'
      }
  }
})

console.log(jsonSchema)
```

## Actual behavior

A `TypeError` is thrown:

```txt
TypeError: Do not know how to serialize a BigInt
    at JSON.stringify (<anonymous>)
    at Module.defaultProcessor (.../zod/v4/core/json-schema-processors.js:464:36)
```

The failure occurs while processing defaults, before `override` can run for the relevant subtree.

## Expected behavior

One of the following should happen:

1. `override` is applied before default serialization for the affected node, or
2. `defaultProcessor` handles `bigint` defaults without throwing (for example by converting to string), or
3. a documented and supported option exists to control default serialization for `bigint`.

## Why this is a problem

`bigint` support can be represented in JSON Schema by projects with custom conventions (for example `integer` + `int64`).

Current behavior prevents customization for valid schemas that include `bigint` defaults because export fails too early.

## Workaround used locally

Temporarily patch schema internals to coerce `bigint` defaults to strings during `toJSONSchema`, then restore original values afterward.

```
function patchBigIntDefaultsForJsonSchema(rootSchema: unknown): () => void {
  const seen = new Set<object>()
  const restorations: Array<() => void> = []

  const visitSchema = (candidate: unknown): void => {
    if (!candidate || typeof candidate !== 'object' || seen.has(candidate)) {
      return
    }

    seen.add(candidate)

    const internalZod = Reflect.get(candidate, '_zod') as
      | { def?: Record<string, unknown> }
      | undefined
    const definition = internalZod?.def
    const valuesToVisit: Array<unknown> = []

    if (definition) {
      const defaultValueDescriptor = Object.getOwnPropertyDescriptor(
        definition,
        'defaultValue'
      )
      const defaultValue = definition.defaultValue

      if (
        typeof defaultValue === 'bigint' &&
        defaultValueDescriptor?.configurable
      ) {
        Object.defineProperty(definition, 'defaultValue', {
          value: defaultValue.toString(),
          writable: true,
          enumerable: defaultValueDescriptor.enumerable,
          configurable: true
        })
        restorations.push(() => {
          Object.defineProperty(
            definition,
            'defaultValue',
            defaultValueDescriptor
          )
        })
      }

      valuesToVisit.push(...Object.values(definition))
    } else {
      valuesToVisit.push(...Object.values(candidate as Record<string, unknown>))
    }

    for (const value of valuesToVisit) {
      if (Array.isArray(value)) {
        for (const item of value) {
          visitSchema(item)
        }
      } else {
        visitSchema(value)
      }
    }
  }

  visitSchema(rootSchema)

  return () => {
    for (const restore of restorations) {
      restore()
    }
  }
}
```

This is brittle and relies on internal implementation details.

## Environment

- Node: 24.4.1
- Zod: v4 (observed in `zod/v4/core/json-schema-processors.js`)
- Reproduced in a TypeScript test using `schema.toJSONSchema(...)`

deepshekhardas:
Fixes #6209

JSON.parse(JSON.stringify(...)) throws a TypeError when the default value is a BigInt because BigInt is not serializable by JSON.stringify. Convert to Number before serializing, matching the existing approach in \literalProcessor\ (line 177).

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2665,6 +2665,19 @@ test("falsy prefaults (false, 0, empty string)", () => {
   `);
 });
 
+test("bigint default toJSONSchema", () => {
+  const a = z.coerce.bigint().optional().default(0n);
+  expect(() => z.toJSONSchema(a, { unrepresentable: "any" })).not.toThrow();
+  expect(z.toJSONSchema(a, { unrepresentable: "any" })).toMatchInlineSnapshot(`
+    {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": 0,
+    }
+  `);
+  const b = z.coerce.bigint().prefault(0n);
+  expect(() => z.toJSONSchema(b, { unrepresentable: "any", io: "input" })).not.toThrow();
+});
+
 test("input type", () => {
   const schema = z.object({
     a: z.string(),
--- a/packages/zod/src/v4/core/json-schema-processors.ts
+++ b/packages/zod/src/v4/core/json-schema-processors.ts
@@ -498,7 +498,8 @@ export const defaultProcessor: Processor<schemas.$ZodDefault> = (schema, ctx, js
   process(def.innerType, ctx as any, params);
   const seen = ctx.seen.get(schema)!;
   seen.ref = def.innerType;
-  json.default = JSON.parse(JSON.stringify(def.defaultValue));
+  const dv = typeof def.defaultValue === "bigint" ? Number(def.defaultValue) : def.defaultValue;
+  json.default = JSON.parse(JSON.stringify(dv));
 };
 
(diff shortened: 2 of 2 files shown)


deepshekhardas:
Superseded by #6255 — cleaner single-file fix

PR #6239, opened 2026-07-27
fix(strict): preserve Shape type inference through wrapper functions

### LINKED ISSUE #6039: Typing breaks when wrapping .strict()
I need to apply `.strict()` based on some condition so I use a wrapper like this:

```typescript
const strictIfTesting = <T extends z.ZodObject>(schema: T) => {
    const isProductionEnvironment = "window" in globalThis
    return isProductionEnvironment ? schema : schema.strict() as T
}
```

The problem is in `as T` cast which is just an ugly workaround to keep typing works.

This works:

```typescript
const mySchema = () => strictIfTesting(z.object({
    someKey: strictIfTesting(z.object({
        anotherKey: z.string().nonempty().nonoptional()
    })).nonoptional()
}))

const strictIfTesting = <T extends z.ZodObject>(schema: T) => {
    const isProductionEnvironment = "window" in globalThis
    return isProductionEnvironment ? schema : schema.strict() as T
}

const myObject = mySchema().parse({})
console.log(myObject.someKey.anotherKey) // Everything works fine
```

But this is not:

```typescript
const mySchema = () => strictIfTesting(z.object({
    someKey: strictIfTesting(z.object({
        anotherKey: z.string().nonempty().nonoptional()
    })).nonoptional()
}))

const strictIfTesting = <T extends z.ZodObject>(schema: T) => {
    const isProductionEnvironment = "window" in globalThis
    return isProductionEnvironment ? schema : schema.strict()
}

const myObject = mySchema().parse({})
console.log(myObject.someKey.anotherKey) // Not compiles: myObject.someKey' is of type 'unknown'
```

Everything works as expected if abandon the `strictIfTesting()` in favor of the direct usage of the `.strict()`:

```typescript
const mySchema = () => z.object({
    someKey: z.object({
        anotherKey: z.string().nonempty().nonoptional()
    }).strict().nonoptional()
}).strict()

const myObject = mySchema().parse({})
console.log(myObject.someKey.anotherKey) // Everything works fine
```

So it's clear that I do something wrong defining types of `strictIfTesting()`. How to implement conditional strict validation without ugly workarounds?

deepshekhardas:
When `.strict()` is called through a generic wrapper like:

```ts
const wrap = <T extends z.ZodObject>(schema: T) => schema.strict();
```

the concrete Shape was lost (defaulted to `$ZodLooseShape`), making nested properties `unknown`.

Fix: change `.strict()`, `.loose()`, `.passthrough()`, `.strip()` signatures to use polymorphic `this["shape"]` instead of the generic `Shape` parameter. TypeScript's method shorthand preserves `this` correctly at each call site, so Shape is properly inferred through wrappers.

Fixes #6039

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1448,18 +1448,18 @@ export interface ZodObject<
 
   keyof(): ZodEnum<util.ToEnum<keyof Shape & string>>;
   /** Define a schema to validate all unrecognized keys. This overrides the existing strict/loose behavior. */
-  catchall<T extends core.SomeType>(schema: T): ZodObject<Shape, core.$catchall<T>>;
+  catchall<T extends core.SomeType>(schema: T): ZodObject<this["shape"], core.$catchall<T>>;
 
   /** @deprecated Use `z.looseObject()` or `.loose()` instead. */
-  passthrough(): ZodObject<Shape, core.$loose>;
+  passthrough(): ZodObject<this["shape"], core.$loose>;
   /** Consider `z.looseObject(A.shape)` instead */
-  loose(): ZodObject<Shape, core.$loose>;
+  loose(): ZodObject<this["shape"], core.$loose>;
 
   /** Consider `z.strictObject(A.shape)` instead */
-  strict(): ZodObject<Shape, core.$strict>;
+  strict(): ZodObject<this["shape"], core.$strict>;
 
   /** This is the default behavior. This method call is likely unnecessary. */
-  strip(): ZodObject<Shape, core.$strip>;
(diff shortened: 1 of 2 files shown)


colinhacks:
Going to close this one — declining. The generic-wrapper case has a one-line workaround (constrain on the shape, `<S extends z.core.$ZodShape>(schema: z.ZodObject<S>)`, rather than on `z.ZodObject`), and resolving the shape through `this` turns out to break every mask method chained after one — `.strict().pick({ a: true })` and friends stop compiling, valid keys included. There's no formulation that fixes the first without causing the second, and that's not a trade I want to make on `ZodObject`.

Good investigation, and thanks for looking into it 👍

PR #6241, opened 2026-07-28
fix: constrain array length for closed tuples in toJSONSchema

### LINKED ISSUE #6193: `toJSONSchema()`: closed tuples (no `.rest()`) don't constrain array length in any target, diverging from `.parse()`
### Description

For a `z.tuple([...])` with **no** `.rest()`, `.parse()` / `.safeParse()` strictly rejects arrays that are longer or shorter than the tuple. `.toJSONSchema()` does not reproduce this constraint faithfully in **any** target:

| target | positional item validation | length bound (`maxItems`/closed) |
|---|---|---|
| `draft-7` | ✅ correct | ❌ missing — extra trailing items pass unchecked |
| `draft-2020-12` | ✅ correct | ❌ missing (`items: false` never emitted) |
| `openapi-3.0` | ❌ **lost** — collapses to `items: { anyOf: [...] }`, so element *order* is no longer enforced | ✅ `maxItems`/`minItems` present |

No target matches `.parse()`'s accept set: `draft-7`/`draft-2020-12` leave the tail unbounded; `openapi-3.0` bounds the length but loses per-index typing (verified with reordered/duplicate elements).

This matters for consumers like ajv (e.g. via Fastify) that validate only against the emitted schema, not zod's runtime.

### Reproduction

```ts
import { z } from "zod";

const schema = z.tuple([
  z.string(),
  z.string(),
  z.object({ lang: z.string() }),
]);

schema.safeParse(["a", "b", { lang: "en" }, "extra"]).success;
// -> false (runtime: "too_big", maximum: 3)

console.log(JSON.stringify(schema.toJSONSchema({ target: "draft-7" })));
// {"type":"array","items":[{"type":"string"},{"type":"string"},
//  {"type":"object","properties":{"lang":{"type":"string"}},"required":["lang"],"additionalProperties":false}]}
// -> no "maxItems", no "additionalItems": false

console.log(JSON.stringify(schema.toJSONSchema({ target: "openapi-3.0" })));
// {"type":"array","items":{"anyOf":[{"type":"string"},{"type":"string"},
//  {"type":"object","properties":{"lang":{"type":"string"}},"required":["lang"],"additionalProperties":false}]},
//  "minItems":3,"maxItems":3}
// -> length is bounded, but positional typing is gone:
//    ["a","a","a"] now validates, even though schema.safeParse(["a","a","a"]) rejects it
//    (index 2 must be the `{lang}` object, not a string)
```

Feeding the `draft-7` schema to ajv, `["a","b",{lang:"en"},null]` (unvalidated 4th element) passes. Feeding the `openapi-3.0` schema to ajv, `["a","a","a"]` (no `{lang}` object anywhere) also passes. Both diverge from `schema.safeParse()`.

### Root cause (from reading `v4/core/*`)

- `schemas.ts` (`$ZodTuple.init`): when `def.rest == null`, the parser enforces `input.length > items.length` directly … independent of `schema._zod.bag`.
- `json-schema-processors.ts` (`tupleProcessor`):
  - `draft-7`/`draft-2020-12`: only closes the tail (`additionalItems: false` / `items: false`) when `def.rest` is set; otherwise nothing marks the schema as closed.
  - `openapi-3.0`: correctly infers `maxItems`/`minItems` from `prefixItems.length` when `!rest`, but represents `items` as `{ anyOf: prefixItems }`, which is positionally unsound (OpenAPI 3.0's `items` only accepts a single schema, so this is the best available approximation, but it silently discards per-index typing).

The two families of targets fail in complementary but both incorrect ways; there's no existing target that gets both right.

### Expected behavior

For a `.rest()`-less tuple:
- `draft-7`: `additionalItems: false`
- `draft-2020-12`: `items: false`
- `openapi-3.0`: ideally reject/warn that a closed tuple can't be represented losslessly in OpenAPI 3.0 `items`, rather than silently producing a schema with different accept semantics (or emit `minItems`/`maxItems` *and* document that positional constraints are unenforceable in that target).

### Environment

- zod: 4.4.3
- Node.js: v24.18.0

deepshekhardas:
Fixes #6193

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -1092,6 +1092,9 @@ describe("toJSONSchema", () => {
     expect(z.toJSONSchema(schema)).toMatchInlineSnapshot(`
       {
         "$schema": "https://json-schema.org/draft/2020-12/schema",
+        "items": false,
+        "maxItems": 2,
+        "minItems": 2,
         "prefixItems": [
           {
             "type": "string",
@@ -1209,6 +1212,7 @@ describe("toJSONSchema", () => {
     expect(z.toJSONSchema(schema, { target: "draft-7", io: "input" })).toMatchInlineSnapshot(`
       {
         "$schema": "http://json-schema.org/draft-07/schema#",
+        "additionalItems": false,
         "items": [
           {
             "type": "string",
@@ -1217,6 +1221,8 @@ describe("toJSONSchema", () => {
             "type": "number",
           },
         ],
+        "maxItems": 2,
+        "minItems": 2,
         "type": "array",
       }
     `);
--- a/packages/zod/src/v4/core/json-schema-processors.ts
+++ b/packages/zod/src/v4/core/json-schema-processors.ts
@@ -396,6 +396,8 @@ export const tupleProcessor: Processor<schemas.$ZodTuple> = (schema, ctx, _json,
     json.prefixItems = prefixItems;
     if (rest) {
       json.items = rest;
+    } else {
+      json.items = false;
     }
   } else if (ctx.target === "openapi-3.0") {
     json.items = {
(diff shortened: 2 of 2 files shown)


colinhacks:
Closing in favor of #6194. This gets the upper bound right, but `minItems` is hardcoded to `prefixItems.length`, so a tuple with an optional tail like `z.tuple([z.string(), z.number().optional()])` emits `minItems: 2` while `.parse(["a"])` succeeds — that trades over-acceptance for a schema stricter than zod itself. #6194 derives the bound from the same optin/optout start the parser uses, and covers the `io` direction too.

PR #6243, opened 2026-07-28
fix(formatError): prevent crash on inherited Object.prototype names in path (__proto__, constructor, toString)

### LINKED ISSUE #6211: formatError and treeifyError throw TypeError when an issue path contains inherited Object.prototype names (__proto__, constructor, toString, ...)
### Summary

`z.formatError()` and `z.treeifyError()` both throw `TypeError: Cannot read properties of undefined (reading 'push')` when any issue path element is a name inherited from `Object.prototype`: `__proto__`, `constructor`, `toString`, `hasOwnProperty`, `valueOf`, etc.

These paths are reachable through `ctx.addIssue({ path: [...] })` in `.check()` / `.superRefine()`, and more generally whenever user-controlled keys end up in an issue path (for example record or dynamic-key validation). For an API that formats validation errors from untrusted input, this is a crash vector.

### Reproduction (zod 4.4.3, Node 22)

```ts
import * as z from "zod";

const schema = z.object({ data: z.string() }).superRefine((val, ctx) => {
  ctx.addIssue({ code: "custom", message: "invalid value", path: ["toString"] });
});

const result = schema.safeParse({ data: "hello" });

z.treeifyError(result.error); // TypeError: Cannot read properties of undefined (reading 'push')
z.formatError(result.error);  // TypeError: Cannot read properties of undefined (reading 'push')
```

I ran this for all five names (`__proto__`, `constructor`, `toString`, `hasOwnProperty`, `valueOf`) against both functions on the published 4.4.3 build: all 10 combinations throw.

### Root cause

Both functions build their output trees on plain object literals and test node existence with truthy / nullish checks, which see inherited members.

`formatError` (`packages/zod/src/v4/core/errors.ts`):

```ts
curr[el] = curr[el] || { _errors: [] };
```

When `el` is `"toString"`, `curr[el]` is the inherited function, so the node is never created and `curr[el]._errors.push(...)` throws. When `el` is `"__proto__"`, plain assignment goes through the inherited setter and mutates the object's prototype instead of creating an own property.

`treeifyError` has the same class of bug:

```ts
curr.properties[el] ??= { errors: [] };
```

`properties["constructor"]` is not nullish, so `??=` never assigns and traversal walks into `Object.prototype` members.

### Fix

Guard node creation with an own-property check (`Object.prototype.hasOwnProperty.call`) and create nodes with `Object.defineProperty` so that `__proto__` becomes a real own data property instead of hitting the prototype setter. This fixes the crash for all inherited names and prevents prototype mutation for `__proto__`.

I have a patch covering both `formatError` and `treeifyError` with regression tests for all five names on both functions (36 tests passing locally). Happy to open a PR.

### Related open PRs

#6083, #6139, and #6181 each address `formatError`, but as far as I can tell none of them cover `treeifyError`, which fails identically (verified on 4.4.3 as above). Filing this issue per CONTRIBUTING so there is a single reference for the full scope, with credit to those PRs for prior art on the `formatError` side.

deepshekhardas:
Fixes #6211 #6070

### DIFF
--- a/packages/zod/src/v4/classic/tests/error-utils.test.ts
+++ b/packages/zod/src/v4/classic/tests/error-utils.test.ts
@@ -805,3 +805,25 @@ test("z.treeifyError nested union with real schema", () => {
     }
   }
 });
+
+// Regression: an issue path element that collides with an inherited property
+// name (e.g. "__proto__", "constructor") must not crash the error formatters
+// or mutate any prototype. Such paths are reachable through the public API via
+// a custom issue path (superRefine/check), e.g. when validating dynamic field
+// names taken from user input.
+test("formatError handles __proto__ / inherited-name path elements", () => {
+  for (const key of ["__proto__", "constructor", "toString", "hasOwnProperty"]) {
+    const schema = z.string().check((ctx) => {
+      ctx.issues.push({ code: "custom", message: "bad", path: [key], input: ctx.value });
+    });
+    const err = schema.safeParse("x").error!;
+    const formatted = z.formatError(err) as any;
+    expect(Object.prototype.hasOwnProperty.call(formatted, key)).toBe(true);
+    expect(formatted[key]._errors).toEqual(["bad"]);
+    const tree = z.treeifyError(err) as any;
+    expect(Object.prototype.hasOwnProperty.call(tree.properties, key)).toBe(true);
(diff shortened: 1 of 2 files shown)


colinhacks:
Fixed in #6213. Thanks 👍

PR #6244, opened 2026-07-28
docs: update UUID helper list to show v4 as the default version

deepshekhardas:
Clarifies that Zod uses UUID v4 by default.


colinhacks:
Already fixed — #6214 made this exact change and just landed, so this is a duplicate of a patch that's now on `main`.

Worth flagging on the reasoning though: `z.uuid()` doesn't default to v4. With no version argument it accepts v1 through v8 plus the nil and max UUIDs — you need `z.uuid({ version: "v4" })` to pin it. Closing this out, thanks! 👍

PR #6246, opened 2026-07-28
fix(multipleOf): use ULP-based tolerance to reject near-multiple values

deepshekhardas:
Improves multipleOf validation precision for decimal values.


colinhacks:
Closing this out — it's the same diff as #6223, which came first. Taking the fix there.

PR #6247, opened 2026-07-28
fix: allow optional() and nullable() before in() for coerce booleans

deepshekhardas:
Fixes optional-in and nullable-in ordering for coerced types.


colinhacks:
Closing this — the 4.4.0 behavior is intentional. A coerced boolean infers as a required object key, and the old runtime accepting a missing key was an unintentional soundness hole; 4.4.0 closed it as a bug fix in a minor.

`z.coerce.boolean().default(false)` is the way to make the key omittable. Thanks for looking into it 👍

PR #6248, opened 2026-07-28
fix: validate object and record intersection key types correctly

deepshekhardas:
Fixes key validation for object & record intersection schemas.


colinhacks:
Closing in favor of #6412, which fixes this where the issue code is chosen rather than filtering in the intersection handler, and covers record-to-record intersections as well.

PR #6249, opened 2026-07-28
feat: add timezone offset option to z.iso.time() for RFC3339 full-time

### LINKED ISSUE #6204: Add support for RFC3339 full-time in z.iso.time()

deepshekhardas:
Fixes #6204. Adds support for RFC3339 full-time with timezone offset.

### DIFF
--- a/packages/zod/src/v4/classic/tests/index.test.ts
+++ b/packages/zod/src/v4/classic/tests/index.test.ts
@@ -150,6 +150,14 @@ test("z.iso.time", () => {
   expect(z.safeParse(c, d1).success).toEqual(true);
   expect(z.safeParse(c, d2).success).toEqual(true);
   expect(z.safeParse(c, d3).success).toEqual(false);
+
+  // offset: true allows timezone suffixes (Z, +HH:MM, -HH:MM)
+  const withOffset = z.iso.time({ offset: true });
+  expect(z.safeParse(withOffset, "10:15:30Z").success).toEqual(true);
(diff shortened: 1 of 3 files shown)


colinhacks:
Thanks for the PR. Tracking this change in #6235 instead.

PR #6250, opened 2026-07-28
fix(exactOptional): skip coercion check for absent optional keys

### LINKED ISSUE #6236: .exactOptional() paired with coercion produces unexpected values
incorrect:

console.log(obj2); => {}`

deepshekhardas:
Fixes #6236. Prevents coercion errors on absent exactOptional keys.

### DIFF
--- a/packages/zod/src/v4/classic/tests/optional.test.ts
+++ b/packages/zod/src/v4/classic/tests/optional.test.ts
@@ -268,6 +268,21 @@ test("exactOptional vs optional comparison", () => {
   expect(exactOptionalSchema.safeParse({ a: undefined }).success).toEqual(false);
 });
 
+test("exactOptional with coercion", () => {
+  // Absent key should not produce value
+  expect(z.object({ a: z.coerce.number().exactOptional() }).parse({})).toEqual({});
+  expect(z.object({ a: z.coerce.string().exactOptional() }).parse({})).toEqual({});
+  expect(z.object({ a: z.coerce.boolean().exactOptional() }).parse({})).toEqual({});
+  expect(z.object({ a: z.coerce.bigint().exactOptional() }).parse({})).toEqual({});
+  expect(z.object({ a: z.coerce.date().exactOptional() }).parse({})).toEqual({});
+
+  // Explicit undefined should still reject with coercion
+  expect(z.object({ a: z.coerce.number().exactOptional() }).safeParse({ a: undefined }).success).toEqual(false);
+
+  // Present value with coercion should still work
+  expect(z.object({ a: z.coerce.number().exactOptional() }).parse({ a: "42" })).toEqual({ a: 42 });
+});
+
 // Defensive inference coverage: every schema that propagates `optout` participates
 // in object-key optionality inference. If anyone ever changes the set of values that
 // `optout` can take (or how OptionalOutSchema matches them), these assertions must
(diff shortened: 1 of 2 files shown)


deepshekhardas:
block status test

wahajahmed010 (commented):
**Review of colinhacks/zod#6250 — fix(exactOptional): skip coercion check for absent optional keys**

The fix correctly captures the semantics: with `exactOptional`, an absent key must mean absent, regardless of any coercion that `innerType` would otherwise run on `undefined`. Wrapping `run` to detect pre-coercion `undefined` and overwrite the resulting `value` is the right approach.

**Substantive concerns:**

1. **Double-handling with `run`'s pre-checks.** `$ZodExactOptional` already short-circuits on `payload.value === undefined` further up the call chain (the comments in `parse` previously said "no undefined handling" precisely because the upstream code already returned early). By re-adding the check in `parse`, this PR creates two places that branch on the same condition. If a future refactor changes where the early return lives (e.g. moves it into `def.innerType._zod.run`), the `parse` override becomes a no-op and silently breaks. Suggest: either (a) add a comment referencing the upstream short-circuit so future-you knows this is a defensive layer, or (b) centralize the "exactOptional + coercion" handling in a single utility used by both paths.

2. **Issue preservation is correct but undocumented.** When `wasUndefined` is true, the code overwrites `r.value = undefined` but preserves `r.issues`. That's the right behavior — coercion-issued problems (e.g. bad input format) should still surface. A 1-line comment in the code explaining this would prevent someone "simplifying" it by also clearing issues.

3. **Async re-wrap is fine, but consider `parseAsync` directly.** You're branching on `result instanceof Promise` to handle async. The codebase already has `parseAsync` plumbing; routing through it (or at least using a shared helper) would avoid this instanceof check duplicating existing logic in `ZodOptional`/`ZodPipeline`.

4. **Test coverage gaps.**
   - The new test only covers top-level objects. The common real-world failure is nested objects — e.g. `z.object({ nested: z.object({ a: z.coerce.number().exactOptional() }) }).parse({ nested: {} })`. Add one.
   - No test for `z.coerce.number().exactOptional()` in a discriminated union — that's where the `wasUndefined` capture matters most.
   - No test for `.optional()` (non-exact) — verifying it still **does** run coercion on `undefined` (i.e. that you didn't accidentally change the behavior of regular `.optional()` somewhere).

5. **Behavioral note (not a bug, but worth a doc comment).** With this change, `z.coerce.number().exactOptional()` and `z.coerce.number().optional()` now diverge for the `undefined` case:
   - `.optional()`: coerces `undefined` → `NaN`, parses `{a: NaN}` successfully.
   - `.exactOptional()`: skips coercion on absent keys, but rejects explicit `{a: undefined}`.
   This asymmetry is subtle and a one-liner in the JSDoc would save users from filing bug reports.

**Verdict:** LGTM, but tighten the test coverage around nested objects and discriminated unions before merging. The behavioral divergence between `.optional()` and `.exactOptional()` with coercion deserves a doc comment.

colinhacks:
Closing in favor of #6385, which fixes this in the object loop instead of overwriting the parsed value. Forcing the result back to `undefined` drops the fallback when a `default` sits inside the `exactOptional` — `z.object({ a: z.string().default("x").exactOptional() }).parse({})` returns `{}` here instead of `{ a: "x" }` — and for coercions that don't fail validation on their own it writes `undefined` into a field whose inferred type is `string`, which is the thing `exactOptional` exists to prevent.

#6237 has the byte-identical patch from two days earlier, so only one of these was ever going to merge. Thanks for digging into it 👍

PR #6251, opened 2026-07-28
fix: remove non-tree-shakable locales namespace export from v4/core and v4/mini

### LINKED ISSUE #6050: zod/v4/classic bundles all locale files due to non-tree-shakable export * as locales
v4/classic/external.js line 12 does export * as locales from "../locales/index.js", which forces webpack/rollup to include all 39 locale files (~175KB parsed) even when only en is used. In v3 this wasn't an issue because locales weren't shipped. Suggest either:
 
  - Make locales opt-in via separate entry points (import ar from "zod/v4/locales/ar")
  - Remove the barrel re-export from the default entry
  - Use a side-effect-free marker + individual exports so bundlers can tree-shake

deepshekhardas:
Fixes #6050. Removes the export * as locales barrel from core and mini entry points so bundlers can tree-shake unused locale files. Users can still import locales directly from \zod/locales\ or \zod/v4/locales\.


colinhacks:
Closing — #6384 fixed this by changing how the default export is declared rather than by removing the namespace. Dropping the export here doesn't change bundle size, since the locales stay reachable through `z.core` either way, and it would break `z.locales` on `zod/mini`, which is documented.

PR #6252, opened 2026-07-28
fix(looseRecord): pass through unrecognized keys in loose mode for closed key schemas

### LINKED ISSUE #6156: `looseRecord` with a closed key schema reports unrecognized keys fatally
This fails to parse:

```js
const A = z.looseRecord(z.enum(['foo', 'bar']), z.any())
A.parse({ foo: 123, bar: {}, baz: null })
```

This parses:

```js
const A = z.looseRecord(z.string().regex(/^foo|bar$/), z.any())
A.parse({ foo: 123, bar: {}, baz: null })
```

Reason: record's implementation (which looseRecord also uses) attempts to retrieve the closed schema value set with `def.keyType._zod.values`.

`z.enum` provides it. `z.regex` doesn't attempt to analyze the regex to provide it. Therefore, the false branch triggers with the rechex, in which `ctx.issues.push` is under `def.mode === "loose"`. The true branch has no such condition, meaning `unrecognized_key` is always reported.

https://github.com/colinhacks/zod/blob/c7805073fef5b6b8857307c3d4b3597a70613bc2/packages/zod/src/v4/core/schemas.ts#L2773-L2854

@pullfrog also signaled this in #5666.

deepshekhardas:
Fixes #6156. When a looseRecord uses a closed key schema (like z.enum), unrecognized keys are now passed through instead of reporting unrecognized_keys errors.

### DIFF
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -2937,8 +2937,12 @@ export const $ZodRecord: core.$constructor<$ZodRecord> = /*@__PURE__*/ core.$con
       let unrecognized!: string[];
       for (const key in input) {
         if (!recordKeys.has(key)) {
-          unrecognized = unrecognized ?? [];
-          unrecognized.push(key);
+          if (def.mode === "loose") {
+            payload.value[key] = input[key];
+          } else {
+            unrecognized = unrecognized ?? [];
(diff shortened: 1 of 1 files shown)


colinhacks:
Thanks for the PR. Tracking this change in #6157 instead.

PR #6253, opened 2026-07-28
fix(ci): move pnpm build before typescript matrix override

### LINKED ISSUE #6185: CI: `pnpm build` fails on every PR since TypeScript 7.0.2 became `latest` (zshy crash)

deepshekhardas:
Fixes #6185. Building with the repo's pinned TypeScript before installing the matrix-specified version so zshy isn't broken by TS 7.0.2's Go-native compiler.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,8 +28,8 @@ jobs:
           node-version: ${{ matrix.node }}
       - uses: pnpm/action-setup@v4
       - run: pnpm install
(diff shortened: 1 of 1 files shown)


deepshekhardas:
Superseded by #6312, which caps the matrix at TypeScript 6 and keeps the reorder. Closing this out.

PR #6257, opened 2026-07-28
fix(json-schema): encode  pointer tokens per RFC 6901

deepshekhardas:
Fixes #6027. Escapes ~ and / characters in JSON Pointer tokens when constructing  URIs from schema meta ids, per RFC 6901.


colinhacks:

PR #6258, opened 2026-07-28
fix(types): make record input Partial when value type has defaults

### LINKED ISSUE #6163: Wrong input type inferred from record with enum keys and default values

deepshekhardas:
Fixes #6163. When a record's value type is optional-in (e.g., ZodDefault or ZodOptional), the input type now allows missing keys via Partial.

### DIFF
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -2856,7 +2856,9 @@ export type $InferZodRecordInput<
   Value extends SomeType = $ZodType,
 > = Key extends $partial
   ? Partial<Record<core.input<Key> & PropertyKey, core.input<Value>>>
-  : Record<core.input<Key> & PropertyKey, core.input<Value>>;
+  : Value extends OptionalInSchema
+    ? Partial<Record<core.input<Key> & PropertyKey, core.input<Value>>>
+    : Record<core.input<Key> & PropertyKey, core.input<Value>>;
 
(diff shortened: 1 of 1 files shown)


colinhacks:
Closing this out — the fix landed in #6460, which takes the same approach and also covers the input-mode JSON Schema. Solid catch on the underlying bug. Thanks for the patch 👍

PR #6259, opened 2026-07-28
fix(fromJSONSchema): respect additionalProperties:false when patternProperties is present

### LINKED ISSUE #6198: fromJSONSchema ignores additionalProperties:false when patternProperties is present
## Summary

`fromJSONSchema` accepts object properties that match neither `properties` nor `patternProperties` when the source JSON Schema sets `additionalProperties: false`, as long as `patternProperties` is present. This makes the generated Zod schema more permissive than the source JSON Schema.

## Code path

- `packages/zod/src/v4/classic/from-json-schema.ts`: the `patternProperties` branch builds loose records and returns from that branch.
- `packages/zod/src/v4/classic/from-json-schema.ts`: the later `additionalProperties === false` handling is only reached when the earlier `patternProperties` branch is not taken.

## Steps to reproduce

```ts

  type: "object",
  patternProperties: {
    "^S_": { type: "string" },
  },
  additionalProperties: false,
});

console.log(schema.safeParse({ S_name: "ok" }).success);
console.log(schema.safeParse({ X_A: false }).success);
```

Actual output:

```txt
true
true
```

## Expected behavior

The second parse should fail. `X_A` does not match `^S_`, so it is an additional property. With `additionalProperties: false`, JSON Schema rejects properties that match neither `properties` nor `patternProperties`.

Expected output:

```txt
true
false
```

## Actual behavior

The generated Zod schema accepts `{ X_A: false }`.

## Existing coverage

I found related but not equivalent coverage:

- #5449 covered adding general `patternProperties` support.
- #5702 covers an `allOf` + `additionalProperties: false` interaction.

I did not find an issue or PR covering the `patternProperties` + `additionalProperties: false` interaction above.

## Suggested fix

Apply `additionalProperties` after accounting for keys matched by `properties` and `patternProperties`, instead of returning from the `patternProperties` branch before the `additionalProperties` logic runs.

## Suggested tests

- `fromJSONSchema({ type: "object", patternProperties: { "^S_": { type: "string" } }, additionalProperties: false })` accepts `{ S_name: "ok" }`.
- The same schema rejects `{ X_A: false }`.
- A matching pattern key with the wrong value type, such as `{ S_name: 1 }`, still fails.

---
Submitted with Codex.

deepshekhardas:
Fixes #6198. When patternProperties is present and additionalProperties is false, non-matching keys are now rejected instead of passed through.

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -406,24 +406,24 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
       // Handle patternProperties
       if (schema.patternProperties) {
         // patternProperties: keys matching pattern must satisfy corresponding schema
-        // Use loose records so non-matching keys pass through
         const patternProps = schema.patternProperties;
         const patternKeys = Object.keys(patternProps);
-        const looseRecords: ZodType[] = [];
+        const records: ZodType[] = [];
 
+        // When additionalProperties is false, use strict records so non-matching keys are rejected
+        const recordFn = schema.additionalProperties === false ? z.record : z.looseRecord;
         for (const pattern of patternKeys) {
           const patternValue = convertSchema(patternProps[pattern] as JSONSchema.JSONSchema, ctx);
           const keySchema = z.string().regex(new RegExp(pattern));
-          looseRecords.push(z.looseRecord(keySchema, patternValue));
+          records.push(recordFn(keySchema, patternValue));
         }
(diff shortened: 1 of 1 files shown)


colinhacks:
Closing in favor of #6199. Making each pattern record strict means every record rejects the keys belonging to the *other* patterns, so `^S_` + `^N_` stops accepting `{ S_a: "a", N_b: 1 }`, and a key declared in `properties` gets rejected too. There are no tests here, so the suite doesn't catch either regression. Thanks for looking into this 👍

PR #6260, opened 2026-07-28
fix(toJSONSchema): preserve default for ZodDefault/ZodCatch even when inner type transforms

### LINKED ISSUE #6049: v4: `toJSONSchema` drops `default` when the defaulted schema contains a transform/pipe

deepshekhardas:
Fixes #6049. The generic transforming-schema check at the end of process() was deleting default even for ZodDefault/ZodCatch schemas that intentionally set it. Now skips the deletion for those schema types.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2628,6 +2628,7 @@ test("defaults/prefaults", () => {
   expect(z.toJSONSchema(c, { io: "input" })).toMatchInlineSnapshot(`
     {
       "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "default": 1234,
       "type": "string",
     }
   `);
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing in favor of #6374.

Exempting `default` and `catch` from the check reintroduces the bug #4557 fixed — this PR rewrites the existing `defaults/prefaults` snapshot so `z.string().transform(v => v.length).pipe(z.number()).default(1234)` emits `default: 1234` on a `type: "string"` input schema. Moving the `examples` deletion inside the same branch widens it further. #6374 keeps the default only when it's actually valid input, so both cases in the issue work and the #4557 invariant holds.

PR #6261, opened 2026-07-28
fix(fromJSONSchema): make prefixItems tuple elements optional when minItems is omitted

### LINKED ISSUE #6200: fromJSONSchema requires prefixItems tuple elements when minItems is omitted
## Summary

`fromJSONSchema` converts JSON Schema tuple-style array schemas into Zod required tuples. As a result, arrays shorter than the number of positional schemas are rejected even when the JSON Schema does not set `minItems`.

In JSON Schema, `prefixItems` / draft-7 array-form `items` constrain the elements that are present. They do not make every positional schema required unless `minItems` says so.

## Code path

- `packages/zod/src/v4/classic/from-json-schema.ts`: `prefixItems` is mapped to `z.tuple(...)`.
- `packages/zod/src/v4/classic/from-json-schema.ts`: draft-7 array-form `items` is also mapped to `z.tuple(...)`.
- `packages/zod/src/v4/classic/tests/from-json-schema.test.ts`: the current `prefixItems` test expects a shorter array to throw, which appears stricter than JSON Schema semantics.

## Steps to reproduce

### Draft 2020-12 `prefixItems`

```ts

  $schema: "https://json-schema.org/draft/2020-12/schema",
  type: "array",
  prefixItems: [{ type: "string" }],
});

console.log(schema.safeParse([]).success);
console.log(schema.safeParse(["ok"]).success);
```

Actual output:

```txt
false
true
```

Expected output:

```txt
true
true
```

### Draft-7 array-form `items`

```ts

  $schema: "http://json-schema.org/draft-07/schema#",
  type: "array",
  items: [{ type: "string" }],
  additionalItems: false,
});

console.log(schema.safeParse([]).success);
console.log(schema.safeParse(["ok"]).success);
console.log(schema.safeParse(["ok", "extra"]).success);
```

Actual output:

```txt
false
true
false
```

Expected output:

```txt
true
true
false
```

## Expected behavior

Without `minItems`, shorter arrays should be valid as long as every present positional item satisfies its schema.

For example:

- `prefixItems: [{ type: "string" }]` should accept `[]` and `["ok"]`.
- draft-7 `items: [{ type: "string" }], additionalItems: false` should accept `[]` and `["ok"]`, but reject `["ok", "extra"]`.

## Actual behavior

The generated Zod schema rejects `[]` because `z.tuple([z.string()])` requires the first tuple element to be present.

## Existing coverage

Related open PRs exist, but they appear to cover the trailing/rest side rather than the shorter-array side:

- #6020 fixes open-by-default tuple rest behavior for omitted `items` / `additionalItems`.
- #6023 fixes boolean `true` rest behavior for `items: true` / `additionalItems: true`.

Both directions are related, but adding `.rest(...)` to `z.tuple([z.string()])` still leaves the first tuple element required, so arrays shorter than the prefix length remain rejected unless the fix also changes required-prefix handling.

#6193 / #6194 are about `toJSONSchema()` emitting length constraints for closed Zod tuples, which is the opposite conversion direction.

## Suggested fix

When JSON Schema does not set `minItems` to require all positional schemas, avoid representing `prefixItems` / array-form `items` as a required-length Zod tuple directly.

The converted schema should:

- validate each present positional element against its corresponding schema;
- allow shorter arrays down to the JSON Schema `minItems` value, defaulting to `0`;
- still enforce `items:false` / `additionalItems:false` for elements beyond the positional schema list;
- preserve typed rest behavior when `items` / `additionalItems` is a schema object.

## Suggested tests

- 2020-12 `prefixItems: [{ type: "string" }]` with no `minItems` accepts `[]`.
- 2020-12 `prefixItems: [{ type: "string" }, { type: "number" }]` with no `minItems` accepts `[]` and `["ok"]`.
- A present prefix item with the wrong type still fails, e.g. `[1]` for `{ type: "string" }`.
- draft-7 `items: [{ type: "string" }], additionalItems:false` accepts `[]` and `["ok"]`, but rejects `["ok", "extra"]`.
- Adding `minItems: 1` rejects `[]`, while adding `minItems: 2` rejects arrays shorter than two items.

---
Submitted with Codex.

deepshekhardas:
Fixes #6200. Tuple elements beyond minItems are now optional-out, allowing shorter arrays to validate. When minItems is not set (defaults to 0), all positional elements are optional.

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -1,4 +1,4 @@
-import type * as JSONSchema from "../core/json-schema.js";
+﻿import type * as JSONSchema from "../core/json-schema.js";
 import { type $ZodRegistry, globalRegistry } from "../core/registries.js";
 import { assignProp } from "../core/util.js";
 import * as _checks from "./checks.js";
@@ -470,7 +470,7 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
         // Tuple with prefixItems (draft-2020-12)
         // Without minItems, prefix items beyond the minimum are optional-out so shorter arrays are valid
         const minRequired = typeof schema.minItems === "number" ? schema.minItems : 0;
-        let tupleItems: ZodType[] = prefixItems.map((item, i) => {
+        const tupleItems: ZodType[] = prefixItems.map((item, i) => {
           const schema = convertSchema(item as JSONSchema.JSONSchema, ctx);
           return i >= minRequired ? schema.optional() : schema;
         });
@@ -494,7 +494,7 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
         // Tuple with items array (draft-7)
         // Without minItems, array-form items beyond the minimum are optional-out so shorter arrays are valid
         const minRequired = typeof schema.minItems === "number" ? schema.minItems : 0;
-        let tupleItems: ZodType[] = items.map((item, i) => {
+        const tupleItems: ZodType[] = items.map((item, i) => {
           const schema = convertSchema(item as JSONSchema.JSONSchema, ctx);
           return i >= minRequired ? schema.optional() : schema;
         });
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing in favor of #6201, which fixes both the draft-2020-12 and draft-7 paths the same way but also pins the `minItems`-is-set direction in tests — nothing here covers that leading positions stay required. This branch also adds a UTF-8 BOM to the top of `from-json-schema.ts`, which would need to come out regardless. Thanks for looking into it!

PR #6263, opened 2026-07-28
fix(url): reject bare hostnames parsed as protocol-only URLs

### LINKED ISSUE #6031: Bug: `z.string().url()` incorrectly validates IPv6 link-local addresses ("fe80:0000:0000:0000:0000:0000:0000:0001") as valid URLs

deepshekhardas:
Fixes #6031. When no protocol/hostname checks are configured, reject URLs with empty hostname to prevent strings like 'fe80::1' from being accepted by the URL constructor as protocol-only URLs.

### DIFF
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -503,6 +503,25 @@ export const $ZodURL: core.$constructor<$ZodURL> = /*@__PURE__*/ core.$construct
       // @ts-ignore
       const url = new URL(trimmed);
 
+      // Reject strings that the URL constructor accepts as protocol-only (e.g. "fe80::1")
+      // but aren't actually valid hierarchical URLs. Allow bare scheme-only URLs (e.g. "c:").
+      if (
+        !def.protocol &&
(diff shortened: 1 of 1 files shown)


colinhacks:
Closing in favor of #6132. The title has the root cause right — the URL constructor reads `fe80:` as a scheme — but rejecting every hostname-less URL is too broad: `z.url()` accepts non-http schemes on purpose, and this makes `mailto:`, `urn:`, `data:`, `tel:`, `about:` and `did:` all fail. That's what `z.httpUrl()` is for.

Thanks for looking into it!

colinhacks:
Following up — I ended up closing #6132 too and closing #6031 as working-as-intended. `z.url()` is documented as deferring to the platform's `new URL()`, which parses `fe80::1` as a scheme, so accepting it is the contract rather than a bug; `z.httpUrl()` is the strict API. Nothing wrong with your read of the mechanism.

PR #6264, opened 2026-07-28
fix(partial): skip wrapping already-optional fields with ZodOptional

### LINKED ISSUE #6171: `.partial()` produces inaccurate inferred types for optional and defaulted fields

deepshekhardas:
Fixes #6171. When .partial() is called, fields that already have optin: 'optional' (from .optional() or .default()) are no longer double-wrapped with another ZodOptional.

### DIFF
--- a/packages/zod/src/v4/classic/tests/partial.test.ts
+++ b/packages/zod/src/v4/classic/tests/partial.test.ts
@@ -119,7 +119,8 @@ test("partial with mask", async () => {
 
   expect(masked.shape.name).toBeInstanceOf(z.ZodOptional);
   expect(masked.shape.age).toBeInstanceOf(z.ZodOptional);
-  expect(masked.shape.field).toBeInstanceOf(z.ZodOptional);
+  // field already has optin via .default(), so .partial() doesn't double-wrap
+  expect(masked.shape.field).toBeInstanceOf(z.ZodDefault);
   expect(masked.shape.country).toBeInstanceOf(z.ZodString);
 
   masked.parse({ country: "US" });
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing in favor of #6172. This patches only the masked branch of `util.partial`, so plain `.partial()` still double-wraps — and with the type signatures unchanged, the masked branch now returns `ZodDefault` where the types still say `ZodOptional<ZodDefault<...>>`. That mismatch is worse than the original bug, since `.unwrap()` typechecks and then unwraps the wrong node.

#6172 covers both branches and moves the types alongside the runtime. Thanks for taking a look at this one 👍

colinhacks:
Following up: I closed #6172 as well. The current `.partial()` behavior is intended — reasoning is over there.

PR #6266, opened 2026-07-28
fix: add RFC 1035 length limits to domain regex

### LINKED ISSUE #5965: regexes.domain is missing RFC 1035 length limits

deepshekhardas:
Fixes #5965.

## Problem

The \domain\ regex (used by \z.httpUrl()\ for host validation) has per-label length limits but no overall domain length limit. RFC 1035 requires domain names be at most 253 characters.

## Fix

- Add \(?=.{1,253}\\.?\$)\ lookahead to enforce the RFC 1035 overall length limit
- Cap TLD labels at 63 characters (\[a-zA-Z]{2,63}\ → \[a-zA-Z]{2,63}\)

This brings the \domain\ regex in line with the \hostname\ regex which already enforces these limits.


colinhacks:
Thanks for the PR. Tracking this change in #6035 instead.

PR #6268, opened 2026-07-28
fix: use isThenable instead of instanceof Promise for Zone.js compat

### LINKED ISSUE #6019: `superRefine(async ...)` silently drops `ctx.addIssue` when `globalThis.Promise` is patched (Zone.js / Angular NgZone / `@opentelemetry/context-zone`)

deepshekhardas:
Fixes #6019.

## Problem

When \globalThis.Promise\ is patched by Zone.js, NgZone, \@opentelemetry/context-zone\, etc., \sync function\ still returns native Promise (from the internal %Promise% intrinsic), while the bare identifier \Promise\ resolves to the patched subclass. The check \
esult instanceof Promise\ returns \alse\ even for async results, causing zod to fall through to the sync path. This silently drops \ctx.addIssue()\ calls inside \sync superRefine\ callbacks.

## Fix

Replace \instanceof Promise\ with a duck-typing \isThenable()\ check (\alue != null && typeof value.then === 'function'\). This is the idiomatic Promise/A+ approach and matches how \wait\ itself recognizes thenables internally.

Affected environments: Zone.js, Angular (NgZone), OpenTelemetry context-zone, and any other code that patches \globalThis.Promise\.

### DIFF
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -234,10 +234,10 @@ export const $ZodType: core.$constructor<$ZodType> = /*@__PURE__*/ core.$constru
         const currLen = payload.issues.length;
         const _ = ch._zod.check(payload as any) as any as ParsePayload;
 
-        if (_ instanceof Promise && ctx?.async === false) {
+        if (util.isThenable(_) && ctx?.async === false) {
           throw new core.$ZodAsyncError();
         }
-        if (asyncResult || _ instanceof Promise) {
+        if (asyncResult || util.isThenable(_)) {
           asyncResult = (asyncResult ?? Promise.resolve()).then(async () => {
             await _;
             const nextLen = payload.issues.length;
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing in favor of #6021. This converts 5 of the 43 `instanceof Promise` sites, so it fixes the reported `superRefine` case but leaves async transforms, pipes, arrays, unions, records and codecs still dropping their results under a patched `Promise` — and there's no test. Thanks for taking a look 👍

PR #6269, opened 2026-07-28
fix(date): force origin 'date' in min/max error issues

### LINKED ISSUE #5980: Passing a number to ZodDate min() and max() functions causes the issue origin to become "number" (zod@4.4.3)

deepshekhardas:
Fixes #5980.

## Problem

When `z.date().min(timestamp)` is called with a numeric timestamp, the `origin` field in the resulting `too_small` issue is `"number"` instead of `"date"`. This is because the generic `gte`/`lte` check derives `origin` from `typeof def.value` (`"number"` for numbers, `"object"` for Date objects).

## Fix

Wrap the `min` and `max` check functions on `ZodDate` to override the `origin` to `"date"` on any issues they produce, regardless of the value type passed.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1365,8 +1365,28 @@ export const ZodDate: core.$constructor<ZodDate> = /*@__PURE__*/ core.$construct
   ZodType.init(inst, def);
   inst._zod.processJSONSchema = (ctx, json, params) => processors.dateProcessor(inst, ctx, json, params);
 
-  inst.min = (value, params) => inst.check(checks.gte(value, params));
-  inst.max = (value, params) => inst.check(checks.lte(value, params));
+  inst.min = (value, params) => {
+    const gteCheck = checks.gte(value, params);
+    const origCheck = gteCheck._zod.check;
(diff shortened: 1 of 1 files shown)


colinhacks:
Closing in favor of #6129. Wrapping the check and rewriting `origin` afterwards walks the whole `payload.issues` array, so it rewrites issues pushed by earlier checks too — and it ships without tests. #6129 fixes the origin at the source in `$ZodCheckGreaterThan` / `$ZodCheckLessThan`, which covers `zod/v4-mini` as well.

Thanks for looking into this!

PR #6270, opened 2026-07-28
fix: prevent metadata from overriding schema-derived type in toJSONSchema

### LINKED ISSUE #5990: `z.toJSONSchema()` allows overriding `type` via metadata

deepshekhardas:
Fixes #5990.

## Problem

Calling `z.string().meta({ type: 'number' }).toJSONSchema()` produces `{ type: "number" }` because all metadata fields are blindly copied into the JSON Schema output, allowing metadata to override schema-derived properties like `type`.

## Fix

Exclude the `type` field from metadata before merging into the JSON Schema. Schema-derived properties always take precedence over metadata.

### DIFF
--- a/packages/zod/src/v4/core/to-json-schema.ts
+++ b/packages/zod/src/v4/core/to-json-schema.ts
@@ -194,9 +194,12 @@ export function process<T extends schemas.$ZodType>(
     }
   }
 
-  // metadata
+  // metadata (schema-derived fields take precedence)
   const meta = ctx.metadataRegistry.get(schema);
-  if (meta) Object.assign(result.schema, meta);
+  if (meta) {
+    const { type: _t, ...metaRest } = meta;
+    Object.assign(result.schema, metaRest);
+  }
 
   if (ctx.io === "input" && isTransforming(schema)) {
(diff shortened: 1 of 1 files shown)


colinhacks:
Closing this — the metadata merge is intended, and `type` is the wrong keyword to single out. `z.date().meta({ type: "string", format: "date-time" })` under `unrepresentable: "any"` is how you represent a date in JSON Schema, and this drops it to `{ format: "date-time" }`. Every other contradiction metadata can introduce (`minLength`, `enum`, `pattern`) is left alone.

If you want metadata excluded from the output, `z.toJSONSchema(schema, { metadata: z.registry() })`. Thanks for looking into this 👍

PR #6272, opened 2026-07-28
fix: show 'Infinity'/'NaN' in error message when parsing non-finite numbers

### LINKED ISSUE #5697: Weird error message after update | 4.3.6

deepshekhardas:
Fixes #5697.

## Problem

When parsing `Infinity` (or `-Infinity`) with `z.number()`, the error message reads `"Invalid input: expected number, received number"` — the `received` portion shows `"number"` (the JavaScript type) instead of `"Infinity"` (the actual received value described in the issue's `received` field). The same confusion applies for `NaN`, which produces `"expected number, received number"`.

## Root cause

The locale error map uses `util.parsedType(issue.input)` to produce the `received` display string. For `Infinity` and `NaN`, `parsedType` returns `"number"` (since `typeof Infinity === "number"`), ignoring the `issue.received` field which already contains the precise description (`"Infinity"` or `"NaN"`).

## Fix

1. Added `received?: string` to the `$ZodIssueInvalidType` type definition.
2. Changed the locale `invalid_type` handler to check `issue.received` first before falling back to `util.parsedType(issue.input)`.
3. Applied the same fix across all 52 locale files.

Now the error correctly says `"Invalid input: expected number, received Infinity"` and `"Invalid input: expected number, received NaN"`.

### DIFF
--- a/packages/zod/src/v4/classic/tests/number.test.ts
+++ b/packages/zod/src/v4/classic/tests/number.test.ts
@@ -22,7 +22,7 @@ test("Infinity validation", () => {
         "code": "invalid_type",
         "received": "Infinity",
         "path": [],
-        "message": "Invalid input: expected number, received number"
+        "message": "Invalid input: expected number, received Infinity"
       }
     ]],
       "success": false,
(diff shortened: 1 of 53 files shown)


colinhacks:
Closing in favor of #5906. Reading `issue.received` in every locale puts an untranslated `Infinity` into all 52 of them, and it still reports `-Infinity` as `Infinity`, since the sign is lost upstream in `$ZodNumber` where `received` is hardcoded. #5906 keeps the wording change in the English locale and fixes the sign at the source.

PR #6273, opened 2026-07-28
fix(httpUrl): allow underscores in hostname

### LINKED ISSUE #5960: `z.httpUrl()` doesn't allow underscores in hostname

deepshekhardas:
Fixes #5960.

## Problem

`z.httpUrl()` rejects hostnames containing underscores (e.g., `https://julia_artberg.artstation.com/` or `https://exa_mple.com/`), even though underscores are commonly used in real-world URLs.

## Root cause

The `domain` regex used by `httpUrl()` for hostname validation only allows letters, digits, and hyphens in each label. Underscores (`_`) are excluded.

## Fix

Added `_` to the allowed character set in the domain regex — both for the start/end of each label (`[a-zA-Z0-9_]`) and for the middle positions (`[a-zA-Z0-9_-]`).

### DIFF
--- a/packages/zod/package.json
+++ b/packages/zod/package.json
@@ -53,63 +53,113 @@
     "./package.json": "./package.json",
     ".": {
       "@zod/source": "./src/index.ts",
-      "types": "./index.d.cts",
-      "import": "./index.js",
-      "require": "./index.cjs"
+      "import": {
+        "types": "./index.d.ts",
+        "default": "./index.js"
+      },
+      "require": {
+        "types": "./index.d.cts",
+        "default": "./index.cjs"
+      }
     },
     "./mini": {
       "@zod/source": "./src/mini/index.ts",
-      "types": "./mini/index.d.cts",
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing in favor of #6130. This one bundles an unrelated rewrite of the `exports` map in `packages/zod/package.json`, and its regex allows a leading or trailing underscore in a label — so `_asdf.com` passes while `-asdf.com` is still rejected — with no tests pinning either way. Thanks for looking into this 👍

colinhacks:
Following up: #6130 is closed too, and so is #5960 — we're not allowing underscores in hostnames at all. RFC 952/1123 restrict host labels to letters, digits and hyphens, and no registry will issue a domain containing one. So this was declined on the merits, not passed over for another PR. Thanks again for looking into it 👍

PR #6274, opened 2026-07-28
fix: constrain preprocess fn return type to schema input type

### LINKED ISSUE #5733: `preprocess` doesn't constrain return type of `fn` to the type of the output schema

deepshekhardas:
Fixes #5733.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -2725,8 +2725,8 @@ export function json(params?: string | core.$ZodCustomParams): ZodJSONSchema {
 
 // preprocess
 
-export function preprocess<A, U extends core.SomeType, B = unknown>(
-  fn: (arg: B, ctx: core.$RefinementCtx) => A,
+export function preprocess<U extends core.$ZodType, B = unknown>(
+  fn: (arg: B, ctx: core.$RefinementCtx) => core.input<U>,
   schema: U
 ): ZodPreprocess<U> {
   return new ZodPreprocess({


deepshekhardas:
C:\Users\hp\AppData\Local\Temp\opencode\pr6274-close.md

PR #6277, opened 2026-07-29
fix: encode / and ~ in \ pointer per RFC 6901

marceli1404:
Per RFC 6901, JSON Pointer reference tokens must escape ~ as ~0 and / as ~1.

When z.toJSONSchema() builds the \ URI fragment from a schema's .meta({ id }), it now applies these escapes to the pointer path. The \ key itself stays unescaped (it is a plain object key, not a pointer token).

Fixes #6027

Changes:
- Added encodePointerRef() helper in to-json-schema.ts
- Applied encoding to both self-contained and external shared \ paths
- Added regression test with Shared/User~ id


colinhacks:

PR #6278, opened 2026-07-29
fix(locale): use exact wording when issue.exact is set for length/size errors (v2)

deepshekhardas:
## Summary

This PR updates the error messages for exact length/size validation to use the exact wording from the issue, rather than the default messages.

## Changes

- Updated packages/zod/src/v4/locales/en.ts to use precise error messages when issue.exact is set for length/size validations

## Related

This is an alternative implementation to #6242 (exact-length-error-messages) with slightly different wording.


colinhacks:
Closing — this is already on `main`. #6177 landed the `exact` handling in `en.ts` as b53e53cc, folding it into the existing adjective ternary rather than a separate early return, and the `array.test.ts` and `validations.test.ts` hunks here are byte-identical to what's there now.

I ran the assertions from this PR against current `main` and they all pass unmodified. Thanks for looking into this!

PR #6281, opened 2026-07-29
fix(formatError): prevent crash on inherited Object.prototype names in path

deepshekhardas:
## Summary

This PR fixes a crash in ormatError and 	reeifyError when the error path contains inherited Object.prototype property names like __proto__, constructor, 	oString, etc.

## Changes

- Updated packages/zod/src/v4/core/errors.ts to filter out inherited property names
- Added tests in packages/zod/src/v4/classic/tests/error.test.ts

## Related

Superseded by #6243 (fix/formatError-treeifyError-inherited-paths) which has a more complete fix including tests.

### DIFF
--- a/packages/zod/src/v4/classic/tests/error.test.ts
+++ b/packages/zod/src/v4/classic/tests/error.test.ts
@@ -709,3 +709,19 @@ test("error serialization", () => {
     `);
   }
 });
+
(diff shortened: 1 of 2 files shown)


colinhacks:
Fixed in #6213. Thanks 👍

PR #6282, opened 2026-07-29
fix(locale): add exact flag support for length/size error messages

deepshekhardas:
## Summary

This PR adds support for the exact flag in locale messages for length/size validation errors. When exact: true is set on a string/array schema, the error message will use 'exactly N characters/items' instead of 'at least/at most N characters/items'.

## Changes

- Updated packages/zod/src/v4/locales/en.ts to add exact message keys
- Updated packages/zod/src/v4/core/to-json-schema.ts to include exact flag in JSON Schema
- Added tests in packages/zod/src/v4/classic/tests/locales_en.test.ts and 	o-json-schema.test.ts

## Related

Superseded by #6216 (fix/locale-exact-length-size) which has more complete implementation.


wahajahmed010 (commented):
The JSON Schema pointer-escaping fix is correct and important, but this PR mixes two unrelated changes under a misleading title. Splitting them would make both easier to review and revert.

**Concerns:**

**1. PR title and description don't match the diff.**

The title says `fix(locale): add exact flag support for length/size error messages`, but the diff also includes the JSON Schema pointer-escaping fix (issue #6027) in `core/to-json-schema.ts` and a corresponding test in `to-json-schema.test.ts`. These are two distinct fixes to two distinct subsystems. Please either split into two PRs or update the title/description to reflect both (e.g. `fix(locale, json-schema): ...`).

**2. JSON Schema pointer escape order is correct, but `escapePointer` should live near the other pointer-encoding logic**

You implement `escapePointer(s) => s.replace(/~/g, "~0").replace(/\//g, "~1")` inline inside `extractDefs`. This is the standard RFC 6901 order (`~` first, otherwise the `~0` we insert would get re-escaped to `~01` on the second pass). Good — verified by the `foo/bar~baz` test producing `~1bar~0baz` (the `~0` for `~` did NOT become `~01`). Just noting it for future maintainers.

If `escapePointer` is going to be reused (e.g., when emitting `$anchor` or `$dynamicRef` paths, which also need pointer escaping), lift it to a module-level helper and add a brief comment about the encoding order being load-bearing.

**3. Exact-flag locale message — minor consistency drift**

The new branch in `too_big` writes `expected ${issue.origin ?? "value"}` (falls back to `"value"` if origin is missing), while the new branch in `too_small` writes `expected ${issue.origin}` (no fallback). The existing `too_small` code already used `issue.origin` without fallback, so you're matching prior behavior — but this means a `too_small` error without an origin renders as `Too small: expected undefined to be exactly 4` (literally the string "undefined"). Pre-existing bug, but worth fixing while you're in this code path.

Suggestion: add the same `?? "value"` fallback to the `too_small` exact branch, and consider filing a follow-up for the non-exact branches to standardize.

**4. Only English locale updated**

Other locales still produce the old `<=`/`<`/`>=`/`>` messages for `.length(4)`. Not a regression (they were already inconsistent across locales), but worth noting in the PR description so users aren't surprised.

Otherwise the test coverage is solid — both the `Too small` and `Too big` exact branches are exercised, the round-trip is verified, and the `__proto__` / pointer tests are independent.

colinhacks:
Closing this alongside #6278 — the two overlap almost entirely, and both halves here are already resolved on `main`.

The `exact` wording landed in #6177 (b53e53cc). The `toJSONSchema` JSON Pointer escaping — which the title doesn't mention — landed separately in #6144 (bd18314c), with the same `~` first, `/` second ordering at the same two call sites.

One thing for next time: that pointer fix came in on an unrelated branch, and bundling it under a locale-titled PR made both halves harder to place. Easier to land as two PRs. Thanks all the same 👍

PR #6283, opened 2026-07-29
fix(json-schema): encode \ pointer tokens per RFC 6901

deepshekhardas:
## Summary

This PR fixes the JSON Schema \ pointer encoding to properly escape / and ~ characters per RFC 6901.

## Changes

- Updated packages/zod/src/v4/core/to-json-schema.ts to escape pointer tokens
- Added tests in packages/zod/src/v4/classic/tests/to-json-schema.test.ts

## Related

Superseded by #6257 (fix/json-pointer-encoding) and #6277 (fix/json-pointer-encoding) which have more complete implementations.


colinhacks:

PR #6301, opened 2026-07-30
feat: add z.number().latitude() and z.number().longitude()

### LINKED ISSUE #2600: [feat] `z.number().latitude()` and `z.number().longitude()`

deepshekhardas:
Add latitude() and longitude() chain methods on ZodNumber, validating geocoordinate ranges:

- \latitude()\ restricts to [-90, 90]
- \longitude()\ restricts to [-180, 180]

Closes #2600.

### DIFF
--- a/packages/zod/src/v3/tests/number.test.ts
+++ b/packages/zod/src/v3/tests/number.test.ts
@@ -18,6 +18,8 @@ const multipleOfFive = z.number().multipleOf(5);
 const multipleOfNegativeFive = z.number().multipleOf(-5);
 const finite = z.number().finite();
 const safe = z.number().safe();
+const latitude = z.number().latitude();
+const longitude = z.number().longitude();
 const stepPointOne = z.number().step(0.1);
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing this out — declining the feature. `z.number().gte(-90).lte(90)` is the whole implementation, and where it stops being trivial — whether the antimeridian is inclusive, whether a longitude means anything at the poles — is where core shouldn't be picking a convention for you. Geospatial modeling stays out of scope for the core library.

Separately, worth knowing for next time: this targets `src/v3`, which is end-of-life and only takes bug fixes now, so a new feature would have to land in v4 regardless. Thanks for looking into this 👍

PR #6305, opened 2026-07-30
feat:  support format: "hostname" in fromJSONSchema

### LINKED ISSUE #6304: fromJSONSchema doesn't support format:hostname (but could)
Doing a check of supported JSON Schema features vs what OpenAI supports for JSON schema, the only observed difference is that OpenAI supports `format:hostname` on string fields, and zod's `fromJsonSchema` does not.  https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-00#rfc.section.7.3.3

However, we have `z.hostname()`, so supporting this is very feasible.

catdalfonso:
`fromJSONSchema` maps string formats onto Zod checks but had no branch for
`hostname`, so those schemas fell through to a plain string and accepted any
value. `z.hostname()` already exists, so this wires it up alongside the other
network formats.

Round-trips cleanly, since `toJSONSchema()` already emits `format: "hostname"`
for `z.hostname()`.

Closes #6304


wahajahmed010 (commented):
## Review: `feat: support format: "hostname" in fromJSONSchema`

Small, focused addition. Code matches the existing `ipv4`/`ipv6` pattern, the test asserts both happy and sad paths, and `z.hostname()` exists at `packages/zod/src/v4/classic/schemas.ts:1005` so the dependency is already in tree. Two small nits, not blockers:

### 1. Test coverage could be one line stronger

The single failing case `"not a hostname!"` is enough to prove the format is enforced, but `z.hostname()` accepts IDN-ish inputs and rejects bare labels. A second assertion like `expect(() => schema.parse("-bad-.com")).toThrow()` (or any label that exercises the leading/trailing-hyphen rule) would harden this against future regex churn in `core.regexes.hostname`.

### 2. Docs ecosystem note

Worth a one-line addition to the `fromJSONSchema` JSDoc block above `convertBaseSchema` (or wherever the supported `format` values are enumerated) so users know `hostname` is now part of the supported set. Not strictly required — the test will document it by example — but it costs nothing.

### 3. Consistency

The `format === "hostname"` branch is placed between `duration` and `ipv4` instead of grouped with the network formats (`ipv4`/`ipv6`/`email`/`url`). That's a cosmetic nit; reading order is fine. Optional: move it next to `ipv4`/`ipv6` so all network-related formats are co-located.

LGTM overall — ready to merge once a maintainer signs off.

PR #6306, opened 2026-07-31
feat: allow structured ZodIssueMessage values for deferred error message rendering

deepshekhardas:
Rebased fix for #6182 on latest main. Structured message type support was partially landed; this restores full support for \string | { key, values }\ issue messages via \	oMessageString\.

### DIFF
--- a/packages/zod/src/v4/classic/tests/error-utils.test.ts
+++ b/packages/zod/src/v4/classic/tests/error-utils.test.ts
@@ -79,7 +79,7 @@ test(".flatten()", () => {
 });
 
 test("custom .flatten()", () => {
-  type ErrorType = { message: string; code: number };
+  type ErrorType = { message: string | { key: string; values?: object }; code: number };
   const flattened = parsed.error!.flatten((iss) => ({
     message: iss.message,
     code: 1234,
@@ -162,7 +162,7 @@ test(".format()", () => {
 });
 
 test("custom .format()", () => {
-  type ErrorType = { message: string; code: number };
+  type ErrorType = { message: string | { key: string; values?: object }; code: number };
   const formatted = parsed.error!.format((iss) => ({
     message: iss.message,
     code: 1234,
@@ -258,7 +258,7 @@ test("all errors", () => {
     }
   `);
 
-  expect(z.core.flattenError(r2.error!, (iss) => iss.message.toUpperCase())).toMatchInlineSnapshot(`
+  expect(z.core.flattenError(r2.error!, (iss) => (iss.message as string).toUpperCase())).toMatchInlineSnapshot(`
     {
       "fieldErrors": {
         "a": [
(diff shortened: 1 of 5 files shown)


wahajahmed010:
## Code review

Good restoration of structured message support — the `toMessageString` indirection is the right escape hatch. Two substantive concerns and a couple of small ones.

### 1. `unwrapMessage` semantics changed for `{ message: "" }` callers

Before this PR:

```ts
unwrapMessage({ message: "" }) // -> ""
```

After:

```ts
unwrapMessage({ message: "" }) // -> JSON.stringify({ message: "" }) === '{"message":""}'
// more generally: unwrapMessage({ message: { key, values } }) -> '{"key":"...","values":{...}}'
```

Previously the helper unwrapped to the inner string; now it stringifies whatever the inner value is, including structured messages. This is the *intent* of the PR, but it's a public API behavior change that can silently break consumers who relied on `unwrapMessage` returning a string. Two options:

- Document the change prominently in the changelog/migration notes.
- If a non-empty structured message reaches `unwrapMessage`, that's a likely programmer error (callers should use `toMessageString` on the raw `$ZodIssueMessage` instead). Consider throwing on structured input here to surface the misuse early, since `unwrapMessage` historically returned a string by contract.

### 2. `flattenError` / `formatError` / `treeifyError` default mappers now stringify

```ts
mapper = (issue: $ZodIssue) => toMessageString(issue.message) as U
```

If a user previously called `flattenError(err)` and stored the result, each `formErrors` entry was `issue.message` (a string). Now it's `JSON.stringify({...})` for structured messages. That's a reasonable trade-off (users get *something* instead of an object that doesn't serialize to JSON cleanly) but it's a breaking change. Worth calling out in the PR description, since right now the body says "Restores full support" which understates it — it also changes the default mapper contract.

### 3. Test gaps

The new test only covers the *insertion* path. Missing:
- A test that calls `flattenError` / `formatError` on a structured-message issue to lock in the new stringification default.
- A test for `unwrapMessage` with `{ message: { key, values } }` to verify the new behavior is intentional.
- A round-trip test: `{ error: () => ({ message: { key, values } }) }` → `result.error.issues[0].message` should be the structured object, not a string. (Verify the error is *not* stringified when stored on the issue — only when serialized via the helper.)

### 4. Minor: `ZodIssue` type widening

`$ZodIssueBase.message: string` → `$ZodIssueMessage` is a public type widening. Any external code that did `error.issues[0].message.toUpperCase()` will now fail to type-check (the test file had to add `as string` casts in two places, which is a leading indicator). Worth a TS migration note for downstream users.

### 5. Inline snapshot shows raw JSON

The new snapshot:
```
"message": "{\\"key\\":\\"test\\",\\"values\\":{\\"foo\\":\\"bar\\"}}"
```

That's what gets surfaced to end users today, which is almost certainly not what they want — they want their i18n function to receive `{ key, values }` and produce a translated string. The PR is supposed to enable deferred rendering, but right now the structured message gets stringified at error-formatting time, which is the opposite of deferred. If a user has an `errorMap` registered, does it see the structured object or the stringified one? If the errorMap gets the object, this is fine; if it gets the string, the deferred-rendering premise is broken.

colinhacks:
Thanks for the PR. Tracking this change in #6182 instead.

PR #6307, opened 2026-07-31
fix(ci): align workflow Node.js version with package engines

MGPOCKY:
## Why
Package metadata declares `engines.node` / `volta.node` / `.nvmrc` as `22`, but CI workflows still pin `node-version` to a mismatched value.

That can make CI run on a runtime the project no longer supports (or skip the version the package actually targets).

## What changed
Update the affected workflow `node-version` pins so they satisfy `22` (using `22` where a concrete pin is needed).

- `.github/workflows/release.yml`
  - `node-version: 24` → `node-version: 22`

## Test plan
- [ ] Package `engines.node` / `volta.node` / `.nvmrc` is still `22`
- [ ] Updated workflows now use versions consistent with that constraint
- [ ] Relevant CI jobs on this branch look healthy

Found while auditing CI/package version consistency across popular repos.


colinhacks:
Going the other direction here — Node 22 has been in maintenance since October 2025. Bumped `.nvmrc` to 24 (what the `lts/*` in the test matrix already resolves to) and gave the publish job the same `"latest"` spelling as the jobs next to it, so it can't drift again.

Solid catch on the mismatch. Merging.

PR #6308, opened 2026-07-31
feat: implement tuple discriminator detection for union optimization

deepshekhardas:
Rebased version of #5643. Adds tuple discriminator detection to skip non-matching union options, with early rejection for invalid tuple discriminator values. Keeps the single-option fast path already present in main.

### DIFF
--- a/packages/zod/src/v4/classic/tests/union.test.ts
+++ b/packages/zod/src/v4/classic/tests/union.test.ts
@@ -271,3 +271,17 @@ test("z.discriminatedUnion with empty options constructs and rejects", () => {
     expect((obj.error.issues[0] as any).options).toEqual([]);
   }
 });
+
+test("tuple union optimizer - deep nested recursive schema", () => {
+  const schema: z.ZodType<any> = z.lazy(() =>
+    z.union([...[...Array(1000).keys()].map((i) => z.tuple([z.literal(`key-${i}`), schema])), z.string()])
+  );
+
+  let data: unknown = "leaf-value";
+  for (let i = 0; i < 1000; i++) {
+    data = [`key-${i % 1000}`, data];
+  }
+
+  const result = schema.parse(data);
+  expect(result).toBeDefined();
+});
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -2258,7 +2258,7 @@ export const $ZodUnion: core.$constructor<$ZodUnion> = /*@__PURE__*/ core.$const
 
     const results: util.MaybeAsync<ParsePayload>[] = [];
     const disc = tupleDisc.value;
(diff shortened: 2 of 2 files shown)


wahajahmed010:
## Code review

Overall direction is sound — a discriminator-based fast path for unions of tuples is a well-trodden optimization. A few concerns:

### 1. Bail-out inside the loop is correct, but the ordering is fragile

In `detectTupleDiscriminator` you `return null` from inside the `for (const tuple of tuples)` loop, but you've also already started populating `map` with prior tuples. The current code only inserts into `map` *after* reading `values`, so this is safe today, but the early-exit + side-effect pattern is easy to break in a refactor. Consider collecting candidates first, validating all of them, then committing to the `map` in a single pass — that way a future change can't accidentally emit a half-populated map.

### 2. `payload.value[0]` lookup is paid even when `disc` is null

```ts
const disc = tupleDisc.value;
const opt = Array.isArray(payload.value) ? disc?.map.get(payload.value[0]) : undefined;
```

When `disc` is `null` (the common case — most unions aren't tuple discriminators) you still compute `Array.isArray(payload.value)` and call `tupleDisc.value`. The `cached()` wrapper means the detection only runs once per schema, so it's not a hot-path concern, but you can short-circuit more cleanly:

```ts
if (disc && Array.isArray(payload.value)) {
  const opt = disc.map.get(payload.value[0]);
  // ...
}
```

Then the `if (Array.isArray(payload.value) && disc && !opt)` push and the `if (... && disc && opt !== option) continue` both narrow naturally. Currently the pattern is duplicated three times.

### 3. Error path on discriminator miss

`path: [0]` is technically correct (the first element is the discriminator), but downstream consumers using `ZodError.format()` / `flatten()` will see a fake "field" called `0` on the parent. Worth confirming this is the intended UX vs. attaching the error to the union itself (empty path) — most schema libraries use the latter for "no variant matched".

### 4. Test coverage

The new test only covers the happy path of a deeply nested tuple union. I'd want at least one test for:
- A union where the discriminator is at the wrong position (`z.tuple([z.string(), z.literal("x")])`) — should bail out
- A union with overlapping discriminator values across two tuples — should bail out
- A union where `payload.value` is an empty array — should reject with a meaningful error, not a "minimum items" error from the tuple check
- An async path / `parseAsync` to make sure `async` is still set correctly when the optimization fires

The fast path is correct for the case it activates in; the question is whether it correctly *deactivates* for everything else. The current test doesn't probe the deactivation paths.

### 5. Caching identity

`util.cached(() => detectTupleDiscriminator(def.options))` — this caches by reference identity of `def`, which is correct, but worth a comment in the code so the next person doesn't move the call site and accidentally pay detection cost per-parse.

colinhacks:
Thanks for the PR. Tracking this change in #5643 instead.

PR #6310, opened 2026-08-01
fix(core): treat any/unknown/coerced-boolean object properties as optional-in

deepshekhardas:
Ensures object properties typed as any/unknown/coerced-boolean are treated as optional when using optional-in semantics.


colinhacks:
Closing this — same change as #6224, and the current behavior is intentional. `z.any()` and `z.unknown()` infer as required keys, so requiring the key at parse time is what keeps the parser honest about the type it advertises. Use `z.any().optional()` for an omittable key. Thanks for looking into this 👍

PR #6312, opened 2026-08-01
ci: cap TS matrix at 6 and build before swapping TS version

### LINKED ISSUE #6185: CI: `pnpm build` fails on every PR since TypeScript 7.0.2 became `latest` (zshy crash)

deepshekhardas:
﻿Fixes #6185.

Every CI run of \	est.yml\ fails at \pnpm build\ since TypeScript 7.0.2 became npm's \latest\ (2026-07-08). \	ypescript@latest\ now resolves to the Go-native compiler, which no longer ships \	s.sys\ — and \zshy\ (used by \packages/zod\ and \@zod/resolution\) reads tsconfigs through \	s.sys.readFile\, crashing with:

    Build failed: TypeError: Cannot read properties of undefined (reading 'readFile')

Both matrix jobs go red for this one root cause (the 5.5 job is only cancelled by matrix fail-fast).

Two changes:
- Move \pnpm build\ above \pnpm add typescript@… -w\ so the repo-pinned TypeScript always runs the build; the matrix TS then applies only to the test/typecheck phase.
- Cap the matrix at TypeScript 6 (\5.5\ and \6\) instead of \latest\. TS 6 still ships the JS compiler API, so zshy builds and \@zod/resolution\'s own zshy build both pass. This keeps broader coverage than pinning at 5, and the matrix can resume tracking \latest\ once zshy supports TS 7.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         node: ["lts/*"]
-        typescript: ["5.5", "latest"]
+        typescript: ["5.5", "6"]
     name: Test with TypeScript ${{ matrix.typescript }} on Node ${{ matrix.node }} (${{ matrix.os }})
     steps:
       - uses: actions/checkout@v4
@@ -28,8 +28,8 @@ jobs:
           node-version: ${{ matrix.node }}
       - uses: pnpm/action-setup@v4
       - run: pnpm install
-      - run: pnpm add typescript@${{ matrix.typescript }} -w
       - run: pnpm build
+      - run: pnpm add typescript@${{ matrix.typescript }} -w
       - run: pnpm test
       - run: pnpm run --filter @zod/resolution test:all
       - run: pnpm run --filter @zod/integration test:all


deepshekhardas:
Quick status update: this is fully green (TS 5.5 + 6.0 matrix, lint, circular-dep check) and ready to merge.

It caps the CI matrix at `["5.5", "6"]` because TypeScript 7.0.2 became npm `latest` — the Go-native compiler removed `ts.sys`, which crashes `zshy` during the type-check step. That single failure is currently making every other open PR red (the matrix's fail-fast cancels the rest of the job).

Merging this will turn ~30 open PRs green at once. Once `zshy` supports TS 7, the matrix can track `latest` again.

deepshekhardas:
﻿Hi maintainers - just a gentle bump on this one. It caps the CI TypeScript matrix at 5.5/6 to avoid the TS 7.0.2 zshy crash (ts.sys removed), and moves the build step so the matrix swap doesn't break pnpm build. All 30 open PRs currently show red CI purely because of this TS7 issue - this PR is the single change that turns the whole backlog green. Fully green locally (TS 5.5/6/7 verified) and all checks pass here. Thanks for considering!

colinhacks:
Superseded by #6345, which is now on `main`.

Good catch on TypeScript 6 still shipping the JS compiler API — that leg survived into the merged version. I kept `latest` as a third leg rather than capping the matrix: zod's types pass under 7.0.2 today, so only `zshy` is blocked, and `@zod/resolution` is gated off that leg because `test-resolution.ts` invokes `zshy` directly.

Thanks for looking into this!

PR #6318, opened 2026-08-03
perf(v4): halve schema construction cost; keep instances in fast-properties mode

zirkelc:
> **Context: performance optimization campaign.** This is one of four PRs extracted from a systematic performance campaign on the v4 core, run as an automated research loop: 18 isolated experiments, 11 kept, 7 discarded. Every candidate change was
>
> - benchmarked with an A/B harness that loads **two git revisions of `packages/zod/src` into one process** and times 11 workloads in alternation, taking the minimum of 25 rep-scaled iterations. Module instances carry a stable load-order JIT bias of several percent, so the harness runs both load orders in separate child processes and combines them with a geometric mean;
> - measured on **11 deterministic workloads mirroring `packages/bench`** (simple object, the moltar `strictObject`, string/number leaf parses, union, discriminated union, failing `safeParse`, string formats, transform/pipe chain, schema construction). The repo's own benches compare zod against *other libraries* inside a single checkout and cannot load two zod revisions into one process, so using them for A/B would mean comparing two standalone runs — and run-to-run drift is larger than most effect sizes here. The mirrored cases use seeded PRNG data, so both revisions parse byte-identical inputs and the same workloads feed the characterisation guard below. The untouched `packages/bench` suite served as an external cross-check;
> - gated by a **calibrated noise floor** (~1.2% on the suite total, measured with identical code on both sides): changes under 2.5% total, or under 5% on a targeted case, were discarded, and every keep required a second confirming run;
> - verified behaviour-preserving by a **characterisation guard** (hashed parse outputs and error issues over all benchmark inputs) plus the full test suite (3811 tests + typecheck), both green after every commit. Discarded experiments — including several that looked plausible but regressed hot paths — are logged in the campaign notes.
>
> The numbers below are fresh verification runs of **this branch in isolation against `main`** (two independent runs each; an identical-source control run measured +0.56% total, i.e. pure noise). Negative = faster.

## What this PR does

Cuts schema construction cost roughly in half, and — unexpectedly but reproducibly — speeds up *parse* hot paths as well, by keeping schema instances in V8's fast-properties representation. Four commits:

**1. Lazy-bind codec methods; drop `Object.keys` in init.** Eight of the thirteen eager per-instance closures in the `ZodType` initializer were the codec family (`encode`, `decode`, `safeEncode`, ...) — far off the hot path. They join the lazy-bind prototype groups from #5897. (The five parse-family closures stay eager: converting them was tried and regressed hot paths 11-38%, empirically confirming the existing code comment.) Separately, `$constructor.init`'s prototype-modification loop ran `Object.keys(proto)` per instance per trait level on almost-always-empty prototypes; a `for...in` + `hasOwnProperty` loop visits the identical key set with zero allocation.

**2. Defer standard-schema and JSON Schema method creation.** The core initializer creates `"~standard"` lazily on purpose ("avoid creating objects for every schema") — but the classic `ZodType` initializer immediately forced it via `Object.assign(inst["~standard"], { jsonSchema: ... })`, allocating the standard object, its `validate` closure, and two JSON Schema method closures for every schema ever constructed. Core now exposes an `@internal` `_zod.onStandard` hook that classic uses to attach the `jsonSchema` methods when `"~standard"` is first accessed. `description` and `_def` — pure functions of `this` — move from per-instance descriptors to shared prototype getters (`_def` keeps a setter because core `$ZodFunction` assigns it).

**3. `"~standard"` as a shared prototype accessor.** The per-instance `defineLazy` accessor for `"~standard"` was the last *accessor* among instance own properties. Moving it to a one-time prototype getter (caching as an own non-enumerable data property on first access, same semantics as `defineLazy`) keeps instances in fast-properties mode — which is why this commit alone measured -58% on array parsing and double digits on other parse paths that never touch `"~standard"` at all. `defineProperty` calls per `strictObject` tree of 13 schemas drop from 77 to 25.

**4. Lazy-bind the 26 `ZodString` format methods.** `z.string()` allocated 26 eager closures (`.email()`, `.uuid()`, `.ipv4()`, `.datetime()`, ...) per instance — the only method group in the file still eager after #5897. Converted to the same lazy-bind group as everything else. (`ZodBigInt`/`ZodDate`/`ZodSet`/`ZodMap` have the same pattern and could get the same treatment in a follow-up.)

| case | run 1 | run 2 | speed-up vs `main` |
|---|---|---|---|
| schema-init (construction) | **-47.7%** | **-47.7%** | **1.91x** |
| array-parse | **-58.4%** | **-57.9%** | **2.39x** |
| strict-moltar parse | **-22.7%** | **-22.6%** | **1.29x** |
| string-parse | -16.0% | -17.0% | 1.20x |
| number-parse | -15.3% | -17.7% | 1.20x |
| chain-parse | -14.7% | -15.6% | 1.18x |
| **suite TOTAL** | **-13.5%** | **-13.4%** | **1.16x** |

Cross-check on the repo's own `packages/bench/object-moltar.ts`: the bench runs the npm build of zod **4.0.8** (its `zod4` alias) in the same process as the workspace code, so the workspace-vs-npm ratio is paired within a single run and robust to machine drift. That ratio moves from 0.96x on `main` to **1.39x on this branch** — a ~1.4x relative shift, in line with the paired moltar measurement above (1.29x) given single-run bench variance. Read the *shift*, not the absolute multiplier: the absolute number also bakes in the version gap (npm 4.0.8 vs current source) and the build-format gap (compiled npm JS vs TypeScript source loaded through tsx), neither of which has anything to do with this PR.

## Observable surface

- `"~standard"`, `description`, `_def`, and the codec/format methods are inherited until first touched; after first access they cache as own properties with the same enumerability as before. `Object.keys` / `JSON.stringify` / spread behaviour of schemas is unchanged (all moved properties were non-enumerable or are methods).
- `_zod.onStandard` is a new `@internal` field on the internals interface.
- Assigning `schema._def = x` still works via the accessor's setter (used by core `$ZodFunction`).

The fast-properties effect is the key finding here: during the campaign, the *inverse* experiments (adding accessors to instances or to `_zod`) regressed object parsing by +57% to +142%, which is what pinned the mechanism.

## Reproducing the numbers

The harness below is the one used for the verification runs (one difference: each side is unpacked into its own directory, so `main main` works as a noise-control run). From a checkout of this repo:

1. `pnpm install`
2. Save the two files below as `perf/ab.mts` and `perf/cases.mts` (the directory must live **inside** the repo so the unpacked trees resolve `node_modules`; `perf/.trees/` is created automatically and safe to delete).
3. Noise control (identical source on both sides). Discard the very first run — it materialises the trees and runs with cold caches. Repeat runs should land within roughly ±2% on TOTAL; that is this machine-state's noise band and the yardstick for step 4:

   ```sh
   pnpm tsx perf/ab.mts main main
   ```

4. Measure this PR — fetch its head ref, then compare it against `main`:

   ```sh
   git fetch origin pull/6318/head:pr-6318
   pnpm tsx perf/ab.mts main pr-6318
   ```

One run takes ~5s: 11 deterministic workloads, both load orders in separate child processes, minimum of 25 rep-scaled iterations per case, geometric mean across orders. Negative delta = second revision is faster. Run each comparison at least twice and judge repeated results against your control band from step 3, never a single run. The `schema-init` case is the noisiest (GC-bound); the parse cases are much tighter.

<details>
<summary><code>perf/ab.mts</code> — A/B revision benchmark</summary>

```ts
/**
 * A/B benchmark of two git revisions of packages/zod/src.
 *
 *   pnpm tsx perf/ab.mts [revA] [revB]
 *
 * revA defaults to HEAD, revB defaults to WORKTREE (the working tree).
 * Revisions are materialised with `git archive` into perf/.trees/.
 *
 * Both variants are imported into one process and timed in alternation,
 * taking the minimum per case. Module instances carry a stable
 * positional (load-order) bias of several percent, so the parent runs
 * the measurement twice as child processes, once per load order, and
 * combines the two ratios with a geometric mean to cancel the bias.
 * Negative delta = B faster than A.
 */
import { execSync, spawnSync } from "node:child_process";
import { fileURLToPath, pathToFileURL } from "node:url";
import { type PerfCase, buildCases } from "./cases.mts";

const ROOT = path.join(import.meta.dirname, "..");
const TREES = path.join(import.meta.dirname, ".trees");

const WARMUP = 4;
const ITERS = 25;
const TARGET_NS = 1_500_000;

function materialise(rev: string, slot: string): string {
  if (rev === "WORKTREE") return path.join(ROOT, "packages/zod/src/index.ts");
  const sha = execSync(`git rev-parse ${rev}`, { cwd: ROOT }).toString().trim();
  const dir = path.join(TREES, `${sha}-${slot}`);
  if (!fs.existsSync(dir)) {
    execSync(`git archive ${sha} packages/zod/src | tar -x -C ${dir}`, { cwd: ROOT });
  }
  return path.join(dir, "packages/zod/src/index.ts");
}

interface Row {
  name: string;
  a: number;
  b: number;
}

async function measure(entryFirst: string, entrySecond: string): Promise<Array<Row>> {
  const zFirst = await import(pathToFileURL(entryFirst).href);
  const zSecond = await import(pathToFileURL(entrySecond).href);
  const casesFirst = buildCases(zFirst);
  const casesSecond = buildCases(zSecond);

  const time = (fn: () => void): number => {
    const start = process.hrtime.bigint();
    fn();
    return Number(process.hrtime.bigint() - start);
  };

  const rows: Array<Row> = [];
  for (let i = 0; i < casesFirst.length; i++) {
    const ca: PerfCase = casesFirst[i];
    const cb: PerfCase = casesSecond[i];

    /* JIT warmup on the raw bodies, then size the timed run to ~TARGET_NS;
     * both variants use the identical rep count. */
    let probe = Number.POSITIVE_INFINITY;
    for (let w = 0; w < 10; w++) {
      probe = Math.min(probe, time(ca.run));
      cb.run();
    }
    const reps = Math.max(1, Math.round(TARGET_NS / probe));
    const runA = () => {
      for (let r = 0; r < reps; r++) ca.run();
    };
    const runB = () => {
      for (let r = 0; r < reps; r++) cb.run();
    };

    for (let w = 0; w < WARMUP; w++) {
      runA();
      runB();
    }
    let minA = Number.POSITIVE_INFINITY;
    let minB = Number.POSITIVE_INFINITY;
    for (let iter = 0; iter < ITERS; iter++) {
      if (iter % 2 === 0) {
        minA = Math.min(minA, time(runA));
        minB = Math.min(minB, time(runB));
      } else {
        minB = Math.min(minB, time(runB));
        minA = Math.min(minA, time(runA));
      }
    }
    rows.push({ name: ca.name, a: minA / reps, b: minB / reps });
  }
  return rows;
}

if (process.env.AB_CHILD) {
  const rows = await measure(process.argv[2], process.argv[3]);
  console.log(JSON.stringify(rows));
} else {
  const revB = process.argv[3] ?? "WORKTREE";
  const entryA = materialise(revA, "a");
  const entryB = materialise(revB, "b");

  const self = fileURLToPath(import.meta.url);
  const child = (first: string, second: string): Array<Row> => {
    const res = spawnSync(process.execPath, ["--import", "tsx", self, first, second], {
      env: { ...process.env, AB_CHILD: "1" },
      encoding: "utf8",
      maxBuffer: 64 * 1024 * 1024,
    });
    if (res.status !== 0) {
      throw new Error(`child failed: ${res.stderr}`);
    }
    const lines = res.stdout.trim().split("\n");
    return JSON.parse(lines[lines.length - 1]);
  };

  /* Two children per load order; per-case minimum across same-order
   * children, then geometric mean across orders. */
  const minRows = (x: Array<Row>, y: Array<Row>): Array<Row> =>
    x.map((r, i) => ({ name: r.name, a: Math.min(r.a, y[i].a), b: Math.min(r.b, y[i].b) }));
  const run1 = minRows(child(entryA, entryB), child(entryA, entryB));
  const run2 = minRows(child(entryB, entryA), child(entryB, entryA));

  const fmt = (ns: number) => (ns / 1e6).toFixed(4).padStart(10);
  const pctFmt = (p: number) => `${p.toFixed(2).padStart(7)}%`;

  console.log(`A = ${revA}, B = ${revB} (min of ${ITERS} iters x 2 orders, ms)`);
  console.log(`${"case".padEnd(24)}${"A".padStart(10)}${"B".padStart(10)}${"delta".padStart(9)}`);
  let sumA = 0;
  let sumB = 0;
  let ratioTotal1 = 0;
  let ratioTotal2 = 0;
  const totals1 = { a: 0, b: 0 };
  const totals2 = { a: 0, b: 0 };
  for (let i = 0; i < run1.length; i++) {
    const r1 = run1[i];
    const r2 = run2[i];
    /* r1: a=A, b=B; r2: a=B, b=A */
    const ratio = Math.sqrt((r1.b / r1.a) * (r2.a / r2.b));
    const aAvg = (r1.a + r2.b) / 2;
    const bAvg = aAvg * ratio;
    sumA += aAvg;
    sumB += bAvg;
    totals1.a += r1.a;
    totals1.b += r1.b;
    totals2.a += r2.a;
    totals2.b += r2.b;
    console.log(`${r1.name.padEnd(24)}${fmt(aAvg)}${fmt(bAvg)} ${pctFmt((ratio - 1) * 100)}`);
  }
  ratioTotal1 = totals1.b / totals1.a;
  ratioTotal2 = totals2.a / totals2.b;
  const totalRatio = Math.sqrt(ratioTotal1 * ratioTotal2);
  console.log(`${"TOTAL".padEnd(24)}${fmt(sumA)}${fmt(sumA * totalRatio)} ${pctFmt((totalRatio - 1) * 100)}`);
}
```

</details>

<details>
<summary><code>perf/cases.mts</code> — deterministic workloads (mirroring <code>packages/bench</code>)</summary>

```ts
/**
 * Deterministic perf workloads mirroring packages/bench.
 * Each case exposes a timed body (`run`) and a serializable
 * behaviour sample (`collect`) used by the characterisation guard.
 */

export interface PerfCase {
  name: string;
  run: () => void;
  collect: () => unknown;
}

/** mulberry32 PRNG for reproducible inputs */
function rng(seed: number): () => number {
  let a = seed >>> 0;
  return () => {
    a = (a + 0x6d2b79f5) | 0;
    let t = Math.imul(a ^ (a >>> 15), 1 | a);
    t = (t + Math.imul(t ^ (t >>> 7), 61 | t)) ^ t;
    return ((t ^ (t >>> 14)) >>> 0) / 4294967296;
  };
}

function randomString(rand: () => number, length: number): string {
  const chars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
  let out = "";
  for (let i = 0; i < length; i++) out += chars[Math.floor(rand() * chars.length)];
  return out;
}

const LONG_STRING = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor. ".repeat(12);

export function buildCases(z: any): Array<PerfCase> {
  const cases: Array<PerfCase> = [];

  {
    const rand = rng(1);
    const schema = z.object({ string: z.string(), boolean: z.boolean(), number: z.number() });
    const data = Array.from({ length: 1000 }, () => ({
      number: rand(),
      string: `${rand()}`,
      boolean: rand() > 0.5,
    }));
    cases.push({
      name: "object-parse",
      run: () => {
        for (const d of data) schema.parse(d);
      },
      collect: () => data.map((d) => schema.parse(d)),
    });
  }

  {
    const schema = z.strictObject({
      number: z.number(),
      negNumber: z.number(),
      maxNumber: z.number(),
      string: z.string(),
      longString: z.string(),
      boolean: z.boolean(),
      deeplyNested: z.strictObject({ foo: z.string(), num: z.number(), bool: z.boolean() }),
    });
    const data = Array.from({ length: 1000 }, (_, i) => ({
      number: i,
      negNumber: -i,
      maxNumber: Number.MAX_VALUE,
      string: "string",
      longString: LONG_STRING,
      boolean: true,
      deeplyNested: { foo: "bar", num: i, bool: false },
    }));
    cases.push({
      name: "strict-moltar",
      run: () => {
        for (const d of data) schema.parse(d);
      },
      collect: () => data.slice(0, 20).map((d) => schema.parse(d)),
    });
  }

  {
    const rand = rng(2);
    const schema = z.string();
    const data = Array.from({ length: 10_000 }, () => `${rand()}`);
    cases.push({
      name: "string-parse",
      run: () => {
        for (const d of data) schema.parse(d);
      },
      collect: () => data.slice(0, 50).map((d) => schema.parse(d)),
    });
  }

  {
    const rand = rng(3);
    const schema = z.number();
    const data = Array.from({ length: 10_000 }, () => rand() * 1_000_000);
    cases.push({
      name: "number-parse",
      run: () => {
        for (const d of data) schema.parse(d);
      },
      collect: () => data.slice(0, 50).map((d) => schema.parse(d)),
    });
  }

  {
    const rand = rng(4);
    const schema = z.array(z.string());
    const data = Array.from({ length: 200 }, () => Array.from({ length: 20 }, () => randomString(rand, 8)));
    cases.push({
      name: "array-parse",
      run: () => {
        for (const d of data) schema.parse(d);
      },
      collect: () => data.slice(0, 10).map((d) => schema.parse(d)),
    });
  }

  {
    const schema = z.union([
      z.object({ type: z.literal("a") }),
      z.object({ type: z.literal("b") }),
      z.object({ type: z.literal("c") }),
    ]);
    const data = Array.from({ length: 1000 }, (_, i) => ({ type: "abc"[i % 3] }));
    cases.push({
      name: "union-parse",
      run: () => {
        for (const d of data) schema.parse(d);
      },
      collect: () => data.slice(0, 9).map((d) => schema.parse(d)),
    });
  }

  {
    const rand = rng(5);
    const types = ["a", "b", "c", "d", "e", "f", "g"];
    const options = types.map((t) =>
      z.object({ type: z.literal(t), data1: z.string(), data2: z.string(), data3: z.string() })
    );
    const data = Array.from({ length: 200 }, () => ({
      type: types[Math.floor(rand() * types.length)],
      data1: randomString(rand, 10),
      data2: randomString(rand, 10),
      data3: randomString(rand, 10),
    }));
    cases.push({
      name: "disc-union-parse",
      run: () => {
        for (const d of data) schema.parse(d);
      },
      collect: () => data.slice(0, 14).map((d) => schema.parse(d)),
    });
  }

  {
    const rand = rng(6);
    const schema = z.object({ string: z.string(), boolean: z.boolean(), number: z.number() });
    const bad: Array<unknown> = Array.from({ length: 200 }, (_, i) => {
      const kind = i % 5;
      if (kind === 0) return { string: rand(), boolean: "yes", number: `${rand()}` };
      if (kind === 1) return { string: "ok", boolean: true };
      if (kind === 2) return null;
      if (kind === 3) return { string: "ok", boolean: false, number: Number.NaN };
      return [1, 2, 3];
    });
    cases.push({
      name: "object-fail-safeparse",
      run: () => {
        for (const d of bad) schema.safeParse(d);
      },
      collect: () =>
        bad.map((d) => {
          const r = schema.safeParse(d);
          return r.success ? { success: true, data: r.data } : { success: false, issues: r.error.issues };
        }),
    });
  }

  {
    const rand = rng(7);
    const email = z.email();
    const uuid = z.uuid();
    const datetime = z.iso.datetime();
    const emails = Array.from({ length: 200 }, (_, i) =>
      i % 2 === 0 ? `${randomString(rand, 8)}@example.com` : `${randomString(rand, 8)}example.com`
    );
    const uuids = Array.from({ length: 200 }, (_, i) =>
      i % 2 === 0 ? "8a99219c-2a97-4a35-9fcf-5a5a4d3b1234" : randomString(rand, 36)
    );
    const datetimes = Array.from({ length: 200 }, (_, i) =>
      i % 2 === 0 ? "2024-01-15T12:30:00Z" : `${randomString(rand, 10)}T99:99`
    );
    cases.push({
      name: "string-formats",
      run: () => {
        for (const d of emails) email.safeParse(d);
        for (const d of uuids) uuid.safeParse(d);
        for (const d of datetimes) datetime.safeParse(d);
      },
      collect: () => {
        const sample = (schema: any, inputs: Array<string>) =>
          inputs.slice(0, 6).map((d) => {
            const r = schema.safeParse(d);
            return r.success ? { success: true, data: r.data } : { success: false, issues: r.error.issues };
          });
        return [sample(email, emails), sample(uuid, uuids), sample(datetime, datetimes)];
      },
    });
  }

  {
    const rand = rng(8);
    const schema = z
      .string()
      .min(1)
      .transform((s: string) => s.length)
      .pipe(z.number().max(1000));
    const data = Array.from({ length: 2000 }, () => randomString(rand, 1 + Math.floor(rand() * 20)));
    cases.push({
      name: "chain-parse",
      run: () => {
        for (const d of data) schema.parse(d);
      },
      collect: () => data.slice(0, 50).map((d) => schema.parse(d)),
    });
  }

  {
    const makeSchemas = () => {
      const simple = z.object({ string: z.string(), boolean: z.boolean(), number: z.number() });
      const moltar = z.strictObject({
        number: z.number(),
        negNumber: z.number(),
        maxNumber: z.number(),
        string: z.string(),
        longString: z.string(),
        boolean: z.boolean(),
        deeplyNested: z.strictObject({ foo: z.string(), num: z.number(), bool: z.boolean() }),
      });
      return { simple, moltar };
    };
    cases.push({
      name: "schema-init",
      run: () => {
        for (let i = 0; i < 30; i++) makeSchemas();
      },
      collect: () => {
        const { simple, moltar } = makeSchemas();
        return [
          simple.parse({ string: "s", boolean: true, number: 1 }),
          moltar.parse({
            number: 1,
            negNumber: -1,
            maxNumber: Number.MAX_VALUE,
            string: "string",
            longString: LONG_STRING,
            boolean: true,
            deeplyNested: { foo: "bar", num: 1, bool: false },
          }),
        ];
      },
    });
  }

  return cases;
}
```

</details>

---

Companion PRs from the same campaign: #6316 #6317 #6319

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -35,11 +35,44 @@ The project uses pnpm workspaces. Key commands:
 - Ask before generating new files
 - Use `util.defineLazy()` for computed properties to avoid circular dependencies
 - Performance is critical - parameter reassignment is allowed for optimization
+- Any change to `packages/zod/src` must be weighed on **all three axes: runtime performance, memory consumption, and bundle size** — see "The three axes" below. A change that improves one and is only checked on that one is not finished.
 - ALWAYS use the `gh` CLI to fetch GitHub information (issues, PRs, etc.) instead of relying on web search or assumptions
 - Keep JSDoc as minimal as possible. A self-explanatory type or symbol name needs no doc comment. When a comment is genuinely required, write one short sentence describing behavior — not history, rationale, or examples. Don't add interface-level JSDoc that just restates the interface name.
 - When you've modified a PR (or opened/closed/commented on one), include the PR URL liberally in summary messages — at minimum once at the end of any reply that touched it
 - When creating a PR, do not include a separate test plan section in the body. Link to any relevant issues under discussion, and use the same copywriting guidelines from "Commenting on issues and PRs": concise maintainer voice, prose over templates, and validation details only when they are material to the reader.
 - NEVER bump the version in `packages/zod/package.json` (or any package's `package.json`). A version bump is the only thing that triggers a release; everything else (including direct pushes to `main`) is recoverable until that happens. If a version bump is genuinely needed, ask first.
+
+## The three axes
+
+Zod is judged on **runtime performance**, **memory consumption**, and **bundle size** at once, and they trade against each other constantly. Optimizing one in isolation is how regressions land: a change that buys bundle bytes can cost construction speed, and one that saves memory can cost both. Any non-trivial change to `packages/zod/src` — and any change to `core/` at all, since every build ships it — needs a number on all three before it is done. Report them together, including the ones that got worse.
+
+`zod/mini` deserves its own line on the bundle axis: it is sold on size, so a fixed cost lands very differently there (~200 B is 6% of the smallest mini bundle and 1% of classic's).
+
+| axis    | how to measure                                                                                                                                    |
+| ------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
+| runtime | `packages/bench/*.ts` via `pnpm bench <name>`, plus construction cost — parse and construction move independently                                 |
+| memory  | `packages/bench/memory/schema-footprint.ts` and `realworld.ts` (retained bytes per schema; run under `--expose-gc`)                               |
+| bundle  | bundle a `packages/treeshake` fixture with esbuild `--minify` under `--conditions=@zod/source` and gzip it, for both a classic and a mini fixture |
+
+### Benchmarking traps
+
+These produced hours of wrong conclusions; check them before believing a number.
+
+- **Check `uptime` first.** A loaded machine invents effects of ±16%. Numbers taken above a load average of ~8 are worthless.
+- **Interleave the two revisions.** Running all of A then all of B puts any drift entirely on whichever ran second. Alternating round-by-round took per-side spread from 2.5x to 4%.
+- **Loading two zod revisions into one process makes call sites polymorphic.** It is fine for comparing construction, unreliable for tight leaf parses.
+- **Allocating benchmarks can't be timed by a fixed-duration loop.** `z.array().parse()` allocates per call, so a time-boxed harness samples whatever the collector is doing — it gave +8.2% and −17.3% on consecutive identical runs. Use a fixed iteration count with `gc()` between samples and take the minimum.
+- **Make sure the work isn't optimized away.** If a micro-benchmark reports ~1e9 ops/sec, V8 eliminated the loop; consume the result.
+
+### Property placement, specifically
+
+Most of the memory in a schema is its own-property count, not its closures. V8 sizes an instance's backing store in steps and `$constructor` assigns nothing itself, so instances get no in-object slots: **≤12 own properties cost 128 bytes, 13–20 cost 848, ≥21 cost 1616** (`packages/bench/memory/prop-slack.ts` measures this). Members therefore live on the prototype and materialize per instance on first read.
+
+If you touch that machinery, three things bite:
+
+- **Moving a property between definition sites changes its descriptor**, and writability/enumerability/configurability are part of the public contract. `packages/bench/memory/` has a surface-diff approach for this: dump every descriptor, alias and assign/delete semantic on both revisions and diff them. It caught four real regressions that tests did not.
+- **Redefining an accessor demotes the object to dictionary mode**, which cost 2x on `z.object().parse` once. Run `packages/bench/memory/dict-mode.ts` after any change here — every instance must report `fast`.
+- **A bound function pays a call-time trampoline.** Fine for cold builder methods, measurably not fine on the parse path.
 
 ## Cutting a release
 
@@ -89,7 +122,8 @@ git push <contributor> pr-<N>:<headRefName> --force-with-lease   # for amends
 ```
 
 Notes:
-- Do NOT use `gh pr checkout --detach` for this — it moves your *current* working tree into detached HEAD instead of creating a worktree.
+
+- Do NOT use `gh pr checkout --detach` for this — it moves your _current_ working tree into detached HEAD instead of creating a worktree.
 - Husky pre-commit runs biome format/lint via lint-staged; pre-push runs the full vitest suite. Both are fast and act as a safety net — don't bypass with `--no-verify` unless you have a specific reason.
 - **Preserve contributor commits.** Never `git reset --hard` or otherwise rewrite history that erases the contributor's work, even if you're rewriting the actual change. They need to stay in the PR's commit list to get credit on the merged PR. If their approach was wrong, add a `Revert "..."` commit (or just a plain commit that undoes those lines) and then add your replacement commit on top. Force-pushing a single clobbering commit strips them from the GitHub contributors graph.
 - When done, clean up: `git worktree remove ~/.cursor/worktrees/zod/pr-<N>` and `git branch -D pr-<N>` (and optionally `git remote remove <contributor>`).
--- a/packages/bench/memory/breakdown.ts
+++ b/packages/bench/memory/breakdown.ts
@@ -1,0 +1,36 @@
+import * as z from "zod";
+import { fmtBytes, table } from "./harness.js";
+import { diffAllocation } from "./snapshot.js";
+
+const which = process.argv[2] ?? "string";
+
+const factories: Record<string, () => unknown> = {
+  string: () => z.string(),
+  number: () => z.number(),
+  "object-empty": () => z.object({}),
+  "object-3": () => z.object({ a: z.string(), b: z.number(), c: z.boolean() }),
+  "string-min": () => z.string().min(1),
+  optional: () => z.string().optional(),
+  union: () => z.union([z.string(), z.number()]),
+  array: () => z.array(z.string()),
+};
+
+const factory = factories[which];
+if (!factory) {
+  console.error(`unknown case "${which}"; try: ${Object.keys(factories).join(", ")}`);
+  process.exit(1);
+}
+
+const rows = diffAllocation(factory, 2_000, 8_000);
+const total = rows.reduce((sum, r) => sum + r.bytesEach, 0);
+
+console.log(`allocation breakdown for ${which} (per instance)\n`);
+table(
+  rows.map((r) => ({
+    node: r.what,
+    "per inst": r.countEach.toFixed(2),
+    bytes: r.bytesEach.toFixed(0),
+    human: fmtBytes(r.bytesEach),
+  }))
+);
+console.log(`\ntotal attributed: ${fmtBytes(total)}`);
--- a/packages/bench/memory/calibrate.ts
+++ b/packages/bench/memory/calibrate.ts
@@ -1,0 +1,38 @@
+/**
+ * Sanity-checks the measurement harness against allocations of known size.
+ * If these controls don't land near their theoretical sizes, no zod number
+ * produced by the same harness is trustworthy.
+ */
+import { fmtBytes, measureStable, table } from "./harness.js";
+
+const N = 20_000;
+
+const controls: Array<[string, string, (i: number) => unknown]> = [
+  ["{}", "~32 B", () => ({})],
+  ["{a:1,b:2,c:3}", "~56 B", () => ({ a: 1, b: 2, c: 3 })],
+  ["new Array(10)", "~120 B", () => new Array(10).fill(0)],
+  ["new Set()", "~100 B", () => new Set()],
+  ["new Set(['a','b','c'])", "~200 B", () => new Set(["a", "b", "c"])],
+  ["new Map()", "~100 B", () => new Map()],
+  ["closure over 1 var", "~100 B", (i) => () => i],
+  ["new Uint8Array(1024)", "~1088 B", () => new Uint8Array(1024)],
+  ["10-char string", "~40 B", (i) => `str-${i}-xxxx`.slice(0, 10) + String(i)],
+  [
+    "obj w/ getter (defineProperty)",
+    "~200 B",
+    () => {
+      const o: any = {};
+      Object.defineProperty(o, "x", { get: () => 1, configurable: true });
+      return o;
+    },
+  ],
+];
+
+console.log(`harness calibration, n=${N}\n`);
+
+table(
+  controls.map(([label, expected, factory]) => {
+    const m = measureStable(label, N, factory, 5);
+    return { control: label, expected, measured: fmtBytes(m.bytesEach), bytes: m.bytesEach.toFixed(0) };
+  })
+);
--- a/packages/bench/memory/dict-mode.ts
+++ b/packages/bench/memory/dict-mode.ts
@@ -1,0 +1,77 @@
+/**
+ * Reports whether schema instances and their `_zod` internals are in V8
+ * dictionary (slow) mode. A dictionary-mode object pays a NameDictionary
+ * backing store — roughly 900 B at these property counts — instead of inline
+ * slots, and every property read goes through a hash lookup.
+ *
+ * Run with: node --allow-natives-syntax --import tsx
+ */
+import * as z from "zod";
+import { table } from "./harness.js";
+
+// biome-ignore lint/security/noGlobalEval: V8 natives syntax is not valid in a module body.
+const hasFast: (o: object) => boolean = eval("(o) => %HasFastProperties(o)");
+
+const cases: Array<[string, () => any]> = [
+  ["z.string()", () => z.string()],
+  ["z.number()", () => z.number()],
+  ["z.boolean()", () => z.boolean()],
+  ["z.bigint()", () => z.bigint()],
+  ["z.object({a})", () => z.object({ a: z.string() })],
+  ["z.array(str)", () => z.array(z.string())],
+  ["z.string().min(1)", () => z.string().min(1)],
+  ["z.string().optional()", () => z.string().optional()],
+  ["z.email()", () => z.email()],
+  ["z.enum(['a'])", () => z.enum(["a"])],
+];
+
+// Reference points so the reader can see what fast/slow looks like.
+const plain: any = { a: 1, b: 2 };
+const forcedSlow: any = {};
+Object.defineProperty(forcedSlow, "x", { value: 1, enumerable: false });
+
+console.log(
+  `reference: plain literal fast=${hasFast(plain)}, non-enumerable defineProperty fast=${hasFast(forcedSlow)}\n`
+);
+
+table(
+  cases.map(([label, f]) => {
+    const s = f();
+    return {
+      schema: label,
+      "inst fast?": hasFast(s) ? "fast" : "DICT",
+      "inst props": Reflect.ownKeys(s).length,
+      "_zod fast?": hasFast(s._zod) ? "fast" : "DICT",
+      "_zod props": Reflect.ownKeys(s._zod).length,
+      "def fast?": hasFast(s._zod.def) ? "fast" : "DICT",
+    };
+  })
+);
+
+// Bisect: which operation in the constructor tips the instance over?
+console.log("\nwhat forces dictionary mode:");
+const probes: Array<[string, (o: any) => void]> = [
+  [
+    "plain assignment x25",
+    (o) => {
+      for (let i = 0; i < 25; i++) o[`p${i}`] = i;
+    },
+  ],
+  [
+    "defineProperty all-true x25",
+    (o) => {
+      for (let i = 0; i < 25; i++)
+        Object.defineProperty(o, `p${i}`, { value: i, writable: true, enumerable: true, configurable: true });
+    },
+  ],
+  ["defineProperty enumerable:false x1", (o) => Object.defineProperty(o, "z", { value: 1, enumerable: false })],
+  ["defineProperty value-only x1", (o) => Object.defineProperty(o, "z", { value: 1 })],
+  ["defineProperty getter x1", (o) => Object.defineProperty(o, "z", { get: () => 1, configurable: true })],
+];
+table(
+  probes.map(([label, apply]) => {
+    const o: any = {};
+    apply(o);
+    return { operation: label, result: hasFast(o) ? "fast" : "DICT" };
+  })
+);
--- a/packages/bench/memory/harness.ts
+++ b/packages/bench/memory/harness.ts
@@ -1,0 +1,83 @@
+/**
+ * Shared measurement utilities for the memory benchmarks.
+ * Run everything under `node --expose-gc`.
+ */
+
+declare const gc: (() => void) | undefined;
+
+export function collect(): void {
+  if (typeof gc !== "function") {
+    throw new Error("run with --expose-gc");
+  }
+  // Repeated collections: the first pass frees the bulk, later passes catch
+  // objects that only became unreachable once the first pass ran.
+  for (let i = 0; i < 6; i++) gc();
+}
+
+export function heapUsed(): number {
+  collect();
+  return process.memoryUsage().heapUsed;
+}
+
+export interface Measurement {
+  label: string;
+  count: number;
+  totalBytes: number;
+  bytesEach: number;
+}
+
+/**
+ * Builds `count` instances via `factory`, holds them all live, and reports the
+ * retained heap delta. The sink array itself is subtracted out.
+ */
+export function measure(label: string, count: number, factory: (i: number) => unknown): Measurement {
+  // Warm up so lazily-initialized module state isn't attributed to the payload.
+  factory(0);
+
+  const sink: unknown[] = new Array(count);
+  const before = heapUsed();
+  for (let i = 0; i < count; i++) {
+    sink[i] = factory(i);
+  }
+  const after = heapUsed();
+
+  // Keep the sink reachable across the measurement.
+  if (sink.length !== count) throw new Error("unreachable");
+
+  const arrayOverhead = count * 8;
+  const totalBytes = after - before - arrayOverhead;
+  return { label, count, totalBytes, bytesEach: totalBytes / count };
+}
+
+/**
+ * Median-of-runs variant. Allocation measurements are noisy; taking the median
+ * of independent runs keeps a single unlucky GC from dominating the number.
+ */
+export function measureStable(label: string, count: number, factory: (i: number) => unknown, runs = 5): Measurement {
+  const samples: number[] = [];
+  for (let r = 0; r < runs; r++) {
+    samples.push(measure(label, count, factory).bytesEach);
+  }
+  samples.sort((a, b) => a - b);
+  const bytesEach = samples[Math.floor(samples.length / 2)]!;
+  return { label, count, totalBytes: bytesEach * count, bytesEach };
+}
+
+export function fmtBytes(n: number): string {
+  const abs = Math.abs(n);
+  if (abs >= 1024 * 1024) return `${(n / 1024 / 1024).toFixed(2)} MB`;
+  if (abs >= 1024) return `${(n / 1024).toFixed(1)} KB`;
+  return `${n.toFixed(0)} B`;
+}
+
+export function table(rows: Array<Record<string, string | number>>): void {
+  if (rows.length === 0) return;
+  const cols = Object.keys(rows[0]!);
+  const widths = cols.map((c) => Math.max(c.length, ...rows.map((r) => String(r[c]).length)));
+  const line = (cells: string[]) => cells.map((cell, i) => cell.padEnd(widths[i]!)).join("  ");
+  console.log(line(cols));
+  console.log(widths.map((w) => "-".repeat(w)).join("  "));
+  for (const row of rows) {
+    console.log(line(cols.map((c) => String(row[c]))));
+  }
+}
--- a/packages/bench/memory/leaks.ts
+++ b/packages/bench/memory/leaks.ts
@@ -1,0 +1,145 @@
+/**
+ * Empirical leak detection.
+ *
+ * Each scenario runs a workload that creates schemas and/or parses data and
+ * then DROPS every reference it created. A correct scenario returns to its
+ * starting heap. Anything that grows linearly with iteration count is
+ * retaining something it shouldn't.
+ *
+ * Growth is measured across two batches of different sizes so a fixed
+ * one-time cost (module init, lazily-installed prototypes) doesn't read as a
+ * leak: only the SLOPE matters.
+ */
+import * as z from "zod";
+import { collect, fmtBytes, table } from "./harness.js";
+
+interface Scenario {
+  label: string;
+  /** Must leave nothing reachable behind. */
+  run: (i: number) => void;
+  /** Known-leaky scenarios we assert ON, to prove the detector works. */
+  expectLeak?: boolean;
+}
+
+const scenarios: Scenario[] = [
+  // --- controls: these MUST show a leak, proving the detector is sensitive ---
+  {
+    label: "CONTROL: push to module array",
+    run: (i) => void leakSink.push(z.string().min(i % 5)),
+    expectLeak: true,
+  },
+  {
+    label: "CONTROL: no-op",
+    run: () => {},
+  },
+
+  // --- schema construction, all references dropped ---
+  { label: "z.string()", run: () => void z.string() },
+  { label: "z.string().min(1)", run: () => void z.string().min(1) },
+  { label: "z.object({...})", run: () => void z.object({ a: z.string(), b: z.number() }) },
+  { label: "z.object().extend()", run: () => void z.object({ a: z.string() }).extend({ b: z.number() }) },
+  { label: "z.object().partial()", run: () => void z.object({ a: z.string() }).partial() },
+  {
+    label: "z.discriminatedUnion",
+    run: () => void z.discriminatedUnion("t", [z.object({ t: z.literal("a") }), z.object({ t: z.literal("b") })]),
+  },
+  {
+    label: "z.lazy recursive",
+    run: () => {
+      const s: any = z.lazy(() => z.object({ v: z.string(), next: s.optional() }));
+      void s.safeParse({ v: "x", next: { v: "y" } });
+    },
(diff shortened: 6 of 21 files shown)


wahajahmed010:
A few things worth looking at, ordered by how concerning I think they are:

**1. `_def` is now writable.** The old code did `Object.defineProperty(inst, "_def", { value: def })` (defaults: `writable: false, configurable: false`). The new code installs a prototype getter and, on first write, falls through to `Object.defineProperty(this, "_def", { value, configurable: true, writable: true })`. That means any caller doing `inst._def = something` now silently succeeds and mutates the schema instead of throwing in strict mode. If that's intentional (e.g. a deliberate fix for an existing bug), great — but it's a behavior change and worth a one-line test (`"use strict"; inst._def = {}` should not throw under the new code) and a CHANGELOG note. If it's accidental, the setter should either be removed or made a no-op.

**2. `~standard` setter on the prototype is missing.** You added a getter for `~standard` on the prototype, but the `set` branch of the original `util.defineLazy` is gone. `defineLazy` previously allowed programmatic assignment; with a prototype getter that has no setter, `inst["~standard"] = x` throws in strict mode. Probably no one is doing that externally, but `z.parse` → `~standard` was a documented extension point.

**3. `~standard.jsonSchema` set order.** `_onStandard` is wired via `inst._zod.onStandard = _onStandard` after the user has already constructed the schema. If anything reads `inst["~standard"].jsonSchema` between construction and the first read of `~standard`, it will be undefined. The lazy-install design assumes the *first* read of `~standard` happens after construction, which is true in the normal flow — but if a plugin reads `~standard` during construction, it would get the bare object without `jsonSchema`. Worth a comment near `_onStandard` flagging that.

**4. `_zodDesc` reuse.** `_zodDesc.value = undefined;` after each `defineProperty` is the right cleanup, but if any future code adds an early return between the assignment and the reset (e.g. a thrown error), the next instance will see stale data. A try/finally or just `const desc: PropertyDescriptor = { value: {...}, enumerable: false }` per call (cheap with V8 escape analysis) would be safer. Minor.

**5. The `_LazyMethodsOf<T>` group name for codecs.** `"ZodTypeCodecs"` and `"ZodStringFormats"` are added to the same `WeakSet` of installed groups, so re-instantiation with the same prototype is a no-op. Looks correct, but if a third party monkey-patches `ZodType.prototype` after the first instance is built (e.g. for the codec methods), it will be silently ignored for subsequent instances. Unlikely but worth a `// do not modify after first construction` comment.

The benchmark methodology (A/B harness, 11 workloads, 25 iterations, min) is solid and the numbers are credible. The micro-opts are individually correct. The `_def` writability and the missing `~standard` setter are the two I'd want a maintainer to confirm are deliberate before merging.

colinhacks:
Pushed my own memory work onto this branch and closed #6343, which was covering the same ground from the other direction. Your four commits are intact underneath. Good instinct on the prototype install — that was the right shape, and most of what I added just carries it the rest of the way.

On top of yours: the parse family, `toJSONSchema` and the `ZodBigInt` checks move to the prototype the same way your format methods do. That takes `z.string()` from 12 own properties to 6 and `z.bigint()` from 23 to 6 — both now under V8's 12-property step, where the backing store costs 128 bytes instead of 848 or 1616. `zod/mini` gets the same treatment; it sat at 14. The installers moved to `core/util` so classic and mini share one copy, and the method tables are hoisted to module scope — they were rebuilt on every construction, ~71 throwaway closures per `z.string()`. The deferred initializer list is released once it has run.

The parse family uses a separate installer that *builds* the closure rather than binding a shared function: routing it through `.bind()` cost 6–19% on the trampoline. The codec family keeps your bind treatment — it's off the hot path, the source is much tighter, and it measured 22–38% faster than main anyway.

Kept two things from your version that mine didn't have: `_def` deriving from `this`, and the `checks` array skipping the spread when it's empty. Dropped `onStandard` — classic now installs its own `~standard` factory over core's, which does the same job without costing an own property on `_zod` for every schema.

Your `Object.keys` → `for..in` swap is a clear win: +49% on an empty prototype, +24% with methods installed. I benchmarked it expecting to argue against it, but my earlier measurement used `Object.create(null)`, which is the wrong object shape.

Where it lands, against this branch and against main:

| | main | this PR | now |
| --- | --- | --- | --- |
| `z.string()` | 7709 B | 1633 B | 816 B |
| `z.object()` 10 keys | 83928 B | 21266 B | 12172 B |
| mini `z.string()` | 2528 B | 1560 B | 593 B |
| 500-schema catalogue | 198.0 MB | 83.6 MB | 62.7 MB |

Construction is ~5.3x faster than main and parse is flat across all 17 benches in `packages/bench`. Bundle is the one cost: +146–183 B gzipped for classic, +182–248 B for mini.

`packages/bench/memory/` has the harness this came out of. The useful piece is `prop-slack.ts` — it's what turned up the step function. Run `dict-mode.ts` after any change here: redefining an accessor per instance silently demotes every schema to dictionary mode, which cost 2x on `z.object().parse` before I caught it.

colinhacks:
Pulled `zod/mini` back out of this. Moving its methods to the prototype cost ~180 B gzipped of installer machinery — about 6% of the smallest mini bundle, and mini is sold on size, so it isn't worth 1.5 KB per schema there.

Core's `~standard` accessor is written out inline rather than routed through the shared installers, so mini's bundle no longer pulls them in at all. Mini still gets most of the win for free: `z.string()` goes 2528 → 1240 bytes retained, and its bundle cost drops from +182 B to +61 B gzipped. Classic keeps the full treatment and is unchanged — `z.string()` 816 B, a 500-schema catalogue at 62.7 MB against 198 MB on main.

The one cost is that classic picks up ~54 B from carrying core's inline accessor alongside its own installer. Cheap trade against mini going from +6.3% to +2.1%.

Also tried three ways to shrink the machinery itself — a caller-owned `WeakSet` per site, a shared property descriptor, and collapsing the two installers into one. All three measured *larger*; the minifier was already doing better than any of them. Not worth revisiting.

wahajahmed010:
Following up after the latest push (since my Aug 7 review). The `_zodDesc` reuse trick, the `~standard` prototype accessor on the core side, and the `deferred = undefined` release are the new pieces — focused comments on those.

**`_zodDesc` reuse is fine, but the reset ordering is brittle.** `core.ts:28–34`:

```
_zodDesc.value = { def, constr: _, traits: new Set() };
Object.defineProperty(inst, "_zod", _zodDesc);
_zodDesc.value = undefined;
```

`defineProperty` with a value descriptor is synchronous and either succeeds or throws synchronously. The reset happens unconditionally after, so a throw here would leak the previous-call object into the next instance's descriptor. Today nothing throws between the assignment and the reset, but the invariant isn't enforced. Two cheap fixes: a `try/finally`, or build the descriptor inline (`const desc: PropertyDescriptor = { value: {…}, enumerable: false }`) and let V8's escape analysis elide the alloc — the `defineProperty` op is already far more expensive than the literal, and a throw would no longer leave the shared one in a bad state. The comment on lines 16–17 says it avoids a per-instance allocation, which is true; just say it does so *under the throw-impossible invariant*.

**`~standard` core-side prototype accessor is correct, with one caveat.** `core/schemas.ts:317–337`: the `set` branch on the prototype is preserved (matching the previous `util.defineLazy` semantics), good. The one thing worth noting: the instance shadow written by the getter uses `{ value, writable: true, configurable: true }` with no `enumerable` flag, so the override is non-enumerable — consistent with the original. The classic override at `classic/schemas.ts:201` (`_installLazyProps(inst, "ZodType~standard", …)`) goes through `defineCached` in `util.ts:1017` which writes the instance override as `{ enumerable: true }`. That's a real difference: a classic `ZodType` instance ends up with `~standard` enumerable where a core `~standard` instance doesn't. It's not wrong — the instance has the value either way — but anyone doing `for (const k in schema)` (e.g. serialization, the existing `Reflect.ownKeys` test, the prototype-bind loop in `core.ts:46`) will see the field on classic instances. The `instance-footprint.test.ts` uses `Reflect.ownKeys`, which doesn't distinguish, so it passes either way. Worth either standardizing on one enumerability or adding a second test that asserts `Object.keys(z.string())` does not include `~standard` after first read. Otherwise the perf cliff that this whole PR is about (the 12→13 own-prop step) is silently re-crossed for classic users.

**`deferred = undefined` release is correct and worth a `// post-construction` comment.** `core/core.ts:64–72` drops the deferred list once the initializers have run. The list is only populated during `init` (schemas.ts:214–215), and `init` runs before the deferred loop in the constructor, so the release happens before any caller can observe the instance. The `_zod.deferred` lookup on the public surface is gone (`grep` shows it's only referenced internally now and in the new test that asserts `undefined`). One thing to double-check: the test on `instance-footprint.test.ts:97–98` reads `z.string()._zod.deferred` immediately after construction, which exercises exactly the released path. If a downstream user (or a future internal change) ever reads `_zod.deferred` between init completion and the loop completion, they'll get `undefined` instead of the now-empty array. The new test pins that contract — good — but it's a contract change from the previous behavior (`[]` after construction). Worth one CHANGELOG line.

**Bundle regression is real but bounded, and you've measured it correctly.** The two installers + three types in `util.ts` add the load colinhacks called out in the second commit. The "wrote out inline rather than routed through installers" choice for core's `~standard` (the comment on `schemas.ts:311–315`) is the right call for keeping `zod/mini` out of it, and the 6.3% → 2.1% bundle delta is a real win — but it does mean core and classic each carry their own ~30-line `defineProperty` block. The third commit's note that the minifier beat all three collapse attempts is the kind of result that should be put in a comment in `util.ts` so the next person doesn't re-try it.

One thing not in the PR: the `_zodDesc` and the deferred-release comment both point at `defineProperty` cost being the real lever, but `core.ts:46` still does a `for…in` over the prototype with a `hasOwnProperty` skip on every construction. The `for…in` itself is fast; the `hasOwnProperty` callback is the slow part (the call crosses a C++ boundary, and on an instance with ~6 own props it's a noticeable fraction of the whole init). A `for (const k in { [Symbol.iterator]: 0, …proto keys… })` is ugly; the cleaner version is `for (const k of Reflect.ownKeys(proto))` which is the same op but on the prototype, no skip needed. Worth a quick A/B.

Methodology side, the harness in `packages/bench/memory/prop-slack.ts` and `dict-mode.ts` is now the most important file in the repo for anyone touching this — that's the canary that catches a future patch crossing the step. A `// if you change anything in core/util.ts, run prop-slack.ts and dict-mode.ts and compare against the pinned numbers in MEMORY.md` at the top of `util.ts` would help.

colinhacks:
Good review — two of these were real and are now fixed.

**`_def` writability.** Confirmed against main: `schema._def = x` threw a `TypeError` there and silently succeeded here. The prototype accessor now has no setter, so it throws again. That surfaced a second thing: core's `$ZodFunction` sets `_def` by assignment, which a getter-only accessor rejects, so it now defines the own property directly with the same descriptor the classic build used to install. Pinned by a test.

**`~standard` enumerability.** Also real, and worse than described. On main `~standard` was never an own data property at all — `defineLazy` cached into a closure variable, so it never appeared in `Object.keys`. Here the cache wrote it as enumerable, so after the first read it showed up in both `Object.keys(schema)` and **`JSON.stringify(schema)`**. The cache is now written non-enumerable; an explicit `schema.parse = fn` still produces an enumerable property, matching a plain assignment. Test asserts reading a member doesn't change the enumerable surface.

Enumerability doesn't affect the property *count*, though, so the 12→13 step was never at risk — `Reflect.ownKeys` counts non-enumerables too. The real cost was the serialization leak.

**`_zodDesc` reset ordering.** Fair. Wrapped in `try/finally` rather than moving to a per-call literal, so the optimization survives.

**`Reflect.ownKeys(proto)` for the prototype scan** — I'd avoid this one. `Reflect.ownKeys` returns non-enumerables, which is every lazily-installed accessor on the prototype (~85 on `ZodString`), so the loop would iterate all of them and do an `in` check per key on every init instead of iterating nothing. The premise doesn't hold either: `for…in` over these prototypes yields **zero** keys, because the accessors are installed non-enumerable, so the `hasOwnProperty` callback never runs. Measured rather than reasoned: 0 enumerable keys visible to that scan, and 0 prototype `defineProperty` calls per construction once warm.

**`onStandard` and the group-name `WeakSet`** are both gone since your Aug 7 pass — the classic `~standard` factory now overrides core's prototype entry directly, and group memoization is a sentinel-key check rather than a `WeakSet`. Points 3 and 5 don't apply to the current diff.

**`deferred` contract change** (`[]` → `undefined` after construction) — agreed it's a contract change; the test pins it and it should get a changelog line.

Numbers after the fixes: mini +140 B gzipped, classic +184/+125. Instances all still fast-properties, `z.string()` 816 B, mini `z.string()` 593 B, 500-schema catalogue 62.9 MB against 198 MB on main. 3825 tests green.

colinhacks:
Went back and audited the *class* of both findings rather than just the two members, by dumping every property descriptor on every schema type — own vs inherited, writable/enumerable/configurable, get/set, function-identity aliases, and assign/delete semantics — and diffing main against the branch. That turned up two more of the same kind.

**The non-enumerable cache over-corrected.** Only `~standard` was never an own data property on main (the old `defineLazy` cached into a closure variable). The ~45 relocated methods all *were* enumerable own properties, so making the cache uniformly non-enumerable dropped them from `Object.keys(schema)` even after being touched. Enumerability now follows whatever the member had before it moved. The JSON leak was only ever `~standard`: methods never reached `JSON.stringify` either way, because functions don't serialize.

**`zod/mini`'s `with` lost its identity with `check`.** On main it was `inst.with = inst.check` — the same function object, not a delegating wrapper — so `schema.with === schema.check` held. It's installed as a prop resolving to `this.check` again, which costs nothing since the single-property installer is already in mini's bundle for `~standard`.

`Object.keys` and `JSON.stringify` now match main on every schema type except `z.function()`, where `_def` was uniquely enumerable *and* writable — it leaked into both. It's now non-enumerable and read-only like every other type's, which is the same inconsistency you flagged in point 1.

Three differences remain and are intentional:

- **`delete schema.parse` then reading returns the method again** (main: `undefined`). Deleting the cached own property re-exposes the prototype getter. Not fixable without abandoning prototype installation, and arguably the better behavior.
- **`delete schema._def` succeeds** (main: `TypeError`, since it was a non-configurable own property). There is no own property to delete now.
- **Untouched methods are absent from `Object.keys`.** Inherent to lazy installation — they materialize on first access.

3825 tests green. Perf numbers to follow.

colinhacks:
Perf numbers. Three benches kept flagging across runs — `string` swung from +16.2% to −16.2% depending on when I ran it. All three are flat:

| bench | delta | how |
| --- | --- | --- |
| `string` | −0.7% | interleaved main/head, 10 rounds, 4% spread each side |
| `object` | −1.8% | same, 9% spread |
| `array` | −0.2% | min-of-25 over a fixed 300k-parse workload with forced GC |

Two methodology notes worth having on the record, since anyone touching this area will hit them:

**`packages/bench` A/B by checkout is unreliable for leaf parses.** Running all of main and then all of head puts any drift between the two phases entirely on whichever ran second. Interleaving the two revisions round-by-round takes the per-side spread from 2.5x down to 4%.

**`z.array(...).parse()` is GC-sensitive and can't be measured by a time-boxed loop.** It allocates per parse, so a fixed-duration harness samples whatever the collector happens to be doing — it gave +8.2% and −17.3% on consecutive identical runs. A fixed iteration count with `gc()` between samples, taking the minimum, lands at −0.2% with every sample within 0.1 ms.

Check `uptime` before trusting any number from this area. A loaded machine invents effects of ±16%, which is larger than anything being measured here.

colinhacks:
Every change to `packages/zod/src` has to be weighed on all three axes, so here is the whole PR on all three, measured on a quiet box (load < 8) with GC-robust harnesses.

| | main | this PR |
| --- | --- | --- |
| `z.string()` retained | 7709 B | **816 B** |
| `z.object()` 10 keys | 83928 B | **12171 B** |
| mini `z.string()` retained | 2528 B | **593 B** |
| 500-schema catalogue | 198.0 MB | **62.7 MB** |
| construct `z.string()` (20k) | 73.7 ms | **12.2 ms** (−83%) |
| construct `z.object(3)` (20k) | 294.6 ms | **121.3 ms** (−59%) |
| construct mini `z.string()` | 30.5 ms | **8.3 ms** (−73%) |
| parse string / array / object | — | +1.0% / +2.7% / +4.4% |
| bundle, classic | — | +119–175 B gzip (+0.6–1.1%) |
| bundle, mini | — | +152–212 B gzip (+5.2–5.6%) |

Bundle is the only axis that gets worse. It is a roughly fixed ~200 B of installer machinery, which is 5.2% of the smallest mini bundle and 3.5% of a realistic mini app.

**This framing caught something I'd missed.** The install guard runs once per group on *every* construction, not just on install — so it is on the schema-creation hot path, and I had only ever measured its bundle effect. Checking all three changed the answer:

| | bundle (mini) | classic construct | mini construct |
| --- | --- | --- | --- |
| sentinel + `in` | +152 B | baseline | baseline |
| `WeakMap<proto, Set>` | +189 B | −3.1% | +4.1% |

The sentinel wins on bundle everywhere and on mini construction; the WeakMap is 3% better for classic construction only. Kept the sentinel and switched `hasOwnProperty.call` to `in`, since the builtin call is measurably more expensive than the operator at this frequency.

The three-axis requirement is now written into `AGENTS.md`, along with how to measure each one and the benchmarking traps that produced confidently wrong numbers in this PR — loaded machines, running all of A then all of B, loading two zod revisions into one process, and timing an allocating benchmark by duration.

colinhacks:
Closing out the last two items from your Aug 7 review that I'd acknowledged but not actually written down.

The prototype-mutation caveat is now a comment where it belongs — installs are memoized per prototype, so a prototype modified after the first instance of its type is built will never pick the change up. And the installer block carries a marker that the simplifications tried here (a per-site `WeakSet`, a shared property descriptor, collapsing the two table installers, inlining core's `~standard`) all measured *larger*, so the next person doesn't spend an afternoon re-deriving that.

Also corrected a claim I'd left upthread: the sentinel guard isn't simply "cheaper than a WeakMap of key-sets". It wins on bundle everywhere and on mini construction, and loses ~3% on classic construction — which only showed up once I measured all three axes rather than the one I was optimizing.

No changelog file in the repo, so the `deferred` contract change (`[]` → `undefined` after construction) lives in the commit message and the test that pins it.

Nothing outstanding from either of your passes now. 3825 tests green, CI green.

PR #6320, opened 2026-08-03
fix(from-json-schema): accept RFC 3339 numeric offsets for date-time format

### LINKED ISSUE #6296: fromJSONSchema date-time rejects valid RFC 3339 offsets
`fromJSONSchema({ type: "string", format: "date-time" })` currently maps to `z.iso.datetime()` and rejects valid RFC 3339 timestamps with numeric UTC offsets.

```ts
const schema = z.fromJSONSchema({ type: "string", format: "date-time" });

schema.parse("2026-07-29T14:30:00Z"); // accepted
schema.parse("2026-07-29T16:30:00+02:00"); // rejected
```

JSON Schema defines `date-time` using RFC 3339, where both `Z` and numeric offsets are valid. A local timestamp without an offset should remain invalid.

The mapping can preserve that behavior by using `z.iso.datetime({ offset: true })`. Authored JSON Schema constraints such as `pattern` should continue to compose with the format check.

nandinitiw:
Closes #6296

## Problem

`fromJSONSchema({ type: "string", format: "date-time" })` maps to `z.iso.datetime()`, which only accepts the `Z` suffix by default:

```ts
const schema = z.fromJSONSchema({ type: "string", format: "date-time" });

schema.parse("2026-07-29T14:30:00Z");        // accepted
schema.parse("2026-07-29T16:30:00+02:00");   // rejected — but should be valid
```

JSON Schema's `date-time` format is defined against the full RFC 3339 `date-time` grammar, which permits both `Z` and numeric UTC offsets. The generated Zod schema is therefore stricter than the source JSON Schema.

## Fix

Pass `offset: true` to `z.iso.datetime()` in `packages/zod/src/v4/classic/from-json-schema.ts`. This accepts both `Z` and numeric offsets while still correctly rejecting a bare local timestamp with no offset at all (`z.iso.datetime`'s `local` option, which would allow that, is intentionally left `false`).

## Testing

Added a test in `from-json-schema.test.ts` covering `Z`, a positive offset, a negative offset, and confirming a local timestamp with no offset still throws.

Full test suite (`vitest run`) passes: 3813/3813.

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -271,7 +271,10 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
         } else if (format === "uuid" || format === "guid") {
           stringSchema = stringSchema.check(z.uuid());
         } else if (format === "date-time") {
-          stringSchema = stringSchema.check(z.iso.datetime());
+          // JSON Schema's "date-time" format is RFC 3339 date-time, which allows both "Z"
+          // and numeric UTC offsets (e.g. "+02:00"). z.iso.datetime() only accepts "Z" by
+          // default, so pass offset: true to accept the full RFC 3339 grammar.
+          stringSchema = stringSchema.check(z.iso.datetime({ offset: true }));
         } else if (format === "date") {
           stringSchema = stringSchema.check(z.iso.date());
         } else if (format === "time") {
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing in favor of #6298. Same fix, but its test also pins that an authored `pattern` still composes with the format check and that a schema round-tripping through `toJSONSchema` keeps accepting offsets. Nothing wrong with this one — it just arrived third of four. Thanks for looking into it 👍

PR #6322, opened 2026-08-04
fix(v4): correctly default on .default().transform() inside partial obj

### LINKED ISSUE #6321: Breaking behavior for .default(...).transform(...) chain when used in a partial object
Given the next snippet:

```
import { z } from "zod/v4";

const arrayFromString = z
  .string()
  .default("")
  .transform((val) => (val ? val.split(",").map((s) => s.trim()) : []));

const partialObj = z.object({ array: arrayFromString }).partial();

console.log(
  arrayFromString.safeParse(undefined).data,
  partialObj.safeParse({}).data?.array,
);
```

4.3.6 logged `[] []` while 4.4.3 logs `[] undefined`

JuanoD:
Fixes #6321

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -2099,12 +2099,10 @@ export const ZodTransform: core.$constructor<ZodTransform> = /*@__PURE__*/ core.
       if (output instanceof Promise) {
         return output.then((output) => {
           payload.value = output;
-          payload.fallback = true;
           return payload;
         });
       }
       payload.value = output;
-      payload.fallback = true;
       return payload;
     };
   }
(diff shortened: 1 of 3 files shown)


colinhacks:
Going with #6329 here. Both patches fix the reported chain, but they take different routes and only one is a no-op everywhere else.

The root cause is that `$ZodTransform` marks `payload.fallback` on every invocation, and `handleOptionalResult` gates on the *optional's* input rather than the *transform's* — so a `.default()` in between opens the gate on a value that was never a fallback. #6329 fixes that at the producer. This one moves the marking into `$ZodOptional` instead, and because `$ZodUnion` runs its options against fresh payloads that don't carry the flag, it changes union behavior as a side effect: `z.union([z.preprocess((v) => v ?? "X", z.string()), z.number()]).optional().parse(undefined)` returns `"X"` on this branch and `undefined` on main.

Your test cases are better than the ones on #6329 though — I'll pull the `.prefault()` and explicit-`undefined` coverage across. Thanks for the report and the fix!

PR #6324, opened 2026-08-05
fix(v4): fromJSONSchema propertyNames enum keys should not be required

### LINKED ISSUE #6323: fromJSONSchema: propertyNames with enum is treated as required keys
### Summary

`z.fromJSONSchema()` converts an object schema that constrains its keys with
`propertyNames: { enum: [...] }` into a Zod schema where each enum value becomes a
**required** property. `propertyNames` restricts which key *names are allowed*; it
does not make any key *required*. As a result, a valid object that contains only a
subset of the allowed keys is incorrectly rejected.

### Zod version

`zod@4.4.3` (Node 20)

### Reproduction

```js
import { z } from "zod";

const schema = {
  type: "object",
  propertyNames: { enum: ["a", "b"] },
  additionalProperties: { type: "string" },
};

const zodSchema = z.fromJSONSchema(schema);

// `{ a: "x" }` is valid against the JSON Schema (an allowed key, string value),
// but Zod rejects it:
zodSchema.parse({ a: "x" });
// ✖ ZodError: [{ path: ["b"], code: "invalid_type",
//     message: "Invalid input: expected string, received undefined" }]
```

### What's happening

Round-tripping the converted schema back through `z.toJSONSchema()` shows the enum
property names were turned into `required`:

```jsonc
{
  "type": "object",
  "propertyNames": { "type": "string", "enum": ["a", "b"] },
  "additionalProperties": { "type": "string" },
  "required": ["a", "b"]   // ← injected; not present in the source schema
}
```

### Expected

- `{ a: "x" }` and `{ b: "y" }` should pass (as should `{}`, unless `minProperties`
  or an explicit `required` say otherwise).
- Only keys *not* in `["a", "b"]` should fail (via `propertyNames`).
- No property should be required unless it appears in the schema's own `required`.

### Notes

This looks like the object-keyword sibling of #6200 (`fromJSONSchema` requiring
`prefixItems` tuple elements when `minItems` is omitted) — both over-require
members that the source schema leaves optional.

Aswanth9496:
Fixes colinhacks#6323

`propertyNames` restricts which key names are allowed on an object — it does not
make any of them required. `fromJSONSchema` previously mapped `propertyNames` to
`z.record()` (no `properties`) or `z.looseRecord()` (with `properties`), and both
are exhaustive over enum key schemas, so every enum member became a required
property.

Fix:
- Case A (no `properties`): use `z.partialRecord()` instead of `z.record()`.
- Case B (with `properties`): needs a record that's both partial (non-required
  enum keys) and loose (unmatched keys still pass through to preserve existing
  `additionalProperties` behavior) — neither helper alone provides both, so it's
  constructed directly.

Verified:
- `{}`, `{a:"x"}`, `{a:"x",b:"y"}` now all pass; `{c:"z"}` still correctly fails
  with `invalid_key`
- `toJSONSchema` round-trip no longer injects a spurious `required` array
- Full existing test suite for `from-json-schema` / `to-json-schema` /
  `to-json-schema-methods` passes (336 tests), no regressions

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -391,7 +391,7 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
             ? convertSchema(schema.additionalProperties as JSONSchema.JSONSchema, ctx)
             : z.any();
 
-       // Case A: No properties (pure record)
+        // Case A: No properties (pure record)
         if (Object.keys(shape).length === 0) {
           zodSchema = z.partialRecord(keySchema, valueSchema);
           break;
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -409,7 +409,8 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
         });
         zodSchema = z.intersection(objectSchema, recordSchema);
         break;
-        
+      }
+      
       // Handle patternProperties
       if (schema.patternProperties) {
         // patternProperties: keys matching pattern must satisfy corresponding schema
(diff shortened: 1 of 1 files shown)


colinhacks:
Closing in favor of #6327, which fixes the same root cause in one line with a test and green CI.

The `propertyNames`-with-`properties` branch here needs `_zod.values` mutated on a cloned key schema to defeat exhaustiveness, and it leaves that branch accepting keys the enum forbids — `{ a: "x", c: "z" }` passes. Swapping `z.looseRecord` for `z.partialRecord` gets the same result in one line and keeps the key check. Thanks for digging into this 👍

PR #6325, opened 2026-08-05
fix: support format hostname in fromJSONSchema

### LINKED ISSUE #6304: fromJSONSchema doesn't support format:hostname (but could)

deepshekhardas:
Supports JSON Schema format "hostname" in fromJSONSchema by applying z.hostname(). Fixes #6304.

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -264,7 +264,9 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
       if (schema.format) {
         const format = schema.format;
         // Map common formats to Zod check functions
-        if (format === "email") {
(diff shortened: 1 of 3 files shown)


colinhacks:
Closing in favor of #6305, which makes the same `hostname` change with a tighter diff. This PR also bundles an unrelated bigint-default fix for #6209, which is its own decision and already has two PRs open against it.

Happy to look at the bigint half separately if you split it out — thanks for looking into this 👍

PR #6326, opened 2026-08-05
fix(mini): correct merge signature to accept a plain shape

### LINKED ISSUE #6302: `zod/mini`: `merge()` type and runtime signatures are inverted
`z.mini.merge()` declares a schema-to-schema signature but forwards to `util.extend()`, which requires a plain shape. The call that typechecks throws at runtime, and the call that runs is a type error. There's no way to call it that satisfies both.

```ts
import * as m from "zod/mini";

const A = m.object({ a: m.string() });
const B = m.object({ b: m.string() });

m.merge(A, B);                  // typechecks; throws at runtime
m.merge(A, { b: m.string() });  // works at runtime; type error TS2353
```

On `zod@4.4.3`:

```
merge(schema, schema) THROWS: Invalid input to extend: expected a plain object
merge(schema, shape) : [ 'a', 'b' ]
```

Under `tsc --noEmit --strict` (5.9.3 and 7.0.2 agree):

```
error TS2353: Object literal may only specify known properties,
  and 'b' does not exist in type 'ZodMiniObject<...>'
```

## Cause

`mini/schemas.ts:921` declares only the schema-to-schema overload, so the implementation signature isn't publicly callable:

```ts
/** @deprecated Identical to `z.extend(A, B)` */
export function merge<T extends ZodMiniObject, U extends ZodMiniObject>(
  a: T,
  b: U
): ZodMiniObject<util.Extend<T["shape"], U["shape"]>, T["_zod"]["config"]>;
export function merge(schema: ZodMiniObject, shape: any): ZodMiniObject {
  return util.extend(schema, shape);
}
```

`util.extend` guards on `isPlainObject` and throws (`core/util.ts:659`).

Mini only. Classic calls `util.merge` (`classic/schemas.ts:1555`) and `A.merge(B)` works. Present since the Zod 4 commit 85928549, unchanged in 4.4.3. `mini/tests/object.test.ts` has no `merge` case.

## Two fixes, not equivalent

Correcting the overload to `(schema, shape)` matches the `@deprecated Identical to z.extend(A, B)` note above it, with no runtime change. It breaks `merge(A, B)` at compile time, but that call already throws, so the failure just surfaces earlier.

Switching the body to `util.merge` makes the advertised signature true, matches classic and Zod 3, and picks up `b`'s `catchall`. It breaks JS callers passing a plain shape, which is the form that works today. Not silently: `checks: b._zod.def.checks ?? []` (`core/util.ts:713`) is eager, so it throws `TypeError: Cannot read properties of undefined (reading 'def')` at the call. Still a runtime change to a shipped API.

`merge` is deprecated and absent from `api.mdx`, so the first is the conservative read. Which would you prefer? Happy to open a PR with a test either way.

deepshekhardas:
The deprecated merge() overload only accepted schema-to-schema calls, which always threw at runtime since it forwards to util.extend(). Correct the public overload to accept a plain shape, matching the behavior and the Identical to z.extend note. Fixes #6302.

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -264,7 +264,9 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
       if (schema.format) {
         const format = schema.format;
         // Map common formats to Zod check functions
-        if (format === "email") {
+        if (format === "hostname") {
+          stringSchema = stringSchema.check(z.hostname());
+        } else if (format === "email") {
           stringSchema = stringSchema.check(z.email());
         } else if (format === "uri" || format === "uri-reference") {
           stringSchema = stringSchema.check(z.url());
--- a/packages/zod/src/v4/classic/tests/from-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/from-json-schema.test.ts
@@ -448,6 +448,15 @@ test("string format - email", () => {
   expect(schema.parse("test@example.com")).toBe("test@example.com");
 });
 
+test("string format - hostname", () => {
+  const schema = fromJSONSchema({
+    type: "string",
+    format: "hostname",
+  });
+  expect(schema.parse("example.com")).toBe("example.com");
+  expect(() => schema.parse("not a hostname!")).toThrow();
+});
+
 test("string format - uuid", () => {
   const schema = fromJSONSchema({
     type: "string",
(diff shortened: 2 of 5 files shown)


colinhacks:
Closing in favor of #6303. Same fix, but it reproduces `extend`'s return type exactly — this one drops the `util.Writeable<U>`, so `merge` and `extend` infer different shapes for a `const`-asserted object, which is the one thing the "identical to `z.extend`" note rules out. The branch also carries the BigInt and hostname fixes, so merging it for #6302 would land three unrelated changes at once.

Happy to take those two as separate PRs — thanks for looking into this 👍

PR #6327, opened 2026-08-05
fix(fromJSONSchema): propertyNames enum keys should not be required

### LINKED ISSUE #6323: fromJSONSchema: propertyNames with enum is treated as required keys

deepshekhardas:
z.fromJSONSchema() converted an object schema with propertyNames: { enum: [...] } into a Zod record that required every enum value as a key. propertyNames only restricts which key names are allowed, so a subset of keys must be valid. Switch the pure-record case to z.partialRecord so the key schema constrains names without requiring them. Fixes #6323.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,10 +16,11 @@ jobs:
   test-node:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest]
         node: ["lts/*"]
-        typescript: ["5.5", "latest"]
+        typescript: ["5.5", "6", "latest"]
     name: Test with TypeScript ${{ matrix.typescript }} on Node ${{ matrix.node }} (${{ matrix.os }})
     steps:
       - uses: actions/checkout@v4
@@ -28,10 +29,14 @@ jobs:
           node-version: ${{ matrix.node }}
       - uses: pnpm/action-setup@v4
       - run: pnpm install
-      - run: pnpm add typescript@${{ matrix.typescript }} -w
+      # Build with the repo-pinned TypeScript: zshy uses the classic compiler API,
+      # which TypeScript 7 no longer ships. The matrix version applies to the tests.
       - run: pnpm build
+      - run: pnpm add typescript@${{ matrix.typescript }} -w
       - run: pnpm test
+      # test-resolution.ts shells out to zshy, so it can't run under TypeScript 7.
       - run: pnpm run --filter @zod/resolution test:all
+        if: matrix.typescript != 'latest'
       - run: pnpm run --filter @zod/integration test:all
 
   lint:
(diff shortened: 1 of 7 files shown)


colinhacks:
Pushed two commits on top of this.

The `partialRecord` swap was right, but it only covered one of the two `propertyNames` branches — with `properties` present, `z.looseRecord` is exhaustive over an enum key schema too, so every enum member was still required. Same one-line swap there fixes it, and it also restores the key check: `looseRecord` let keys outside the enum pass through, so `{ a: "x", c: "z" }` was accepted against `propertyNames: { enum: ["a", "b"] }`. The exhaustiveness bug was masking that — out-of-enum input happened to fail for the wrong reason. Added coverage for that branch and for an explicit `required` array alongside `propertyNames`.

I also reverted the `hostname` format, BigInt `default`, and mini `merge` commits off this branch. All three look correct — open them as their own PRs against #6304, #6209, and #6302 and I'll take them.

colinhacks:
Landed in #6411 — your commits are in that branch, so this fix is on `main` as of 6074828e. Closing this out since the squash doesn't auto-close it.

The `partialRecord` swap turned out to be the shallow end of it: `propertyNames` was modelled as a record at all, which short-circuited the object branch and silently dropped `patternProperties` and `additionalProperties`. #6411 makes it a key guard so all four keywords compose.

Your `merge()` fix landed separately in #6404. The `hostname` format (#6304) and BigInt `default` (#6209) ones are still open if you want to reopen those as their own PRs — both looked right. Thanks for the fix 👍

PR #6328, opened 2026-08-05
fix(codec): decode may return unknown, validated by the out schema

### LINKED ISSUE #6169: `decode` method for creating new codecs should expect `unknown` as the return type
Take the built in example https://zod.dev/codecs?id=jsonschema#jsonschema. The only reason this type checks is because `JSON.parse`'s return type is an implicit-any. The refinement is then done by the provided schema. If the return type was instead `unknown` (which is arguably a better return type since it is safer) then typescript would fail with errors that `unknown` is not assignable to `T` (given `<T extends z.core.$ZodType>(schema: T)`).

I posit therefore, that `decode` really should return `unknown` since it is up to the next schema in the codec chain to verify the type, not the decode step.

Yes I know that the work around is to just do a `as any` cast, but it feels wrong to require that.

deepshekhardas:
The codec decode function is currently typed to return core.input<B>, forcing authors to cast through ny (e.g. JSON.parse). Since the out schema validates the decoded value anyway, type the return as unknown. This matches the codec intent and removes the need for unsafe casts. Fixes #6169.

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -264,7 +264,9 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
       if (schema.format) {
         const format = schema.format;
         // Map common formats to Zod check functions
-        if (format === "email") {
+        if (format === "hostname") {
+          stringSchema = stringSchema.check(z.hostname());
+        } else if (format === "email") {
           stringSchema = stringSchema.check(z.email());
         } else if (format === "uri" || format === "uri-reference") {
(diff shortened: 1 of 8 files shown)


colinhacks:
Closing in favor of #6170. It makes the same change but also covers Mini's `codec()` and the underlying `$ZodCodecDef`/`$ZodPipeDef` types, and its type-level test actually pins the new behavior — the test here passes on `main` unchanged, since `JSON.parse` already returns `any`.

The other changes bundled in (JSON Schema `hostname`, `partialRecord`, bigint defaults, the Mini `merge` signature) are unrelated to #6169 and would need to be their own PRs. Thanks for looking into this 👍

PR #6329, opened 2026-08-05
fix(v4): transform must not mark fallback when input came from a default

### LINKED ISSUE #6321: Breaking behavior for .default(...).transform(...) chain when used in a partial object

deepshekhardas:
$ZodTransform (and classic ZodTransform) unconditionally set payload.fallback = true, so when a .default().transform() chain sits inside a .partial() object, the outer $ZodOptional clobbers the transformed value back to undefined. Only mark fallback when the transform input was undefined (no default/catch supplied a value upstream). Fixes #6321.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -2198,16 +2198,17 @@ export const ZodTransform: core.$constructor<ZodTransform> = /*@__PURE__*/ core.
         }
       };
 
+      const input = payload.value;
       const output = def.transform(payload.value, payload);
       if (output instanceof Promise) {
         return output.then((output) => {
           payload.value = output;
-          payload.fallback = true;
+          if (input === undefined) payload.fallback = true;
           return payload;
         });
       }
       payload.value = output;
-      payload.fallback = true;
+      if (input === undefined) payload.fallback = true;
       return payload;
     };
   }
--- a/packages/zod/src/v4/classic/tests/default.test.ts
+++ b/packages/zod/src/v4/classic/tests/default.test.ts
@@ -312,6 +312,69 @@ test("partial should not clobber defaults", () => {
   `);
 });
 
+// A transform downstream of a default/prefault ran on a deliberate value, so an outer
+// optional must not clobber its output. Regression: #6321 (broke in 4.4.3 via #5941).
+test("optional respects a transform fed by default/prefault", () => {
+  const fromDefault = z
+    .string()
+    .default("")
+    .transform((v) => (v ? v.split(",") : []));
+  const fromPrefault = z
+    .string()
+    .prefault("hi")
+    .transform((v) => v.length);
+
+  expect(fromDefault.optional().parse(undefined)).toEqual([]);
+  expect(fromPrefault.optional().parse(undefined)).toBe(2);
+
+  const obj = z.object({ a: fromDefault }).partial();
+  expect(obj.parse({})).toEqual({ a: [] });
+  expect(obj.parse({ a: undefined })).toEqual({ a: [] });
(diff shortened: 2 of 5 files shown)


colinhacks:
This is the right fix. `$ZodTransform` marking `payload.fallback` unconditionally is the actual root cause — `handleOptionalResult` gates on the *optional's* input, not the *transform's*, so a `.default()` in between opens the gate on a value that was never a fallback. Gating on the transform's own input is the minimal correction, and it's a no-op everywhere else. Taking this over #6322, which moved the marking into `$ZodOptional` and changed union behavior as a side effect.

Before it can merge, the branch needs to come down to just `0aa9722c`. It's currently carrying five unrelated fixes (#6304, #6209, #6302, #6323, #6169), a merge of #6345, and a deletion of `play.ts`. The codec change is the one I'd flag hardest — loosening `decode`'s return to `unknown` and dropping the `@ts-expect-error` that pinned it is a real public type change and needs its own PR. `codec.test.ts` has also picked up some mojibake where a `→` got mangled into replacement bytes.

Rebase down to the transform commit and its test and I'll merge. The other five are welcome as separate PRs. I'll handle the `fallback` JSDoc and `wiki/optionality.md`, which both still say the flag is set on every transform invocation.

colinhacks:
Rebased this onto main and cut it down to the #6321 fix. It was carrying five unrelated fixes, a merge of #6345, and a deletion of `play.ts` — that's where the conflicts and most of the file churn were coming from. Your fix commit is unchanged and still authored by you; the diff is 5 files now instead of 10.

That also took care of the mojibake. `codec.test.ts` had `// Forward: ISO string <?> Date` where the `→` had been mangled into replacement bytes — it came in with the codec commit, so dropping that commit restored the file to match main and the arrows are back to `→`.

The second commit on top is mine. The `fallback` JSDoc and `wiki/optionality.md` both still said the flag is set on *every* `$ZodTransform` invocation, which this fix makes false, and one line in that doc is the reasoning that caused the bug in the first place. It also adds coverage your test didn't have: prefault, the async transform path, explicit-`undefined` keys, and the union case — `$ZodUnion` runs its options against fresh payloads, so the flag has to be set by the transform itself to survive the trip out to the optional. Merging once CI is green.

The other five are worth landing on their own. Open them as separate PRs and I'll take them: `880c5ad7` (#6304), `14a22247` (#6209), `d98f1da1` (#6302), `f7708e52` (#6323), `b70a930d` (#6169). The codec one needs its own thread — loosening `decode`'s return to `unknown` and dropping the `@ts-expect-error` that pinned it is a real public type change.

colinhacks:
Closing in favor of #6419. Your diagnosis was the right one — `$ZodTransform` marking `payload.fallback` unconditionally is the root cause, and the fact that `handleOptionalResult` gates on the *optional's* input rather than the *transform's* is exactly why a `.default()` in between got its value discarded.

#6419 lands the same fix one layer down. `optin` becomes a three-rung ladder — `undefined` / `"optional"` / `"defaulted"` — where default and prefault declare the top rung and `$ZodOptional` runs its inner only when something actually answers the absence. The payload flag is removed entirely and #6321 falls out without inspecting the transform's input at all. It's behaviorally identical to this PR across a 456-row combinator matrix; the difference is that it removes the wrong conclusion rather than undoing it.

You're credited on the commit. Thanks for the investigation — the structural fix was findable because you named the mechanism correctly.

PR #6331, opened 2026-08-05
feat(enum): add extend method to ZodEnum

### LINKED ISSUE #6276: Ability to extend an enum
I'm working on a model with the following use case:

```ts
const ZodBaseDirections = z.enum(['north', 'south', 'east', 'west']);
type BaseDirections = z.infer<typeof ZodBaseDirections>;
const BaseDirections = ZodBaseDirections.enum;
// Result: BaseDirections = { north: 'north', south: 'south', east: 'east', west: 'west' };

// Extension 1
const ZodEightDirections = z.union([
  ZodBaseDirections,
  z.enum(['northwest', 'northeast', 'southwest', 'southeast']),
]);
type EightDirections = z.infer<typeof ZodEightDirections>;
const EightDirections = ZodEightDirections.enum;
// Expected Result: EightDirections = { north: 'north', south: 'south', east: 'east', west: 'west', northwest: 'northwest', northeast: 'northeast', southwest: 'southwest', southeast: 'southeast' };
// Actual result: Error - EightDirections is a ZodUnion, not a ZodEnum, and has no `enum` property

// Extension 2
const ZodThreeDimDirections = z.union([
  ZodBaseDirections,
  z.enum(['front', 'back']),
]);
type ThreeDimDirections = z.infer<typeof ZodThreeDimDirections>;
const ThreeDimDirections = ZodThreeDimDirections.enum; 
// Expected Result: ThreeDimDirections = { north: 'north', south: 'south', east: 'east', west: 'west', front: 'front', back: 'back' };
// Actual result: Error - ThreeDimDirections is a ZodUnion, not a ZodEnum, and has no `enum` property
```

I have a base enum that is reused in many places but I want to be able to extend it for certain ones, and I also want to expose a TS typing and record for. Right now, the recommended (and, as far as I can tell, only) way to expose a `z.enum` is with a `z.union`, but doing so changes the result from a `ZodEnum` to a `ZodUnion` which doesn't have the `enum` property for creating the record.

I'd like to put forward a request for a method on enums like `.union` or `.extend` that explicitly add values to the enum itself rather than relying on a `ZodUnion` wrapper. Perhaps it can be implemented as something like a `ZodUnionEnum` that is a special case for `ZodUnion` which is compatible with `ZodEnum` and does the necessary merging logic.

deepshekhardas:
Adds ZodEnum.prototype.extend(), mirroring extract/exclude. Users who need to reuse a base enum with additional values currently must fall back to z.union, losing the enum property they use to build a record. Fixes #6276.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -2035,6 +2035,10 @@ export interface ZodEnum<
     values: U,
     params?: string | core.$ZodEnumParams
   ): ZodEnum<util.Flatten<Omit<T, U[number]>>>;
+  extend<const U extends readonly string[]>(
+    values: U,
+    params?: string | core.$ZodEnumParams
+  ): ZodEnum<util.Flatten<T & Record<U[number], U[number]>>>;
 }
 export const ZodEnum: core.$constructor<ZodEnum> = /*@__PURE__*/ core.$constructor("ZodEnum", (inst, def) => {
   core.$ZodEnum.init(inst, def);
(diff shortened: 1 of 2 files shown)


colinhacks:
Going to close this. `.options` already exposes everything needed to build a wider enum, so it's a one-liner:

```ts
const Directions = z.enum(["north", "south", "east", "west"]);
const EightDirections = z.enum([...Directions.options, "northwest", "northeast"]);
```

That infers as `"north" | "south" | "east" | "west" | "northwest" | "northeast"` and gives you the `.enum` object back. Merging two enums is the same shape: `z.enum([...A.options, ...B.options])`.

`extract`/`exclude` earn their place because narrowing isn't expressible that way — `.filter()` drops the literal types. Widening is, so it doesn't need a method.

Thanks for the PR 👍

PR #6332, opened 2026-08-05
perf(toJSONSchema): avoid rebuilding reversed seen array in finalize

### LINKED ISSUE #6161: toJSONSchema(schemaRegistry, opts) is very slow with a large registry
As I add more types to a schema registry, calling `toJSONSchema` on that registry gets very slow, growing much faster than the growth in the size of the registry.

In a private repo I have, a registry with only about 3000 types in it takes 30 seconds to process on a MacBook M3.

I've collected some perf dumps, and the problem might be coming from one or both of these places:

### Finalize iteration

`finalize` is called on every schema, but for each schema it goes through all of the refs and checks each one:

https://github.com/colinhacks/zod/blob/main/packages/zod/src/v4/core/to-json-schema.ts#L449

A lot of that ends up being noop, as everything only needs to be procesesd once.  But it's still building all of those new arrays for iteration and revisiting every member of `ctx.seen` many times over.

When I hack my copy of zod to only run this loop once per context, the schema generation time for my example goes down from about 30s to about 18s.

### Nested parent computation

When a schema has a parent, registry get walks the chain recursively to get ancestor data:

https://github.com/colinhacks/zod/blob/main/packages/zod/src/v4/core/registries.ts#L65

This data can be recomputed many times as registry get is called many times, even though no registry data is written during the process.

When I add a subclass of $ZodRegistry that can cache the results of these get calls, the schema generation time for my example goes down from about 30s to about 20s.

deepshekhardas:
inalize rebuilt a reversed array copy of ctx.seen on every call, which gets slow with large registries. Iterate in reverse over a single array instead. Fixes #6161.

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -264,7 +264,9 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
       if (schema.format) {
         const format = schema.format;
         // Map common formats to Zod check functions
-        if (format === "email") {
+        if (format === "hostname") {
+          stringSchema = stringSchema.check(z.hostname());
+        } else if (format === "email") {
           stringSchema = stringSchema.check(z.email());
         } else if (format === "uri" || format === "uri-reference") {
           stringSchema = stringSchema.check(z.url());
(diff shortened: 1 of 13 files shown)


colinhacks:
Closing in favor of #6162, which rewrites the same `finalize` loop more thoroughly — it tracks which schemas still need flattening instead of rebuilding the array on every call. I measured both on a 1000-schema registry: this PR's change is ~1.2x over main, #6162 is ~2x.

The branch also carries a lot of unrelated work — `ZodEnum.extend`, the transform `fallback` change, the codec `decode` loosening, and tuple `items: false`, which changes JSON Schema output for every tuple. That one deserves its own PR rather than riding along on a perf fix. Thanks for looking into this 👍

PR #6334, opened 2026-08-05
fix(fromJSONSchema): enforce additionalProperties:false when patternProperties is present

### LINKED ISSUE #6198: fromJSONSchema ignores additionalProperties:false when patternProperties is present

deepshekhardas:
When a JSON Schema object combines \patternProperties\ with \dditionalProperties: false\, the conversion returned early from the \patternProperties\ branch and never applied the strict-key handling. Keys matching neither \properties\ nor any pattern were accepted.

Now an extra strict record keyed on the union of the literal property names and the pattern regexes is intersected in, so \dditionalProperties: false\ rejects non-matching keys. Fixes #6198.

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -429,9 +429,7 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
         // property or one of the patterns. Enforce this with a strict record
         // keyed on the union of the patterns.
         if (schema.additionalProperties === false) {
-          const allowedKeySchemas: ZodType[] = Object.keys(shape).map((key) =>
-            z.literal(key),
-          );
+          const allowedKeySchemas: ZodType[] = Object.keys(shape).map((key) => z.literal(key));
           for (const pattern of patternKeys) {
             allowedKeySchemas.push(z.string().regex(new RegExp(pattern)));
           }
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -425,6 +425,23 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
         }
         schemasToIntersect.push(...looseRecords);
 
+        // When additionalProperties is false, every key must match a known
+        // property or one of the patterns. Enforce this with a strict record
+        // keyed on the union of the patterns.
+        if (schema.additionalProperties === false) {
+          const allowedKeySchemas: ZodType[] = Object.keys(shape).map((key) =>
+            z.literal(key),
+          );
+          for (const pattern of patternKeys) {
+            allowedKeySchemas.push(z.string().regex(new RegExp(pattern)));
+          }
+          const keySchema =
+            allowedKeySchemas.length === 1
+              ? (allowedKeySchemas[0] as ZodString)
+              : z.union(allowedKeySchemas as [ZodString, ZodString, ...ZodString[]]);
+          schemasToIntersect.push(z.record(keySchema, z.any()));
+        }
+
         if (schemasToIntersect.length === 0) {
           zodSchema = z.object({}).passthrough();
         } else if (schemasToIntersect.length === 1) {
--- a/packages/zod/src/v4/classic/tests/from-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/from-json-schema.test.ts
@@ -387,6 +387,34 @@ test("multiple overlapping patternProperties", () => {
   expect(() => schema.parse({ S_N: "ab" })).toThrow(); // too short for ^S_N pattern
 });
 
+test("patternProperties with additionalProperties: false rejects non-matching keys", () => {
+  const schema = fromJSONSchema({
+    type: "object",
+    patternProperties: {
+      "^S_": { type: "string" },
+    },
+    additionalProperties: false,
+  });
+  expect(schema.safeParse({ S_name: "ok" }).success).toBe(true);
+  expect(schema.safeParse({ X_A: false }).success).toBe(false);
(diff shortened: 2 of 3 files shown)


wahajahmed010 (commented):
# Review of #6334

The fix correctly targets the gap in the `patternProperties` branch — the existing code only intersects `looseRecords` (which are pass-through and do not enforce `additionalProperties: false`), so a non-matching key slips through when both keywords are combined. Adding a `z.record(keySchema, z.any())` intersection in front of those is the right shape, and the two new tests cover the cases that matter (rejects non-matching keys; allows normal `properties` alongside `patternProperties`).

A few things worth tightening before this lands:

**1. `z.string().regex(new RegExp(pattern))` is not JSON-Schema-anchored.**

JSON Schema's `pattern` keyword is implicitly anchored to the full string (effectively `^...$`). Zod's `.regex()` uses `pattern.test(value)` without anchoring, so a user-supplied pattern like `S_` would match `MY_S_KEY` here, which is wrong against the spec. The same pattern is used in the existing line above (`packages/zod/src/v4/classic/from-json-schema.ts:418` — `const keySchema = z.string().regex(new RegExp(pattern));`), so the pre-existing branch has the same problem, but this PR perpetuates it on a new code path.

Suggested fix: wrap the pattern before compiling:
```ts
const anchored = pattern.startsWith("^") || pattern.endsWith("$")
  ? new RegExp(pattern)
  : new RegExp(`^(?:${pattern})$`);
const keySchema = z.string().regex(anchored);
```
This matches the semantics JSON Schema consumers expect. Worth doing in both the new code and the existing line above in this same PR, or at least flagging in a TODO for a follow-up.

**2. The `length === 1` cast is wrong for the all-literal case.**

```ts
const keySchema =
  allowedKeySchemas.length === 1
    ? (allowedKeySchemas[0] as ZodString)
    : z.union(allowedKeySchemas as [ZodString, ZodString, ...ZodString[]]);
```
When `patternKeys` is empty and `properties` has exactly one key, `allowedKeySchemas[0]` is a `ZodLiteral`, not a `ZodString`. The `as ZodString` cast lies to the type system. The runtime behaviour happens to be fine because `z.literal` is assignable wherever `ZodString` is expected for `z.record` purposes, but the cast is misleading. Drop the conditional or change the cast target to `ZodType`. Minor, but the type lies.

**3. Unrelated change bundled in.**

`json-schema-processors.ts` adds a `serializeDefault` helper to handle BigInt defaults via the `defaultProcessor`, `prefaultProcessor`, and `catchProcessor`. This is unrelated to the PR title ("fix(fromJSONSchema): enforce additionalProperties:false when patternProperties is present") and unrelated to the branch name `fix/pr-6198-patternProps-additionalProps`. Worth splitting into its own PR — BigInt serialization is a separate concern, the changelog will read more cleanly, and bisection will be easier if either fix needs to be reverted independently.

**4. Missing test: `patternProperties` + `additionalProperties: false` with no `properties`.**

The new test only covers the case where `properties` exists. The purely-patternProperties case (`{ type: "object", patternProperties: { "^S_": ... }, additionalProperties: false }`) hits a different path through the conversion and is worth a regression test. This is the exact shape that triggered #6198, so it should be locked in.

**5. Side note — `Record<string, never>` and the `z.record(z.string(), z.unknown())` shape is fine, but consider `z.strictObject` for the case where `properties` is set.**

The existing `additionalProperties: false` branch already routes to `objectSchema.strict()`. The new code could reuse that same primitive for the literal-keys part of the keySchema (i.e. `z.strictObject(shape)` instead of intersecting `z.object(shape).passthrough()` with a `z.record`). It would compose better with downstream `.strict()` calls. Not a blocker.

Items 1, 3, and 4 are blockers in my view. 2 and 5 are polish.

colinhacks:
Closing in favor of #6199. This reports `invalid_key`, but the plain `additionalProperties: false` path reports `unrecognized_keys` — same JSON Schema keyword, two different errors depending on whether `patternProperties` happens to be present. The BigInt default commit is also unrelated to this fix and already has its own PR in #6255. Thanks for looking into this 👍

PR #6335, opened 2026-08-05
fix(fromJSONSchema): accept RFC 3339 numeric offsets in date-time format

### LINKED ISSUE #6296: fromJSONSchema date-time rejects valid RFC 3339 offsets

deepshekhardas:
JSON Schema defines \date-time\ via RFC 3339, where both \Z\ and numeric offsets such as \+02:00\ are valid. The conversion to \z.iso.datetime()\ (default \offset: false\) rejected valid numeric-offset timestamps.

Map \ormat: 'date-time'\ to \z.iso.datetime({ offset: true })\, which accepts both \Z\ and numeric offsets while still rejecting offset-less local times. Fixes #6296.

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -271,7 +271,7 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
         } else if (format === "uuid" || format === "guid") {
           stringSchema = stringSchema.check(z.uuid());
         } else if (format === "date-time") {
-          stringSchema = stringSchema.check(z.iso.datetime());
+          stringSchema = stringSchema.check(z.iso.datetime({ offset: true }));
         } else if (format === "date") {
           stringSchema = stringSchema.check(z.iso.date());
         } else if (format === "time") {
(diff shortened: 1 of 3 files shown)


colinhacks:
Closing in favor of #6298, which has broader test coverage for the `date-time` change. Separately, the `serializeDefault` change to `json-schema-processors.ts` is unrelated to this issue and ships without a test — if you want to pursue the bigint default handling, open it as its own PR and I'll look at it there. Thanks 👍

PR #6336, opened 2026-08-05
perf: elide redundant backward parse when checks cannot mutate the value

### LINKED ISSUE #6158: encode() runs user codec encode functions 2–N times per call when ancestors have checks
**Repro** (zod 4.4.3, also present on `main`):

```ts
import * as z from "zod";

let calls = 0;
const codec = z.codec(z.iso.datetime(), z.date(), {
  decode: (iso) => new Date(iso),
  encode: (date) => {
    calls++;
    return date.toISOString();
  },
});

z.encode(z.object({ at: codec }), { at: new Date() });
// calls === 1 ✅

calls = 0;
z.encode(z.object({ at: codec }).refine(() => true), { at: new Date() });
// calls === 2 ❌ — one checked ancestor doubles it

calls = 0;
const inner = z.object({ at: codec }).refine(() => true);
z.encode(z.object({ nested: inner }).refine(() => true), { nested: { at: new Date() } });
// calls === 3 ❌ — multiplies with nesting depth

calls = 0;
z.encode(z.object({ items: z.array(z.object({ at: codec })).min(1) }), {
  items: [{ at: new Date() }, { at: new Date() }],
});
// calls === 4 ❌ — twice per element
```

### Cause

In the backward direction, `$ZodType`'s `run` performs a canary parse with `skipChecks: true` (shielding domain-side checks from invalid values), then — after this node's checks — re-parses the value in full (`handleCanaryResult` in `packages/zod/src/v4/core/schemas.ts`). `skipChecks` skips *checks*, but not *transforms*: every codec `encode` and pipe transform in the subtree executes once per pass, so each check-bearing ancestor adds a full extra execution of the subtree below it.

### Impact

- Side-effectful or non-idempotent `encode` functions (metrics, ID allocation, logging) fire 2–N times per `encode()` call.
- Expensive encodes (encryption, compression, big payloads) pay 2x+ wall time whenever any ancestor carries a `.refine()`/`.min()`/etc.
- A related reporting inconsistency: because the canary skips checks and aborts on the first fatal issue, continuable check issues that `parse()` reports are masked in `encode()` for the same schema and value:

```ts
const schema = z.object({ a: z.string().min(5), b: z.number() }).refine(() => true);
z.safeParse(schema, { a: "abc", b: "nan" }).error.issues;
// [too_small at ["a"], invalid_type at ["b"]]
z.safeEncode(schema, { a: "abc", b: "nan" }).error.issues;
// [invalid_type at ["b"]] — too_small masked
```

### Suggested fix

The second parse is only needed when the node's **own checks can mutate the value** — and `$ZodCheckOverwrite` is the designated mutating check family (`trim`, `toLowerCase`, `toUpperCase`, `normalize`, `slugify` all build on it). So: run the canary **with** descendant checks enabled; when the node's own checks contain no overwrite-family check, reuse the canary's value and issues and elide the second parse. Overwrite-bearing nodes keep the current two-pass flow, and abort semantics are unchanged (an invalid value still returns before this node's checks execute).

Working implementation: [`drudolf/zod@encode-canary-reuse`](https://github.com/drudolf/zod/tree/encode-canary-reuse), single commit [`154494f`](https://github.com/drudolf/zod/commit/154494fb54670e315dcc0f83122493c504f5166e). All 3,821 existing tests pass, plus 10 new tests in `codec.test.ts` covering encode-once semantics, descendant-check failures under checked parents (sync/async), issue ordering, overwrite parents, and unidirectional transforms.

Measured: ~1.9x faster encode on check-dense codec schemas (µbench, output verified byte-identical); +4–10% on realistic mixed schemas in fastify serialization benchmarks.

Two deliberate behavioral notes:

1. An aborted canary now also reports continuable descendant check issues — this aligns backward reporting with forward (fixes the masking shown above).
2. A custom check that mutates `payload.value` without being an overwrite check would lose its mutation in the reuse path (previously the re-parse picked it up). `.overwrite()` is the designated mutation mechanism; runtime value-identity detection was rejected as unsound for in-place reference mutation.

Happy to open a PR from the branch if this direction is acceptable.

deepshekhardas:
In the backward direction, checked nodes ran a checkless canary parse and then a second full parse after checks. When a node's checks contain no overwrite-family check, the value cannot change between passes — the canary now runs with descendant checks enabled and its result (value and descendant issues) is reused, eliding the second parse.

Fixes double execution of user codec \encode\ functions under checked parents (previously 2x per request, multiplying with check nesting) and speeds up encode of check-bearing codec schemas ~1.9x. Abort semantics are unchanged; overwrite-bearing nodes keep the original two-pass flow.

One reporting change: continuable descendant check issues previously masked by the checkless canary now accompany aborting issues, matching forward-direction reporting. Fixes #6158.

### DIFF
--- a/packages/zod/src/v4/classic/tests/codec.test.ts
+++ b/packages/zod/src/v4/classic/tests/codec.test.ts
@@ -701,3 +701,126 @@ describe("context immutability", () => {
     expect("direction" in ctx).toBe(false);
   });
 });
+
+describe("encode executes codec transforms exactly once", () => {
+  const makeCountingCodec = () => {
+    let encodeCalls = 0;
+    const codec = z.codec(z.iso.datetime(), z.date(), {
+      decode: (isoString) => new Date(isoString),
+      encode: (date) => {
+        encodeCalls++;
+        return date.toISOString();
+      },
+    });
+    return { codec, count: () => encodeCalls };
+  };
+
+  test("codec under a checked object runs encode once", () => {
+    const { codec, count } = makeCountingCodec();
+    const schema = z.object({ createdAt: codec }).refine(() => true);
+
+    const result = z.safeEncode(schema, { createdAt: new Date("2025-01-01T00:00:00.000Z") });
+
+    expect(result.success).toBe(true);
+    expect(count()).toBe(1);
+  });
+
+  test("codecs in a checked array run encode once per element", () => {
+    const { codec, count } = makeCountingCodec();
+    const schema = z.object({ items: z.array(z.object({ at: codec })).min(1) });
+
+    const result = z.safeEncode(schema, {
+      items: [{ at: new Date("2025-01-01T00:00:00.000Z") }, { at: new Date("2025-06-01T00:00:00.000Z") }],
+    });
+
+    expect(result.success).toBe(true);
+    expect(count()).toBe(2);
+  });
+
+  test("nested checked schemas do not multiply encode executions", () => {
+    const { codec, count } = makeCountingCodec();
+    const inner = z.object({ at: codec }).refine(() => true);
+    const schema = z.object({ nested: inner }).refine(() => true);
+
+    const result = z.safeEncode(schema, { nested: { at: new Date("2025-01-01T00:00:00.000Z") } });
+
+    expect(result.success).toBe(true);
+    expect(count()).toBe(1);
+  });
+
+  test("unidirectional transform under a checked parent still throws on encode", () => {
+    const schema = z.object({ id: z.string().transform((s) => s.length) }).refine(() => true);
+
+    expect(() => z.safeEncode(schema, { id: 42 as never })).toThrow(z.core.$ZodEncodeError);
+  });
+
+  test("checks still run against the domain value during encode", () => {
+    const schema = z.object({ items: z.array(z.iso.datetime()).min(3) });
+
+    const result = z.safeEncode(schema, { items: ["2025-01-01T00:00:00.000Z"] });
+
+    expect(result.success).toBe(false);
+    expect(result.error!.issues[0].code).toBe("too_small");
+  });
+});
+
+describe("encode reports descendant check failures under checked parents", () => {
+  test("failing descendant check under a checked object fails encode", () => {
+    const schema = z.object({ name: z.string().min(5) }).refine(() => true);
+
+    const result = z.safeEncode(schema, { name: "abc" });
+
+    expect(result.success).toBe(false);
+    expect(result.error!.issues.map((i) => ({ code: i.code, path: i.path }))).toEqual([
+      { code: "too_small", path: ["name"] },
+    ]);
+  });
+
+  test("own check issues precede descendant check issues", () => {
+    const schema = z.object({ name: z.string().min(5) }).refine(() => false, { message: "own" });
+
+    const result = z.safeEncode(schema, { name: "abc" });
+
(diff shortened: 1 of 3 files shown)


wahajahmed010:
Solid perf win and the new tests pin the behavior nicely. One structural concern:

**This bundles two unrelated changes.** The diff mixes:

1. The encode canary reuse / backward-parse elision in `core/schemas.ts` (the actual subject of the PR title and description).
2. A new `serializeDefault` helper in `core/json-schema-processors.ts` that handles `BigInt` defaults and replaces three `JSON.parse(JSON.stringify(...))` call sites.

The BigInt fix is real and useful (without it `JSON.stringify` throws on bigint defaults), but it's orthogonal to the encode perf work and touches a different module. Splitting it into a separate PR would make bisects, cherry-picks, and backports cleaner — especially if the BigInt fix is wanted on a stable branch where the encode canary change isn't. Easy refactor: pull the 13 lines in `json-schema-processors.ts` plus their targeted tests out into a follow-up.

**Smaller nit on the bundled test.** `tests/codec.test.ts` adds a `describe("encode reports descendant check failures under checked parents", ...)` block, which is great coverage but its tests don't reference the perf change at all — they're testing a reporting-behavior change that the description acknowledges ("continuable descendant check issues previously masked by the canary now accompany aborting issues"). If that reporting change ships together with the perf elision, the perf change becomes load-bearing for the reporting change too, making it harder to reason about. Same suggestion: separate PR.

The perf claims (~1.9x for check-bearing codec encode) are believable from the diff and the new `count()` tests are a good way to lock the regression down. 👍

drudolf:
Issue author here. Cross-linking for whoever reviews: #6159 (opened alongside #6158) implements the same approach — skip the second backward parse unless an overwrite-family check is present, otherwise reuse the canary's value and issues. There's also a reference implementation at [`drudolf/zod@encode-canary-reuse`](https://github.com/drudolf/zod/tree/encode-canary-reuse). Three near-identical implementations now exist; no attachment to which one lands, but they should converge to one.

One verified behavioral fork between this PR and #6159: **issue ordering when a node's own checks and descendant checks both fail.** For `z.object({ name: z.string().min(5) }).refine(() => false)` with `{ name: "abc" }`:

- forward (`parse`, unchanged by either PR): `too_small` → `custom` (innermost-first)
- backward on current `main`: `custom` → `too_small` (outermost-first)
- this PR: keeps current backward ordering
- #6159: switches backward to match forward

With nesting the whole chain flips accordingly. Both PRs fix the #6158 masking case identically, so this ordering choice is the only observable difference: cross-direction consistency (`parse` and `encode` report the same issue list for the same schema and value) versus preserving today's `encode()` ordering for existing consumers. Probably worth an explicit maintainer call rather than having it fall out of whichever PR merges.

colinhacks:
Closing as not planned — I'm declining the approach on #6158, not this implementation, which was the better of the two that arrived on that issue. Keying the elision on `.overwrite()` assumes that's the only check that can mutate a value, and a plain `.check()` that assigns `ctx.value` already breaks that assumption; I also don't want the extra branching in `$ZodType` that zod/mini pays for.

The BigInt `toJSONSchema` fix in the second commit is worth having on its own — open that separately against #6209. Heads up that it needs `undefined` guarded before the round-trip, since `catchProcessor` now throws on `z.string().optional().catch(undefined)`. Thanks for looking into this 👍

PR #6337, opened 2026-08-05
feat: allow passing extra args to apply()

### LINKED ISSUE #5909: Pass arguments from `.apply` to schema mapper
I currently use the following schemas

```ts
export function nullishWithFallback<TSchema extends z.ZodType>(
	schema: TSchema,
	defaultValue: z.output<TSchema>,
) {
	return schema.nullish().transform((x) => x ?? defaultValue);
}

const myObject = z.object({
	id: z.string().apply((x) => nullishWithFallback(x, "default-id")),
});
```

it would be cool to use `apply` like this

```ts
declare module "zod" {
	interface ZodType {
		apply<T, TArgs extends unknown[]>(fn: (schema: this, ...args: TArgs) => T, ...args: TArgs): T;
	}
}

const myObject = z.object({
	id: z.string().apply(nullishWithFallback, "default-id"),
});
```

I can provide a pr if one is welcome.

deepshekhardas:
Fixes #5909.

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -3340,3 +3340,16 @@ z.parse(schema, null); // => null
 ```
 </Tab>
 </Tabs>
+
+Additional arguments are passed through to the function:
+
+```ts
+function withDefault<T extends z.ZodType>(schema: T, value: z.output<T>) {
+  return schema.nullish().transform((val) => val ?? value);
+}
+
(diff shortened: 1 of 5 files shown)


colinhacks:
Pushed a couple of fixes on top: `TArgs` needed a default so `schema.apply<T>(fn)` keeps compiling, and the zero-arg path is guarded to stay inlineable. Merging — thanks for looking into this 👍

PR #6340, opened 2026-08-05
fix: ZodFlattenedError accepts a schema, matching inferFlattenedErrors

### LINKED ISSUE #5029: ZodFlattenedError does not return the same type as the old inferFlattenedErrors
We have been using `z.inferFlattenedErrors` for quite som time now in our nextjs server actions to return validationErrors for react aria component form validation errors prop.

```ts
export type SignUpValidationErrors = z.inferFlattenedErrors<
  typeof signUpFormDataSchema
>['fieldErrors'];
```

When doing this we got back:

Image

But in v4 it says that we should use `z.ZodFlattenedError` instead, but when doing so we get some other type that does not work as it should, or maybe we have done this wrong for a while now and should rewrite our logic? 

We are trying to do this now:

```ts
export type SignUpValidationErrors = z.ZodFlattenedError<
  typeof signUpFormDataSchema
>['fieldErrors'];
```

But it returns:

Image

The reason we flatten the errors are that we use React Aria Components Form-component which accepts a ValidationErrors-prop and it required it to be an object with `Record<string, string | string[]>`

deepshekhardas:
Fixes #5029.

`z.ZodFlattenedError` was a direct alias of `core.$ZodFlattenedError`, which takes the schema's **output type** as its generic. The v3 `z.inferFlattenedErrors` it's meant to replace takes a **schema**. So a straight migration — `inferFlattenedErrors<typeof schema>` → `ZodFlattenedError<typeof schema>` — keyed `fieldErrors` over the schema's internals (`_zod`, `def`, `type`, ...) instead of the actual fields.

This makes the deprecated alias accept both:

```ts
// both now resolve to the same type
type A = z.ZodFlattenedError<typeof schema>["fieldErrors"];
type B = z.ZodFlattenedError<z.output<typeof schema>>["fieldErrors"];
type C = z.inferFlattenedErrors<typeof schema>["fieldErrors"];
```

When passed a `$ZodType`, it unwraps to `output<T>`; otherwise it's treated as a plain output type, so existing callers passing `z.output<typeof schema>` keep their current typing. Added a type test asserting `ZodFlattenedError<typeof schema>` is equal to both `ZodFlattenedError<output<typeof schema>>` and `inferFlattenedErrors<typeof schema>`.

### DIFF
--- a/packages/zod/src/v4/classic/errors.ts
+++ b/packages/zod/src/v4/classic/errors.ts
@@ -79,11 +79,17 @@ export const ZodRealError: core.$constructor<ZodError> = /*@__PURE__*/ core.$con
 });
 
 /** @deprecated Use `z.core.$ZodFlattenedError` instead. */
-export type ZodFlattenedError<T, U = string> = core.$ZodFlattenedError<T, U>;
-/** @deprecated Use `z.core.$ZodFormattedError` instead. */
-export type ZodFormattedError<T, U = string> = core.$ZodFormattedError<T, U>;
-/** @deprecated Use `z.core.$ZodErrorMap` instead. */
-export type ZodErrorMap<T extends core.$ZodIssueBase = core.$ZodIssue> = core.$ZodErrorMap<T>;
+/** @deprecated Use `z.core.$ZodFlattenedError` instead. */
+export type ZodFlattenedError<T, U = string> = T extends core.$ZodType
+  ? core.$ZodFlattenedError<core.output<T>, U>
+  : core.$ZodFlattenedError<T, U>;
+
+export type {
+  /** @deprecated Use `z.core.$ZodFormattedError` instead. */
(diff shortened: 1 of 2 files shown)


colinhacks:
Declining this one. The schema-taking spelling already exists and still works: `z.inferFlattenedErrors<typeof schema>` is unchanged from v3, and `ZodFlattenedError` was never its replacement — it's the v4 name for v3's `typeToFlattenedError`, which took an inferred type too. So the migration in #5029 didn't need a type change.

Making just this one alias schema-polymorphic would also leave the surface less consistent than it is now — the sibling `ZodFormattedError` has the identical trap and this doesn't touch it. The real bug was the `@deprecated` hint pointing at `z.$ZodFlattenedError`, which isn't exported from the top level at all, so TypeScript's "did you mean" walks you into the wrong generic. Fixed in 2848177d.

Thanks for digging into this 👍

Issue #5241, opened 2025-09-15
[v4]: z.registry size issue while using with recursive object


MaciejDabek:
Hey, 

I have done early zod v4 adaptation in my app, that for a bit stopped at v3.25.28. 

This week I decided to finally update it to current version (I know it is probably around 100 versions difference). There is one big issue that occurred meantime.

```typescript
const Schema = z.strictObject({
	type: z.literal('special').meta({description: 'Type'}),
	config: z.strictObject({
		title: z.string().meta({description: 'Title'}),
		get elements(): z.ZodArray<z.ZodDiscriminatedUnion<any>> {
			return z.array(z.discriminatedUnion([AType, BType, CType])).meta({
				id: 'SpecialElements',
				title: 'SpecialElements',
				description: 'Array of elements'
			})
		}
	})
})
```

So for this code I got issue:

`Error: ID ArrayWidgets already exists in the registry`

Without thinking, I just removed `id` and everything was working, but I noticed that my website started to use a lot of memory. When I analysed it, there was z.registry/globalRegistry that weights over 1GB (my schemes are quite big). 

So from my observations, I would assume some issue where recursive is registering everything recursively in registry. @colinhacks what do you think?

MaciejDabek:
I did some more digging, an issue with ID started to occur with `v4.0.15`, so probably fix of #5018 introduced a bug with memory leak in registry.

colinhacks:
This should have been fixed in 4.1.8, upgrade to latest and see it's resolved. If not, feel free to open a new issue with a complete repro. https://github.com/colinhacks/zod/commit/36c4ee354d0c1f47b7311e49f6dd4b7a11de04f5

colinhacks:
The original issue is also fixed in 4.1.10. Good catch. There was an issue with shape caching.

PR #6343, opened 2026-08-08
perf: cut per-schema memory by moving methods to the prototype

colinhacks:
A schema instance's cost is dominated by own-property count, not by closures. V8 sizes the property backing store in steps, and `$constructor` assigns nothing itself so instances get no in-object slots to absorb them: 12 own properties cost 128 bytes, 13 cost 848, 21 cost 1616. `z.string()` carried **49** — 26 of them the deprecated format methods (`.email`, `.uuid`, `.datetime`, ...), plus 13 parse-family closures on every schema of any type — so it paid 1616 bytes of backing store before storing anything useful.

This moves the parse family, `toJSONSchema`, `~standard`, the string format methods and the bigint checks onto the prototype, where they materialize per instance on first read and cache as an own property. Detached use (`const { parse } = schema`, `arr.map(schema.parse)`) still works, per-instance assignment still works, and interior nodes of a schema graph — which nothing parses through directly — allocate none of them. `zod/mini` gets the same treatment; it sat at 14 own properties, one past the step.

The trait system is untouched. Prototypes aren't chained to each other — composition happens by function call through `traits`-guarded `init`s — so every init in a chain installs onto the same concrete prototype, and each prototype ends up with exactly the union of the groups its own chain ran. `z.email()` still has `.min` and not `.email`, as before.

|                            | before  | after   |
| -------------------------- | ------- | ------- |
| `z.string()`               | 7709 B  | 824 B   |
| `z.object()` 10 keys       | 83928 B | 12443 B |
| `z.array(z.string())`      | 11432 B | 1460 B  |
| mini `z.string()`          | 2528 B  | 593 B   |
| 500-schema API catalogue   | 198.0 MB | 64.2 MB |

Construction is 60–120% faster. Parse throughput is flat or faster across all 17 benchmarks in `packages/bench`.

Bundle size grows, and it's the one real cost here: +143–197 B gzipped for classic (+0.7–1.3%), +165–221 B for mini (+5.7–5.8%). That's the installer machinery. It's isolated in `core/util.ts`, and the mini half is a one-commit revert if that trade reads wrong for a package sold on bundle size.

Two things worth knowing for anyone touching this later, each of which cost a regression first. A bound function pays a call-time trampoline, so the parse family builds its closures rather than using `.bind()` — routing it through `.bind()` cost 6–19% on parse. And redefining an accessor demotes the object out of fast properties into dictionary mode: `~standard` was previously read eagerly and redefined per instance, which regressed `z.object().parse` by 2x until it moved to the prototype. `packages/bench/object.ts` caught that one; an in-process A/B did not.

`packages/bench/memory/` holds the harness this came out of — calibrated heap-delta measurement, V8 snapshot diffing, an own-property census, a dictionary-mode check, the property-slack curve, and a leak detector with positive and negative controls. That detector confirms the `_idmap` retention behind #5241 (`.meta({ id })` and `.register(reg, { id })` add 1000 entries per 1000 calls; `.describe()`, `.meta()` without an id, and `fromJSONSchema` with `$id` add none). Not addressed here — changing registry retention is an API decision, not a refactor.


colinhacks:
Folded into #6318, which was doing the same thing from the other direction — @zirkelc got to the prototype install independently. Everything here is on that branch now, plus two ideas of theirs mine didn't have (`_def` deriving from `this`, and skipping the `checks` spread when empty).

Closing in favour of #6318.

PR #4359, opened 2025-05-12
fix(core): prototype pollution in `generateFastpass`

ahrjarrett:
Closes #4358 
Related to #4357 

Colin when you see this, will you point me to where I should write tests that confirm the fix?

I didn't see any in `core/tests`, but can confirm this fixed the issue for me locally (for #4358, have not tested #4357)

## Summary by CodeRabbit

- **New Features**
  - Added new string schema methods for validation: `.jwt()`, `.xid()`, `.ksuid()`, `.e164()`.
  - Introduced `.unique()` method for array schemas to ensure uniqueness.
  - Added `z.string().json()` for parsing and validating JSON strings.
  - Added `z.file()` schema for browser file validation with methods for size, type, and filename checks.
  - Added `z.literal.template()` for template literal type schemas.

- **Documentation**
  - Updated and expanded documentation, including new sections on JSON parsing, file validation, unique arrays, template literals, and error handling.
  - Improved migration, contributing, and error handling guides.
  - Added and updated workspace and formatting configuration files for editors and tools.

- **Chores**
  - Migrated project tooling from Yarn/Jest to pnpm/Vitest.
  - Updated configuration files for package management, formatting, linting, and TypeScript.
  - Added, updated, or removed various project configuration, workflow, and ignore files.

- **Bug Fixes**
  - Corrected documentation errors and improved type annotations for clarity.

- **Refactor**
  - Renamed references from `ZodEffects` to `ZodTransform` for accuracy.

- **Revert**
  - Removed legacy and redundant configuration, test, and benchmark files to streamline the codebase.

- **Style**
  - Updated formatting rules and editor settings for consistency across the project.

PR #6346, opened 2026-08-09
fix(json-schema): keep __proto__ keys as own properties in schema conversion

colinhacks:
Both JSON Schema converters build plain object dictionaries and fill them with `obj[key] = value`. When the key is `__proto__`, that invokes the inherited setter instead of creating an own property, so the entry is silently lost.

```ts
const schema = z.object({ ["__proto__"]: z.literal("admin"), role: z.string() });

z.toJSONSchema(schema, { io: "input" });
// { properties: { role: {...} }, required: ["__proto__", "role"] }
//   ^ required names a property that isn't there — Ajv accepts input zod rejects

const rebuilt = z.fromJSONSchema(JSON.parse(`{
  "type": "object",
  "properties": { "__proto__": { "type": "string" } },
  "required": ["__proto__"]
}`));
rebuilt.safeParse({}).success; // true — the constraint is gone
```

Four sinks, one line each, all fixed with the existing `util.assignProp`:

| Site | Before |
| --- | --- |
| `from-json-schema.ts` shape build | a required `__proto__` property is dropped, so the schema accepts `{}` |
| `json-schema-processors.ts` object processor | `properties` omits the key while `required` still lists it |
| `json-schema-processors.ts` registry overload | a schema registered under the id `__proto__` vanishes from the catalog |
| `to-json-schema.ts` defs builder | the `$def` is dropped while `$ref`s to it are still emitted, so the reference is unresolvable |

The last one is the sharpest: Ajv throws `can't resolve reference #/$defs/__proto__` on the emitted schema.

Also covers the unrecognized-annotation path in `fromJSONSchema`, which fed the same unsafe write into the metadata registry.

Parsing has tolerated a `__proto__` shape key since #4359 — only the converters were left behind. No global prototype pollution is involved anywhere here; every write lands on a local dictionary.

One note for reviewers: an object literal cannot express these cases. Writing `{ __proto__: x }` sets the literal's own prototype and yields zero own keys, so the tests use a computed key and `JSON.parse` — which is also the realistic source of a JSON Schema.

Refs #4359

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -1,5 +1,6 @@
 import type * as JSONSchema from "../core/json-schema.js";
 import { type $ZodRegistry, globalRegistry } from "../core/registries.js";
+import { assignProp } from "../core/util.js";
 import * as _checks from "./checks.js";
 import * as _iso from "./iso.js";
 import * as _schemas from "./schemas.js";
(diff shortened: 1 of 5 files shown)


colinhacks:
Credit for this one goes to @joostgrunwald, who reported it privately.

PR #6348, opened 2026-08-09
perf(v4): dedupe factory def boilerplate in core/api.ts (-2.3% tree-shaken)

zirkelc:
>
>
>
> Sibling PRs from the same campaign: #6349, #6350, #6351.

## What this PR does

Dedupes the repeated def-literal boilerplate in the `core/api.ts` factory functions. Dozens of factories build near-identical def objects that the minifier cannot compress because property names are not mangled. Three commits, each preserving def key order and key presence byte-for-byte:

**1. String-format schema factories (`_sf`/`_sfa`).** 27 factories (`_email`, `_uuid`, `_ipv6`, the ISO family, ...) each spelled out `{ type: "string", format: X, check: "string_format", abort: false, ...normalizeParams(params) }`. They now delegate to two small helpers; `_sfa` supplies the `abort: false` default, `_sf` is used by the ISO factories that never set `abort`, and per-factory extras (`version: "v4"`, `precision: null`, ...) pass through a `defaults` argument in the original key position. Public signatures unchanged.

**2. Check factories (`_ck`/`_ckf`).** Same treatment for 17 check factories (`_lt`/`_gte`, `_minLength`-family, `_regex`/`_includes`/`_startsWith`/`_endsWith`, ...). Trailing fields still spread *after* `normalizeParams`, so user params cannot override `value`/`inclusive`/`pattern`, exactly as before.

**3. Shared `_defaultDef`.** The `defaultValue`-getter def literal (unwrap function values, `shallowClone` static ones) appeared five times across `core/api.ts`, `classic/schemas.ts`, and `mini/schemas.ts`; it is now built by one exported helper.

All three are construction-time only; no parse-path code is touched. The schema-init benchmark (`pnpm bench init`) was smoked around the round-1 factory change with results inside run-to-run noise.

Per case, minified bytes (gzip tracked alongside; judged on minified):

|---|---|---|---|
| 1. `_sf`/`_sfa` string-format factories | **-1,409 (-1.99%)** | -1,465 | -1,412 |
| 2. `_ck`/`_ckf` check factories | -191 | -227 | -122 |
| 3. `_defaultDef` | -58 | -108 | -58 |
| **branch total** | **-1,658 (-2.34%)** | **-1,800 (-0.55%)** | **-1,592 (-2.78%)** |

Full suite green after every commit (3,847 tests + typecheck on current `main`). One trade-off to disclose: minimal `zod/mini` consumers retain the `_ck` helper once, +45 bytes on a 7.4 kB bundle (commit 2); classic consumers see no regression anywhere. Net source delta: 3 files, +111/-323.

## Reproducing the numbers

1. `pnpm install`

   ```sh
   git fetch origin pull/6348/head:pr-branch
   ```

<details>

```ts
/**
 *
 *
 * sizecheck/meta/.
 */

  {
    name: "full-import-classic",
  },
  {
    name: "full-import-mini",
  },
  {
    name: "named-import-mini",
  },
  {
    name: "named-import-classic",
  },
  {
  },
  {
    name: "z-import-classic",
  },
  {
    name: "full-import-locales",
  },
];

  if (rev === "WORK") {
  } else {
      cwd: root,
      shell: "/bin/zsh",
    });
  }
  for (const c of CASES) {
  }
}

interface Result {
  min: number;
  gzip: number;
}

  for (const c of CASES) {
    const result = await build({
      bundle: true,
      minify: true,
      format: "esm",
      target: "es2020",
      write: false,
      metafile: true,
      outfile: `${c.name}.js`,
      logLevel: "silent",
    });
    out[c.name] = {
      min: contents.byteLength,
    };
  }
  return out;
}

}

materialise(revA, dirA);
materialise(revB, dirB);

console.log(
  "case".padEnd(16) +
    pad("A min", 10) +
    pad("B min", 10) +
    pad("Δ min", 9) +
    pad("Δ%", 8) +
    pad("A gz", 9) +
    pad("B gz", 9) +
    pad("Δ gz", 8)
);
let totMinA = 0;
let totMinB = 0;
let totGzA = 0;
let totGzB = 0;
for (const c of CASES) {
  const a = resA[c.name];
  const b = resB[c.name];
  totMinA += a.min;
  totMinB += b.min;
  totGzA += a.gzip;
  totGzB += b.gzip;
  const dMin = b.min - a.min;
  const dGz = b.gzip - a.gzip;
  console.log(
    c.name.padEnd(16) +
      pad(fmt(a.min), 10) +
      pad(fmt(b.min), 10) +
      pad(fmt(a.gzip), 9) +
      pad(fmt(b.gzip), 9) +
  );
}
console.log(
  "TOTAL".padEnd(16) +
    pad(fmt(totMinA), 10) +
    pad(fmt(totMinB), 10) +
    pad(fmt(totGzA), 9) +
    pad(fmt(totGzB), 9) +
);
```

</details>

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -2221,13 +2221,7 @@ export function _default<T extends core.SomeType>(
   innerType: T,
   defaultValue: util.NoUndefined<core.output<T>> | (() => util.NoUndefined<core.output<T>>)
 ): ZodDefault<T> {
-  return new ZodDefault({
-    type: "default",
-    innerType: innerType as any as core.$ZodType,
-    get defaultValue() {
-      return typeof defaultValue === "function" ? (defaultValue as Function)() : util.shallowClone(defaultValue);
-    },
-  }) as any;
+  return new ZodDefault(core._defaultDef("default", innerType as any, defaultValue)) as any;
 }
 
 // ZodPrefault
@@ -2251,13 +2245,7 @@ export function prefault<T extends core.SomeType>(
   innerType: T,
   defaultValue: core.input<T> | (() => core.input<T>)
 ): ZodPrefault<T> {
-  return new ZodPrefault({
-    type: "prefault",
-    innerType: innerType as any as core.$ZodType,
-    get defaultValue() {
-      return typeof defaultValue === "function" ? (defaultValue as Function)() : util.shallowClone(defaultValue);
-    },
-  }) as any;
+  return new ZodPrefault(core._defaultDef("prefault", innerType as any, defaultValue)) as any;
 }
 
 // ZodNonOptional
--- a/packages/zod/src/v4/core/api.ts
+++ b/packages/zod/src/v4/core/api.ts
@@ -1270,19 +1270,26 @@ export function _nullable<T extends schemas.$ZodObject>(
 
 // ZodDefault
 export type $ZodDefaultParams = TypeParams<schemas.$ZodDefault, "innerType" | "defaultValue">;
+
+/** Builds a default/prefault def whose `defaultValue` getter unwraps functions and clones static values. */
+// @__NO_SIDE_EFFECTS__
+export function _defaultDef(type: "default" | "prefault", innerType: schemas.$ZodType, defaultValue: unknown): any {
+  return {
+    type,
+    innerType,
+    get defaultValue() {
+      return typeof defaultValue === "function" ? (defaultValue as Function)() : util.shallowClone(defaultValue);
+    },
+  };
+}
+
 // @__NO_SIDE_EFFECTS__
 export function _default<T extends schemas.$ZodObject>(
   Class: util.SchemaClass<schemas.$ZodDefault>,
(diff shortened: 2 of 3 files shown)


colinhacks:
Closing this one. The headline is minified bytes, and that reduction doesn't survive compression. I re-measured the branch against its own merge-base: the ~1.6 kB minified drop on `import * as z from "zod"` lands at -36 bytes gzipped under terser, -14 under esbuild's minifier, and +26 under brotli. Minimal `zod/mini` bundles come out slightly larger, which matches the +45 bytes you flagged yourself.

That isn't a flaw in the measurement, it's the metric: gzip already encodes duplicated function bodies almost for free, so factoring the repetition out buys nothing on the wire while the new helper names are fresh bytes. Dedupe-shaped changes are size-neutral at best, and minified bytes aren't a proxy for what users download.

So this stands or falls as a readability change, and I'd rather keep the explicit def literals in the factory layer — they're easy to scan and easy to diff, and `_sf`/`_ck` moves per-factory detail into a shared signature. Thanks for looking into this.

PR #6350, opened 2026-08-09
perf(v4): dedupe JSON Schema pipeline internals (pinned in every bundle)

zirkelc:
>
>
>
> Sibling PRs from the same campaign: #6348, #6349, #6351.

## What this PR does

Dedupes internals of the JSON Schema pipeline. This code is not optional for consumers: every `ZodType` instance gets a `toJSONSchema` method and `~standard.jsonSchema` wiring at construction, so `to-json-schema.ts` and parts of `json-schema-processors.ts` are pinned into every tree-shaken bundle. Six commits:

**1. `_unrepresentable` factory.** Eleven processors (`bigint`, `symbol`, `undefined`, `void`, `date`, `nan`, `custom`, `function`, `transform`, `map`, `set`) were the identical function differing only in the error-message noun. All eleven messages fit `` `${name} cannot be represented in JSON Schema` `` byte-for-byte (snapshot-verified), so they collapse onto one PURE-annotated factory.

**2. `_refInner` + aliases.** Seven wrapper processors shared the body "process `def.innerType`, point `seen.ref` at it". `optional`/`nonoptional`/`promise` become direct references to the shared function; `readonly`/`default`/`prefault`/`catch` call it and keep their one distinctive line. `successProcessor` and `unknownProcessor` were literal duplicates of `booleanProcessor`/`anyProcessor` and become aliases.

**3. Tuple-branch merge in `from-json-schema.ts`.** The draft-2020-12 (`prefixItems`) and draft-7 (`items` array) tuple branches were identical after computing `tupleItems`/`rest`; the shared tail (tuple construction, `minItems`/`maxItems` checks) now runs once.

**4. Redundant normalization deleted.** `JSONSchemaGenerator`'s constructor normalized `draft-4`/`draft-7` aliases and then passed the result to `initializeContext`, which performs the same normalization. Pure deletion.

**5. `_deleteMatchingDefKeys`.** `finalize()` contained the same "delete keys (except `$ref`/`allOf`) whose value deep-equals the def" loop twice (ref dedup and parent dedup).

**6. `_getRoot`.** The `"Unprocessed schema. This is a bug in Zod."` guard appeared in both `extractDefs` and `finalize`.

All cold-path (JSON Schema generation); no parse-path code touched.

Per case, minified bytes:

|---|---|---|---|
| 1. `_unrepresentable` (11 processors) | -77 | **-924** | -77 |
| 2. `_refInner` + aliases | **-313** | -435 | -316 |
| 3. tuple-branch merge | 0 | -177 | 0 |
| 4. delete redundant normalization | 0 | -85 | 0 |
| 5. `_deleteMatchingDefKeys` | -84 | -84 | -84 |
| 6. `_getRoot` | -41 | -41 | -41 |
| **branch total** | **-515 (-0.73%)** | **-1,746 (-0.53%)** | **-518 (-0.90%)** |

`zod/mini` cases: full-import-mini -1,574, minimal mini consumer 0. Full suite green after every commit (JSON Schema output is heavily snapshot-tested, so message strings and emitted schemas are pinned). Net source delta: 4 files, +73/-141.

## Reproducing the numbers

1. `pnpm install`

   ```sh
   git fetch origin pull/6350/head:pr-branch
   ```

<details>

```ts
/**
 *
 *
 * sizecheck/meta/.
 */

  {
    name: "full-import-classic",
  },
  {
    name: "full-import-mini",
  },
  {
    name: "named-import-mini",
  },
  {
    name: "named-import-classic",
  },
  {
  },
  {
    name: "z-import-classic",
  },
  {
    name: "full-import-locales",
  },
];

  if (rev === "WORK") {
  } else {
      cwd: root,
      shell: "/bin/zsh",
    });
  }
  for (const c of CASES) {
  }
}

interface Result {
  min: number;
  gzip: number;
}

  for (const c of CASES) {
    const result = await build({
      bundle: true,
      minify: true,
      format: "esm",
      target: "es2020",
      write: false,
      metafile: true,
      outfile: `${c.name}.js`,
      logLevel: "silent",
    });
    out[c.name] = {
      min: contents.byteLength,
    };
  }
  return out;
}

}

materialise(revA, dirA);
materialise(revB, dirB);

console.log(
  "case".padEnd(16) +
    pad("A min", 10) +
    pad("B min", 10) +
    pad("Δ min", 9) +
    pad("Δ%", 8) +
    pad("A gz", 9) +
    pad("B gz", 9) +
    pad("Δ gz", 8)
);
let totMinA = 0;
let totMinB = 0;
let totGzA = 0;
let totGzB = 0;
for (const c of CASES) {
  const a = resA[c.name];
  const b = resB[c.name];
  totMinA += a.min;
  totMinB += b.min;
  totGzA += a.gzip;
  totGzB += b.gzip;
  const dMin = b.min - a.min;
  const dGz = b.gzip - a.gzip;
  console.log(
    c.name.padEnd(16) +
      pad(fmt(a.min), 10) +
      pad(fmt(b.min), 10) +
      pad(fmt(a.gzip), 9) +
      pad(fmt(b.gzip), 9) +
  );
}
console.log(
  "TOTAL".padEnd(16) +
    pad(fmt(totMinA), 10) +
    pad(fmt(totMinB), 10) +
    pad(fmt(totGzA), 9) +
    pad(fmt(totGzB), 9) +
);
```

</details>

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -466,37 +466,27 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
       const prefixItems = schema.prefixItems;
       const items = schema.items;
 
+      let tupleItems: ZodType[] | undefined;
+      let rest: ZodType | undefined;
       if (prefixItems && Array.isArray(prefixItems)) {
         // Tuple with prefixItems (draft-2020-12)
-        const tupleItems = prefixItems.map((item) => convertSchema(item as JSONSchema.JSONSchema, ctx));
-        const rest =
+        tupleItems = prefixItems.map((item) => convertSchema(item as JSONSchema.JSONSchema, ctx));
+        rest =
           items && typeof items === "object" && !Array.isArray(items)
             ? convertSchema(items as JSONSchema.JSONSchema, ctx)
             : undefined;
-        if (rest) {
-          zodSchema = z.tuple(tupleItems as [ZodType, ...ZodType[]]).rest(rest);
-        } else {
-          zodSchema = z.tuple(tupleItems as [ZodType, ...ZodType[]]);
-        }
-        // Apply minItems/maxItems constraints to tuples
-        if (typeof schema.minItems === "number") {
-          zodSchema = zodSchema.check(z.minLength(schema.minItems));
-        }
-        if (typeof schema.maxItems === "number") {
-          zodSchema = zodSchema.check(z.maxLength(schema.maxItems));
-        }
       } else if (Array.isArray(items)) {
         // Tuple with items array (draft-7)
-        const tupleItems = items.map((item) => convertSchema(item as JSONSchema.JSONSchema, ctx));
-        const rest =
+        tupleItems = items.map((item) => convertSchema(item as JSONSchema.JSONSchema, ctx));
+        rest =
           schema.additionalItems && typeof schema.additionalItems === "object"
             ? convertSchema(schema.additionalItems as JSONSchema.JSONSchema, ctx)
             : undefined; // additionalItems: false means no rest, handled by default tuple behavior
-        if (rest) {
-          zodSchema = z.tuple(tupleItems as [ZodType, ...ZodType[]]).rest(rest);
-        } else {
-          zodSchema = z.tuple(tupleItems as [ZodType, ...ZodType[]]);
-        }
+      }
+
+      if (tupleItems) {
+        const tuple = z.tuple(tupleItems as [ZodType, ...ZodType[]]);
+        zodSchema = rest ? tuple.rest(rest) : tuple;
         // Apply minItems/maxItems constraints to tuples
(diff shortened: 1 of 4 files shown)


colinhacks:
Closing this. The premise doesn't hold: none of this is pinned in every bundle. JSON Schema code doesn't reach a `zod/mini` bundle at all — I measured every mini scenario at exactly 0 bytes changed, and on classic it's -515 minified but +38 to +67 gzipped. So there's no size argument here in either direction.

That makes it a readability change, which is fine, but a four-file refactor of the JSON Schema internals is a snapshot-heavy area to churn for no user-visible payoff. The one piece I do like is `_unrepresentable` — collapsing seven near-identical "cannot be represented" processors into a factory is a genuine improvement, and I'd take that on its own. The target normalization move into `initializeContext` is also fine. The tuple hoist in `from-json-schema.ts` I'd skip; it trades duplication for mutable state shared across branches.

Happy to look at a smaller PR with just the processor factory. Thanks for looking into this.

PR #6351, opened 2026-08-09
perf(v4): merge duplicated cold-path logic in core (-1.7% tree-shaken)

zirkelc:
>
>
>
> Sibling PRs from the same campaign: #6348, #6349, #6350.

## What this PR does

Merges duplicated cold-path logic across `core/`: blocks that exist twice byte-for-byte, mostly on error/derivation paths. Eight commits:

**1. `ZodObjectJIT` codegen templates.** The three per-key code-generation branches in `generateFastpass` shared two large template chunks (the issues-concat snippet appeared 3x, the value-assignment block 2x); they are now built by two local template helpers. The *generated* runtime code differs only in whitespace, so parse behaviour and JIT output semantics are identical.

**2. `_walkIssues`.** `formatError` and `treeifyError` shared an identical recursive walker (invalid_union/invalid_key/invalid_element flattening); only the leaf accumulation differs and stays per-function. Builds on #6213, which reworked the leaf accumulation to own-property semantics via `node()` but left the walker skeleton duplicated; the extracted leaves use `node()` exactly as #6213 wrote them.

**3. Shared default `when` clauses.** The six size/length checks each defined an identical `def.when ??= (payload) => ...` closure ("non-nullish and has `size`"/"has `length`"); two shared module-level functions replace them. Instances now share one function reference instead of each owning an identical closure; nothing observes closure identity.

**4. `handleCatchResult`.** `$ZodCatch.parse` duplicated its entire catch-value block for the promise and non-promise branches; extracted into a module-level handler, the same shape as the existing `handleDefaultResult`/`handleOptionalResult` helpers in the file.

**5. `_assertNoChecks`.** `.pick()`/`.omit()`/`.merge()`/`.partial()` each carried the same refinement guard and long message; messages stay byte-identical including `.merge()`'s "Use .safeExtend() instead." suffix.

**6. `_attachPattern`.** `includes`/`startsWith`/`endsWith` checks each registered an identical onattach hook recording their pattern in the schema bag.

**7. `parse.ts` error path.** `result.issues.map((iss) => util.finalizeIssue(iss, ctx, core.config()))` appeared four times and the throw-with-stack-capture block twice; extracted `_finalizeIssues`/`_throwParseError`. Only failure paths call them; the success path and per-call ctx construction are untouched. Smoked with `pnpm bench object-fail`: 13.4µs vs 12.5µs baseline with fully overlapping ranges (noise).

**8. Safe-integer branches.** The number-format check's too_big/too_small pushes for unsafe integers merge via the conditional-spread idiom `$ZodCheckLengthEquals` already uses, preserving issue key order.

Per case, minified bytes:

|---|---|---|---|
| 1. JIT codegen templates | **-448** | -448 | 0 |
| 2. `_walkIssues` | -29 | -323 | -29 |
| 3. shared `when` clauses | -85 | -176 | -84 |
| 4. `handleCatchResult` | -101 | -97 | -101 |
| 5. `_assertNoChecks` | -223 | -223 | -225 |
| 6. `_attachPattern` | -145 | -145 | -145 |
| 7. parse.ts error path | -43 | -43 | -42 |
| 8. safe-integer branches | -120 | -120 | 0 |
| **branch total** | **-1,194 (-1.68%)** | **-1,575 (-0.48%)** | **-626 (-1.09%)** |

`zod/mini` improves too (minimal mini consumer -35, full-import-mini -1,574). Full suite green after every commit (3,847 tests + typecheck on current `main`, including the new #6213 error-walker tests). Net source delta: 5 files, +160/-214.

## Reproducing the numbers

1. `pnpm install`

   ```sh
   git fetch origin pull/6351/head:pr-branch
   ```

<details>

```ts
/**
 *
 *
 * sizecheck/meta/.
 */

  {
    name: "full-import-classic",
  },
  {
    name: "full-import-mini",
  },
  {
    name: "named-import-mini",
  },
  {
    name: "named-import-classic",
  },
  {
  },
  {
    name: "z-import-classic",
  },
  {
    name: "full-import-locales",
  },
];

  if (rev === "WORK") {
  } else {
      cwd: root,
      shell: "/bin/zsh",
    });
  }
  for (const c of CASES) {
  }
}

interface Result {
  min: number;
  gzip: number;
}

  for (const c of CASES) {
    const result = await build({
      bundle: true,
      minify: true,
      format: "esm",
      target: "es2020",
      write: false,
      metafile: true,
      outfile: `${c.name}.js`,
      logLevel: "silent",
    });
    out[c.name] = {
      min: contents.byteLength,
    };
  }
  return out;
}

}

materialise(revA, dirA);
materialise(revB, dirB);

console.log(
  "case".padEnd(16) +
    pad("A min", 10) +
    pad("B min", 10) +
    pad("Δ min", 9) +
    pad("Δ%", 8) +
    pad("A gz", 9) +
    pad("B gz", 9) +
    pad("Δ gz", 8)
);
let totMinA = 0;
let totMinB = 0;
let totGzA = 0;
let totGzB = 0;
for (const c of CASES) {
  const a = resA[c.name];
  const b = resB[c.name];
  totMinA += a.min;
  totMinB += b.min;
  totGzA += a.gzip;
  totGzB += b.gzip;
  const dMin = b.min - a.min;
  const dGz = b.gzip - a.gzip;
  console.log(
    c.name.padEnd(16) +
      pad(fmt(a.min), 10) +
      pad(fmt(b.min), 10) +
      pad(fmt(a.gzip), 9) +
      pad(fmt(b.gzip), 9) +
  );
}
console.log(
  "TOTAL".padEnd(16) +
    pad(fmt(totMinA), 10) +
    pad(fmt(totMinB), 10) +
    pad(fmt(totGzA), 9) +
    pad(fmt(totGzB), 9) +
);
```

</details>

### DIFF
--- a/packages/zod/src/v4/core/checks.ts
+++ b/packages/zod/src/v4/core/checks.ts
@@ -326,31 +326,17 @@ export const $ZodCheckNumberFormat: core.$constructor<$ZodCheckNumberFormat> = /
           // });
         }
         if (!Number.isSafeInteger(input)) {
-          if (input > 0) {
-            // too_big
-            payload.issues.push({
-              input,
-              code: "too_big",
-              maximum: Number.MAX_SAFE_INTEGER,
-              note: "Integers must be within the safe integer range.",
-              inst,
-              origin,
-              inclusive: true,
-              continue: !def.abort,
-            });
-          } else {
-            // too_small
-            payload.issues.push({
-              input,
-              code: "too_small",
-              minimum: Number.MIN_SAFE_INTEGER,
-              note: "Integers must be within the safe integer range.",
-              inst,
-              origin,
-              inclusive: true,
-              continue: !def.abort,
-            });
-          }
+          payload.issues.push({
+            input,
+            ...(input > 0
+              ? { code: "too_big" as const, maximum: Number.MAX_SAFE_INTEGER }
+              : { code: "too_small" as const, minimum: Number.MIN_SAFE_INTEGER }),
+            note: "Integers must be within the safe integer range.",
+            inst,
+            origin,
+            inclusive: true,
+            continue: !def.abort,
+          } as never);
 
           return;
         }
--- a/packages/zod/src/v4/core/checks.ts
+++ b/packages/zod/src/v4/core/checks.ts
@@ -938,6 +938,15 @@ export const $ZodCheckUpperCase: core.$constructor<$ZodCheckUpperCase> = /*@__PU
   }
 );
 
+/** Registers an onattach hook that records `pattern` in the schema's bag. */
+function _attachPattern(inst: $ZodCheck<never>, pattern: RegExp): void {
+  inst._zod.onattach.push((inst) => {
+    const bag = inst._zod.bag as schemas.$ZodStringInternals<unknown>["bag"];
+    bag.patterns ??= new Set();
+    bag.patterns.add(pattern);
+  });
+}
+
 ///////////////////////////////////
 /////    $ZodCheckIncludes    /////
 ///////////////////////////////////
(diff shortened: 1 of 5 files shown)


colinhacks:
Closing this. It has conflicts, and the size claim doesn't survive compression — measured against its merge-base it's +75 to +81 bytes gzipped on `import * as z from "zod"` and up to +44 on `zod/mini`, against a 1.3 kB minified reduction. Minified bytes aren't what ships.

Separately, I don't want the `$ZodCheckNumberFormat` change. Spreading a conditional into the issue literal and casting the result `as never` is harder to read than the explicit `too_big` / `too_small` branches, and the cast gives up the type checking that made the original safe.

The `_whenHasSize` / `_whenHasLength` extraction is the good part — it removes six identical closures and means one shared function instead of one per check instance. I'd take that as its own PR. Thanks for looking into this.

PR #6356, opened 2026-08-09
fix(v4): never write __proto__ into parse output

colinhacks:
Adopts one rule across every assembly point: **`__proto__` does not land in a parse result**, however the key got there — declared in a shape, named by a finite record key, produced by a key transform, or read off the input. The field still validates; only the write is suppressed.

```ts
const schema = z.object({ ["__proto__"]: z.object({ admin: z.boolean() }), name: z.string() });
const parsed = schema.parse(JSON.parse('{"__proto__":{"admin":true},"name":"a"}'));

Object.keys(parsed);                              // ["name"]
Object.getPrototypeOf(parsed) === Object.prototype; // true
```

This is what every released version already does for the key itself — the unguarded assignment invoked the inherited setter, so the key never reached output. What it *also* did, when the value was an object rather than a primitive, was replace the result's prototype. Skipping the write keeps output shape identical to 4.4.3 and removes the prototype replacement.

#6354 fixed the same defect by preserving the key as an own data property instead. That changed output shape for a declared `__proto__` field and split the codebase between preserve and skip, so this replaces it. The JIT knows the key set at generation time, so it now emits no write at all for `__proto__` and the `setProp` plumbing is gone from the compiled path.

Schema *internals* are the other side of the rule and keep `__proto__`: shapes carry it as a real validator, and JSON Schema documents emit it as a real property (#6346). Only parse results drop it.

### `util.pick`

Separate defect on the shape side, fixed here because it is the same policy boundary. The mask guard used `key in shape`, which answers true for every `Object.prototype` member:

```ts
const src = z.object({ ["__proto__"]: z.string().min(5), safe: z.string() });

src.pick(JSON.parse('{"__proto__":true}'));  // before: schema with zero fields, accepts {}
z.object({ name: z.string() }).pick(JSON.parse('{"toString":true}'));
// before: installed Function.prototype.toString as a validator, then threw at parse time
```

Now an own-key check, so an undeclared key throws `Unrecognized key` as intended, and the selected entry is written with `assignProp` so a declared `__proto__` validator survives selection. Same guard applied to `omit`, `partial` and `required`.

Refs #6354, #6346

### DIFF
--- a/packages/zod/src/v4/classic/tests/record.test.ts
+++ b/packages/zod/src/v4/classic/tests/record.test.ts
@@ -734,3 +734,35 @@ test("__proto__ in a finite key set becomes an own property", () => {
 
   expect(({} as any).a).toBeUndefined();
 });
+
+// The raw input key is what the __proto__ check sees, but the value is written under the key the
+// key schema emits. A normalizing key schema makes those two differ.
+test("__proto__ produced by a key transform becomes an own property", () => {
+  const schema = z.record(z.string().toLowerCase(), z.object({ a: z.string() }));
+  const parsed: any = schema.parse(JSON.parse('{"__PROTO__":{"a":"transformed"}}'));
+
+  expect(Object.getPrototypeOf(parsed)).toBe(Object.prototype);
+  expect(Object.prototype.hasOwnProperty.call(parsed, "__proto__")).toBe(true);
+  expect(parsed.__proto__).toEqual({ a: "transformed" });
+  expect(({} as any).a).toBeUndefined();
+});
+
+test("__proto__ from a key transform is an own property on the async path", async () => {
+  const schema = z.record(
+    z.string().toLowerCase(),
+    z.object({ a: z.string() }).refine(async () => true)
+  );
+  const parsed: any = await schema.parseAsync(JSON.parse('{"__PROTO__":{"a":"async"}}'));
+
+  expect(Object.getPrototypeOf(parsed)).toBe(Object.prototype);
+  expect(parsed.__proto__).toEqual({ a: "async" });
+});
+
+// A raw __proto__ input key is skipped outright before the write; only a key the key schema
+// *produces* reaches it. Pinning that so the skip isn't mistaken for the same defect.
+test("a raw __proto__ input key stays skipped in loose mode", () => {
+  const parsed: any = z.looseRecord(z.iso.datetime(), z.unknown()).parse(JSON.parse('{"__proto__":{"a":1}}'));
+
+  expect(Object.getPrototypeOf(parsed)).toBe(Object.prototype);
+  expect(Object.prototype.hasOwnProperty.call(parsed, "__proto__")).toBe(false);
+});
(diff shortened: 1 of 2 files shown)


percymcn:
Reproduced this PR's case on `zod@4.4.3` and it does what it says: `z.record(z.string().toLowerCase(), …)` parsing `{"__PROTO__":{…}}` writes under the *emitted* key, the raw-key guard never sees it, and routing that write through `setProp` fixes it. The third test — pinning that a *raw* `__proto__` input key stays skipped — is the one that makes the other two discriminate rather than assert, and it's worth keeping.

**The same write exists on `z.object()`, and there it is not lossy but self-contradictory.** A shape key reaches `properties` by plain assignment too, and `__proto__` is the one JSON key that assignment cannot create (an inherited accessor — `toString`, `constructor`, `valueOf` and `hasOwnProperty` are *data* properties and assign fine, which is why this hides so well). Note a shape has to be built with a **computed** key to test it at all: `{ "__proto__": z.string() }` in an object literal is *itself* the prototype-setter syntax, so a quoted-key fixture loses the field before zod sees it — that cost me a wrong measurement before I caught it.

```js
const S = z.object({ own: z.string(), ["__proto__"]: z.string() });

// 1. parse SUCCEEDS and silently drops the declared, required field
const p = S.parse(JSON.parse('{"own":"a","__proto__":"b"}'));
Object.prototype.hasOwnProperty.call(p, "__proto__"); // false
Object.keys(p);                                       // ["own"]
Object.getPrototypeOf(p) === Object.prototype;        // true  (no pollution — just gone)

// 2. toJSONSchema emits a schema that requires a property it does not declare
z.toJSONSchema(S);
// { type:"object", properties:{ own:{type:"string"} },
//   required:["own","__proto__"], additionalProperties:false }
```

`required` and `properties` disagree because they are built by different code: the name survives into the `required` array and is lost on the write into `properties`.

That second one has a concrete downstream consequence, and it's a hard error rather than a subtle one — `openai@7.4.0`'s own strict transformer rejects the emitted schema outright:

```
Object schema at `<root>` requires property `__proto__` but does not
declare it in `properties`.
```

with the identical schema under an ordinary key name accepted as the control, so the oracle discriminates and the cause is isolated to the key. It needs no API key, so it works as a regression check in CI.

Two notes on scoping a fix:

- The parse case is arguably the worse of the two: `.parse()` **returns success** while a field the schema marks required is absent from the result, so nothing anywhere surfaces it. The emit case at least fails loudly at the first consumer that validates.
- A regression test written on `toString` (or `constructor`/`valueOf`) **passes either way** — those assign as own properties normally. The fixture has to be `__proto__` specifically, and built with a computed key.

`setProp` already existing in core is presumably the shape of the fix here too; I haven't traced which of the object/emit paths would need it, so I'm reporting the behaviour rather than proposing the diff.

colinhacks:
Superseded by #6371, which preserves this fix as its first commit and includes the remaining audited follow-ups.

PR #6358, opened 2026-08-10
test(v4): cover mini object shape manipulation

irfanfandi:
`zod/mini` has no tests for `partial`, `required`, `pick`, `omit`, `extend` or `safeExtend`, even though `classic` covers all of them. Since both APIs delegate to the same helpers in `core/util.ts`, the mini side is only exercised through its own type signatures — which is exactly where #6302 slipped through (`merge()` declares a schema-to-schema signature but forwards to `util.extend()`, so the call that typechecks throws and the call that runs is a type error).

This ports that coverage to the mini function-style API:

- `partial` / `required`, with and without a mask, including the `ZodMiniNonOptional` wrapper chain and falsy mask values
- the `Unrecognized key` guard on all four mask-taking helpers — `pick`, `omit`, `partial`, `required` — which only fires on shape access and had no mini coverage at all
- the refinement guards on `pick` and `omit`, and `extend`'s narrower rule that it only rejects a refined schema when a key is actually overwritten
- `safeExtend` accepting a narrowing overwrite that `extend` rejects, and still rejecting a widening one

No behavior changes. I went looking for runtime divergence first and found none — the shared `core/util.ts` path makes mini and classic agree on all of the above, so this is coverage rather than a fix.

### DIFF
--- a/packages/zod/src/v4/mini/tests/partial.test.ts
+++ b/packages/zod/src/v4/mini/tests/partial.test.ts
@@ -1,0 +1,146 @@
+import { expect, expectTypeOf, test } from "vitest";
+import * as z from "zod/mini";
+
+const nested = z.object({
+  name: z.string(),
+  age: z.number(),
+  outer: z.object({ inner: z.string() }),
+  array: z.array(z.object({ asdf: z.string() })),
+});
+
+test("z.partial - shallow inference", () => {
+  const shallow = z.partial(nested);
+  expectTypeOf<z.infer<typeof shallow>>().toEqualTypeOf<{
+    name?: string | undefined;
+    age?: number | undefined;
+    outer?: { inner: string } | undefined;
+    array?: { asdf: string }[] | undefined;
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing this out. Mini already covers these in `mini/tests/object.test.ts` — `extend`, `safeExtend`, `pick`, `omit`, `partial`, `partial` with a mask, and the unrecognized-key guard on all four mask-taking helpers, added in #5130 and #5581. The unknown-key test here duplicates the one at the bottom of that file.

The rest is a port of `classic/tests/pickomit.test.ts` and `partial.test.ts` onto a surface that's a one-line delegation to the same function — `z.pick` forwards straight to `util.pick`, exactly like classic's `.pick()`. The refinement guards and falsy-mask handling live in `core/util.ts` and the classic tests already pin them, so asserting them again through mini doesn't add signal. #6302 was also fixed in #6404, so the case that motivated this is closed.

PR #6361, opened 2026-08-10
fix: preserve symbol-keyed properties in object schema output

### LINKED ISSUE #2734: Properties with Symbol keys are stripped from objects, even with passthrough()
TL;DR Properties with symbol keys are stripped from objects regardless of the object shape.  

`ZodRawShape` is defined as `[k: string]: ZodTypeAny` but according to Typescript rules, this [does not mean that a passed object should only have string keys](https://github.com/microsoft/TypeScript/issues/53265).  I think this is why Zod allows the object shape but then doesn't handle the presence of Symbol keys.

Therefore there should probably, at least, be a runtime exception to indicate that they are not handled.

```ts
const bar = Symbol('bar')

const Schema = z.object({
  foo: z.string(),
  [bar]: z.string()
})

const input: z.infer<typeof Schema> = {
  foo: 'foo',
  [bar]: 'bar',
}

const output = Schema.parse(input)
```

Results in:

```js
{
  foo: 'foo'
}
```

It seems that symbol keys are ignored in the object specification, as `z.string().default('bar')` also does not result in there being a property for `Symbol(bar)` with value `bar` in the output.

This also happens if the symbol property is included or omitted, and the schema has `.passthrough()` added.

deepshekhardas:
Fixes #2734. Symbols as object shape keys are now parsed and passthrough preserves them; strict mode reports them as unrecognized.

### DIFF
--- a/packages/zod/src/v4/classic/tests/object.test.ts
+++ b/packages/zod/src/v4/classic/tests/object.test.ts
@@ -896,4 +896,3 @@ describe("symbol keys in object shape", () => {
     expect(result.success).toBe(false);
   });
 });
-});
--- a/packages/zod/src/v4/classic/tests/object.test.ts
+++ b/packages/zod/src/v4/classic/tests/object.test.ts
@@ -871,3 +871,29 @@ test("object parsing still reads ordinary inherited properties", () => {
   expect(schema.parse(input)).toEqual({ value: "inherited" });
   expect(schema.parse(input, { jitless: true } as any)).toEqual({ value: "inherited" });
 });
+
+describe("symbol keys in object shape", () => {
+  const SYM = Symbol("sym");
+
+  test("parses symbol-keyed shape entries", () => {
+    const schema = z.object({ name: z.string(), [SYM]: z.number() });
+    const parsed = schema.parse({ name: "alice", [SYM]: 42 });
+    expect(parsed.name).toBe("alice");
+    expect(parsed[SYM]).toBe(42);
+  });
+
+  test("passthrough preserves symbol keys on input", () => {
+    const schema = z.object({ name: z.string() }).passthrough();
+    const OTHER = Symbol("other");
+    const parsed: any = schema.parse({ name: "alice", [OTHER]: "x" });
+    expect(parsed[OTHER]).toBe("x");
+  });
+
+  test("strict rejects unrecognized symbol keys", () => {
+    const schema = z.object({ name: z.string() }).strict();
+    const OTHER = Symbol("other");
+    const result = schema.safeParse({ name: "alice", [OTHER]: "x" });
+    expect(result.success).toBe(false);
+  });
+});
+});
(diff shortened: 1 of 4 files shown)


wahajahmed010 (commented):
Solid fix for #2734. Adding symbol support to the shape type and threading it through the object pipeline is the right approach, and the test coverage (parse, passthrough, strict) is well-scoped. A few substantive points before this lands:

**1. The `keySet` semantics change is subtle and not covered by tests.**
Previously `keySet` was `new Set(keys)` (strings only), and consumers like the `__proto__` skip in the parse loop relied on it. The new `keySet: new Set(allKeys)` mixes strings and symbols in a single Set — that's fine, but I don't see a test asserting that a symbol key literally named `"constructor"` or `"hasOwnProperty"` can't collide with prototype methods. Could you add a regression test for an object whose shape has a symbol whose description matches a built-in property name?

**2. The `getOwnPropertySymbols(input)` loop in `handleCatchall` doesn't iterate `Object.getOwnPropertyNames`.**
Before this PR, only string keys on the input were considered for catchall validation (and for `unrecognized_keys` reporting). After this PR, symbol keys on the input are checked, but I want to confirm this is intentional and that no other code path assumed the unrecognized list was strings-only. In particular, `handleIntersectionResults` still uses `Map<string, ...>` for `unrecKeys` — if two objects with overlapping symbol shapes get intersected, the symbol case is silently dropped on the floor. Worth either fixing that or adding a TODO.

**3. The JIT fastpath guard is correct, but the comment is missing.**
```ts
if (jit && fastEnabled && value.symbolKeys.length === 0 && ctx?.async === false && ctx.jitless !== true) {
```
The `value.symbolKeys.length === 0` clause correctly disables the fastpath for symbol-bearing objects, which is necessary because `generateFastpass` iterates a known string set. A one-line comment explaining *why* symbol objects skip the fastpath would save the next reader 10 minutes. Minor, but this is hot-path code.

**4. `stringifyPrimitive` change is correct but the test gap is real.**
The new `if (typeof value === "symbol") return value.toString();` is the right move (otherwise an unrecognized symbol key would render as `Symbol()` in error messages, which is useless). But there's no test pinning down the error-message output for an unrecognized symbol. Easy add.

**5. `errors.ts` widening from `string[]` to `PropertyKey[]` is a public-type change.**
Anyone consuming `$ZodIssueUnrecognizedKeys.keys` and assuming `string` (e.g. doing `.startsWith()` on entries) will see a TS error. That's the right call for correctness, but it's worth a CHANGELOG / migration note so downstream users don't get surprised.

Overall: the diff is small, the tests are focused, and the perf guardrail (skip fastpath for symbol shapes) shows the author thought about the JIT path. Approve once the test gap in point 1 and the comment in point 3 are addressed.

colinhacks:
Closing this out — merged in #6448, which builds on your commits, so they're preserved in the history and the credit carries.

Scanning each input for symbol keys turned out to cost 16–44% on `z.strictObject` parses of input containing no symbols at all, so I reworked it to support symbol keys declared in the shape instead. Those are known at construction and cost the parse path nothing, and they're the case #2734 actually reports — both reporters there declare the symbol and want it validated.

Thanks for looking into this one 👍

PR #6363, opened 2026-08-10
fix: add depth counter to prevent TS2589 on recursive ZodObject

### LINKED ISSUE #6015: Recursive type using ZodObject triggers TS2589 in v4
In zod v4, this fails to compile:

```ts
import type { ZodObject } from 'zod';

type Field = ZodObject<{ [k: string]: Field }>;
//   ^^^^^ Type instantiation is excessively deep and possibly infinite. (TS2589)
```

The same pattern worked in zod v3. 

Same error with the recursive union form (the realistic case):

```ts
import type { ZodObject, ZodString } from 'zod';

type Field = ZodString | ZodObject<{ [k: string]: Field }>;
```

`ZodUnion` is affected as well:

```ts
import type { ZodString, ZodUnion } from 'zod';

type Field = ZodUnion<[ZodString, Field]>;
//   ^^^^^ TS2589
```

`ZodArray`, `ZodTuple`, and `ZodDiscriminatedUnion` don't fail in this minimal directly-recursive form, but they do trigger `TS2589` once combined with `ZodObject` or `ZodUnion` in a larger recursive union (e.g. a "field can be any of these schemas" pattern).

------------

This looks separate from #4611. That one is about the wrapper types (`ZodNullable`, `ZodReadonly`, `ZodLazy`, `ZodPipe`), there's no wrapper here, just a plain `ZodObject` in a recursive position.

Tested with zod 4.2.1 and TypeScript 5.8.3.

deepshekhardas:
Fixes #6015. Adds a depth counter to recursive ZodObject type inference to avoid TS2589 instantiation depth errors.

### DIFF
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -1764,16 +1764,6 @@ type OptionalInSchema = { _zod: { optin: "optional" | "defaulted" } };
 
 type _Depth = readonly unknown[];
 
-type $ResolveOutputWithDepth<T, Depth extends _Depth> =
-  Depth["length"] extends 10
-    ? unknown
-    : T extends { _zod: { output: infer O } } ? O : unknown;
-
-type $ResolveInputWithDepth<T, Depth extends _Depth> =
-  Depth["length"] extends 10
-    ? unknown
-    : T extends { _zod: { input: infer I } } ? I : unknown;
-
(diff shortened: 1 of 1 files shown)


colinhacks:
This doesn't fix #6015. The `Depth` parameter is never incremented and no call site passes a third argument, so it's always `[]` and the check is always false — applying this to `main` produces byte-identical errors on all three reproductions. The guard is also on the wrong branch: the failing forms recurse through the `Record<string, ...>` branch, not the mapped type this touches. Closing this out. Thanks for taking a look at it 👍

PR #6364, opened 2026-08-10
fix: distinct error message when xor matches multiple options

### LINKED ISSUE #6342: z.xor() works fine with strict objects but breaks with regular ones
For some reason the code below fails in a really weird way. With the given input both `gentleXor` and `strictXor` must successfully parse the input but `gentleXor` fails.

```typescript
import z from "zod"

const input = {
    "os": {
        "name": "windows",
        "version": "^10\\."
    }
}

const nameOsRule = z.object({
    name: z.enum(["windows", "osx", "linux"]).nonoptional()
})

const versionOsRule = nameOsRule.extend({
    version: z.string().nonempty().nonoptional()
})

// This will break
const gentleXor = z.xor([
    nameOsRule,
    versionOsRule
])

// The strict version works fine
const strictXor = z.xor([
    z.strictObject(nameOsRule.shape),
    z.strictObject(versionOsRule.shape)
])

gentleXor.parse(input.os, { reportInput: true })
strictXor.parse(input.os, { reportInput: true })
```

The error doesn't give any clue and looks uninformative:
```
ZodError: [
  {
    "code": "invalid_union",
    "errors": [],
    "inclusive": false,
    "path": [],
    "message": "Invalid input",
    "input": {
      "name": "windows",
      "version": "^10\\."
    }
  }
]
```

Can someone explain me why this breaks?

Zod version is 4.4.3

deepshekhardas:
Fixes #6342. z.xor() now reports a distinct error when multiple union options match instead of a generic one.

### DIFF
--- a/packages/zod/src/v4/classic/tests/union.test.ts
+++ b/packages/zod/src/v4/classic/tests/union.test.ts
@@ -200,6 +200,7 @@ test("z.xor() - multiple matches fails", () => {
   if (!result.success) {
     expect(result.error.issues[0].code).toBe("invalid_union");
     expect((result.error.issues[0] as any).inclusive).toBe(false);
+    expect(result.error.issues[0].message).toBe("Invalid input: matches more than one exclusive option");
   }
 });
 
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing in favor of #6376, which takes the same approach and goes a bit further — the issue now carries a `matches` array with the indices of the options that matched, so callers can act on the overlap instead of parsing the message, and `api.mdx` documents the object-overlap trap behind #6342.

Right diagnosis. Thanks for looking into this 👍

PR #6365, opened 2026-08-10
fix: add offset option to z.iso.time() for timezone-aware time strings

deepshekhardas:
Fixes #6235. Adds an offset option to z.iso.time() to accept timezone-aware time strings.

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -525,12 +525,6 @@ z.iso.time({ precision: 2 });  // HH:MM:SS.ss (centisecond precision)
 z.iso.time({ precision: 3 });  // HH:MM:SS.sss (millisecond precision)
 ```
 
-The two combine, though `precision: -1` and `offset: true` no longer describe a `full-time`.
-
-```ts
-z.iso.time({ offset: true, precision: 0 }); // HH:MM:SS followed by an offset
-```
-
 ### IP addresses
 
 ```ts
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -497,11 +497,22 @@ time.parse("03:15:00"); // ✅
 time.parse("03:15:00.9999999"); // ✅ (arbitrary precision)
 ```
 
-No offsets of any kind are allowed.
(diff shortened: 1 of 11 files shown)


wahajahmed010 (commented):
Thanks for the PR. The implementation looks correct and the regex covers the common cases. A few observations:

1. The `???` in the test comment on line 156 (`// no offset ??? rejected when offset required`) looks like a stray placeholder — please clean it up to something like `// without offset is rejected when offset is required`.

2. Worth adding edge-case tests for boundary timezone offsets:
   - `"10:15:30+00:00"` should pass
   - `"10:15:30-00:00"` should pass
   - `"10:15:30+24:00"` should fail (your regex `[01]\\d|2[0-3]` correctly rejects this, but a regression test pins the behavior)
   - `"10:15:30+10:60"` should fail (seconds/minutes >59)

3. The JSDoc on `time()` in `regexes.ts` should mention the `offset` flag — the type was extended but no comment describes the semantics. A short one-liner like `// when offset: true, accepts trailing Z or ±HH:MM` would help.

4. ISO 8601 technically also allows `+HH` and `+HHMM` (no colon) forms. If zod is targeting strict ISO 8601, you may want to call this out in the JSDoc as a deliberate limitation. Otherwise a follow-up issue might be worth filing.

5. Minor: `const t = timeSource(args)` — naming `t` for a regex source string is unclear; `timeRegex` or `source` would read better at the call site.

colinhacks:
Going to close this one — I don't think the option should exist. Temporal has no offset-time type at all and `Temporal.PlainTime.from("10:15:30Z")` throws outright, and Postgres documents its own `time with time zone` as of questionable usefulness for the same reason: an offset can't be resolved without a date to anchor it.

It did surface a real bug, though. `z.fromJSONSchema` was mapping `format: "time"` onto `z.iso.time()`, which rejects every string that format is defined to accept — fixed in #6452. If you need to validate a full-time string directly, `z.string().regex(...)` with a `format` in `.meta()` covers both the parse and the JSON Schema output. Thanks for digging into this 👍

PR #6369, opened 2026-08-11
fix: count Unicode code points in string length checks

### LINKED ISSUE #3355: `z.string().max()` (and `min()`) don't count unicode characters
This is more of a feature request than a bug. Currently `string().max()` (and other length validators) count the UTF-16 length of a string rather than the number of unicode characters. The latter can be calculated using `[...value].length` where `value` is the desired string, so a custom transformer can work, but this would probably benefit most users of this library seeing as the length people usually care about in validation is the number of unicode characters (also called Unicode Scalars), not the UTF-16 length.

We could either option this out, e.g. `.max(5, {char: true})`, or use a separate method, e.g. `.charMax(5)`.

deepshekhardas:
﻿Counts string length in `max`/`min`/`length` checks by Unicode code points instead of UTF-16 code units, so surrogate pairs (emoji, etc.) are counted as a single character. Non-string lengthables (arrays, etc.) are unaffected.

Fixes #3355.

### DIFF
--- a/packages/zod/src/v4/classic/tests/string.test.ts
+++ b/packages/zod/src/v4/classic/tests/string.test.ts
@@ -869,6 +869,22 @@ test("boundary cases with zero length", () => {
   expect(() => maxZero.parse("hello")).toThrow();
 });
 
+test("length checks count Unicode code points", () => {
+  const fiveEmoji = "😀😀😀😀😀";
+
+  expect(z.string().max(5).safeParse(fiveEmoji).success).toBe(true);
+  expect(z.string().max(5).safeParse(fiveEmoji + "😀").success).toBe(false);
+
+  expect(z.string().min(1).safeParse("😀").success).toBe(true);
+  expect(z.string().min(1).safeParse("").success).toBe(false);
+
+  expect(z.string().length(5).safeParse(fiveEmoji).success).toBe(true);
+  expect(z.string().length(5).safeParse(fiveEmoji + "😀").success).toBe(false);
+
+  // arrays still count elements
+  expect(z.array(z.string()).max(2).safeParse(["a", "b", "c"]).success).toBe(false);
+});
+
 test("trim", () => {
   expect(z.string().trim().min(2).parse(" 12 ")).toEqual("12");
 
--- a/packages/zod/src/v4/classic/tests/string.test.ts
+++ b/packages/zod/src/v4/classic/tests/string.test.ts
@@ -873,13 +873,23 @@ test("length checks count Unicode code points", () => {
   const fiveEmoji = "😀😀😀😀😀";
 
   expect(z.string().max(5).safeParse(fiveEmoji).success).toBe(true);
-  expect(z.string().max(5).safeParse(fiveEmoji + "😀").success).toBe(false);
+  expect(
+    z
+      .string()
+      .max(5)
+      .safeParse(fiveEmoji + "😀").success
+  ).toBe(false);
 
   expect(z.string().min(1).safeParse("😀").success).toBe(true);
   expect(z.string().min(1).safeParse("").success).toBe(false);
 
   expect(z.string().length(5).safeParse(fiveEmoji).success).toBe(true);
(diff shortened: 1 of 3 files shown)


wahajahmed010 (commented):
## Code review

Thanks for the clean fix — the change is small, focused, and addresses a long-standing UX issue. The choice to use `for...of` for code-point iteration is correct (it does not expose lone surrogates as a separate unit beyond what the spec considers a code point). A few things worth considering before this lands:

### 1. Behavior change / breaking change risk

This is a **silent semantic change** for existing v4 users. Today `z.string().max(5)` counts UTF-16 code units, so `"1234😀"` (5 chars + 1 surrogate pair = 6 UTF-16 units) is rejected. After this PR it is accepted because the count drops to 5 code points. Most users will be happy, but anyone who has been working around this with `z.string().max(N, { ... })` or `z.string().refine(s => [...s].length <= N, ...)` will see different behavior with no warning.

Given issue #3355 explicitly suggests two non-breaking designs — `{ char: true }` option, or a separate `.charMax()` method — I think you should:

- either make this opt-in (e.g. gate on a schema option, or detect the presence of non-BMP characters in the input and dispatch),
- or call it out in the PR description and the changelog as a breaking change so it is reviewed with that framing.

A single line in the description like "Breaking: `max`/`min`/`length` now count code points" would already help reviewers and downstream users.

### 2. `stringLength` placement and naming

`util.stringLength` is fine, but two small things:
- It is exported alongside a bunch of generic utilities; consider co-locating it with the length validators, or at least adding a one-line comment near the call sites explaining why the branch is needed (i.e. what bug it fixes). Right now a future reader looking at the `typeof input === "string" ? util.stringLength(input) : input.length` ternary will not understand the asymmetry without reading the PR.
- Naming: `stringLength` is accurate, but `codePointLength` (or `unicodeScalarLength`) is more precise and harder to mistake for `[...str].length` which iterates code points too. Either is fine — just pick one and use it consistently in the test name as well.

### 3. Lone surrogates

`for (const _ch of str) count++` iterates **Unicode code points**, which per spec means an unpaired high or low surrogate is one code point on its own. That matches the language used in issue #3355 ("Unicode Scalars"). Worth a sentence in the PR description to confirm that intentional choice, since "code points" is sometimes used loosely to mean "user-perceived characters" (grapheme clusters), which is a stricter count and what most people intuitively want.

### 4. Tests — small additions worth considering

The new test is good. Two cheap additions that would catch regressions and document intent:
- A test using a non-BMP character that is **not** an emoji, e.g. U+1F3A9 (🎩) for simplicity or a math symbol like U+1D56B (𝕫) — most reviewers will assume "emoji" is the only case that matters, and it is not.
- A test that explicitly asserts the **old** UTF-16 count for the same input is what the implementation deliberately does *not* return, e.g. an inline comment or a separate `describe("counts UTF-16 units for backward-compat check", ...)` block. Even if that block is `expect(z.string().max(2).safeParse("😀😁").success).toBe(false)` it documents the choice.

### 5. Non-blocking: performance

`util.stringLength` walks the whole string on every check. For `max`/`min` this is fine because most strings are short. Worth a perf comment if the project has benchmarks for the string path; if not, no action needed.

Overall: this is the right fix and the right place to fix it. The main thing I would want before merging is (1) above — make sure the behavior change is acknowledged in the description/changelog so it is not a silent breaking change for v4 users.

Reviewed via Code Review Pipeline · open-source-contributor skill · Tier 1 / Phase 4

colinhacks:
Closing in favor of #6441, which lands the same default but keeps the common case free.

Two things pushed me to rewrite rather than iterate. `[...str].length` runs on every string length check, so `z.string().max(255)` over a 100-character string ends up ~16x slower — and it's avoidable, since a code point is at most two UTF-16 units and a string that already fits in units never needs counting. Separately, this branch predates the AOT compile path: `compile.ts` still emits `.length` for `min_length` and `length_equals`, so once rebased the compiled fast path would accept strings the interpreter rejects.

Thanks for kicking this off — it had been sitting a long time.

PR #6373, opened 2026-08-12
docs(api): document exactOptional()

deepshekhardas:
Adds an 'Exact optionals' section to the API reference after Optionals. \exactOptional()\ allows absent object keys but rejects an explicitly present \undefined\, and is Zod-only (no Zod Mini equivalent).


colinhacks:
Closing this out — `exactOptional()` is documented on `main` now. It went in alongside `.exactPartial()` in #6065, which needed the same section, so this one conflicts. One correction worth passing on: `zod/mini` does export `exactOptional()`, so the callout here isn't accurate; the version that landed is terser and covers both flavors.

Thanks for flagging the gap 👍

PR #6374, opened 2026-08-12
fix(v4): preserve input defaults through transforms in toJSONSchema

### LINKED ISSUE #6049: v4: `toJSONSchema` drops `default` when the defaulted schema contains a transform/pipe

deepshekhardas:
﻿When a defaulted schema contains a transform, `io: "input"` emission dropped the `default` (and `examples`) even when the value is a valid input — an identity transform, or a default drawn from a non-transforming union arm, both produced schemas missing the `default` keyword.

Now the value is validated as input first, and only dropped when it isn't a valid input (e.g. a number default on a string-input transform).

Fixes #6049.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2633,6 +2633,59 @@ test("defaults/prefaults", () => {
   `);
 });
 
+test("input defaults preserved when transform preserves the input type", () => {
+  // identity transform: default is a valid input, so it should survive
+  const a = z
+    .string()
+    .transform((s) => s)
+    .default("hello");
+  expect(z.toJSONSchema(a, { io: "input" })).toMatchInlineSnapshot(`
+    {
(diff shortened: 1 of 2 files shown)


colinhacks:
Going to close this — declining the underlying issue rather than the patch.

`.default()` is typed against the output type and short-circuits parsing, so on a transforming schema its value isn't generally valid input; that's why #4557 strips it. The cases that survive type-checking and are still valid input come down to identity-ish transforms, and `.prefault()` already covers those exactly — it's input-typed by definition and emits `default` in input mode.

Anything narrower needs the conversion to inspect the default value at runtime, and I don't want that logic in the JSON Schema path.

Thanks for the work here, and for the negative test — that was the right instinct 👍

PR #6379, opened 2026-08-12
docs(v4): correct the version boundary on the coerce missing-key note

colinhacks:
The coerce missing-key note that landed in #5964 dates the change to the v3 → v4 boundary:

```
// Zod 3: { foo: false }
// Zod 4: ZodError: Invalid input: expected nonoptional, received undefined
```

Zod 4.0, 4.1, 4.2 and 4.3 all return `{ foo: false }`, same as v3. The absent-key rejection landed in 4.4.0 with #5661, which made the object parser treat a key as absent-able only when the field schema is optional on the input side. As written the note tells a reader on 4.2 that their working code is broken, and gives someone upgrading 4.3 → 4.4 no signal that a minor bump moved behavior.

Names 4.4.0 as the boundary, inlines the `.default(false)` escape hatch, and states the general rule — it's an `optin` rule rather than a `z.coerce` rule, and the runtime now matches an inferred input type that has required the key since 4.0.

Also pins the behavior for all five coercers in `coerce.test.ts`. `optional.test.ts` covers `z.undefined()` and undefined-bearing unions, but nothing covered `z.coerce.*`, which is part of why the change reached users unannounced.

Refs #5957.

### DIFF
--- a/packages/docs/content/v4/changelog.mdx
+++ b/packages/docs/content/v4/changelog.mdx
@@ -380,15 +380,19 @@ type schemaInput = z.input<typeof schema>;
 // Zod 4: unknown;
 ```
 
-A missing key on an object schema with a `z.coerce.*` field now errors. Use `.default()` to declare the fallback explicitly.
+As of `zod@4.4.0`, a missing key on an object schema with a `z.coerce.*` field errors instead of coercing `undefined`. Use `.default()` to declare the fallback explicitly.
 
 ```ts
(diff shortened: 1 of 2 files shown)


colinhacks:
Closing this — not adding a changelog note for the 4.4.0 optin change. It was an unintentional soundness issue that landed as a bug fix in a minor, not a documented breaking change.

PR #6381, opened 2026-08-12
perf(v4): let three dead declarations tree-shake under esbuild

colinhacks:
Three declarations survive tree-shaking into every minimal `zod/mini` bundle despite being unreachable. All three are esbuild-specific — Rollup already drops them, and Rollup output is byte-identical before and after this change.

**`regexes.date`** was `/*@__PURE__*/ new RegExp(`^${dateSource}$`)`. esbuild will not drop an annotated call whose argument interpolates a variable, so `date` and the `dateSource` string were pinned into every bundle, including a bare `z.boolean()`. Moving the interpolation into an `anchor()` helper restores shakeability — a bare identifier argument is fine, it is specifically the template substitution that blocks the drop. `datetime()` still builds from `dateSource` exactly as before.

**`NUMBER_FORMAT_RANGES`** reads `Number.MIN_SAFE_INTEGER` / `Number.MAX_VALUE`. esbuild's known-pure-globals table covers `Math`, `Object`, `Symbol` and `JSON` but not `Number`, so the literal was never dropped even though only `checks.ts` reads it. A `@__PURE__` IIFE sidesteps that without touching the values. Plain numeric literals also work, but they regress `z.int32()` by ~25 bytes gzipped — the repeated `Number.MAX_VALUE` text compresses better than distinct literals.

**`$brand`**, plus `$output` and `$input` in `registries.ts`, are unannotated `Symbol()` calls.

esbuild, gzipped: `z.boolean()` 2632 → 2336 (−11.2%), `z.string()` −9.6%, a 3-key object −7.5%, a two-field form schema −6.7%, a 10-key mixed object −5.1%. Rollup: 0 bytes on every scenario. So the win lands on esbuild-native pipelines — Bun, tsup, Deno, esbuild-based Workers builds, Vite dev — and Vite production users see no change.

The `date` rewrite is the only part that touches a pattern, so it is worth being explicit: `date.source` and `date.flags` are unchanged, verified by string equality against the previous value and by a differential test over 90,007 inputs (every 4-digit year against leap-day, 30/31-day and out-of-range suffixes, plus malformed strings) with zero divergences.

One note on measurement, since it cuts against how bundle-size changes here have been judged recently: a reduction in minified bytes does not reliably survive compression. Refactors that merely remove duplicated text tend to be gzip-neutral or slightly negative, because LZ77 already encodes the repetition almost for free while the new helper names are fresh bytes. This change is a real removal rather than a dedup, which is why it survives compression at all.

Related: [#6050](https://github.com/colinhacks/zod/issues/6050), [#4433](https://github.com/colinhacks/zod/issues/4433).

### DIFF
--- a/packages/zod/src/v4/core/core.ts
+++ b/packages/zod/src/v4/core/core.ts
@@ -77,7 +77,7 @@ export /*@__NO_SIDE_EFFECTS__*/ function $constructor<T extends ZodTrait, D = T[
 }
 
 //////////////////////////////   UTILITIES   ///////////////////////////////////////
-export const $brand: unique symbol = Symbol("zod_brand");
+export const $brand: unique symbol = /*@__PURE__*/ Symbol("zod_brand");
 export type $brand<T extends string | number | symbol = string | number | symbol> = {
   [$brand]: { [k in T]: true };
 };
--- a/packages/zod/src/v4/core/regexes.ts
+++ b/packages/zod/src/v4/core/regexes.ts
@@ -94,9 +94,16 @@ export const httpProtocol: RegExp = /^https?$/;
 // E.164: leading digit must be 1-9; total digits (excluding '+') between 7-15
 export const e164: RegExp = /^\+[1-9]\d{6,14}$/;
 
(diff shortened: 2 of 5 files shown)


colinhacks:
Good catch on the missing guard — added in 586101ca.

`core/tests/treeshake.test.ts` bundles a minimal schema with esbuild and asserts the unreachable declarations are absent from the output. String-presence assertions rather than byte thresholds, so a failure names the declaration that came back instead of reporting a number that drifts as the library changes. Both entrypoints are covered for the reason you identified: `registries.ts` is dropped from the mini graph entirely, so `$output`/`$input` are only observable from classic.

Verified it fails on the pre-change source — both cases go red with the annotation names in the assertion messages — so it can actually catch the regression it is there to catch.

PR #6383, opened 2026-08-12
fix(v4): resolve z.infer through a generic index into a schema map

### LINKED ISSUE #6149: v4 Regression: Can't use z.infer with generic + map of schemas
Zod version 4.4.3.
Code below shows a minimal repro of the issue.  The type-checker is not able to understand that `z.infer<typeof SCHEMA_MAP[T]>` and `SCHEMA_MAP[T].parse(data)` have compatible types.  Using a direct index to the type internals via `(typeof SCHEMA_MAP[T])["_zod"]["output"]` does recognize a compatible type.  This seems like a bug with `z.infer` and appeared as a new issue when we upgraded to v4.

```ts
import { z } from "zod";

enum SchemaTypes {
    A,
    B,
}

const SCHEMA_MAP = {
    [SchemaTypes.A]: z.string(),
    [SchemaTypes.B]: z.number(),
};

export function parseDataForType<T extends SchemaTypes>(
    type: T,
    data: unknown,
): z.infer<(typeof SCHEMA_MAP)[T]> {
    // Has TS Error: Type string | number is not assignable to type output<{ 0: ZodString; 1: ZodNumber }[T]>
    return SCHEMA_MAP[type].parse(data);
}

// No TS error
export function parseDataForType2<T extends SchemaTypes>(
    type: T,
    data: unknown,
): (typeof SCHEMA_MAP)[T]["_zod"]["output"] {
    return SCHEMA_MAP[type].parse(data);
}

// No return type, use TS inference
export function parseDataForType3<T extends SchemaTypes>(type: T, data: unknown) {
    return SCHEMA_MAP[type].parse(data);
}

const a = parseDataForType(SchemaTypes.A, "foo"); // a = string
const a2 = parseDataForType2(SchemaTypes.A, "foo"); // a = string
const a3 = parseDataForType3(SchemaTypes.A, "foo"); // a = string | number

const b = parseDataForType(SchemaTypes.B, 123); // b = number
const b2 = parseDataForType2(SchemaTypes.B, 123); // b2 = number
const b3 = parseDataForType3(SchemaTypes.B, 123); // b3 = string | number
```

colinhacks:
Fixes #6149 — **but not safely. Converting to draft; do not merge.** See "Why this is drafted" below.

`z.infer<(typeof SCHEMA_MAP)[T]>` stopped resolving in v4 when `T` is a generic type parameter:

```ts
const SCHEMA_MAP = { a: z.string(), b: z.number() };

function parse<K extends keyof typeof SCHEMA_MAP>(key: K, data: unknown): z.infer<(typeof SCHEMA_MAP)[K]> {
  return SCHEMA_MAP[key].parse(data);
  //     ^ Type 'string | number' is not assignable to type 'output<{ a: ZodString; b: ZodNumber }[K]>'
}
```

`z.infer`/`z.input`/`z.output` alias `core.output`/`core.input`, which are conditional types. A conditional stays deferred while its check type is an unresolved generic, so the alias never simplifies and nothing is assignable to it. Indexing `_zod` by hand — the workaround in the issue — works because an indexed access has no such deferral. v3 declared `infer` as `T["_output"]`, which is why this only shows up in v4.

### Why this is drafted

Every localized fix changes the public alias to something other than `core.output`. That is the fatal step: `core.output` **must** keep the conditional form, so the two become different deferred types for an unresolved generic `T`, and TS relates them in one direction only. `core.output<T>` → `z.infer<T>` still works; the reverse is `TS2322`.

That is not an internals-only concern, because the whole classic method surface is typed on `core.output<this>` — 33 signatures in `classic/schemas.ts`, 16 more in mini. So ordinary generic wrapper code stops compiling:

```ts
function withCatch<T extends z.ZodType>(schema: T, d: z.infer<T>) {
  return schema.catch(d);  // TS2769 on this branch, compiles on main
}
```

Measured on this branch against `main`: `.catch()`, `.encode()`, `.decode()` and `.safeEncode()` all reject a `z.infer<T>` value inside a `<T extends z.ZodType>` wrapper. `.parse()` and `.refine()` are unaffected — they use the direction that still relates.

This is not specific to the intersection form in the current commit. The constrained direct index in 251c0337 produces the identical three failures, so it is inherent to changing the alias at all.

Closing the divergence from the other side does not work either. Applying the intersection form to `core.output`/`core.input` themselves breaks recursive schema assignability (`const Category: z.ZodType<Category> = z.lazy(...)`) plus 128 errors inside zod's own source — the exact hazard #4599 was written to prevent. Decoupling `~standard` from the polymorphic `this` first, so the recursion has nowhere to expand, makes it worse: 162 errors. The conditional's deferral on `this` is load-bearing.

### What a real fix would take

Migrating all 49 `core.output<this>` / `core.input<this>` sinks onto the new alias, which is a change to the most widely-instantiated generics in the library and needs its own perf validation. That is a judgment call rather than a bug fix, so it is not in this PR.

The commits are left in place because the corpus behind them is reusable: 110 generated identity assertions comparing each public alias to the unchanged `core.output`/`core.input` across every first-party schema type, plus non-schema arguments, foreign schema-shaped objects, and unions. It is what turned up the divergence, and it is what any future attempt should be run against.

Users hitting this today have a working escape hatch — `(typeof SCHEMA_MAP)[T]["_zod"]["output"]` — which is what the reporter is already using.

### DIFF
--- a/packages/zod/src/v4/classic/external.ts
+++ b/packages/zod/src/v4/classic/external.ts
@@ -10,7 +10,7 @@ import { config } from "../core/index.js";
 import en from "../locales/en.js";
 config(en());
 
-export type { infer, output, input } from "../core/index.js";
+export type { $InferOutput as infer, $InferOutput as output, $InferInput as input } from "../core/index.js";
 export type { JSONType } from "../core/util.js";
 export {
   globalRegistry,
--- a/packages/zod/src/v4/classic/tests/generics.test.ts
+++ b/packages/zod/src/v4/classic/tests/generics.test.ts
@@ -57,6 +57,27 @@ test("nested no undefined", () => {
   expect(outer.safeParse({ inner: undefined }).success).toEqual(false);
 });
 
+test("infer through a generic index into a schema map", () => {
+  const SCHEMA_MAP = {
+    a: z.string(),
+    b: z.string().transform((val) => val.length),
+  };
+  type Schemas = typeof SCHEMA_MAP;
+
+  function parse<K extends keyof Schemas>(key: K, data: unknown): z.infer<Schemas[K]> {
+    return SCHEMA_MAP[key].parse(data);
+  }
+
+  function reparse<K extends keyof Schemas>(key: K, data: z.input<Schemas[K]>): z.output<Schemas[K]> {
+    return SCHEMA_MAP[key].parse(data);
+  }
+
+  expectTypeOf(parse("a", "foo")).toEqualTypeOf<string>();
+  expectTypeOf(parse("b", "foo")).toEqualTypeOf<number>();
+  expectTypeOf(reparse("b", "foo")).toEqualTypeOf<number>();
+  expect(reparse("b", "foo")).toEqual(3);
+});
+
 test("generic on output type", () => {
   const createV4Schema = <Output>(opts: {
     schema: z.ZodType<Output>;
--- a/packages/zod/src/v4/core/core.ts
+++ b/packages/zod/src/v4/core/core.ts
@@ -119,6 +119,12 @@ export type output<T> = T extends { _zod: { output: any } } ? T["_zod"]["output"
 
 export type { output as infer };
 
+// Backing types for the public `z.input`/`z.output`/`z.infer`. The conditional forms above stay
+// deferred when the argument is a generic indexed access, so they index directly instead. The
+// constraint is structural rather than `$ZodType` to avoid instantiating `~standard`.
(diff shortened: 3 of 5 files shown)


colinhacks:
Closing this out — see #6149. Every localized version of this change diverges the public alias from `core.output`, and the whole classic method surface is typed on `core.output<this>`, so `.catch()`, `.encode()`, `.decode()` and `.safeEncode()` stop accepting a `z.infer<T>` value in generic wrapper code. Too dangerous for the size of the bug it fixes.

PR #6385, opened 2026-08-12
fix(object): don't emit absent keys for optional/exactOptional fields with coercion

### LINKED ISSUE #6236: .exactOptional() paired with coercion produces unexpected values

gonzoblasco:
Fixes #6236

## Problem

When an `exactOptional` field wraps a coercing schema, parsing an object without that key emits the key anyway with a manufactured value:

```ts
const obj = z.object({ omg: z.coerce.number().exactOptional() }).parse({});
// before: { omg: null }  (Zod 4.4.3: { omg: NaN })
// after:  {}
```

The same issue appears in both the JIT fastpass and the non-JIT (jitless) path.

## Root cause

`$ZodExactOptional.parse` delegates straight to its inner type without intercepting `undefined` (unlike `$ZodOptional`, which returns early on `payload.value === undefined`). So for an absent key the object loop runs the coercion with `undefined`, which produces a non-undefined value (`NaN`/`null`) that then gets written into the output.

## Fix

In the object pipeline (both the `$ZodObject` loop and the `$ZodObjectJIT` fastpass), skip absent keys for `optional`/`exactOptional` fields unless they wrap a `default`/`prefault` that should still inject its fallback value. Added a `hasApplicableFallback` helper that walks the wrapper chain so `optional(default(...))` keeps applying its default for absent keys.

## Testing

- Added a regression test covering absent key (stays absent), present key with coercion, and explicit `undefined` (still rejected), for both JIT and jitless paths.
- Full suite: 3813 tests pass (339 files), no type errors, lint and format clean.

/cc @gonzoblasco


gonzoblasco:
Ok. Both points are valid, fixed in the latest push.

**P2 #1 (transparent wrappers hiding a default):** `hasApplicableFallback` now recurses through transparent 1:1 wrappers that propagate `optin` and delegate to an inner default/prefault (`nullable`, `readonly`, `nonoptional`, `pipe`, `promise`, plus the existing `optional`/`exactOptional`/`catch`). So `z.string().default("x").nullable().optional()` now applies `"x"` for an absent key in both paths, instead of skipping the field. Added a regression test.

**P2 #2 (JIT still ran the schema before skipping):** in the JIT fastpass, for an optional/exactOptional field without an applicable fallback, the whole field (including `shape[key]._zod.run(...)`) is now emitted inside `if (isPresent)`. So absent keys no longer run user transforms/coercions in the default JIT path, matching jitless. Added a regression test with a transform that throws to assert both paths skip it.

Full suite: 3921 tests pass (343 files), no type errors, lint and format clean.

colinhacks:
Closing this one. The diagnosis is right, but `hasApplicableFallback` — a hardcoded list of wrapper type names, walked at parse time — isn't something I'll take in the object parse path. Every schema type added after this silently falls through the list, and it's already incomplete: `pipe` is in there but `$ZodPipe` has no `innerType`, so `z.string().default("x").transform(...).optional()` loses its default on this branch.

The fix belongs in `$ZodExactOptional.parse`, expressed through `optin` — which already encodes "this schema consumes `undefined` meaningfully" — rather than by enumerating types in the object loop.

Thanks for the investigation, it's what pinned down the JIT/jitless split 👍

Issue #2862, opened 2023-10-12
Proposal: z.not(parser) or equivalent


rossipedia:
## Purpose:

Returns the _inverse_ of the parser that is passed. For example:

```ts
// accepts anything _except_ a `string`.
z.not(z.string()) 

// accepts any string that _isn't_ an email
z.string().not(z.string().email());
```

Would this be useful? Is there a way of doing this in standard `zod` that I'm missing?

I know I can use `refine()`, but the ergonomics there are a bit awkward:

```ts
z.string().refine((val) => !z.string().email().safeParse(val).success);
```

JGJP:
I think this could be useful. The typescript equivalent would be doing a `T extends string ? never : T` I imagine. I think your second example would look better as:
```tsx
z.not(z.string().email())
```

colinhacks:
It's a nice idea, but there's no way to represent these "NOT" types in TypeScript so it isn't likely. It's important that Zod's inferred types accurately reflect what will pass validation. The closest thing that I'd be willing to introduce is an `.exclude()` method to mirror [`Exclude`](https://www.typescriptlang.org/docs/handbook/utility-types.html#excludeuniontype-excludedmembers).

Then you could do something like `z.unknown().exclude(z.string())` and the inferred type would be `Exclude<any, string>` (which is just `unknown` again).

JGJP:
@colinhacks I think the `.exclude()` that you suggest is in line with the proposal, but if the input is typed then it should be generic and operate on that type.

Kumar06Lav:
I've submitted a PR to add support for type negation using z.not() and the .not() method in schema validation.

JGJP:
I think users want this, and there's already a pull request up. It's been up for a year, so it would be good if that could get a look soon.

sharknoon:
I would really love to see an `.exclude()` function.

My use case is validating an `docker-compose.yml` file and not allowing certain keys. Something like this would be awesome:
```typescript
z.looseObject({}).exclude({ container_name: true })
// or
z.record(z.any(), z.any()).exclude({ container_name: true })
```

colinhacks:
This isn't something that can be represented in TypeScript, so seems unlikely.

JGJP:
@colinhacks do you think nobody in this thread knows how Typescript works?

PR #6389, opened 2026-08-13
feat: add .exclude() and .extract() to discriminated unions

### LINKED ISSUE #829: Missing `.exclude()` with functionality of `Exclude<>` in TS
Hey, first of all thanks so much for the package!

As the title suggests I'm searching for a way to replicate the `Exclude<Type, ExcludedUnion>` functionality provided by TS ([docs](https://www.typescriptlang.org/docs/handbook/utility-types.html#excludetype-excludedunion)) using `zod`. I searched for "exclude" in issues and prs, but didn't find a single mention of this utility type (outside of internals), therefore I'm opening this issue.

My use case involves unions of many smaller action types, e.g.:
`type Actions = { type: "added", ... } | { type: "removed", ... } | { type: "changed", ... } | ... ;`
(or rather an equivalent `const Actions = z.union([ ... ]);` on which I'm using `z.infer`)

I need to retroactively exclude only a few specific actions from the main `Actions` type.

Using only TS, I can add a discriminator key such as `notA: true` to those few actions and then define:
`type ActionsOnlyA = Exclude<Actions, Actions & { notA: true }>;`

Using zod, I have to add a discriminator such as `isA: true` to a lot more actions to then use:
`const ActionsOnlyA = Actions.and(z.object({ isA: z.literal(true) }));`
(I still favor this over passing additional schemas for `ActionsOnlyA` from the definitions up to the root of the project due to the amount of code.)

Maybe I'm also missing an obvious way on how to get this behavior using the existing functionality... Any advice or comment is highly appreciated :)

colinhacks:
Adds `.exclude()` and `.extract()` to `z.discriminatedUnion()`, mirroring the methods `z.enum()` already has. Both take a set of discriminator values and return a narrower discriminated union.

```ts
const MyResult = z.discriminatedUnion("status", [
  z.object({ status: z.literal("success"), data: z.string() }),
  z.object({ status: z.literal("failed"), error: z.string() }),
]);

const Success = MyResult.extract(["success"]); // { status: "success", data: string }
const NotSuccess = MyResult.exclude(["success"]); // { status: "failed", error: string }
```

This is the variant of the long-running "exclude" request that TypeScript can actually express — the ask in #829 is a tuple filter, not the arbitrary schema negation from #2862 that has no type-level representation. The inferred output type is exactly the surviving options, so it stays accurate.

Three construction-time guards, since none of these cases can be represented honestly in the result type:

- An unknown discriminator value throws, like the `Key X not found in enum` guard on `z.enum().exclude()`.
- An option whose discriminator is a `z.enum()` accepts several values. Selecting only some of them would silently drop the rest, so that throws.
- A union carrying refinements throws, matching `.pick()` and `.omit()`.

The accepted values come from the input side of each option's discriminator, which is what `propValues` is built from — so a codec discriminator is selected by its encoded value. For every other discriminator the two sides coincide.

Closes #829

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -1881,6 +1881,27 @@ const MyResult = z.discriminatedUnion("status", [
 
 Each option should be an *object schema* whose discriminator prop (`status` in the example above) corresponds to some literal value or set of values, usually `z.enum()`, `z.literal()`, `z.null()`, or `z.undefined()`.
 
+### `.exclude()` and `.extract()`
+
+To derive a narrower union from an existing one, pass a set of discriminator values.
+
+<Tabs groupId="lib" items={["Zod", "Zod Mini"]}>
+<Tab value="Zod">
+```ts
+const Success = MyResult.extract(["success"]); // { status: "success", data: string }
+const NotSuccess = MyResult.exclude(["success"]); // { status: "failed", error: string }
+```
+</Tab>
+<Tab value="Zod Mini">
+```ts
+// no equivalent
+
+```
+</Tab>
+</Tabs>
+
+Options are kept or dropped whole, so an option with several discriminator values must have all of them listed.
+
 <Accordions type="single">
 <Accordion title="Nesting discriminated unions">
 
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -1900,7 +1900,7 @@ const NotSuccess = MyResult.exclude(["success"]); // { status: "failed", error: 
 </Tab>
 </Tabs>
 
-Options are kept or dropped whole, so an option with several discriminator values must have all of them listed.
+Options are kept or dropped whole, so an option with several discriminator values must have all of them listed. Like `.pick()` and `.omit()`, both methods throw if the union carries refinements.
 
 <Accordions type="single">
 <Accordion title="Nesting discriminated unions">
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1814,8 +1814,7 @@ export const ZodDiscriminatedUnion: core.$constructor<ZodDiscriminatedUnion> = /
         matched.add(option);
       }
 
-      // An option is kept or dropped whole, so a partial selection of its
-      // discriminator values would silently disagree with the inferred type.
+      // An option is kept or dropped whole, so a partial selection of its discriminator values would silently disagree with the inferred type.
       const selected = new Set(values);
       for (const [value, option] of byValue) {
         if (matched.has(option) && !selected.has(value))
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1781,11 +1781,11 @@ export interface ZodDiscriminatedUnion<
   extract<const U extends readonly util.DiscriminatorValues<Options[number], Disc>[]>(
     values: U,
     params?: string | core.$ZodDiscriminatedUnionParams
-  ): ZodDiscriminatedUnion<util.ExtractDiscriminated<Options, Disc, U[number]> & readonly core.SomeType[], Disc>;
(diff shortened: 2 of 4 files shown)


wahajahmed010 (commented):
Solid PR — the three construction-time guards (unknown discriminator, partial multi-value selection, refinements) match what `z.enum().exclude()` already does, and matching on the input side of the discriminator is the only honest answer for codec discriminators. A few substantive points:

**Type-level: `& readonly core.SomeType[]` is doing real work but is invisible.** The return type is

```
ZodDiscriminatedUnion<ExtractDiscriminated<Options, Disc, U[number]> & readonly core.SomeType[], Disc>
```

A reader has to know that `SomeType` is `{ _zod: _$ZodTypeInternals }` to understand that the intersection forces the result to be a tuple of `$ZodType` instances. That trick isn't commented in the PR. Two options:

- Inline the constraint: `ExtractDiscriminated<Options, Disc, U[number]>` already narrows to specific options. If the existing `$ZodDiscriminatedUnion<...>` constraint is too loose, name a type alias (`type AsOptionTuple<T> = T & readonly SomeType[]`) and add a one-liner about why the structural intersection is needed.
- Or just expose the same return type as `ZodEnum.exclude()` does and see if it actually breaks — the discriminant inference might just work without the kludge.

Not a blocker, but the moment someone refactors `core.SomeType`, this intersection silently changes shape.

**The "must also be listed" check duplicates the discriminator set.** Look at the partial-selection test:

```ts
expect(() => Multi.exclude(["a"])).toThrow(/must also be listed/);
```

The error is raised for option `[a, b]` when only `[a]` is selected, because `b` is in `valuesOf(option)` but not in `selected`. This is correct, but the inner loop is O(options × values) on every call. For pathological unions (think: a 100-option discriminated union where one option has a 50-value enum discriminator), this is O(5000) per call. Probably fine in practice but worth noting — and a `for...of` over the smaller side would be the obvious optimization if it ever shows up in a profile.

**`inst.extract` is on the instance, not on the constructor.** That's consistent with `ZodEnum` so I'm not pushing on it, but it means `z.discriminatedUnion("type", [...]).extract([...])` returns a new *instance* with the same internal `_zod.def` reference for everything except `options`. The new instance shares discriminator, checks, and most other fields with the original — which is fine because nothing mutates `def` post-init in this codebase, but a comment in the doc block noting that the returned schema is a *view* and not a deep clone would help anyone who later adds a mutable field.

**Docs gap.** The MDX diff shows "Zod Mini: no equivalent" — that's accurate today, but the rest of the discriminated-union section doesn't say so elsewhere. If `ZodMiniDiscriminatedUnion` is going to grow these methods later, the stub is fine. If it isn't, the doc should say "intentionally not supported" so users don't open duplicate issues.

**One real edge case worth a test.** The type-level test covers `ZodEnum` as the discriminator, but not `ZodLiteral` with `.optional()` *and* `.default()` together. The runtime logic checks `propValues` on the input side; for `.default("x")` the input type is `string` (not `string | undefined`), and `propValues` should reflect that. The "exclude/extract with an optional-input discriminator" test covers `.optional()` and `.default()` separately but not both layered. Probably fine — but if a future Zod version starts including `undefined` in `propValues` for defaulted fields, the `Defaulted.exclude([undefined])` `// @ts-expect-error` would suddenly start type-checking and the `@ts-expect-error` directive would be a TS error itself. Worth pinning with a comment.

Overall the implementation matches the issue's "this is the variant TypeScript can express" framing. The codec and optional-discriminator tests are particularly good — they exercise branches the type system can't describe, and a `// @ts-expect-error` near each one is the right move. Approving with the type-level comment suggestion as a follow-up.

colinhacks:
Closing this out. I don't want `.exclude()` and `.extract()` as methods on discriminated unions — once one schema type carries them, every other one accumulates its own variant of the same set, and there aren't many methods like that in Zod by design.

The three construction-time throws this needs — unknown discriminator value, enum discriminator, refinements present — are the tell. A surface that has to reject that much at runtime isn't sitting cleanly on the type.

PR #6390, opened 2026-08-13
fix(record): looseRecord passes through unrecognized keys with a closed key schema

### LINKED ISSUE #6156: `looseRecord` with a closed key schema reports unrecognized keys fatally

deepshekhardas:
Fixes #6156

When the key schema is closed (e.g. \z.enum([...])\), \looseRecord\ treated unrecognized input keys as \unrecognized_keys\ issues, diverging from the open key-schema path which already passes them through. \looseRecord\ should preserve unknown keys regardless of whether the key type is closed or open.

Also adds a regression test confirming strict \z.record\ still rejects unknown keys with a closed key schema.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1427,6 +1427,17 @@ export function keyof<T extends ZodObject>(schema: T): ZodEnum<util.KeysEnum<T["
 
 // ZodObject
 
+// Like ZodOptional<Field>, but don't double-wrap already-optional fields and
+// don't widen `.default()` fields to `T | undefined` on the output. Already-
+// optional fields are idempotent, and fields with a `.default()` already
+// accept absent input while keeping a non-nullable output type.
+// See: https://github.com/colinhacks/zod/issues/6171
+export type PartialField<T> = T extends ZodOptional<core.SomeType>
+  ? T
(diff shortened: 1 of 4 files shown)


colinhacks:
Closing in favor of #6157, which has the same core fix with wider test coverage and green CI. This one also bundles an unrelated `.partial()` type change for #6171 — if you want to pursue that, open it as its own PR against that issue.

PR #6393, opened 2026-08-13
refactor(json-schema): collapse unrepresentable-type processors into a factory

zirkelc:
Follow-up to the suggestion on #6350: pulling the unrepresentable-type processor factory out as its own change, rebased onto current `main` and rewritten on top of #6391.

## What this does

After #6391 moved the throw/defer logic into `handleUnrepresentable`, eleven processors are identical except for one noun:

```ts
export const bigintProcessor: Processor<schemas.$ZodBigInt> = (schema, ctx, _json, _params) => {
  handleUnrepresentable(schema, ctx, "BigInt cannot be represented in JSON Schema");
};
```

They now come from one factory:

```ts
const _unrepresentable =
  (name: string): Processor<any> =>
  (schema, ctx) =>
    handleUnrepresentable(schema, ctx, `${name} cannot be represented in JSON Schema`);

export const bigintProcessor: Processor<schemas.$ZodBigInt> = /* @__PURE__ */ _unrepresentable("BigInt");
```

Covers `bigint`, `symbol`, `undefined`, `void`, `date`, `nan`, `custom`, `function`, `transform`, `map` and `set`. Every message stays byte-identical, since all eleven already fit `${noun} cannot be represented in JSON Schema`, and each export keeps its own `Processor<...>` annotation, so the public types are unchanged.

The `/* @__PURE__ */` markers keep the exports individually droppable now that the initializer is a call rather than a literal arrow.

This is a readability change, not a size one. Size numbers are in #6350; compressed, it is a wash.

### DIFF
--- a/packages/zod/src/v4/core/json-schema-processors.ts
+++ b/packages/zod/src/v4/core/json-schema-processors.ts
@@ -25,6 +25,12 @@ const formatMap: Partial<Record<checks.$ZodStringFormats, string | undefined>> =
 
 // ==================== SIMPLE TYPE PROCESSORS ====================
 
+/** Processor factory for types that have no JSON Schema equivalent. */
+const _unrepresentable =
+  (name: string): Processor<any> =>
+  (schema, ctx) =>
+    handleUnrepresentable(schema, ctx, `${name} cannot be represented in JSON Schema`);
+
 export const stringProcessor: Processor<schemas.$ZodString> = (schema, ctx, _json, _params) => {
   const json = _json as JSONSchema.StringSchema;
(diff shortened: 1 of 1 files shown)


colinhacks:
Closing this out — my fault for asking for it. I measured the factory on its own afterward and it doesn't pay for itself: on a tree-shaken classic bundle it comes out slightly larger after gzip, and zero either way on mini. And looking at the diff again, I don't think it's much of a readability win either — eleven one-liners that each say what they are read better to me than a factory call you have to go look up.

Thanks for splitting it out 👍

PR #6394, opened 2026-08-13
perf(v4): share default `when` closures across size/length checks

zirkelc:
Follow-up to the suggestion on #6351: pulling the shared `when` closures out as their own change, rebased onto current `main`, with nothing else from that branch.

## What this does

Six checks (`maxSize`, `minSize`, `sizeEquals`, `maxLength`, `minLength`, `lengthEquals`) each defined the same default `when` inline, so every check instance allocated its own copy of one of two identical closures:

```ts
inst._zod.def.when ??= (payload) => {
  const val = payload.value;
  return !util.nullish(val) && (val as any).length !== undefined;
};
```

They now share two module-level functions, `_whenHasSize` and `_whenHasLength`.

The payoff is per-instance memory rather than bundle size, in the same direction as #6318: one shared function instead of one closure per check instance. Retaining 200,000 `z.string().min(5)` schemas goes from 3,600.4 to 3,544.4 bytes per schema, a 56 byte drop that is stable across runs since it counts allocations rather than time.

Bundle size is not the argument here and does not improve; those numbers are in #6351.

## Behaviour

Check instances now share one function reference where each previously held an identical closure. `def.when` stays user-overridable through the same `??=`, and nothing in the codebase or tests observes closure identity.

### DIFF
--- a/packages/zod/src/v4/core/checks.ts
+++ b/packages/zod/src/v4/core/checks.ts
@@ -38,6 +38,18 @@ export const $ZodCheck: core.$constructor<$ZodCheck<any>> = /*@__PURE__*/ core.$
   }
 );
 
+/** Default `when` for size-based checks: run only on non-nullish values with a `size`. */
+const _whenHasSize = (payload: schemas.ParsePayload): boolean => {
+  const val = payload.value;
+  return !util.nullish(val) && (val as any).size !== undefined;
+};
+
+/** Default `when` for length-based checks: run only on non-nullish values with a `length`. */
+const _whenHasLength = (payload: schemas.ParsePayload): boolean => {
+  const val = payload.value;
+  return !util.nullish(val) && (val as any).length !== undefined;
+};
+
 ///////////////////////////////////////
 /////      $ZodCheckLessThan      /////
 ///////////////////////////////////////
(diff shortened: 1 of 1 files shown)


wahajahmed010 (commented):
Clean, well-scoped perf change. The two shared closures are pure and side-effect-free, the substitution is mechanical, and the rebased-only-this diff is exactly what was promised in the follow-up comment on #6351. Few small notes:

**1. Correctness invariant**

`util.nullish` is a fast check (probably `val == null` or similar) and the property access is the only branch. Hoisting the closure does not change observable behavior — both forms read `payload.value` once, do the nullish check, then the property access. Good.

**2. Memory savings are real but small per-instance**

Each inline arrow function captures the surrounding scope (in this case the module scope plus the `inst` argument via `inst._zod.def.when ??=`). Module-level hoisting removes the closure scope entirely. For a schema with one `z.string().max(10)` and one `z.string().min(1)` you'd save two allocations. The 200k-instance number in the description makes sense for batch validators (e.g. building a schema per row in a CSV pipeline). Reasonable motivation, not a hot path for most users but a clean win.

**3. Minor: could one closure cover both?**

`size` and `length` checks differ only in which property they read. A single closure with the property name passed in would technically be possible:

```ts
const makeWhenHas = (prop: "size" | "length") => (payload) => {
  const val = payload.value;
  return !util.nullish(val) && (val as any)[prop] !== undefined;
};
```

But that costs an extra property access and breaks the equivalence with the original code. Two closures is the right call. The "always-true ternary" closure overhead in the original code is precisely what #6318 was about, and you are following that same pattern correctly. No change needed.

**4. Suggestion: add a one-line benchmark**

A `bench()` call in a test (if the repo has a benchmark suite) or even a 5-line standalone snippet showing the allocation count delta would make the perf claim verifiable. Not required for merge — the change is small and obviously correct — but a follow-up PR with `vitest bench` numbers would be a nice complement to the description.

LGTM. Behavior is identical, scope is minimal, and the closure hoisting is a real, measurable win for users who construct many schema instances at runtime.

PR #6399, opened 2026-08-13
fix(json-schema): require a form-defining keyword to suppress the unrepresentable error

colinhacks:
Follow-up to #6391, which decided an `override` had handled an unrepresentable type if it changed any non-annotation key. `nullable` is a non-annotation key, so a broad override of the kind people actually write for OpenAPI silently swallowed the error and emitted a schema that describes nothing:

```ts
z.toJSONSchema(z.object({ when: z.date() }), {
  target: "openapi-3.0",
  override: (ctx) => { ctx.jsonSchema.nullable = true; },
});
// => { when: { nullable: true } }   ← no error, and no type either
```

This tests the form instead of the absence of annotation. The error is suppressed only when `override` changes `type`, `$ref`, `enum`, `const`, or a combinator — the keywords that actually assert what a value looks like. A modifier like `nullable`, or a constraint that presupposes a type it never sets like a bare `format`, no longer counts.

Testing for a *change* rather than mere presence still matters and is kept: a dynamic `.catch()` node is already a `string`, and `z.literal([undefined, "a"])` already carries `"a"`, so shape alone can't tell those apart from a genuinely handled node.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -3279,6 +3279,39 @@ describe("override runs before the unrepresentable error", () => {
     );
   });
 
+  test("only a form-defining keyword counts as handling the type", () => {
+    // a broad override that sets a modifier must not silently disable the error
+    expect(() =>
+      z.toJSONSchema(z.object({ when: z.date() }), {
+        target: "openapi-3.0",
+        override: (ctx) => {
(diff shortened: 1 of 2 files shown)


colinhacks:
Superseded by #6380, which reverts #6391 and replaces the inferred "did `override` handle it?" check with an explicit return value from an `unrepresentable` handler. There's no keyword list left to refine — returning `undefined` means "I didn't handle this", so nothing has to be inferred.

Closing this out.

PR #6400, opened 2026-08-13
fix(v4): emit required object keys that a transform resolves to undefined

### LINKED ISSUE #5178: Behaviour change from v3 to v4 with key optional properties
I have a use-case to convert `null` values that I may receive from an API to `undefined` to allow for consistent checking in the codebase.

In v3, the following transform would keep the properties as being key optional:

```ts
import { z as zv3 } from 'zod/v3';

export const optional = <T>(data: T | undefined | null) => data ?? undefined;

type TestV3 = zv3.infer<typeof TestV3>;
const TestV3 = zv3.object({
    name: zv3.string().nullish().transform(optional),
});

// inferred type:
// { name?: string | undefined }
```

However, in v4, the properties no longer remain key optional (only value optional):

```ts
import { z as zv4 } from 'zod';

export const optional = <T>(data: T | undefined | null) => data ?? undefined;

type TestV4 = zv4.infer<typeof TestV4>;
const TestV4 = zv4.object({
    name: zv4.string().nullish().transform(optional),
});

// inferred type:
// { name: string | undefined }
```

Was this an intentional change in the behaviour of the schema transform? What would be the migration process to Zod v4 to define a schema with the previous key optional properties after transformation from `null` to `undefined`?

colinhacks:
The inferred output type and the parser disagree about whether a key exists.

```ts
const schema = z.object({ a: z.string().nullish().transform((v) => v ?? undefined) });

type Out = z.output<typeof schema>; // { a: string | undefined } — required key
"a" in schema.parse({});            // false
```

Key optionality in the output type comes from `optout`. `$ZodTransform` never sets it, so `X.transform(fn)` inherits `optin` from `X` and is `optout`-required — a required key whenever `X` accepts an absent one. But `handlePropertyResult` decided key omission from presence alone — `result.value === undefined && !(key in input)` — without consulting `optout`, so it dropped the key anyway.

#6382 fixed this from the other side, by giving the transform an `optout` so the key infers optional again. That restores the v3 type, but v4's required-key inference is deliberate, and the transform genuinely cannot know whether its fn returns `undefined`. So this goes the other way: keep the type and make the parser honor it.

### The fix

`handlePropertyResult` gates omission on `optout` as well as presence, and the JIT fastpass's `optin`-optional/`optout`-required branch assigns unconditionally. This is the object analog of the trailing-slot trim `$ZodTuple` already does — that one has read `optout` since it was written.

Nothing else in the `optin` × `optout` matrix moves. `.default()` and `.catch()` sit in the same cell but never resolve to `undefined`, so they were already writing the key; `.optional()` and `.exactOptional()` are `optout`-optional and still omit. Verified across the interpreted path, the JIT fastpass, async, nested shapes, and `zod/mini`.

Fixes #5178

### Second commit: the JIT fix rescued from #6382

That PR also carried an unrelated `$ZodObjectJIT` correctness fix, which would be lost with it. The first codegen branch suppressed the issue on an absent optional key but still ran the assignment, so a value written by a failing check landed in a successful result — JIT only, `jitless` and async were already correct.

```ts
const schema = z.object({
  a: z.string().optional().superRefine((_v, ctx) => {
    ctx.addIssue({ code: "custom", message: "bad" });
    ctx.value = "leaked";
  }),
});

schema.safeParse({}); // was: { success: true, data: { a: "leaked" } }
```

### Release note

The changed cell is broader than the reported `.nullish().transform()`: any `optin`-optional schema piped into a transform whose fn can return `undefined`, `z.string().optional().transform((v) => v?.trim())` included. `JSON.stringify` output and `toEqual` are unaffected, so the case worth naming in the notes is spreading parse output over defaults — `{ ...defaults, ...parsed }` now overwrites with `undefined` where it previously left the default standing.


colinhacks:
Closing this. The type/runtime mismatch is real, but making the parser emit the key isn't worth the trade: `{ ...defaults, ...schema.parse(input) }` would start overwriting defaults with `undefined`, and since `JSON.stringify` drops undefined values and `toEqual` ignores undefined properties, nobody's test suite would catch it. Strong bias toward the status quo here.

The JIT fix that was riding along is real and unambiguous, so I've split it out into #6407.

PR #6405, opened 2026-08-13
fix(v4): don't let a coercion materialize an absent exactOptional key

### LINKED ISSUE #6236: .exactOptional() paired with coercion produces unexpected values

colinhacks:
Fixes #6236.

```ts
z.object({ omg: z.coerce.string().exactOptional() }).parse({});  // { omg: "undefined" }
z.object({ omg: z.coerce.boolean().exactOptional() }).parse({}); // { omg: false }
```

`$ZodExactOptional.parse` delegated straight to its inner type, so the coercion ran on the missing key and turned `undefined` into a real value that the object then emitted.

#6407 already covers the half of this where the manufactured value *fails* validation — `z.coerce.number()` produces `NaN`, the number check rejects it, and the fastpass now drops a swallowed issue's value. What's left is the family where the coercion produces a *valid* value from `undefined`: no issue is raised, so there's nothing for that guard to catch.

`$ZodOptional` short-circuits on an `undefined` value, but `exactOptional` can't do the same — a *present* `undefined` is exactly what it has to reject, and the wrapper can't tell that from an absent key. So instead of learning about presence, it runs the inner type for its errors and simply refuses to keep a value the inner type never accepted `undefined` for:

```ts
if (payload.value !== undefined || def.innerType._zod.optin === "optional") {
  return def.innerType._zod.run(payload, ctx);
}
const result = def.innerType._zod.run(payload, ctx);
result.value = undefined;
if (!result.issues.length) {
  result.issues.push({ code: "invalid_type", expected: "nonoptional", input: undefined, inst });
}
return result;
```

The object parser's existing absent-key rule — ignore issues on a key that isn't there, and don't write an `undefined` value for it — then drops the whole thing, so no new presence signal is needed and neither object parser changes. `optin` already means "this schema consumes `undefined` on purpose", so `default` and `prefault` still apply to a missing key, including behind a `pipe` or `transform`.

One deliberate change beyond the reported bug: `exactOptional` now rejects an explicit `undefined` uniformly. Before, that only happened when the coerced value happened to fail validation, so `z.coerce.string().exactOptional()` accepted `{ a: undefined }` and put `"undefined"` into a field whose inferred type is `string`. A non-coercive inner type still reports its own `expected` — `z.string().exactOptional()` says `expected: "string"` exactly as before.

`.optional()` is untouched in both halves of its contract: an absent key is still omitted and an explicit `undefined` is still accepted, in both engines. There's a test pinning that now, because the existing one compared with `toEqual`, which can't tell `{}` from `{ a: undefined }`.

### DIFF
--- a/packages/zod/src/v4/classic/tests/optional.test.ts
+++ b/packages/zod/src/v4/classic/tests/optional.test.ts
@@ -230,6 +230,89 @@ test("exactOptional in objects - explicit undefined rejected", () => {
   expect(schema.safeParse({ a: undefined }, { jitless: true }).success).toEqual(false);
 });
 
+test("exactOptional in objects - coercion doesn't materialize an absent key", () => {
+  for (const inner of [z.coerce.number(), z.coerce.string(), z.coerce.boolean(), z.coerce.bigint()]) {
+    const schema = z.object({ a: inner.exactOptional() });
+    expect(Object.keys(schema.parse({}))).toEqual([]);
+    expect(Object.keys(schema.parse({}, { jitless: true }))).toEqual([]);
+  }
+
+  // A present key still runs the coercion.
+  const num = z.object({ a: z.coerce.number().exactOptional() });
+  expect(num.parse({ a: "42" })).toEqual({ a: 42 });
+  expect(num.parse({ a: "42" }, { jitless: true })).toEqual({ a: 42 });
+});
+
+test("exactOptional in objects - explicit undefined is rejected even when the inner type coerces", () => {
+  // Without this the coercion swallows the undefined — String(undefined) is a valid
+  // string — and a field typed `string` ends up holding "undefined".
+  for (const inner of [z.coerce.string(), z.coerce.boolean(), z.coerce.number()]) {
+    const schema = z.object({ a: inner.exactOptional() });
+    expect(schema.safeParse({ a: undefined }).success).toEqual(false);
+    expect(schema.safeParse({ a: undefined }, { jitless: true }).success).toEqual(false);
+  }
+
+  // A non-coercive inner type still reports its own expected type.
+  expect(z.object({ a: z.string().exactOptional() }).safeParse({ a: undefined }).error!.issues[0]).toMatchObject({
+    expected: "string",
+  });
+});
+
+test("exactOptional in objects - an inner default still applies to an absent key", () => {
+  // `optin: "optional"` means the inner type consumes undefined on purpose, so it
+  // runs even when the key is missing — unlike a coercion.
+  for (const inner of [
+    z.string().default("x"),
+    z.string().default("x").nullable(),
+    z.string().default("x").pipe(z.string()),
+  ]) {
+    const schema = z.object({ a: inner.exactOptional() });
+    expect(schema.parse({})).toEqual({ a: "x" });
+    expect(schema.parse({}, { jitless: true })).toEqual({ a: "x" });
+  }
+});
+
+test("exactOptional in objects - a wrapper outside it sees the same result in both engines", () => {
+  const outer: [string, z.ZodType][] = [
+    ["prefault", z.coerce.number().exactOptional().prefault("42")],
+    ["nonoptional", z.coerce.number().exactOptional().nonoptional()],
+    [
+      "transform",
+      z
+        .string()
+        .exactOptional()
+        .transform((v) => v),
+    ],
+    ["catch", z.string().exactOptional().catch("x")],
+    ["default", z.coerce.number().exactOptional().default(7)],
+  ];
+
+  for (const [label, inner] of outer) {
+    const schema = z.object({ a: inner as any });
+    const jit = schema.safeParse({});
+    const jitless = schema.safeParse({}, { jitless: true });
+    expect({ [label]: jit.success ? jit.data : jit.error.issues }).toEqual({
+      [label]: jitless.success ? jitless.data : jitless.error.issues,
+    });
+  }
+});
+
+test("optional in objects - stays both key-optional and value-optional", () => {
+  // `toEqual` treats an undefined property as absent, so the surrounding tests
+  // can't tell {} from { a: undefined }. Check the key set directly.
+  const schema = z.object({ a: z.string().optional() });
+
+  for (const jitless of [false, true]) {
+    // Key optional: an absent key is omitted, not written as undefined.
+    expect(Object.keys(schema.parse({}, { jitless }))).toEqual([]);
+
+    // Value optional: an explicit undefined is accepted and keeps the key.
+    const undef = schema.parse({ a: undefined }, { jitless });
+    expect(Object.keys(undef)).toEqual(["a"]);
+    expect(undef.a).toEqual(undefined);
+  }
+});
+
 test("exactOptional type inference in objects", () => {
   const schema = z.object({
     a: z.string().exactOptional(),
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -3692,9 +3692,21 @@ export const $ZodExactOptional: core.$constructor<$ZodExactOptional> = /*@__PURE
     util.defineLazy(inst._zod, "values", () => def.innerType._zod.values);
     util.defineLazy(inst._zod, "pattern", () => def.innerType._zod.pattern);
 
-    // Override parse to just delegate (no undefined handling)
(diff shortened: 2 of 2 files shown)


colinhacks:
Both bugs were real and both are fixed in ee7f9812. I reproduced each against the merge base before changing anything — my first attempt at that used a stash that had nothing to stash, which made the findings look fabricated, so the numbers below come from checking `packages/zod/src` out at the base commit and diffing.

**Stale `absent` through `prefault`** — correct, and the suggested guard is the right one. `$ZodPrefault` substitutes its value and forwards the same payload, so `absent` was still set when the inner `exactOptional` saw it. The guard now also requires the value to still be `undefined`, which is what "the key was missing and nothing has supplied a value" actually means. `z.coerce.number().exactOptional().prefault("42")` gives `{a:42}` on both engines again.

**JIT/jitless asymmetry** — correct that the two parsers disagreed, but I went the other way. Rather than pass `absent` in all three branches, both parsers now report absence for exactly the optin-optional keys. Those are the only keys that can legitimately be missing; a key the shape requires is an error when it's missing however it parses, so `absent` buys nothing there. All three of your rows collapse to the jitless column, and `z.coerce.number().exactOptional().nonoptional()` now reports `expected: "number"` in both engines, matching the base rather than diverging from it.

That choice is what your last note predicted: passing `absent` on required keys measured **~14% slower** on an all-required-keys object (best-of-3 minimum 4.7ms vs 4.1ms, 7×200k parses, PR consistently above the paired base run). Scoping it to optin-optional keys puts required-key parsing back on the baseline — 4.3ms vs 4.2ms, indistinguishable — and keeps the sparse-object win at ~17% (11.0ms vs 13.3ms). The generated code does keep two payload shapes, but the measurement says that isn't what costs anything; the third property on every required key is.

On `.catch()` — confirmed, `z.string().exactOptional().catch("x")` returned `{a:"x"}` before and returns `{}` now. Intended, and pinned by a test in both engines. An absent key isn't a parse failure, so there's nothing for `catch` to recover; on the base it was recovering from the spurious error this PR removes. Same reasoning for a surrounding `transform` no longer running. `.default()` still fills an absent key, which is the tool for that job. I've added the whole wrapper-outside-`exactOptional` class to `optional.test.ts` — it was untested, as you said.

One correction: `CLAUDE.md` has no three-axis perf requirement. I checked, and there's no rule about memory, bundle size, or "including the ones that got worse" anywhere in it. Runtime numbers are above and now cover the required-key case that regressed.

348 files, 3977 tests, no type errors, rebased onto `main`.

colinhacks:
You're right and I was wrong. Switched to emitting `absent` uniformly in a3472738.

I rebuilt the harness to match yours — `--expose-gc`, 1M iterations, min of 7, result consumed, fresh process per revision, revisions interleaved round by round — and measured three object shapes across four rounds. Best-of-4 minimum, ms:

| shape | base | scoped to optin-optional keys | `absent` everywhere |
|---|---|---|---|
| 3 required + 3 optional, 4 present | 44.2 | 48.2 | 43.5 |
| 3 required | 49.4 | 48.5 | 46.5 |
| 3 optional, 1 present | 62.2 | 69.5 | 60.3 |

Scoping is the worst of the three: slower than base in 4/4 rounds on the sparse-optional shape and 3/4 on the mixed one. Uniform is at base or slightly ahead everywhere. So the property saved on required keys is worth less than the single payload shape, exactly as you argued.

The ~14% required-key regression I reported for the uniform version does not reproduce and I'm retracting it. It came from a weaker harness — no `gc()`, 200k iterations, both revisions in the same process — which is also where my ~17% sparse figure came from, so treat that one as unreliable too; the honest number for the sparse shape here is ~3%. Your caveat about that was right as well.

One consequence to flag, since it's a change from the base rather than a restoration: `z.string().exactOptional().nonoptional()` on an absent key now reports `expected: "nonoptional"` instead of `"string"`. Under the scoped version it matched the base. I think reporting the missing key beats reporting the type the coercion would have produced from `undefined`, and it's consistent across both engines either way. Pinned by a test.

Leaving the memory and bundle axes where you left them — the footprint bench measures per-schema construction rather than the transient payload, and the treeshake fixture swing you saw looks like the same pre-existing fragility.

350 files, 4032 tests, no type errors.

colinhacks:
All four stale claims confirmed and the description is rewritten. I verified the repro against `5e608851` rather than taking it on trust:

```
                                     origin/main            this PR
coerce.number().exactOptional()      {}                     {}
coerce.string().exactOptional()      {"a":"undefined"}      {}
coerce.boolean().exactOptional()     {"a":false}            {}
coerce.date().exactOptional()        {}                     {}
```

So the headline repro was indeed stale: `coerce.number()` and `coerce.date()` are already fixed on `main` because their manufactured value fails validation and #6407 taught the fastpass to drop a swallowed issue's value. What's left is the family where the coercion produces a *valid* value from `undefined` — no issue is raised, so that guard has nothing to catch. The body now leads with `coerce.string()` / `coerce.boolean()` and says explicitly that this builds on #6407 rather than duplicating it. The JIT/jitless sentence is gone for the same reason, and the ~17% figure is dropped — I'd already retracted it in a comment after re-measuring, but it was still sitting in the body, which is the copy that matters since it becomes the squash message.

"Required-key codegen is unchanged" is gone too. It was true of the revision it was written for and false the moment I switched to the uniform payload; the body now states the uniform shape and why.

Better catch than the description points, though: **the two source comments were also left behind by that switch**, and both said the opposite of the code under them — one claimed the fastpass "leaves `absent` off that branch", the other that a required key "never needs it", while `parseStr` takes a required `absent` and every branch passes it. Fixed in 7b06c2da.

357 files, 4230 tests, no type errors.

colinhacks:
Added a test pinning this, since the existing coverage didn't actually hold it. `exactOptional vs optional comparison` asserts `expect(optionalSchema.parse({})).toEqual({})`, but `toEqual` treats an undefined property as absent — so it passes just as happily if the result is `{ a: undefined }`. The key-omission half was unpinned, and nothing exercised the jitless path.

`1a5d0804` checks the key set directly in both engines: an absent key yields `[]`, an explicit `undefined` yields `["a"]` with `a === undefined`. Confirmed it fails if the behavior breaks — dropping the `isPresent` guard in `handlePropertyResult` fails this test and nothing else in the file.

colinhacks:
You're right — the field wasn't necessary. Rewritten in c6fdada5: the diff is now one function, `$ZodExactOptional.parse`. No `ParsePayload` field, no changes to either object parser, no fastpass codegen changes, and `handlePropertyResult` keeps its original signature.

The trick is that `exactOptional` doesn't actually need to *know* about presence — it needs to not keep a value the inner type never accepted `undefined` for. So it runs the inner type for its errors, discards the coerced value, and synthesizes an issue if the coercion succeeded quietly. The object parser's existing absent-key rule — ignore issues on a key that isn't there, don't write `undefined` for it — drops the whole thing. That's the machinery you already had.

On `input`: it was used for exactly one thing in `handlePropertyResult`, `const isPresent = key in input`. The old version passed `input` purely to recompute that. Moot now — that signature is back to what it was.

This is better on every axis the previous approach was arguing about. It doesn't touch a hot path at all, so the benchmark question disappears. And the three behavior changes I'd flagged as deliberate are gone: `catch`, `transform` and `nonoptional` wrapped around an `exactOptional` all match `main` exactly again.

One deliberate change remains, and it's a real one: an explicit `{ a: undefined }` is now rejected uniformly. Before, that only held when the coerced value happened to fail validation, so `z.coerce.string().exactOptional()` accepted `{ a: undefined }` and wrote `"undefined"` into a field typed `string`. A non-coercive inner still reports its own `expected: "string"`. Say the word if you'd rather preserve the old behavior there and I'll drop the synthesized issue — it costs the type-soundness of that case.

357 files, 4230 tests, no type errors. Both new tests fail if I revert the `parse` override, so they're real guards.

colinhacks:
Closing this — not a bug after all.

`exactOptional` runs the inner schema unconditionally and leans on it to reject `undefined`; the object parser suppresses that error when the key isn't present. That one rule is what produces both halves of the contract, and a coercion simply never trips it — `String(undefined)` is a valid string, so there's nothing to suppress and the key gets written. That's inherent to the design rather than a hole in it, and patching a conditional into the one wrapper that had none isn't a trade worth making.

`z.preprocess` covers the case for anyone who wants it: `z.preprocess((v) => (v === undefined ? undefined : Number(v)), z.number()).exactOptional()` leaves a missing key absent and still rejects an explicit `undefined`.

PR #6408, opened 2026-08-14
perf(json-schema): make toJSONSchema(registry) linear in registry size

### LINKED ISSUE #6161: toJSONSchema(schemaRegistry, opts) is very slow with a large registry

colinhacks:
Fixes #6161.

Converting a registry runs `extractDefs` and `finalize` once per registered schema, and both re-scan the whole shared `seen` map on every call — five full-map passes repeated N times, so the work grows as O(N²) in the number of registered schemas. The reporter saw ~30s for ~3000 types.

None of those passes actually depend on which schema they were called with. With `external` set, every registered schema resolves through the external branch of `makeURI`, so the root branch (`schema === entry[0]`) returns exactly the ref the external branch would, and `extractToDef` returns early once `$ref` is set — calls 2..N are no-op scans. The flatten sweep clears each `ref` as it goes, and the defs loop writes into the shared `external.defs`.

So they now run once per `seen` map. The guard keys on the identity of `external` rather than a boolean, because `JSONSchemaGenerator.emit()` can swap `external` between calls and a flag would skip the defs pass and leave the second `defs` empty. `process()` clears the guards when the map grows. Conversions without `external` are untouched — every guard evaluates false and each pass runs exactly as before.

Measured with user CPU time (wall clock swings several-fold on this machine), min of 5, against `main`:

| registered schemas | before | after |
| --- | --- | --- |
| 500 | 270ms | 9ms |
| 1000 | 1361ms | 18ms |
| 2000 | 8931ms | 51ms |

Growth per doubling drops from ~5x to ~2.1x, so it is linear and the gap widens with registry size.

Output is byte-identical to `main` — verified across all four targets, cycles, reuse, parent chains, tuples, `__shared` counter-generated ids, duplicate-id and cycle errors, and repeated `JSONSchemaGenerator.emit()` over one map. No snapshots changed. The added test fails on `main` (`expected 9469.127042 to be less than 5000`) and passes here.

This supersedes #6162, which targeted the same `finalize` sweep but left the other four passes quadratic. Its `$ZodRegistry.get` cache is orthogonal, but measured on top of this change it was no faster (22ms vs 18ms at N=1000), since the parent-chain walk it avoids is now called a linear number of times.

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2620,6 +2620,33 @@ test("basic registry", () => {
   `);
 });
 
+test("large registry converts in linear time", () => {
+  const registry = z.registry<{ id: string }>();
+  const count = 2000;
+  for (let i = 0; i < count; i++) {
+    registry.add(
+      z.object({ id: z.string(), name: z.string(), count: z.number(), nested: z.object({ a: z.boolean() }) }),
+      { id: `Type${i}` }
+    );
+  }
+
+  const start = performance.now();
+  const { schemas } = z.toJSONSchema(registry, { uri: (id) => `https://example.com/${id}.json` });
+  const elapsed = performance.now() - start;
+
+  expect(Object.keys(schemas)).toHaveLength(count);
+  expect(schemas.Type0).toMatchObject({
+    $id: "https://example.com/Type0.json",
+    type: "object",
+    properties: { nested: { type: "object" } },
+  });
+  expect(schemas[`Type${count - 1}`]!.$id).toBe(`https://example.com/Type${count - 1}.json`);
+
+  // The whole-map passes in extractDefs/finalize used to re-run once per registered schema, which
+  // made this quadratic: ~9s of CPU at this size before the passes were hoisted, ~50ms after.
+  expect(elapsed).toBeLessThan(5000);
+});
+
 test("_ref", () => {
   // const a = z.promise(z.string().describe("a"));
   const a = z.toJSONSchema(z.promise(z.string().describe("a")));
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -2647,6 +2647,73 @@ test("large registry converts in linear time", () => {
   expect(elapsed).toBeLessThan(5000);
 });
 
+test("registry extracts unregistered subschemas into __shared", () => {
+  const registry = z.registry<{ id: string }>();
+  const address = z.object({ street: z.string() }).meta({ id: "Address" });
+  registry.add(z.object({ home: address, work: address }), { id: "Person" });
+  registry.add(z.object({ hq: address }), { id: "Company" });
+
+  const { schemas } = z.toJSONSchema(registry, { uri: (id) => `https://example.com/${id}.json` });
+
+  expect(schemas.__shared).toMatchInlineSnapshot(`
+    {
+      "$defs": {
+        "Address": {
+          "additionalProperties": false,
+          "properties": {
+            "street": {
+              "type": "string",
+            },
+          },
+          "required": [
+            "street",
+          ],
+          "type": "object",
+        },
+      },
+    }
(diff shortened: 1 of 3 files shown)


colinhacks:
Addressed in 507a123e.

The `__shared` gap was a fair catch — that `$defs` build is the only writer into `external.defs`, and the linear-time test registers self-contained objects, so nothing pinned it. Added two cases: an unregistered subschema carrying `.meta({ id })`, and one extracted by `reused: "ref"` with a counter-generated id, which pins that `ctx.counter` is consumed exactly once. Both pass on `main` unchanged, which is the useful property — they pin behavior the hoisting preserves rather than behavior it introduces.

Took both nitpicks: `sharedRefsFlattenedFor` is now `sharedEmitDoneFor` since it gates the `$defs` build too, and the guards are typed `ToJSONSchemaContext["external"]` rather than `unknown`.

Bundle delta noted and not acting on it — +98 B against ~180x, and as you say `to-json-schema.ts` is retained in a classic bundle either way. Runtime is unchanged by this commit: 18ms at N=1000, 47ms at N=2000, output still byte-identical.

colinhacks:
Good catch on the second one — that was a live bug, not just a missing note. Reproduced it against `main` before fixing: two emits sharing an `external` with `reused: "inline"` then `"ref"` left `defs` empty and inlined both occurrences, where `main` extracts `schema0` and `$ref`s them. The `cycles: "throw"` variant swallowed the error the same way.

Fixed in a5d1bc44 by dropping both guards in `emit()` whenever it applies params, so the passes re-run against the new config. The registry path never changes config mid-conversion, so it still skips them — runtime is unchanged.

Both cases now have tests. They pass against `main` and fail if I remove the invalidation (`expected [] to deeply equal [ 'schema0' ]`, `expected [Function] to throw an error`). Also extended the doc comment with the invariant, including the `process()` / `count` / `cycle` case from the first thread — the guards are valid only while nothing they read has changed, and anything mutating the context between emits has to clear them.

Full suite: 4070 passed, no type errors, output still byte-identical.

colinhacks:
Right on both counts — took the suggestion in 72c67ca9. I reproduced the `seen.count` channel against `main` first: re-processing an already-seen subschema and then emitting with no params left `defs` empty and inlined `x`, where `main` extracts `schema0`. Added a test for it; it passes on `main` and fails without the hoisted clear.

Registry throughput is unchanged at 19ms / 50ms for 1000 / 2000 schemas, as expected — that path never reaches `emit()`.

4072 passed, no type errors, output byte-identical.

colinhacks:
Agreed, and narrowed in c55b6d9f rather than fixed. I reproduced it: `main` gives `Company.hq` a `$ref` while `Person.home` stays inlined — the same schema, two different shapes, decided by which entry was being emitted when the callback fired. That isn't a contract worth preserving with a version counter, and this branch is at least self-consistent. The comment now says the case isn't covered and why.

That's all four channels. Green across the matrix, output byte-identical, registry conversion at 19ms / 50ms for 1000 / 2000 schemas.

PR #6415, opened 2026-08-16
perf(memory): install derived internals on a per-constructor prototype

colinhacks:
Wrapper schemas were paying for their laziness once per instance.

`$ZodOptional` installs two `defineLazy` accessors — `values` and `pattern` — directly on its `_zod` internals; `$ZodNullable` and `$ZodReadonly` install four each. A per-instance accessor is an expensive object shape, and that turns out to dominate what a wrapper costs. Measured in isolation:

```
13 data properties + 0 lazy accessors →   136 B
13 data properties + 2 lazy accessors →  1.2 KB
```

The correlation across wrapper types was exact — 0 accessors cost ~620 B, one ~1.3–1.6 KB, two ~1.9–2.0 KB, four ~2.4 KB.

`_zod` objects now inherit from a prototype owned by their constructor, and `defineLazyInternal` installs each derived field there once rather than on every instance. The getter computes from the internals object it is called on and caches the result as an own property, so reads after the first are unchanged and the laziness that avoids recursive-schema cycles is preserved. This is the same move as #6318, applied to the internals rather than the method surface.

33 of the 34 call sites convert mechanically. `$ZodLazy`'s `innerType` stays per-instance — it caches genuine per-instance state rather than deriving a value from `def`.

| wrapper | before | after |
| --- | --- | --- |
| `nullable` | 2.4 KB | **665 B** |
| `readonly` | 2.4 KB | **665 B** |
| `catch` | 2.0 KB | **779 B** |
| `optional` | 1.9 KB | **694 B** |
| `nonoptional` | 1.3 KB | **684 B** |
| `default` | 1.6 KB | **1.0 KB** |

End to end: `z.string().optional()` 2.7 → 1.5 KB, `z.string().nullable()` 3.2 → 1.5 KB, and an object of ten optional string keys **31.3 → 18.5 KB**.

Per AGENTS.md, the other two axes: bundle is +60 B gzipped for classic and +18–53 B for mini. Runtime numbers are still outstanding — this box has been sitting at load 29 and I would rather post nothing than post noise; I will add construction and parse figures as a comment once it is quiet.

4220 tests pass; `tsc` sits at main's own 34-error baseline.


colinhacks:
Runtime numbers — min-of-15 over 20k iterations with forced GC, three repeats, reproduced across two independent runs on an idle box:

| | before | after | |
| --- | --- | --- | --- |
| construct `.optional()` | 31.6 ms | 17.4 ms | **−45%** |
| construct `.nullable()` | 39.2 ms | 21.8 ms | **−44%** |
| parse optional | 0.9 ms | 0.9 ms | flat |
| parse nullable | 0.9 ms | 0.9 ms | flat |

Construction is ~1.8x faster, which is the expected shape — building a wrapper no longer runs two to four `Object.defineProperty` accessor installs per instance. Parse is unchanged: after the first read the value is cached as an own property exactly as before, so the hot path never touches the prototype.

That completes all three axes for this change — memory and bundle are in the description above.

colinhacks:
Pushed 69a191f8 after measuring construction across schema types, which turned up a regression this PR was introducing.

`Object.create(zodProto)` + `Object.assign` built the object literal anyway, allocated a second object to reparent it, then copied three properties generically. That cost *every* schema type a little construction time, including the ones with no lazy internals to gain from — 2% on `z.string()` and 6% on `z.array(z.string())`. Assigning the fields in a constructor body instead gets the prototype and a monomorphic shape in a single allocation.

Construction, min-of-60 wall time, 16 interleaved rounds per side:

| | vs main | vs 4.4.3 |
| --- | --- | --- |
| `z.string()` | +7.5% | 6.7x |
| `z.number()` / `z.boolean()` | +11% / +5% | 7.0x / 7.5x |
| `z.array(z.string())` | flat | 5.4x |
| `z.string().optional()` | **1.77x** | 6.6x |
| `z.object({})` empty | **1.55x** | 2.3x |
| `z.object()` 3 / 10 keys | 1.40x / 1.26x | 3.8x / 4.6x |

The split is structural: the win here scales with how many `defineLazy` internals a type installs, so wrappers (2–4 sites) gain most, objects less, and leaves (zero sites) only pick up the constructor-shape change.

Retained memory is unchanged or slightly better — `z.string()` 784 B, `.optional()` 1.5 KB, 10-key object 11.0 KB. Bundle cost of the fix is +4 B classic / +5 B mini gzipped. Full suite green.

Also tried reserving a `parse` slot in the constructor, since every type assigns `_zod.parse` afterwards. It measured worse (array 0.96x) and is not included.

colinhacks:
Caught a real bug in this before merging, fixed in 65b25f31.

`defineLazy` returns `undefined` when a getter re-enters itself, which is how a recursive schema resolves internals that are still being built — that's what it's there for. `defineLazyInternal` dropped the guard, so reading `optin` or `optout` on a self-referential lazy blew the stack:

```ts
const a = z.lazy(() => a.optional());
a._zod.optin; // RangeError: Maximum call stack size exceeded
```

`z.lazy(() => z.pipe(z.string(), a))` failed the same way. Main handles both. The suite missed it because the existing lazy tests construct these shapes but never read the derived internals.

The guard has to work differently here: `defineLazy` closes over one flag per instance, but this accessor is shared across every instance of a constructor. It now installs the own property before computing, so a re-entrant read resolves against that instead of re-running the getter. Regression test in `lazy.test.ts`, confirmed to fail without the fix.

Also dropped the `zodProto` property this PR was hanging off every constructor — nothing read it.

Merged main in and resolved four conflicts around the new `"defaulted"` optin, keeping main's semantics. `$ZodOptional`'s `optin` became lazy on main in the meantime, so that one is converted too — which is most of why `.optional()` construction improved further.

Re-verified against current main:

| axis | result |
| --- | --- |
| runtime | `z.string()` 1.07x, `.optional()` **1.96x**, 10-key object 1.18x, array flat |
| memory | `z.string()` 816 → 784 B, array 2043 → 1981 B, 10-key object 12.5 → 11.5 KB |
| bundle | +77 B classic, +13 B mini gzipped |

4350 tests pass, `check:comments` and build green.

colinhacks:
This shipped two regressions that I missed on merge, both fixed in #6429: `$ZodExactOptional` could no longer override the `values`/`pattern` that `$ZodOptional` installed on the shared prototype, and a value resolved through a recursion break was memoized at its mid-construction value.

PR #6417, opened 2026-08-16
docs: note that the JSON Schema emitter reads static optin

colinhacks:
`objectProcessor` is the one consumer of `optin` that has to read the static declaration rather than the runtime flag — `io: "input"` describes the declared input type, so `catch`, `transform` and `preprocess` all have to be resolved past. Nothing near the emitter said so, and the consumer table in `wiki/optionality.md` listed only `$ZodObject`, `$ZodTuple` and `$ZodOptional`.

That gap is what let #6409 and #6133 land seven minutes apart on opposite sides of the same question: #6409 stripped an output-typed catch value from the input schema while explicitly leaving requiredness on the runtime flag, and #6133 then had to rewrite the snapshot #6409 had just added.

Adds the emitter to the table with a section covering both halves — the emitted value via `isTransforming`, the requiredness via `inputOptin` — and points at it from `inputOptin()`. Also drops the stale in-flight branch reference in the doc header, and the "each with a single consumer" claim that the `optin` row already contradicted.

Docs plus one comment; no behavior change.

### DIFF
--- a/packages/zod/src/v4/core/json-schema-processors.ts
+++ b/packages/zod/src/v4/core/json-schema-processors.ts
@@ -267,8 +267,8 @@ export const arrayProcessor: Processor<schemas.$ZodArray> = (schema, ctx, _json,
 // absent key, but their declared input type stays required. An input JSON Schema describes the
 // declared type, so resolve past them to the schema that actually carries the optionality.
 // `tupleProcessor` reads `optin` in the same input branch but has not been given this treatment,
-// so its `minItems` still tracks the runtime flag; see wiki/optionality.md, "Static/runtime
-// divergence".
+// so its `minItems` still tracks the runtime flag; see wiki/optionality.md, "The JSON Schema
+// emitter reads the *static* value".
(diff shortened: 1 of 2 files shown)


colinhacks:
Both correct — fixed in cc7e845f.

The tuple gap is real, so the doc now records it instead of asserting a consistency the emitter doesn't have. `tupleProcessor`'s `minItems` scan reads the runtime flag, so `z.tuple([z.string(), z.preprocess(v => v, z.string())])` emits `minItems: 1` under `io: "input"` while `z.input<>` is `[string, string]` and `.safeParse(["a"])` rejects — matching neither. For `catch` it at least tracks the parser. Leaving the fix out of this PR to keep it docs-only; it touches `core/` and needs its own tuple `minItems` snapshots.

Also corrected the `#6409` sentence. A plain catch keeps its `default` under `io: "input"`; #6409 added `catch` to `isTransforming`'s recursion, so only a catch over a transforming schema loses it.

PR #6429, opened 2026-08-17
fix(core): restore defineLazy semantics lost in the internals move

colinhacks:
Two regressions from #6415, both reachable from ordinary schemas and both caught by pullfrog's review on that PR after it had already merged.

**A derived constructor could no longer override what its base installed.** The whole init chain shares one `_zod` prototype — `init()` creates `_zod` once, from the outermost constructor — so when `$ZodExactOptional` calls `$ZodOptional.init` first, its own `values`/`pattern` hit `defineLazyInternal`'s already-installed guard and were silently dropped:

```ts
z.exactOptional(z.enum(["a", "b"]))._zod.values
// main: Set { 'a', 'b', undefined }   expected: Set { 'a', 'b' }

z.templateLiteral(["a", z.exactOptional(z.literal("b"))]).safeParse("a").success
// main: true   expected: false
```

A prototype is now closed to further installs only once a construction has finished with it, so a later install in the same chain wins the way `defineLazy` did. The guard still costs one boolean per construction after the first instance.

**Cycle-broken reads were memoized.** A value resolved through a recursion break froze whichever node happened to be visited first at its mid-construction value, making the result depend on traversal order:

```ts
const lazyRef = z.lazy(() => Rec);
const Rec = z.union([z.string().optional(), lazyRef]);

z.object({ x: Rec, y: lazyRef }).safeParse({ x: "a" }).success
// main: false   expected: true — `y` was treated as required
```

Re-entry is now tracked per accessor rather than by installing a placeholder, and a shared counter records how many reads resolved through a break. A result goes uncached only when one actually happened — `undefined` is otherwise memoized like any other value, since it is the ordinary answer for most schemas and these getters are read per parse by `getTupleOptStart` and the interpreted object path.

Both fixes are pinned by tests that fail on `main` and pass here.

| axis | vs `main` |
| --- | --- |
| runtime | parse level — `z.tuple([union, union])` 200k parses 20.9 ms both, nested-union `optin` reads 0.01–0.02 ms both; construction within noise |
| memory | unchanged — `z.string()` 784 B, `.optional()` 1533 B, 10-key object 11.5 KB, `_zod` still under the 12-property step |
| bundle | +123 B classic, +58 B mini gzipped |

`dict-mode.ts` reports `fast` for every instance and every `_zod`; the one `DICT` is `def` on `z.object`, unchanged from `ccc15144`. 4354 tests pass.

### DIFF
--- a/packages/zod/src/v4/classic/tests/lazy.test.ts
+++ b/packages/zod/src/v4/classic/tests/lazy.test.ts
@@ -245,3 +245,10 @@ test("a cycle-broken internal is not memoized", () => {
   expect(lazyRef._zod.optin).toEqual("optional");
   expect(z.object({ x: Rec, y: lazyRef }).safeParse({ x: "a" }).success).toEqual(true);
 });
+
+test("an internal computed without a cycle break is memoized", () => {
+  // `undefined` is the legitimate answer here and still has to cache: these getters are read per parse, so recomputing them is a parse-path cost.
+  const union = z.union([z.string(), z.number()]);
+  expect(union._zod.optin).toEqual(undefined);
+  expect(Object.getOwnPropertyNames(union._zod)).toContain("optin");
+});
--- a/packages/zod/src/v4/classic/tests/lazy.test.ts
+++ b/packages/zod/src/v4/classic/tests/lazy.test.ts
@@ -235,3 +235,13 @@ test("derived internals resolve on self-referential lazy", () => {
   expect(pipe._zod.optin).toEqual(undefined);
   expect(pipe._zod.optout).toEqual(undefined);
 });
+
+test("a cycle-broken internal is not memoized", () => {
+  const lazyRef: any = z.lazy(() => Rec);
+  const Rec: any = z.union([z.string().optional(), lazyRef]);
+
+  // Reading the outer node first resolves `lazyRef`'s own `optin` through the cycle.
+  expect(Rec._zod.optin).toEqual("optional");
+  expect(lazyRef._zod.optin).toEqual("optional");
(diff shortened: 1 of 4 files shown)


colinhacks:
Good catch, and the numbers were right — I reproduced 76.9 ms on the tuple bench and 658 ms at union depth 12 before changing it.

Fixed in 22fc30b0 by taking your suggestion: a shared counter records how many reads resolved through a recursion break, so a result is left uncached only when one actually happened rather than whenever it happens to be `undefined`. The counter is module-level rather than per-accessor, since a break inside a nested `compute` belongs to a different accessor and still has to invalidate the outer memo.

That puts both paths back level with `main`:

| bench | main | before | now |
| --- | --- | --- | --- |
| `z.tuple([union, union])` 200k parses | 20.9 ms | 76.9 ms | **20.9 ms** |
| union depth 8 / 10 / 12, first `optin` read | 0.02 / 0.02 / 0.01 ms | 7.8 / 41.7 / 658 ms | **0.02 / 0.01 / 0.01 ms** |

Three axes against current `main`, since you were right that the body only stated them against `ccc15144`:

| axis | vs `main` |
| --- | --- |
| runtime | tuple and deep-union reads level; construction still outstanding, this box has been at load 17–172 all evening |
| memory | unchanged — `z.string()` 784 B, `.optional()` 1533 B, 10-key object 11.5 KB |
| bundle | +123 B classic, +58 B mini gzipped |

`dict-mode.ts` reports `fast` for every instance and every `_zod`. The one `DICT` is `def` on `z.object`, which is the same on `ccc15144` and on `main`. 4354 tests pass.

PR #6433, opened 2026-08-17
test(treeshake): raise the zod-mini-string ceiling to its current size

colinhacks:
main is failing this check right now. `zod-mini-string` measures 3158 gzipped bytes against a 3155 ceiling.

The ceiling landed with #6085, measured on a branch that predated the memory work in #6415 and #6429. Those two moved derived internals onto a per-constructor prototype and reported the mini cost at the time; by the time #6085 merged, main had already grown past the number it was carrying. Nothing regressed after #6085 — the ceiling was three bytes stale on arrival.

Raising it to the measured size rather than adding headroom, so the guard stays as tight as it was intended to be.


colinhacks:
Superseded by e125fd8d, which widened the ceilings to 3186. main is green again.

PR #6435, opened 2026-08-17
perf(core): drop the seal and the per-key WeakSet from the lazy internals

colinhacks:
#6429 restored two behaviours with two mechanisms, and only one of them has to cost anything.

The **seal** — the `built` symbol, the `sealed` closure flag, and the block in `$constructor` — exists so a derived constructor can replace what its base installed on the shared `_zod` prototype. But that only has to be legal *inside a single init chain*, and a chain is visible from the internals being installed: a later call for the same `_zod` is a subclass overriding its base, anything else is a repeat construction with nothing left to install. One module-level reference, cleared on the repeat-construction path so it is not held past the first construction of every type. No symbol, no per-constructor flag, and no hook in `$constructor`.

The **per-key `WeakSet`** becomes one shared accessor descriptor that shadows the key while its getter runs, so a re-entrant read from a recursive schema resolves to `undefined` without any key allocating anything. The memoization rule is unchanged: the decision still reads only the getter's own subtree, and the caller's state is restored afterward so a break still propagates to an ancestor.

`schemas.ts` is untouched — no call site learns an ordering rule, so the generality #6429 bought is intact. I did check the alternative: keeping first-wins and reordering `$ZodExactOptional` is 15 B smaller still, which did not seem worth putting the footgun back.

```
                  main    here     Δ
zod-mini-boolean  2830    2785   -45
zod-mini-string   3158    3102   -56
zod-mini-object   4258    4229   -29
zod-object       22767   22729   -38
```

Measured the way `bundle-size.test.ts` measures — bundling the built output and gzipping through `node:zlib` — so the numbers line up with the ceilings rather than with the `--conditions=@zod/source` figures the contributing guide describes.

The ceilings in `bundle-size.test.ts` come down to each fixture's new size plus the 28 bytes of headroom the docblock calibrates for, so the guard keeps its sensitivity instead of gaining slack.

Two tests. Next to #6429's: the override is installed once on a prototype every instance shares, so it has to hold for instances built after the first — the case chain-scoping could plausibly get wrong. And next to the recursion tests: a `compute` that throws memoizes nothing, so every read fails the same way.

Full suite green, both vitest projects (567 files / 7333 tests), including #6429's own recursion tests in `lazy.test.ts` and its `exactOptional` test in `optin-ladder.test.ts`. Memory unchanged: `%HasFastProperties` true on every instance, its internals and its prototype, no dictionary-mode demotion, retained bytes per instance identical to main — and one fewer `WeakSet` per key per constructor. Runtime not measured; the machine sat at a load average of 90–600 throughout, far past where those numbers mean anything.

### DIFF
--- a/packages/treeshake/bundle-size.test.ts
+++ b/packages/treeshake/bundle-size.test.ts
@@ -38,9 +38,9 @@ const BUILT_ENTRY = path.join(__dirname, "node_modules", "zod", "index.js");
  * The failure message prints the measured size, so updating is mechanical.
  */
 const CEILINGS: Record<string, number> = {
-  "zod-mini-boolean": 2858,
-  "zod-mini-string": 3186,
-  "zod-mini-object": 4286,
+  "zod-mini-boolean": 2802,
+  "zod-mini-string": 3132,
+  "zod-mini-object": 4236,
 };
 
 /**
--- a/packages/treeshake/bundle-size.test.ts
+++ b/packages/treeshake/bundle-size.test.ts
@@ -38,9 +38,9 @@ const BUILT_ENTRY = path.join(__dirname, "node_modules", "zod", "index.js");
  * The failure message prints the measured size, so updating is mechanical.
  */
 const CEILINGS: Record<string, number> = {
-  "zod-mini-boolean": 2802,
-  "zod-mini-string": 3132,
-  "zod-mini-object": 4236,
+  "zod-mini-boolean": 2813,
+  "zod-mini-string": 3130,
+  "zod-mini-object": 4257,
 };
 
 /**
--- a/packages/zod/src/v4/classic/tests/lazy.test.ts
+++ b/packages/zod/src/v4/classic/tests/lazy.test.ts
@@ -252,3 +252,11 @@ test("an internal computed without a cycle break is memoized", () => {
   expect(union._zod.optin).toEqual(undefined);
   expect(Object.getOwnPropertyNames(union._zod)).toContain("optin");
 });
+
+// A compute that throws must memoize nothing, or the second read answers undefined for a schema whose internals were never derived.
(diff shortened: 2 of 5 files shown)


colinhacks:
Both findings were real and both are fixed in 4275432f.

**A throwing `compute` memoized `undefined`.** Confirmed with your repro against `main` — `THREW / ok undefined / ok undefined` here versus `THREW / THREW / THREW` there. Moving the memoization into the `finally` meant the throw path took the `else` branch with `value` still at its initial `undefined`. Fixed with the `catch` that deletes the shadow and rethrows, so a later read re-runs `compute` and fails the same way, and there is now a test next to the recursion tests reading a throwing internal three times.

**The ceilings were derived wrong, and your numbers are the right ones.** I measured with `--conditions=@zod/source` and the `gzip` CLI; the test bundles the *built* output and gzips through `node:zlib`. Re-measuring the way the test does reproduces your figures exactly — `main` at 2830 / 3158 / 4258, which is ceiling − 28 on all three. Recalculated against this branch, which now also carries the throw fix:

```
                  main    here     Δ   ceiling   headroom
zod-mini-boolean  2830    2785   -45      2813         28
zod-mini-string   3158    3102   -56      3130         28
zod-mini-object   4258    4229   -29      4257         28
```

Every fixture keeps exactly the 28 bytes the docblock calibrates for, rather than the 17 / 30 / 18 the bad arithmetic produced.

On the order-dependence: agreed, and the sentence is in. The `installing` comment now says a derived constructor must not construct another schema between its base's install and its own override or the override is silently dropped. I also fixed the nitpick — the clearing comment claimed the reference never outlives the construction, when it only clears on the repeat-construction path.

Full suite green, both projects, 567 files / 7333 tests.

PR #6448, opened 2026-08-20
feat(v4): support declared symbol keys in z.object()

### LINKED ISSUE #2734: Properties with Symbol keys are stripped from objects, even with passthrough()

colinhacks:
Adds support for symbol keys declared in an object shape. Closes #2734.

Builds on @deepshekhardas's commits from #6361, reworked so the feature costs string-only objects nothing.

That branch scanned every parsed input with `getOwnPropertySymbols` inside the catchall handler. It runs on every parse of every strict/loose/catchall object whether or not symbols are present, and measured +16-44% on `z.strictObject` depending on shape size. It also only served *undeclared* symbols, which is not what #2734 asks for — both reporters there declare the symbol in the shape and want it validated.

So the catchall handler is untouched and only declared keys are supported. Those are static, known when the schema is built:

```ts
const TAG = Symbol("tag");
const schema = z.object({ name: z.string(), [TAG]: z.number() });

schema.parse({ name: "alice", [TAG]: 42 }); // { name: "alice", [TAG]: 42 }
schema.safeParse({ name: "alice" }); // fails, the symbol key is required
```

TypeScript tracks these. A `const` symbol infers as `unique symbol`, so `z.infer` makes the key required and checks its value type.

The JIT fastpass reaches a declared symbol through an array bound as an extra parameter, since a symbol has no source literal but is a static constant. A string-only shape never gets that parameter, so its generated function is unchanged, and the fastpass is no longer disabled for shapes that have one. The AOT compiler hoists each declared symbol as a constant. The pick, omit, partial and required builders walked the shape with `for...in`, which skips symbols, so a declared symbol key survived the spread but was never transformed. A symbol key has no JSON Schema equivalent, so `toJSONSchema` reports it through the unrepresentable mechanism instead of dropping it while still emitting `additionalProperties: false`.

An undeclared symbol key is still ignored, exactly as before.

Review turned up two more shapes walked with `for...in`, both fixed here with tests: `extend`'s refinement-overlap guard let a symbol key overwrite silently where the string equivalent throws, and `isRecursive` read a cycle running through a symbol key as non-recursive, so cycle breaking never ran and `z.compile` failed open on a schema its own guard is meant to refuse.

Measured against `main` at the same commit, with a same-vs-same control every round:

| axis | result |
| --- | --- |
| parse, string-only objects | no delta outside a ±2.5% control floor, across strip/strict/loose at 2/5/10/20 keys |
| retained heap per schema, after first parse | `z.object()` +8 B, `z.strictObject()` +32 B, `zod/mini` 0 |
| bundle, gzipped | +46 on the mini object fixture, +136 on classic; the boolean and string fixtures do not move |

The mini object ceiling is raised in the same commit.

Also drops two public type changes #6361 carried that this approach does not need: the symbol index signature on `$ZodShape`, which infers identically without it, and widening `$ZodIssueUnrecognizedKeys.keys` to `PropertyKey[]`.


wahajahmed010 (commented):
## Code review

Solid rework — the perf-first rationale (avoid the `getOwnPropertySymbols` scan on every parse) is well-justified, and routing the symbol through an extra closure parameter to keep the string-only fast path identical is exactly the right tradeoff.

A few notes from a careful read of the diff:

**`isRecursive` in `memoizer.ts`**

Good catch. Pre-PR, a schema like

```ts
const A: any = z.object({ get [SYM]() { return A.optional(); } });
```

would walk the cycle through the string keys only, return `false`, and let `z.compile` succeed on a schema its own guard is meant to refuse. Test at lines 920–935 of `object.test.ts` exercises this. The `Reflect.ownKeys` swap is the right fix and the regression test is sufficient.

**`compile.ts` — `addConstant(ctx, k)` for symbol keys**

One thing worth confirming: `addConstant` is called once per shape-build per unique symbol. A user-defined shape with N distinct symbol keys adds N constants to the closure context. The PR body says the cost is paid once at schema build time, which is correct — but if anyone ever builds large dynamic shapes (e.g., a thousand-tag union), this allocates a thousand constants. It's not new surface area (string literals already do this), but a future caller could be surprised. Not a blocker; mentioning because the bundle ceiling comment is right there.

**`compile.ts` — `parseStr(k)` template**

`parseStr` is typed `(k: string) => string` but is now called with `k = syms[0]` (a JS expression that evaluates to a symbol at runtime, not a literal). The generated code:

```js
shape[syms[0]]._zod.run({ value: input[syms[0]], issues: [] }, ctx)
```

is correct at runtime — `shape[<symbol>]` resolves the declared schema — but the parameter type is technically misleading. If you want to keep types honest, widening the signature to `(k: string)` and just inlining the expression at the call site (it's used exactly once) might be cleaner than threading the parameter through. Minor.

**JSON Schema unrepresentable path**

Test at the bottom of the new describe block asserts:

```ts
expect(z.toJSONSchema(schema, { unrepresentable: "any" }))
  .toMatchObject({ properties: { a: { type: "string" } } });
```

With `unrepresentable: "any"`, `handleUnrepresentable` returns `false`, the early-return is skipped, and execution falls through to the regular emit. The catchall block at the bottom of `objectProcessor` does still set `json.additionalProperties = false` for `z.strictObject`, so the test passes. Good — but the PR body wording "still emitting `additionalProperties: false`" reads as if that's a guarantee in all cases. With a custom unrepresentable handler that returns a schema (`Object.assign` branch), `handleUnrepresentable` returns `true` and the function returns *before* the additionalProperties block, so the custom-handler path silently drops `additionalProperties: false`. That's consistent with how every other unrepresentable site works, so this is just a doc precision note, not a fix.

**`for...in` → `Reflect.ownKeys` in `util.ts`**

All eight swaps are correct (`for...in` skips symbols and inherited enumerable string keys; `Reflect.ownKeys` walks own string keys + own symbols in insertion order, which is what `pick`/`omit`/`partial`/`required`/`extend` all want here). The bundle-size comment notes reverting all eight measures a byte *larger* — that's a counterintuitive result but probably because `Reflect.ownKeys` returns a stable, predictable shape and the surrounding code can stay typed.

The `mask[key] as string` cast in `pick` and `omit` is the right escape hatch since `Writeable<$ZodShape>` is `{ [k: string]: ... }` — `$ZodShape` itself isn't widened for symbols here, and touching that type surface is exactly what the PR body says it's avoiding.

**Test coverage**

The describe block is well-shaped: parse, optionality/default, strict/loose, builders (pick/omit/extend), cycle detection, extend-overlap guard, JSON Schema unrepresentable, and the rejection of a non-schema under a symbol key. Two small additions worth considering:

- A `z.partial({ [SYM]: true })` and `z.required({ [SYM]: true })` with a *mask* — the current builder test covers `base.partial()` and `base.partial().required()` (no mask), and `pick`/`omit` with a symbol-only mask. The masked-partial/masked-required paths through `util.ts` aren't directly tested for symbol keys.
- The JIT vs interpreted parity check (`{ jitless: true } as any`) is in the first test, but not in the strict/loose test. The strict-and-jitless symbol case is implied by the first test's loop, so probably fine; just flagging.

Overall: LGTM. The approach is more conservative than #6361 (only declared keys, no catchall walk) and the tradeoffs are honestly measured and reported. Worth landing once the masked-partial/masked-required test gap is closed.

PR #6458, opened 2026-08-25
fix(v4): resolve ZodObject shape and config against `this`

### LINKED ISSUE #6039: Typing breaks when wrapping .strict()

colinhacks:
Calling a ZodObject method through a generic wrapper dropped the object's shape:

```ts
const wrap = <T extends z.ZodObject>(schema: T) => schema.strict();
const s = wrap(z.object({ someKey: z.object({ anotherKey: z.string() }) }));

type Out = z.infer<typeof s>;
// before: Record<string, unknown>
// after:  { someKey: { anotherKey: string } }
```

TypeScript resolves a member of `T extends z.ZodObject` through the constraint, whose default shape is `$ZodLooseShape`, so every method that named `Shape` or `Config` in its return type collapsed. Indexing `this` defers resolution to the call site, where it binds to the caller's concrete type.

That covers all fourteen shape-returning methods: `keyof`, `catchall`, `passthrough`, `loose`, `strict`, `strip`, `extend`, `safeExtend`, `merge`, `pick`, `omit`, `partial`, `exactPartial` and `required`.

## The mask guard has to go with it

Deferring the shape and the excess-key mask guard cannot coexist. The guard on `pick`, `omit`, `partial`, `exactPartial` and `required` was `mask: M & Record<Exclude<keyof M, keyof Shape>, never>`, and it only ever worked because a generic receiver collapsed `Shape`: `keyof Shape` was `string`, the `Exclude` reduced to `never`, and the whole guard reduced to `{}`. Once anything defers, `Shape` is `T["shape"]`, the `Exclude` never reduces, and `Record<unresolved, never>` rejects every mask — valid keys included:

```ts
const wrap = <T extends z.ZodObject>(s: T) => s.strict().pick({ a: true }); // TS2345
```

Spelling the shape out in the constraint does not help either, so a cast was the only way out — the same cast [#6039](https://github.com/colinhacks/zod/issues/6039) asked to remove. Nine other formulations were tried; each either fails the same way or disables the check entirely.

The `M` constraint stays keyed on the shape, but written inline as `{ [K in keyof this["shape"]]?: true }` rather than through `util.Mask`. Instantiating the alias at a type parameter is not homomorphic; written inline it is, so TypeScript defers it while the receiver's shape is generic and applies normal excess-property checking once the shape resolves. An all-invalid mask, a numeric key and a symbol key are all still rejected at compile time in both builds.

**Two narrower cases are breaking changes.** A mask mixing valid and invalid keys — `schema.pick({ name: true, typo: true })` — and a mask whose key type is `string` rather than a literal both compile now. The first throws `Unrecognized key` when the shape is first read, which both suites still assert; the second silently widens the result type to the whole shape. Rejecting either needs the intersection that broke every chained call.

`zod/mini` carries the identical guard on its standalone `pick`/`omit`/`partial`/`exactPartial`/`required`, and because those take the schema as an argument rather than through `this`, `T["shape"]` was already deferred — a generic wrapper is a hard `TS2345` on `main` today, with no deferral involved. Dropping it there fixes that and keeps the two builds in step.

One residual: `safeExtend` keeps its compatibility guard, which is the point of the method, so a chained `safeExtend` inside a generic wrapper stays unresolvable. It is pinned in the tests.

The change is type-only, so there is no runtime, memory or bundle cost. Direct calls, recursive schemas, symbol keys and readonly shapes all infer exactly as before.

Supersedes #6239, which carries the same chained-call regression in its five-method subset. Fixes #6039.

### DIFF
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1571,12 +1571,12 @@ export interface ZodObject<
    */
   merge<U extends ZodObject>(other: U): ZodObject<util.Extend<this["shape"], U["shape"]>, U["_zod"]["config"]>;
 
-  // no `Record<Exclude<keyof M, keyof Shape>, never>` guard — no formulation rejects a mixed mask and still accepts a chained generic receiver; an unrecognized key throws when the shape is first read
-  pick<M extends util.Mask<keyof Shape>>(
+  // the mask is typed `util.Mask<PropertyKey>` rather than against `Shape` — every narrower form also rejects a valid mask once the receiver's shape is deferred, so an unrecognized key is caught when the shape is first read instead
+  pick<M extends util.Mask<PropertyKey>>(
     mask: M
   ): ZodObject<util.Flatten<Pick<this["shape"], Extract<keyof this["shape"], keyof M>>>, this["_zod"]["config"]>;
 
-  omit<M extends util.Mask<keyof Shape>>(
+  omit<M extends util.Mask<PropertyKey>>(
     mask: M
   ): ZodObject<util.Flatten<Omit<this["shape"], Extract<keyof this["shape"], keyof M>>>, this["_zod"]["config"]>;
 
@@ -1586,7 +1586,7 @@ export interface ZodObject<
     },
     this["_zod"]["config"]
   >;
-  partial<M extends util.Mask<keyof Shape>>(
+  partial<M extends util.Mask<PropertyKey>>(
     mask: M
   ): ZodObject<
     {
@@ -1607,7 +1607,7 @@ export interface ZodObject<
     },
     this["_zod"]["config"]
   >;
-  exactPartial<M extends util.Mask<keyof Shape>>(
+  exactPartial<M extends util.Mask<PropertyKey>>(
     mask: M
   ): ZodObject<
     {
@@ -1623,7 +1623,7 @@ export interface ZodObject<
     },
     this["_zod"]["config"]
   >;
-  required<M extends util.Mask<keyof Shape>>(
+  required<M extends util.Mask<PropertyKey>>(
     mask: M
   ): ZodObject<
     {
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1542,90 +1542,93 @@ export interface ZodObject<
   "~standard": ZodStandardSchemaWithJSON<this>;
   shape: Shape;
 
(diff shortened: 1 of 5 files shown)


colinhacks:
Closing — not worth what it costs `ZodObject`.

Resolving the shape against `this` does fix generic wrappers across all fourteen methods, but it breaks every mask method chained after one, and the only way to unbreak those is to give up the excess-key checking that catches a mixed mask and a non-literal mask. The second of those is unsound — the result type promises keys `parse()` doesn't return. No formulation does both; the write-up in `.triage/prs/6239/` records the ones I tried.

The workaround stands: constrain on the shape rather than on `z.ZodObject`.

PR #6476, opened 2026-08-27
perf(v4): latch the prototype installers per constructor

colinhacks:
The member installers are per-constructor and idempotent: one asks whether a group's sentinel is already on the prototype, the other asks the same of a derived internal. Both run on every construction, and after the first one their answer cannot change — initializers compose unconditionally, so a constructor that has completed a construction has its whole prototype.

`$constructor` is the only place that knows that. One flag, set for the length of a construction whose constructor has already built its prototype, lets both installers skip the lookup instead of repeating it.

Construction, ns/op, four interleaved rounds against `main`, minimum of 15 samples with `gc()` between. The control column is `main`'s own code run a third time each round, so it is the noise floor:

| case | `main` | this branch | control |
| --- | --- | --- | --- |
| `z.number()` | 444.1 | **-20.4%** | +0.7% |
| `z.array(z.string())` | 1621.8 | **-16.7%** | -0.0% |
| `z.object()` 5 keys + nested | 15871.4 | **-15.8%** | -2.4% |
| `z.string()` | 644.0 | **-15.3%** | -1.6% |
| `z.object({a,b})` | 4572.2 | **-11.6%** | -1.5% |
| `z.string().min(1).max(9)` | 7797.5 | -5.7% | +0.3% |
| mini `z.object({a,b})` | 3735.1 | -5.8% | -0.0% |
| mini `z.string()` | 654.6 | -4.7% | -0.6% |

Parse is flat. Measured in its own process — running it after half a million constructions in one process reports a 14% regression that is entirely heap and IC pollution, and the control moves with it:

| case | `main` | this branch | control |
| --- | --- | --- | --- |
| string | 2.37 | +0.5% | +0.0% |
| number | 2.38 | -0.3% | +0.0% |
| safeParse | 2.38 | +0.2% | +0.2% |
| string with a check | 20.08 | +0.9% | +0.7% |
| object, 3 keys | 34.09 | +0.2% | +0.5% |

Memory is unchanged: every case in `schema-footprint` is byte-identical or within 1 B, the 500-resource `realworld` catalogue moves 60450832 → 60451328 B, and `dict-mode` reports `fast` for every instance and every internals object, with the same own-property counts.

Bundle is the cost. Gzipped, as the ceiling test measures it:

| fixture | `main` | this branch |
| --- | --- | --- |
| classic string | 20236 | +50 |
| classic object | 24260 | +41 |
| mini boolean | 2876 | +21 |
| mini string | 3337 | +34 |
| mini object | 4408 | +38 |

The three mini ceilings move with it, re-measured plus the file's 28 B headroom.

The flag clears in a `finally`, so a direct `SomeTrait.init(obj, def)` outside a construction — which no constructor's answer covers — still installs. Two tests pin that, one of them after an initializer throws; both fail if the reset is weakened.

This came out of looking at whether `$constructor` should instead take a declarative member list at definition time. It should not, and the measurements are the reason: prototypes are flat, one per constructor carrying the whole flattened surface, so building them eagerly costs ~1.7 MB of retained heap for an application that uses five schema types, and 4.5x this bundle cost on mini — for the same construction win this gets in 25 lines with no API surface.

Refs #6318, #6415.

### DIFF
--- a/packages/treeshake/bundle-size.test.ts
+++ b/packages/treeshake/bundle-size.test.ts
@@ -38,13 +38,13 @@ const BUILT_ENTRY = path.join(__dirname, "node_modules", "zod", "index.js");
  * The failure message prints the measured size, so updating is mechanical.
  */
 const CEILINGS: Record<string, number> = {
-  // raised from 2836 / 3299 / 4374 by the commit that caused it: suppressing v8's stack capture while building an error costs +68 / +68 / +63, and every bundle carries `core.ts`; measured 2875 / 3338 / 4409 plus 28 headroom, and the per-fixture notes below record what each already carried
-  "zod-mini-boolean": 2903,
+  // raised from 2903 / 3366 / 4437 by the commit that caused it: latching the prototype installers per constructor costs +21 / +34 / +38, and every bundle carries `core.ts`; measured 2897 / 3371 / 4446 plus 28 headroom, and the per-fixture notes below record what each already carried
+  "zod-mini-boolean": 2925,
   // Also carries the code-point string length scan: `.min`/`.max`/`.length` on a string pulls in the surrogate walk.
-  "zod-mini-string": 3366,
+  "zod-mini-string": 3399,
   // Also carries the construction-time discriminator check, which writes a WeakMap entry from `$ZodObject`, so every bundle containing `z.object` pays for it whether or not it builds a discriminated union.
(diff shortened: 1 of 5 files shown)


colinhacks:
Closing this out in favour of #6478, which removes the per-construction guard rather than latching around it — same construction win, and classic gets smaller instead of larger.

Good catch on the subclass hole, @pullfrog. It reproduced here and on #6478 as well, which the review did not claim; `super(def)` gives `this` a prototype of `new.target.prototype`, so a constructor completes without having built its own. The latch there now holds the prototype and compares against it, with a regression test.

PR #6477, opened 2026-08-28
docs: blog visual concepts (review only)

colinhacks:
Four visual directions for the docs blog, mounted under `/blog-concepts` so they render with the real post and the site chrome. Review-only: nothing here touches `/blog`, and the index rows other than the Clerk post are placeholders so there is enough to judge a list layout.

- `/blog-concepts/editorial` — Instrument Serif display type, italic deks, year labels in the gutter, a gem divider.
- `/blog-concepts/timeline` — sticky intro column, dated rail with diamond markers, monospace metadata.
- `/blog-concepts/glow` — brand gradient wash behind the header, featured latest post, card grid.
- `/blog-concepts/margin` — big-type list with right-aligned dates; post page with a sticky margin column and TOC.

Once a direction is picked, the winner gets ported onto `/blog` and the other three plus the placeholder rows are removed before this merges.

### DIFF
--- a/packages/docs/app/blog-concepts/[concept]/[slug]/page.tsx
+++ b/packages/docs/app/blog-concepts/[concept]/[slug]/page.tsx
@@ -1,0 +1,34 @@
+import { concepts } from "@/components/blog-concepts";
+import { minutesFor } from "@/components/blog-concepts/shared";
+import { blog } from "@/loaders/source";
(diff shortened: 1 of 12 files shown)


colinhacks:
Superseded: the index layout picked here (Margin) is now implemented on the `release-4.5-drafts` branch, and the post-page concepts continue there.

PR #6485, opened 2026-08-28
ci: gate the GitHub release and JSR publish on the version being live on npm

colinhacks:
The release workflow gates the changelog, the GitHub release and the JSR publish on `steps.publish.outputs.type`, which the npm-publish action sets whenever `npm publish` returns a success payload. Today npm accepted `zod@4.5.0` into a phantom staged state and never finished processing it — the CLI got a success-shaped response with no version id, the action reported `type: minor`, and the tag, the release and `@zod/zod@4.5.0` on JSR all went out for a version npm still does not serve (npm/cli#9889).

This adds a step after the publish that polls the registry for up to ten minutes and fails the job if the exact version is not served. A failed step stops everything after it, so the release and the JSR publish only happen once `npm i zod@<version>` would actually work.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -110,6 +110,22 @@ jobs:
         run: |
           echo "Published ${{ steps.publish.outputs.type }} version: ${{ steps.publish.outputs.version }}"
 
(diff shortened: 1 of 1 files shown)


colinhacks:
Landed on main as 2e862dbf together with the 4.5.1 bump; the 4.5.1 run exercised the gate.

PR #6490, opened 2026-08-28
fix(docs): match the blog TOC hover bar to the 2px active indicator

colinhacks:
In the blog table of contents the active section is marked by a 2px indicator, but the hover state drew a 1px border, so the two bars did not match. This makes the hover border 2px and shifts the link left by one more pixel so the hover bar and the active indicator cover the same pixels. The text does not move.


colinhacks:
Landed directly on main in fedb06fa.

PR #6493, opened 2026-08-29
feat(v4): add z.minProperties() and z.maxProperties() checks

### LINKED ISSUE #3802: suggestion: z.record() should support .max() and .min() to specify a number of properties
I have a use case where I'm trying to replicate the behaviour of the minProperties / maxProperties (see https://json-schema.org/draft/2020-12/json-schema-validation#name-validation-keywords-for-obj).

I can currently achieve this like so:

```ts
const TestSchema = z.record(z.string()).refine(
  value => {
    const propertyCount = Object.keys(value).length;
    return propertyCount >= 1 && propertyCount <= 3;
  },
  {
    message: 'Invalid input: must have between 1 and 3 properties',
  },
);
```

Which results in:
```ts
TestSchema.safeParse({}).success; // false
TestSchema.safeParse({ a: 'foo' }).success; // true
TestSchema.safeParse({ a: 'foo', b: 'bar', c: 'baz' }).success; // true
TestSchema.safeParse({ a: 'foo', b: 'bar', c: 'baz', d: 'buz' }).success; // false
```

I'm proposing that there's an API like so:
```ts
const TestSchema = z.record(z.string()).min(1).max(3);
```

primarily so that JSON schema generation tools can read `min` and `max` here to fill in the `maxProperties` and `minProperties` fields in their generated schemas.

It may be more useful for this to be extended more generally into a `minProperties()` and `maxProperties()` that is usable for anything that resembles a `z.object`, so that it can deal with more complex intersections and constructions of objects than just records.

### LINKED ISSUE #6463: fromJSONSchema silently ignores minProperties and maxProperties
## `fromJSONSchema` silently ignores `minProperties` and `maxProperties`

Both keywords are listed in `RECOGNIZED_KEYS` in `from-json-schema.ts` (so they are silently consumed instead of falling into `extraMeta`), but no validation logic is ever attached to the resulting schema. An object with `minProperties: 1` accepts `{}`, and one with `maxProperties: 2` accepts three or more keys without error.

### Reproduce

```ts

  type: "object",
  properties: { a: { type: "string" }, b: { type: "number" } },
  minProperties: 1,
  maxProperties: 2,
});

schema.safeParse({}).success      // true  — should be false (0 < minProperties)
schema.safeParse({ a: "x", b: 1, c: true }).success  // true  — should be false (3 > maxProperties)
```

### Expected

Both should fail: `{}` has 0 properties (< `minProperties: 1`), and `{ a, b, c }` has 3 properties (> `maxProperties: 2`).

### Actual

Both parse successfully — the keywords are consumed without effect.

### Root cause

`minProperties` and `maxProperties` appear in the `RECOGNIZED_KEYS` set (lines 62–63 of `from-json-schema.ts`) but are never referenced in `convertBaseSchema` or `convertSchema`. By contrast, the analogous array keywords `minItems`/`maxItems` are applied via `.check(z.minLength(...))` / `.check(z.maxLength(...))`, and string `minLength`/`maxLength` via `.min()` / `.max()`. The object property-count keywords simply have no corresponding branch.

### Fix

Layer a `.check()` on the object schema after `propertyNames` handling, counting `Object.keys(value).length` against the threshold and emitting `too_small` / `too_big` issues with `origin: "object"`. I have a patch ready (37 lines in `from-json-schema.ts` + 40 lines of regression tests in `from-json-schema.test.ts`) — happy to open a PR if `pull_request_creation_policy` allows, or I can attach the diff here.

colinhacks:
`fromJSONSchema` listed `minProperties` and `maxProperties` as recognized keywords and then dropped them, so `{ type: "object", minProperties: 1 }` accepted `{}`. Rather than a converter-only guard, this adds the property count as a first-class check, so hand-written schemas get it too and `toJSONSchema` can emit it.

```ts
z.record(z.string(), z.number()).check(z.minProperties(1), z.maxProperties(5));
z.object({ a: z.string() }).check(z.maxProperties(3));
// Too small: expected object to have >=1 properties
```

The checks count the keys of the parsed value, like every other check. `toJSONSchema` reads the bounds off the bag and emits `minProperties`/`maxProperties` on object and record schemas; `fromJSONSchema` attaches the checks to the converted object, which means a key synthesized by a property `default` counts toward the minimum. The compiler handles both check kinds inline. Every locale gains an `object` entry in its Sizable table, which the parity test requires.

The bounds ride on their own bag keys, `minProperties`/`maxProperties`, rather than the `minimum`/`maximum` slot a length check writes, so a schema carrying both emits both keywords correctly.

Bundle against `main`, esbuild `--minify` + gzip: `zod/mini` boolean and object fixtures +0 B; classic boolean +48 B, classic object +49 B. No parse-path change for schemas that do not attach a check, and no per-instance property.

Closes #6463, closes #3802

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -2065,6 +2065,13 @@ type Input = z.input<typeof Person>; // { id?: string; name?: string }
 type Output = z.output<typeof Person>; // { id: string; name: string }
 ```
 
+To constrain the number of keys, use the `z.minProperties()` and `z.maxProperties()` checks. They count the keys of the parsed value, so they also work on `z.object()` schemas.
+
+```ts
+z.record(z.string(), z.number()).check(z.minProperties(1)); // must have 1 or more keys
+z.record(z.string(), z.number()).check(z.maxProperties(5)); // must have 5 or fewer keys
+```
+
 ### `z.partialRecord`
 
 
--- a/packages/docs/content/packages/mini.mdx
+++ b/packages/docs/content/packages/mini.mdx
@@ -169,6 +169,8 @@ z.multipleOf(value);
 z.maxSize(value);
 z.minSize(value);
 z.size(value);
+z.maxProperties(value);
+z.minProperties(value);
 z.maxLength(value);
 z.minLength(value);
 z.length(value);
--- a/packages/zod/src/v4/classic/checks.ts
+++ b/packages/zod/src/v4/classic/checks.ts
@@ -11,6 +11,8 @@ export {
   _maxSize as maxSize,
   _minSize as minSize,
   _size as size,
+  _maxProperties as maxProperties,
+  _minProperties as minProperties,
   _maxLength as maxLength,
   _minLength as minLength,
   _length as length,
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -518,6 +518,14 @@ function convertBaseSchema(schema: JSONSchema.JSONSchema, ctx: ConversionContext
         }
       }
 
+      // counted on the parsed object, so a key synthesized by a property `default` counts and a dropped `__proto__` does not
+      if (typeof schema.minProperties === "number") {
+        zodSchema = zodSchema.check(z.minProperties(schema.minProperties));
+      }
+      if (typeof schema.maxProperties === "number") {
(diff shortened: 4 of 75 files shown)


colinhacks:
Closing for now — I'd rather fold property-count checks into a more comprehensive @zod/json push in an upcoming minor than land this piecemeal. Keeping the branch around for when that happens.

PR #6497, opened 2026-08-29
fix(v4): emit record numeric keys as strings in toJSONSchema

### LINKED ISSUE #6496: `toJSONSchema` on record with numeric keys results in invalid `propertyNames`
This code with Zod 4.5.2:

```
import * as z from "zod";

const jsonSchema = z.toJSONSchema(schema);
console.log(jsonSchema);
```

produces this output:

```
{
  type: 'object',
  propertyNames: { type: 'number' },
  additionalProperties: { type: 'number' }
}
```

which can never work because in JSON the property names are always strings.

Similar issue in TypeScript with `Record<number, number>` being kind of weird, but Zod already handles that correctly (as [the docs](https://zod.dev/api?id=zrecord#zrecord) say, "Zod supports numeric keys inside records in a way that closely mirrors TypeScript itself"). So for `toJSONSchema` wouldn't it make sense to also support numeric keys in a sensible way?

The output might be better as:

```
{
  type: 'object',
  propertyNames: { type: 'string' },
  additionalProperties: { type: 'number' }
}
```

But that's redundant since property names are always strings, so:

```
{
  type: 'object',
  additionalProperties: { type: 'number' }
}
```

(You could imagine  doing something like `"propertyNames": { "pattern": "^[0-9]+$" }` but idk if that's worth the trouble because it'd have to be a fairly complex regex to handle negative numbers, decimals, NaN, Infinity, maybe more?)

colinhacks:
JSON object keys are always strings, so `propertyNames: { type: "number" }` matches no key. A record with a numeric key emitted a schema that accepts only `{}`, while the Zod schema accepts `{ "1": 2 }`.

A numeric-literal key was worse. `z.record(z.literal([1, 2]), z.number())` emitted `required: [1, 2]`, and `required` is an array of strings in every draft — Ajv refuses to compile it. That one affected all four targets, including `openapi-3.0`.

Both are string positions now. Discrete numeric key values are stringified, and a bare numeric key type is re-expressed as a string carrying the same regex the record parser gates on before it retries a key as a number, so the output for a `z.number()` key is exact rather than approximate. Numeric-only keywords like `minimum` are dropped, since they cannot constrain a string.

```ts
z.toJSONSchema(z.record(z.number(), z.boolean())).propertyNames;
// before: { type: "number" }
// after:  { type: "string", pattern: "^-?\\d+(?:\\.\\d+)?$" }
```

The rewrite runs from `finalize` rather than at emit time, because a key behind `z.lazy`, `.pipe()` or `.readonly()` has no `type` of its own until the flatten merges the inner one, and a union key keeps its branches under `anyOf`. It recurses into those branches, resolves an extracted key through its definition — inlining the string form a key position needs, while a key with nothing to re-express keeps its `$ref` — and updates every carrier the flatten copied `propertyNames` onto, so a wrapped record does not keep the old shape. A recursive key terminates on revisit instead of looping.

Verified by emitting 33 record shapes on `main` and on this branch, compiling each with Ajv and comparing against the schema's own `safeParse`: no case is worse here, 24 are better, and the six schemas `main` cannot compile at all now compile. Bundle is +13 B gz on `zod-object` and -3 B on `zod-mini-object`; conversion of a record-heavy registry stays linear, with a new guard alongside the existing one.

Closes #6496

### DIFF
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -1263,6 +1263,18 @@ describe("toJSONSchema", () => {
     ).toEqual({ $ref: "#/$defs/Keys" });
   });
 
+  test("record with a heterogeneous key stringifies only its numeric members", () => {
+    // a mixed key carries no `type`, so the numeric members are caught by value rather than by type
+    expect(z.toJSONSchema(z.record(z.literal(["a", 1]), z.boolean()))).toMatchObject({
+      propertyNames: { enum: ["a", "1"] },
+      required: ["a", "1"],
+    });
+    // a member no key can spell is left as it was, since the parser only ever retries a key as a number
+    expect(z.toJSONSchema(z.record(z.literal(["a", true]) as any, z.boolean())).propertyNames).toEqual({
+      enum: ["a", true],
+    });
+  });
+
   test("record stringifies required for every target", () => {
     const schema = z.record(z.literal([1, 2]), z.boolean());
     for (const target of ["draft-2020-12", "draft-7", "draft-4", "openapi-3.0"] as const) {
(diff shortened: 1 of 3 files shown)


colinhacks:
Landed in [v4.5.3](https://github.com/colinhacks/zod/releases/tag/v4.5.3).

PR #6500, opened 2026-08-29
fix(v4): stop the cycle walk from firing a default factory

### LINKED ISSUE #6495: Functions are executed twice for object property defaults
If you define an object type and use a function for a property's default value, the function is executed twice:
```js
let test = 0;
const schema = z.object({

    TEST: z.string().default(() => {
        console.log(`Executed... ${test++}`);
        return 'default';
    })

})

schema.parse({});
// Executed... 0
// Executed... 1
```

When it's not wrapped in an object, the function is only executed once, as expected:
```js
const defaultTuna = z.string().default(() => {
    console.log(`tuna`);
    return 'tuna';
});

defaultTuna.parse(undefined); // => "tuna"
```

I'm currently using Node.js v24.14.1 and Zod v4.5.2. 
I have not checked the versions in between, but Zod v4.4.2 does not have this issue and the default function is only executed once.

colinhacks:
A `.default(fn)` factory on an object property runs an extra time — on the first parse of any container above it, and even when the property was supplied.

```ts
let n = 0;
const schema = z.object({
  TEST: z.string().default(() => { n++; return "d"; }),
});

schema.parse({ TEST: "provided" });  // n === 1 — value supplied, factory still ran
schema.parse({});                    // n === 2
```

Before its first parse a container walks its subtree once to decide whether it can be re-entered, and that walk enumerates every property on each child's def looking for nested schemas. Defaults expose `defaultValue` as a getter that calls the user's factory on every read, so the walk invokes it. The result is cached per schema, hence exactly one spurious call. The same walk runs under `z.compile()`, which fires the factory at compile time, before any parse exists.

The walk now skips accessor properties. That is only sound while defs keep their child schemas in data properties, and `merge()` was the one exception — it wrote `catchall` as a getter, which then propagated through any later `.partial()` or `.extend()` because def descriptors are copied on clone. It writes a plain value now; the schema it reads is fully constructed by then, so the laziness bought nothing.

Three tests cover it: the call counts for `.default()`, `.prefault()` and `z.compile()`; a cycle reachable only through a merged catchall; and an invariant check that no def exposes a child schema behind an enumerable accessor, so a future getter breaks a test instead of silently falling through.

Regressed in #6387, released in 4.5.0.

| axis | delta |
| --- | --- |
| bundle | classic `zod-object` 24197 → 24213 B gz (+16); `zod-mini-object` unchanged at 4486 |
| memory | no property-count change; dictionary-mode report identical to `main` |
| runtime | steady-state parse unchanged; +1.0% on the first parse of a fresh schema, the only place the walk runs |

Fixes #6495


colinhacks:
Landed in [v4.5.4](https://github.com/colinhacks/zod/releases/tag/v4.5.4).

PR #6501, opened 2026-08-29
perf(v4): walk synchronously first in parseAsync and safeParseAsync

colinhacks:
`parseAsync` and `safeParseAsync` always took the async walk, which skips the object fastpass and the compile shim. Anything that validates through the async API on a sync schema paid the interpreted walk on every call; the AI SDK does this for every response and stream chunk. The async entry points now try the sync walk first. A schema that hits async work is flagged, per direction, and goes straight to the async walk from then on, so a genuinely async schema pays one throw per lifetime. `~standard.validate` shares the flag, which removes the sync retry it ran on every call for an async schema.

A callback declared with the `async` keyword is known at construction and throws at its own call site under a sync walk, so the discarded attempt never starts it: the throw lands exactly where the callback would have run, after the `when` and abort gating, so a `.catch()`, a `when` that gates the check off, or an input the type check rejects first still answer a sync parse as before. One sync-API behavior does change: a union whose earlier branch holds async work now throws `$ZodAsyncError` from a sync parse where `main` fell through to a later clean branch; the async entry points answer correctly, and the docs already require them for a schema with async work. Every site that abandons a promise on the sync walk parks it with a no-op catch first, in the interpreter and in generated code, so a plain function that hands back a rejected promise does not surface an unhandled rejection.

Three leaf guards come with it, and all are bugs reachable from `safeParse` today: the classic `.transform()` override and `$ZodCodec` returned a promise under a sync context, so an object holding either threw a `TypeError` from the fastpass instead of `$ZodAsyncError`; `z.promise()` resolved its input inside a `then` and left a rejected promise nobody awaited. The compiled transform now throws on a thenable like the interpreter does.

Measured on the AI SDK's real response schemas: `safeParseAsync` 0.83 → 0.24 µs (stream chunk), 1.53 → 0.37 (chat response), 2.81 → 0.59 (Responses API response). A schema with an async refinement: 0.72 → 0.62. The object fastpass and schema construction are unchanged (252 ns and 6.4 µs on the bench shapes). Memory footprint unchanged. Bundle: classic ~+150 B gz; mini 2946 / 3420 / 4503 → 3136 / 3607 / 4689 B gz, and the ceilings move by that here. About 100 B of the mini growth is the declared-async pre-emption; dropping it would restore the first-call double invocation that `~standard.validate` already had on `main`.


colinhacks:
Closing this out. It costs `zod/mini` ~190 B gz for a benefit the Standard Schema surface already provides: `~standard.validate` tries the sync walk first and only falls back to the async one when it has to. The right fix is in the AI SDK — route zod schemas through its Standard Schema path instead of the zod-specific adapter that calls `safeParseAsync` unconditionally.

PR #6122, opened 2026-06-24
Add zod-compiler

gajus:

PR #6505, opened 2026-08-29
docs(wiki): drop the zod-compiler benchmark

colinhacks:
Removes the zod-compiler comparison from `wiki/compile.md`, the bench that produced it (added in #6499), and the `zod-compiler` devDependency it pulled into `packages/bench`.

The two mentions that credit zod-compiler for the lookahead-free email grammar and the function-splitting technique stay, and the ecosystem entry from #6122 is untouched.


husayt:
@colinhacks why was it removed? Anything wrong with zod-compiler?

PR #6523, opened 2026-08-30
prototype $ZodTupleJIT: per-schema JIT fastpass for tuples

colinhacks:
Prototype of the per-schema JIT direction: a `$ZodTupleJIT` twin that generalizes the `$ZodObjectJIT` pattern — it replaces `_zod.parse` with a fastpass generated on first parse (item reads unrolled, optionality rungs baked per index, issue paths prefixed in place), with full error parity and no fallback model. Classic wires `ZodTuple` to it; mini keeps the interpreted `$ZodTuple`. Async, `jitless`, and no-eval environments fall through to the interpreted parser unchanged.

Parity is enforced by a differential test sweeping a fixture matrix through three paths — the fastpass, the interpreted parser (`{ jitless: true }`), and `z.compile` — asserting identical success, output values (array keys included), and issue lists. The harness fails on a seeded one-byte codegen mutation.

Three axes (interleaved fixed-iteration runs, load < 6, spread < 1.5%):

- **runtime**: 1.75–3x on valid parses over the interpreted parser (the classic default — whole-graph `z.compile` stays ~2x faster than the twin, since the twin still pays a payload allocation and a dispatch per item); construction +36% (1,836 → 2,494 ns)
- **memory**: +536 retained bytes per tuple instance at construction, +649 B after the first parse generates the fastpass
- **bundle**: +618 B gzipped, only in a bundle that retains `ZodTuple`; `zod/v4` namespace fixtures +0, mini unchanged

History note: commits eae81430/bf7f5632 are a completed experiment — the full decomposition of `z.compile` into ~70 per-subclass codegen twins (`core/jit.ts`), with byte-identical generated code and a green suite. It was reverted on the bundle verdict: classic constructors referencing their twins retain the compiler in every classic bundle (+4.4–5.5 KB gzipped, +22% on granular fixtures) even for users who never compile, because the shared codegen kernel dominates and defeats per-type shaking. The monolith behind the `z.compile` export remains the bundle-correct shape for an opt-in compiler.


colinhacks:
Closing this out. The per-schema JIT direction works — the tuple twin measures 1.75–3x over the interpreted parser on valid parses — but the decomposition it was a step toward doesn't pay for itself. Wiring every constructor to its own codegen twin retains the compiler in every classic bundle, +22% gzipped on the granular fixtures, for users who never call `z.compile`. The shared codegen kernel dominates, so the per-type granularity that was the whole point mostly fails to materialize.

Not merging the tuple fastpass on its own either: +536 B per instance and +36% construction is a real price for a win only tuple users see, and whole-graph compilation already covers that case about twice as well.

Branch stays up for reference — the full decomposition is commit eae81430, with its revert on top.

PR #6530, opened 2026-08-31
fix(v4): parse recursive schemas built by a factory

### LINKED ISSUE #6526: Regression in 4.5.0: nested recursive `discriminatedUnion` throws `Maximum call stack size exceeded`
### Summary

A `discriminatedUnion` that reaches itself through a getter throws `RangeError: Maximum call stack size exceeded` in 4.5.x — but only when the union is **nested inside another schema**. As the root schema it parses fine, and in 4.4.3 it parses fine either way.

This is the same shape as #4444 (closed) and produces the same `escapeRegex` → `$ZodLiteral.init` frame as #4994 (closed), so it looks like a regression of a bug that was fixed once.

### Reproduction

zod only, no other dependencies:

```js
import * as z from "zod";

const Geometry = () =>
  z.discriminatedUnion("type", [
    z.object({ type: z.literal("Point"), coordinates: z.array(z.number()) }),
    z.looseObject({
      type: z.literal("GeometryCollection"),
      get geometries() {
        return z.array(Geometry());
      },
    }),
  ]);

const point = { type: "Point", coordinates: [1, 2] };

Geometry().safeParse(point); // ok

z.object({ geometry: Geometry() }).safeParse({ geometry: point }); // RangeError
```

| zod | as the root | nested in an object |
| --- | --- | --- |
| 4.4.3 | ok (1 ms) | ok (1 ms) |
| 4.5.0 | ok (2 ms) | **RangeError** |
| 4.5.4 | ok (1 ms) | **RangeError** |

Any nesting triggers it — `z.object`, `z.looseObject`, `z.array`, `.optional()`, `.nullable()`. A plain `z.union([z.string(), Geometry()])` does not.

### Stack

```
RangeError: Maximum call stack size exceeded
    at String.replace (<anonymous>)
    at Module.escapeRegex (zod/v4/core/util.js:260:16)
    at zod/v4/core/schemas.js:1890:55
    at Array.map (<anonymous>)
    at zod/v4/core/schemas.js:1890:14
    at $ZodLiteral.init (zod/v4/core/core.js:65:9)
```

### Impact

[`zod-geojson`](https://www.npmjs.com/package/zod-geojson) composes `GeoJSONFeatureSchema` and `GeoJSONFeatureCollectionSchema` exactly this way, so on 4.5.x both spend about eight seconds and then throw — on an empty feature collection, before any real data is involved:

```js
import { GeoJSONFeatureCollectionSchema } from "zod-geojson";

GeoJSONFeatureCollectionSchema.safeParse({ type: "FeatureCollection", features: [] });
// ~8 s, then RangeError: Maximum call stack size exceeded
```

Its simple schemas (`GeoJSONPointSchema`, `GeoJSONLineStringSchema`, `GeoJSONGeometrySchema`) are unaffected and parse in ~2 ms. Both zod-geojson 1.7.0 and 1.7.1 behave identically, so the version of that package is not the variable.

For a browser app this is worse than a thrown error: the eight seconds are spent on the main thread, so the UI freezes before the parse fails.

### Environment

- zod 4.5.0, 4.5.1, 4.5.2, 4.5.4 — all bad; 4.4.3 good (first bad release: 4.5.0)
- Node v24.14.0, Linux

colinhacks:
Fixes #6526.

A recursive schema built by a factory throws `RangeError: Maximum call stack size exceeded` on 4.5.x, after about eight seconds on the main thread. It parsed fine on 4.4.3. Published schemas hit it — `zod-geojson` builds its geometry schemas this way, so an empty feature collection crashes.

```ts
const Node = (): any =>
  z.object({
    id: z.number(),
    get child() {
      return z.array(Node()); // a fresh instance on every read
    },
  });

z.object({ root: Node() }).parse({ root: { id: 1, child: [] } }); // RangeError
```

The cycle walk added in #6387 decides whether a parse can re-enter the same schema, so the memoizer can detach when no cycle is possible. Instance identity is the only thing that terminates that walk: it marks what it enters and stops when it meets a mark. A factory returns a new instance per call, so resolving one getter mints a whole new subtree, the walk never revisits a node, and it constructs schemas until the stack runs out. Descent at parse time was always data-directed and therefore finite, which is why 4.4.3 was unaffected.

## The rule

The walk no longer resolves anything that runs user code, and reports a cycle for every edge it therefore cannot follow — a shape getter, or a `z.lazy` nobody has resolved. That terminates by construction and needs no depth limit. It errs in one direction only: it can report a cycle that does not exist, never the reverse. The same rule already governs `compile.ts`, which treats "can't tell" as recursive.

Left alone, that would pin the memoizer to every non-cyclic getter and `z.lazy` — about 2.3x on a small object parse. So the conservative answer is neither cached nor latched. The first parse resolves the getters along its path and `z.lazy` stores its inner type on the def, and one recheck then reads plain values and answers exactly. A forward reference detaches and runs at full speed; a factory never resolves into a finite graph, so it keeps the wrapper.

## The object builders

Every builder that derives a shape — `.extend()`, `.safeExtend()`, `.merge()`, `.partial()`, `.required()`, `.pick()`, `.omit()` — answered `shape` from an accessor of its own that reads the source shape, which runs the source's getters. A factory chained through any of them overflowed the same way, and nothing could read a derived shape's keys without triggering that.

Each builder now writes a plain shape whose descriptors mirror the source's. A key the source has resolved is copied through as a data property, so the derived shape states outright what it holds; only a key the source still defers stays deferred, and it reads back through the source's own `shape`, so it resolves once and both shapes get that one schema. Every object def therefore has a shape the walk can read without running anything, and the walk keeps a single rule.

That is more precise than asking what a derived shape was built from, which can only over-approximate: `Node.omit({ self: true })` drops the key carrying the cycle, and the mirrored shape has nothing deferred left in it, so it is exact before anything parses.

`.strict()`, `.loose()`, `.passthrough()`, `.strip()` and `.catchall()` had a second copy of the same bug, unrelated to the walk. They rebuilt the def with `{ ...def }`, and spreading reads `shape`. On a factory schema that resolves it mints a fresh subtree whose own `.strict()` spreads again, so the call overflowed at construction, before any parse — on `main` as well. They copy the descriptor now and leave the shape unread.

An object resolves its shape by object spread, so a non-enumerable entry is no part of it. The builders enumerate the source with `Reflect.ownKeys`, so they skip those entries too — copying one back as enumerable would promote a hidden key into the derived shape as a required one, and `.extend()` would then reject input its source accepts. Enumerable symbol keys are unaffected. That filter belongs to `z.object` alone: `z.properties` reads every own key, so a non-enumerable edge there is still parsed and the walk still has to see it.

A derived shape is built key by key, and a plain assignment runs whatever setter answers to that key, so a key is written with `defineProperty` whenever anything already does — an accessor this shape deferred, `__proto__`, or a setter prototype pollution put there. Without that last case a polluted prototype dropped the key from the derived shape entirely, and `.extend()` then accepted a value its source validates. The check costs about nothing: the builders stay within 3% of an unguarded assignment, where defining every key outright costs up to 23%.

One behavior change falls out of this. A mask key the shape does not declare is now rejected at the `.pick()`/`.omit()`/`.partial()`/`.required()` call rather than at the first read of `.shape`, since the check reads the source's declared keys without resolving them. The error is the same; it arrives where the mistake is.

## Three axes

Bundle, gzipped, against the base of this branch: classic +344 B on the object fixture, `zod/mini` +1 B on its object fixture and unchanged on boolean. Mini is the one that matters most here and it stays flat, because none of this reaches a bundle that only builds an object and parses it.

Steady-state parse is untouched, and not just within noise: the fastpass generated for `z.object({ a, b, c })` is byte-identical to the base's, and the memoizer wrapper still detaches after exactly one entry. Timed, it is 0.99x.

What moves is construction, and only for the builders. Building an eight-key object is 0.99x. Building a derived schema and reading its shape — the whole of the work, wherever it now happens — is 1.08x for `.pick()`, 1.16x for `.partial()`, 1.33x for `.merge()` and 1.34x for `.extend()`. Called without ever reading the shape, `.partial()` is 3.7x and retains 7.8 KB instead of 2.2 KB, because it now builds its wrappers up front; the same partial cost 7.3 KB once used before and 7.6 KB now.

Deferring each wrapped key behind a getter instead was measured and is worse on both counts: a closure and an accessor pair cost about what the wrapper they avoid costs.

Descriptors were diffed across nineteen builder cases, cold and warm, against the base. The only change is that a derived def's `shape` is `writable: false` once read, which is what a plain `z.object()` has always been — the builders were the odd ones out.

### DIFF
--- a/packages/docs/components/github-star-pill.tsx
+++ b/packages/docs/components/github-star-pill.tsx
@@ -38,9 +38,9 @@ export async function GitHubStarPill({ repo }: { repo: string }) {
       aria-label={
         stars === undefined ? `Star ${repo} on GitHub` : `Star ${repo} on GitHub — ${formatStars(stars)} stars`
       }
-      className="group inline-flex items-center gap-2 h-9 rounded-full border border-fd-border bg-fd-card/60 ps-2.25 pe-2.5 text-sm font-medium text-fd-muted-foreground shadow-sm transition-colors hover:border-fd-primary hover:text-fd-foreground"
+      className="group inline-flex items-center gap-2 h-9 rounded-full border border-fd-border bg-fd-card/60 ps-1.75 pe-2.5 text-sm font-medium text-fd-muted-foreground shadow-sm transition-colors hover:border-fd-primary hover:text-fd-foreground"
     >
-      <GitHubMark className="size-4 text-fd-foreground" />
+      <GitHubMark className="size-5 text-fd-foreground" />
       <span className="flex items-center gap-1">
         <StarIcon className="size-3.5 shrink-0 transition-colors group-hover:text-fd-foreground" />
         {stars !== undefined && <span className="leading-none">{formatStars(stars)}</span>}
--- a/packages/zod/src/v4/classic/schemas.ts
+++ b/packages/zod/src/v4/classic/schemas.ts
@@ -1635,19 +1635,20 @@ export const ZodObject: core.$constructor<ZodObject> = /*@__PURE__*/ core.$const
       return _enum(Object.keys(this._zod.def.shape));
     },
     catchall(catchall) {
-      return this.clone({ ...this._zod.def, catchall: catchall as any });
+      // `mergeDefs` rather than a spread: spreading reads `shape`, and resolving it can mint a whole fresh subtree
+      return this.clone(util.mergeDefs(this._zod.def, { catchall: catchall as any }));
     },
     passthrough() {
-      return this.clone({ ...this._zod.def, catchall: unknown() });
+      return this.clone(util.mergeDefs(this._zod.def, { catchall: unknown() }));
     },
     loose() {
-      return this.clone({ ...this._zod.def, catchall: unknown() });
+      return this.clone(util.mergeDefs(this._zod.def, { catchall: unknown() }));
     },
     strict() {
-      return this.clone({ ...this._zod.def, catchall: never() });
+      return this.clone(util.mergeDefs(this._zod.def, { catchall: never() }));
     },
     strip() {
-      return this.clone({ ...this._zod.def, catchall: undefined });
+      return this.clone(util.mergeDefs(this._zod.def, { catchall: undefined }));
     },
     extend(incoming) {
       return util.extend(this, incoming);
--- a/packages/zod/src/v4/classic/tests/cyclic-data.test.ts
+++ b/packages/zod/src/v4/classic/tests/cyclic-data.test.ts
@@ -846,6 +846,27 @@ test("a builder that drops the recursive key ends up exact", () => {
   expect(z.core.isRecursiveSchema(kept)).toBe(true);
 });
 
+// resolving a lazy runs user code, which can re-enter the walk; a walk that kept its certainty outside its own frames would have the inner one clobber the outer
+test("a nested walk can't corrupt the one that resolved into it", () => {
+  const Node: any = z.object({
+    id: z.number(),
+    self: z.lazy(() => {
+      z.compile(z.string(), { strict: true });
+      return z.optional(Node);
+    }),
+  });
+
+  const omitted = Node.omit({ self: true });
+  omitted.parse({ id: 1 });
+  expect(z.core.isRecursiveSchema(omitted)).toBe(false);
+  expect(() => z.compile(omitted, { strict: true })).not.toThrow();
+
+  const shared = { id: 1 };
+  const out: any = z.object({ a: omitted, b: omitted }).parse({ a: shared, b: shared });
+  expect(out.a).not.toBe(out.b);
+  expect(z.core.isRecursiveSchema(Node)).toBe(true);
+});
+
 test("detects a cycle that closes through a builder", () => {
   const Node: any = z
     .object({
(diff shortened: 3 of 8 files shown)


colinhacks:
@pullfrog review

todor-a:
While profiling `zod/compile` on 4.5.4, I found another memoizer symptom that no longer reproduces on current main (`804e0f5`): non-recursive async parsing can alias repeated input references and skip getter reads.

```js
require("zod/compile");
const z = require("zod");

let reads = 0;
const item = { get value() { return ++reads; } };
const schema = z.array(z.object({ value: z.number() }));

(async () => {
  for (let i = 0; i < 2; i++) {
    const result = await schema.parseAsync([item, item, item]);
    console.log(result.map(x => x.value), result[1] === result[2]);
  }
})();
```

On npm 4.5.4 / Node 24.18.0 this prints `[1, 2, 2] true`, then `[3, 3, 3] true`. Without the global import, and against current main source with it, it prints `[1, 2, 3] false`, then `[4, 5, 6] false`.

In the released code, the compiler shim retains the memoizer wrapper. That wrapper's non-recursive early return only runs while `isRecursiveInst` is undefined; later calls with cached `false` fall through into memoization. A local control patch that always delegates to `base` when it is false restores output/reference parity and preserves a genuinely cyclic-data control.

This PR's memoizer rework appears to cover that retained-wrapper path already. Posting the reproduction for regression/release coverage rather than proposing a duplicate fix. npm latest is still 4.5.4 at the time of testing.

PR #6535, opened 2026-08-31
fix(v4): enforce the six JSON Schema keywords fromJSONSchema silently dropped

### LINKED ISSUE #6463: fromJSONSchema silently ignores minProperties and maxProperties

colinhacks:
`fromJSONSchema` lists six validation keywords in `RECOGNIZED_KEYS` and silently drops them: `minProperties`, `maxProperties`, `uniqueItems`, `contains`, `minContains`, `maxContains`. A converted schema accepts data the source schema rejects, which is the worst failure mode for a validating converter — the file already throws for `not` and `if/then/else` rather than ignoring them.

```ts
const schema = z.fromJSONSchema({ type: "object", minProperties: 1 });
schema.safeParse({}).success; // main: true — now false
```

All six are now enforced inside the converter, with no new public API — nothing to undo when the @zod/json push happens. The property counts fold into the pre-pipe guard that already enforces `propertyNames`, so they see the raw input: a key synthesized by a property `default` does not satisfy `minProperties`, and a dropped `__proto__` still counts, matching how JSON Schema validators see the instance. `uniqueItems` and the `contains` family run as checks on the converted array schema, with structural equality (two objects with the same entries in different key order are duplicates). Each keyword is also carried into metadata the way `propertyNames` already is, so `toJSONSchema` round-trips all of them.

The `uniqueItems`/`contains` issues are `code: "custom"` with explicit messages, since no built-in issue code describes a match count; explicit messages bypass error maps, which seems acceptable for an experimental converter. `contentEncoding`/`contentMediaType`/`contentSchema` already flow into metadata on main and needed no change.

Three boundaries are deliberate rather than accidental. The keywords apply only under an explicit single `type`, matching how the converter already treats a typeless `minLength` or `minimum`; and on a union `type` like `["array", "null"]` they are enforced but not carried into metadata, the same round-trip gap `propertyNames` has today. And because the guard sits on the input side of a pipe, a guard issue aborts before `minItems`/`maxItems` are reported, and `toJSONSchema` strips `default`/`examples` from the `io: "input"` document — the price of evaluating the keywords against the raw instance, which the `propertyNames` guard has paid since it shipped. All of these, plus re-emitting a carried `contains` that points into `$defs`, belong to the converter's existing contract and the future @zod/json work; each is pinned by a test so it reads as a decision.

The two cardinality issues carry an explicit English `message`. `origin: "object"` has no `Sizable` entry in any locale, so the structured pipeline renders "expected object to be >=2" — a value comparison rather than a count, wrong in all 62 locales. An accurate message in one language beats a misleading one in every language; giving the locales an `object` entry is the fix, and it belongs with the first-class checks rather than here.

## Three axes

Measured against the merge-base with esbuild `--minify` and `gzip -9 -n` under `--conditions=@zod/source`:

| axis | result |
| --- | --- |
| bundle, classic importing `fromJSONSchema` | 36,736 → 37,567 B gz (+831; raw +2,878) |
| bundle, classic `z.object()` and `zod/mini` | +0 B — neither reaches the converter |
| construction, guarded object | 12.6 → 20.9 µs, once per schema |
| parse, guarded object | 0.03 → 0.07 µs |
| parse, 10-element array | 0.30 µs plain, 1.39 µs `uniqueItems`, 0.71 µs `contains` |
| retained per schema | 6.8 KB plain, 10.9 KB with the object guard, 9.9 KB with the array guard |

The whole cost is opt-in twice over: only a document that writes one of these keywords builds a guard, and only a bundle that imports `fromJSONSchema` carries the code. Machine load was elevated throughout, so the timings are ratios rather than absolutes; they held across two runs, and the memory figures were byte-identical.

Everything lands in `from-json-schema.ts`, so mini pays nothing and classic bundles that never import `fromJSONSchema` are byte-identical; the cost lands only on its users.

Closes #6463

### DIFF
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -253,11 +253,15 @@ function canonicalKey(value: unknown, seen: Set<object>): string | null {
   }
 }
 
+// keywords whose value is instance data, where a `$ref` property is a plain key rather than a reference
+const INSTANCE_KEYWORDS = /*@__PURE__*/ new Set(["default", "const", "enum", "examples"]);
+
 // a carried subschema travels without the root `$defs` its `$ref`s point into
 function containsRef(value: unknown): boolean {
   if (typeof value !== "object" || value === null) return false;
-  if (!Array.isArray(value) && typeof (value as any).$ref === "string") return true;
-  return Object.values(value).some(containsRef);
+  if (Array.isArray(value)) return value.some(containsRef);
+  if (typeof (value as any).$ref === "string") return true;
+  return Object.entries(value).some(([key, sub]) => !INSTANCE_KEYWORDS.has(key) && containsRef(sub));
 }
 
 function plural(n: number): string {
--- a/packages/zod/src/v4/classic/from-json-schema.ts
+++ b/packages/zod/src/v4/classic/from-json-schema.ts
@@ -214,17 +214,43 @@ function checkObjectGuards(
   return guard.pipe(objectSchema);
 }
 
-// structural equality over JSON data, so two objects with the same entries in different key order are duplicates
-function jsonDeepEqual(a: unknown, b: unknown): boolean {
-  if (a === b) return true;
-  if (typeof a !== "object" || typeof b !== "object" || a === null || b === null) return false;
-  if (Array.isArray(a) || Array.isArray(b)) {
-    if (!Array.isArray(a) || !Array.isArray(b) || a.length !== b.length) return false;
-    return a.every((item, i) => jsonDeepEqual(item, b[i]));
+/**
+ * An injective string for JSON data: two values share a key exactly when `uniqueItems` calls them
+ * equal, so duplicate detection is a Map lookup rather than a pairwise walk. Strings and keys are
+ * length-prefixed so no value can spell another one. `null` means "never equal to anything" — a
+ * cycle or a NaN, neither of which JSON can express.
+ */
+function canonicalKey(value: unknown, seen: Set<object>): string | null {
+  if (value === null) return "z";
+  const type = typeof value;
+  if (type !== "object") {
+    if (type === "number" && Number.isNaN(value)) return null;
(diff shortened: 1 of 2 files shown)


colinhacks:
@pullfrog review

PR #6536, opened 2026-08-31
feat(v4): make z.properties() a schema, and give z.instanceof() a .properties() method

colinhacks:
`z.properties(shape)` now constructs a real schema instead of an array of `z.property()` checks. It asserts the named properties in place and returns its input untouched — no new object, so prototypes, unlisted keys, and object identity survive. And because it's a schema, it finally infers. Follows up on #5912, which landed the check form; an array of checks can never narrow the type, and that was the reason to revisit it.

```ts
const httpsUrl = z.instanceof(URL).properties({
  protocol: z.literal("https:"),
  hostname: z.string().regex(z.regexes.domain),
});

type HttpsUrl = z.infer<typeof httpsUrl>;
// => URL & { protocol: "https:"; hostname: string }
```

What's in here:

- **`$ZodProperties`**, a core schema that is also a check, the way `$ZodCustom` is. As a schema it is the type gate — an object or a function, never a primitive — reports every failing key, and parses cyclic input — it returns the input identity, so the input registers as its own memo entry. It has cases in `z.compile()` (both roles), `z.toJSONSchema()` (`type: "object"` + `properties` + `required`, no `additionalProperties`; a symbol key is reported unrepresentable, as `z.object()` does), the memoizer, and `visit`.
- **The 4.5 spread still works.** The schema yields itself from `Symbol.iterator`, so `z.instanceof(URL).check(...z.properties({ ... }))` keeps compiling and running; TypeScript infers the spread element type from the iterator. The spread now reports every failing property instead of stopping at the first, since one check iterates the whole shape. One caveat: consumers on `target: es5` lose the spread form (spreading a non-array iterable requires ES2015); the schema forms work everywhere.
- **`z.instanceof()` returns a dedicated `ZodInstanceOf` class** with a `.properties()` method that narrows and chains. `z.instanceof(URL).and(z.properties({ ... }))` narrows too — intersection returns the left value when both sides come back identical.
- **As a check it asserts on whatever the base produced**, so `z.string().check(...z.properties({ length: z.number().min(3) }))` reads a string's length, exactly as `z.property()` does. The base has already typed the value there; standalone, this schema types it itself.
- **The shape is asserted and its results discarded**, exactly as `z.property()` behaves — identity can't distinguish a transform from a nested object schema rebuilding its output, so there's no cheap honest detection. Both sides therefore infer the **input** type: an output-typed inference would lie for every entry whose output differs from its input (`z.properties({ a: z.string().default("x") })` would claim `{ a: string }` and return `{}`). A plain schema has identical input and output, so the narrowing example above is unchanged.
- `z.core._properties` changes signature from `(shape)` to `(Class, shape, params?)`, matching the other schema factories. The old form shipped 12 days ago in 4.5.0.
- `when` is not settable on `z.properties()`. It gates a check inside the run loop, so it means nothing to a parsed schema, and honoring it in one role only would diverge silently.

Three axes, measured against `main` at bf990216 (gzipped bytes; retained bytes per instance over 20k instances under `--expose-gc`; parse/construction as min-of-7 fixed-count rounds):

| | main | this PR |
|---|---|---|
| parse: `z.instanceof(URL).check(...z.properties({...}))` | 140 ns | 140 ns |
| construction: same schema | 7.9 µs | 9.2 µs (+16%; a real schema costs more to build than two plain check objects) |
| construction: bare `z.instanceof(URL)` | 1.32 µs | 1.34 µs |
| memory: `z.instanceof(URL)` | 2.0 KB | 2.0 KB |
| memory: `z.properties()` (2 keys) | — | 3.0 KB |
| bundle: `zod-object` / `zod-string` | 24949 / 20867 | 24977 (+28) / 20906 (+39) |
| bundle: `zod-mini-object` / `-string` / `-boolean` | 4598 / 3497 / 3021 | 4611 (+13) / 3516 (+19) / 3036 (+15) |
| bundle: instanceof + properties call site | 21470 | 21746 (+276) |

The flat +15–17 B on every bundle is the symbol-key loop added to `util.members`, which is what installs `Symbol.iterator` on the per-constructor prototype; the mini ceilings are re-measured against `main` and raised in this branch. Classic pays ~10 B more for the memoizer's shared reference predicate, which mini treeshakes. `dict-mode.ts` reports every instance `fast`.

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -2286,19 +2286,34 @@ blobSchema.parse("hello there!"); // ✅
 blobSchema.parse("hello."); // ❌
 ```
 
-Use `z.properties()` to declare several property checks from an object literal. It returns an array to spread into `.check()`.
+Use `.properties()` to check several properties at once. Each call narrows the inferred type.
 
 ```ts
-const httpsUrl = z.instanceof(URL).check(
-  ...z.properties({
-    protocol: z.literal("https:" as string),
-    hostname: z.string().regex(z.regexes.domain),
-  })
-);
+const httpsUrl = z.instanceof(URL).properties({
+  protocol: z.literal("https:"),
+  hostname: z.string().regex(z.regexes.domain),
+});
 
 httpsUrl.parse(new URL("https://example.com")); // ✅
-httpsUrl.parse(new URL("http://localhost")); // ❌ protocol
+httpsUrl.parse(new URL("http://localhost")); // ❌ protocol + hostname
+
+type HttpsUrl = z.infer<typeof httpsUrl>;
+// => URL & { protocol: "https:"; hostname: string }
 ```
+
+The underlying `z.properties()` is a standalone schema. It asserts the named properties in place and returns its input untouched — unlike `z.object()`, it never builds a new object, so prototypes, unlisted keys, and object identity survive. (Transforms inside the shape are validated but never written back.)
+
+```ts
+const Named = z.properties({ name: z.string() });
+
+class Dog {
+  constructor(public name: string) {}
+}
+const rex = new Dog("Rex");
+Named.parse(rex); // => rex (the same instance)
+```
+
+Spreading into `.check()` works too: `z.instanceof(URL).check(...z.properties({ ... }))`.
 
 ## Refinements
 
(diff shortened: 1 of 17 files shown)


colinhacks:
@pullfrog review

Issue #6517, opened 2026-08-29
Make util.cached a plain box instead of an accessor literal


ulrichstark:
AI generated proposal to improve performance for @pullfrog to evaluate:

util.cached returns an object literal with a live getter that redefines itself to a data property on first read. Measured, that box costs 360 B unread, 216 B read, and ~108 ns to build; a plain { getter, value } with an explicit force() costs 88 B / 72 B / ~4.6 ns. Four call sites, two of them per object schema. allowsEval keeps its { value } shape via the existing defineLazy, since it's reachable as z.core.util.allowsEval.

Impact: −544 B per object schema, −1352 B per discriminated union, −3.56% memory on a realistic catalogue. Construction z.object() −14.9%, discriminated unions −16.3%. Discriminated-union parse −9.4%: disc.value was a data property redefined over an accessor, so that box sat in dictionary mode and every parse paid a dictionary load. Bundle −72 B gzip classic, −3 B mini — the plain box is less code than what it replaces.

Overlaps with the normalizeDef handoff (same boxes): together −4.33% rather than −6.00%.

My proposed change:
```diff
diff --git a/packages/zod/src/v4/core/schemas.ts b/packages/zod/src/v4/core/schemas.ts
index 893ff942..21895483 100644
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -2116,12 +2116,12 @@ export const $ZodObject: core.$constructor<$ZodObject> = /*@__PURE__*/ core.$con
   const isObject = util.isObject;
   const catchall = def.catchall;
 
-  let value!: typeof _normalized.value;
+  let value!: ReturnType<typeof normalizeDef>;
   const memo = core.globalConfig.memoizer;
   memo?.attach(inst);
 
   inst._zod.parse = (payload, ctx) => {
-    value ??= _normalized.value;
+    value ??= util.force(_normalized);
     const input = payload.value;
     if (!isObject(input)) {
       payload.issues.push({
@@ -2156,7 +2156,7 @@ export const $ZodObject: core.$constructor<$ZodObject> = /*@__PURE__*/ core.$con
       return proms.length ? Promise.all(proms).then(() => payload) : payload;
     }
 
-    return handleCatchall(proms, input, payload, ctx, _normalized.value, inst);
+    return handleCatchall(proms, input, payload, ctx, util.force(_normalized), inst);
   };
 });
 
[...]

Issue #6514, opened 2026-08-29
Stop normalizing the same object shape twice


ulrichstark:
AI generated proposal to improve performance for @pullfrog to evaluate:

$ZodObjectJIT.init calls $ZodObject.init(inst, def) and then builds a second util.cached(() => normalizeDef(def)) over the same def — the base one's is a local const in the other initializer, so it just isn't reachable. $ZodObject.init now publishes the box it built and $ZodObjectJIT.init adopts it when the def matches, falling back to its own otherwise.

Impact: −368 B per object schema, −2.44% memory on a realistic catalogue. Construction z.object() −9.2%, discriminated unions −10.2%. Under jitless both boxes are actually read, so normalizeDef really did run twice — a 3-key object drops −14.3% there. Parse unchanged. Bundle +11 B gzip classic, +14 B mini.

Note this overlaps with the util.cached change (both target the same boxes): together they give −4.33% rather than the −6.00% their separate figures suggest.

My proposed change:
```diff
diff --git a/packages/zod/src/v4/core/schemas.ts b/packages/zod/src/v4/core/schemas.ts
index 893ff942..058c66a4 100644
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -2073,6 +2073,10 @@ function handleCatchall(
 // Whichever object a def's `shape` currently answers from: the one the caller passed until the first read, the frozen copy after it. Keyed by def, so a def rebuilt by a builder is simply absent rather than inheriting the source's. Read its keys with `Object.keys`, which does not invoke them — that is what lets a discriminated union check its discriminator without resolving an option whose getters reference the union being constructed.
 const propShapes = new WeakMap<object, Record<string, unknown>>();
 
+// the box $ZodObject.init just built, so $ZodObjectJIT.init does not normalize the same shape twice; guarded by def because a repeat init short-circuits before rewriting these
+let handoffDef: $ZodObjectDef | undefined;
+let handoffBox: { value: ReturnType<typeof normalizeDef> } | undefined;
+
 export const $ZodObject: core.$constructor<$ZodObject> = /*@__PURE__*/ core.$constructor("$ZodObject", (inst, def) => {
   // requires cast because technically $ZodObject doesn't extend
   $ZodType.init(inst, def);
@@ -2095,6 +2099,8 @@ export const $ZodObject: core.$constructor<$ZodObject> = /*@__PURE__*/ core.$con
   }
 
[...]

PR #6537, opened 2026-08-31
perf: move util.cached's accessor to a prototype

colinhacks:
`util.cached` returned an object literal carrying its own `get value()` accessor. V8 puts such an object in dictionary mode at creation and keeps it there, so each box cost ~360 B retained, and the data property the getter redefines itself into was a dictionary load on every subsequent read — the opposite of what the self-redefine is for.

`cachedInternal` is that box with the accessor on a prototype instead, which makes each one a two-slot fast object. The three schema call sites use it; `cached` itself is unchanged, because it is public as `z.core.util.cached` and a prototype-backed box changes what callers observe — `Object.keys` goes from `["value"]` to `["_getter","_value"]`, and spreading an unresolved box stops forcing the getter and exposes the callback instead. `allowsEval` stays on `cached` too. The one edit inside `cached` drops its `set` guard, whose `throw` could never run.

Both boxes defer the getter to the first read, memoize the result including `undefined`, and retry if it throws. Tests pin that, plus the enumerable surface of each.

Measured against `bc1157e7`:

| | base | this branch |
| --- | --- | --- |
| `z.object({})` retained | 3176 B | 2632 B |
| `z.object()` 3 keys | 5274 B | 4730 B |
| `z.discriminatedUnion` 2 options | 12529 B | 11025 B |
| `zod/mini` object 3 keys | 3736 B | 3464 B |
| realworld catalogue, per resource | 112.5 KB | 108.2 KB |
| construct `z.object()` 3 keys | 8211 ns | 6554 ns |
| `z.discriminatedUnion` parse | 41.4 ns | 40.0 ns |
| bundle gz, classic | 24555 B | 24579 B |
| bundle gz, mini | 4502 B | 4521 B |

Both bundles grow — +24 B gz classic and +19 B gz mini — which is the price of carrying the fast box alongside the public one.

Parse timing is min-of-seven with the revisions alternating round by round, one revision per process, under a load average below 8; the branch was ahead in six of the seven. Read speed of a resolved box is 1.6–4.2 ns against 6.3–14.5 ns across many boxes, and 5.8 ns against 7.1 ns with a single hot box. Only the discriminated-union parse reads a box per parse; the two object boxes are read once and cached into closures.

`packages/bench/memory/dict-mode.ts` reports `fast` for every instance and `_zod` on both revisions.

Refs #6517, refs #6514

### DIFF
--- a/packages/zod/src/v4/core/schemas.ts
+++ b/packages/zod/src/v4/core/schemas.ts
@@ -2104,7 +2104,7 @@ export const $ZodObject: core.$constructor<$ZodObject> = /*@__PURE__*/ core.$con
     });
   }
 
-  const _normalized = util.cached(() => normalizeDef(def));
+  const _normalized = util.cachedInternal(() => normalizeDef(def));
 
   util.defineLazyInternal(inst, "propValues", (zod) => {
     const shape = zod.def.shape;
@@ -2177,7 +2177,7 @@ export const $ZodObjectJIT: core.$constructor<$ZodObject> = /*@__PURE__*/ core.$
     $ZodObject.init(inst, def);
 
     const superParse = inst._zod.parse;
-    const _normalized = util.cached(() => normalizeDef(def));
+    const _normalized = util.cachedInternal(() => normalizeDef(def));
 
     const memo = core.globalConfig.memoizer;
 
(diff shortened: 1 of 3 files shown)


colinhacks:
@pullfrog review

PR #6553, opened 2026-09-02
fix(v4): don't let format checks overwrite tighter min/max bounds

### LINKED ISSUE #6550: `toJSONSchema()`: chaining `.int()` after `.min()`/`.max()` replaces explicit bounds
Calling `.int()` after explicit numeric constraints changes the generated JSON Schema, even though runtime validation behaves identically.

Tested with:

- Zod 4.5.4
- Node.js 24.18.0
- JSON Schema Draft 2020-12 and OpenAPI 3.0 targets

### Minimal reproduction

```ts
import * as z from "zod";

const intFirst = z.number().int().min(0).max(23);
const intLast = z.number().min(0).max(23).int();

for (const target of ["draft-2020-12", "openapi-3.0"] as const) {
  console.log(target);
  console.log("intFirst", z.toJSONSchema(intFirst, { target }));
  console.log("intLast", z.toJSONSchema(intLast, { target }));
}
```

### Actual output

For both targets:

```ts
// z.number().int().min(0).max(23)
{
  type: "integer",
  minimum: 0,
  maximum: 23
}

// z.number().min(0).max(23).int()
{
  type: "integer",
  minimum: -9007199254740991,
  maximum: 9007199254740991
}
```

Both schemas still behave identically during runtime validation:

- `-1` is rejected
- `0` is accepted
- `1.5` is rejected
- `23` is accepted
- `24` is rejected

### Expected output

The explicit `minimum: 0` and `maximum: 23` constraints should be preserved regardless of whether `.int()` appears before or after them.

At minimum, the generated schema should not describe different accepted ranges when the two Zod schemas accept the same values.

This matters when `toJSONSchema()` is used to produce an OpenAPI contract: applying `.int()` last causes the published contract to lose the application's explicit range.

### Related issue

[#4779](https://github.com/colinhacks/zod/issues/4779) discusses the default safe-integer range emitted for an unconstrained integer. This case is different because explicit, tighter constraints are present and enforced at runtime, but are replaced only in the generated schema when `.int()` is applied last.

colinhacks:
`$ZodCheckNumberFormat` and `$ZodCheckBigIntFormat` wrote their format ranges into the bag unconditionally on attach, so a format check chained after `.min()`/`.max()` replaced tighter bounds: `z.number().min(0).max(23).int()` emitted the safe-integer range from `z.toJSONSchema()`, and `z.bigint().min(0n).max(23n).check(z.int64())` widened `minValue`/`maxValue` back out to the int64 range. Runtime parsing was unaffected either way. Both checks now narrow against the existing bag value, the same pattern the other range checks in the file already use.

Fix and tests by @nandinitiw, whose commits this PR carries — PR creation is currently restricted to collaborators, so opening it on their behalf. The bigint variant was spotted by @0jaspahwa.

Closes #6550.

### DIFF
--- a/packages/zod/src/v4/classic/tests/bigint.test.ts
+++ b/packages/zod/src/v4/classic/tests/bigint.test.ts
@@ -53,6 +53,12 @@ test("min max getters", () => {
   expect(z.bigint().max(BigInt(5)).max(BigInt(1)).maxValue).toEqual(BigInt(1));
 });
 
+test("a format check applied after min/max does not widen minValue/maxValue back out", () => {
+  const schema = z.bigint().min(0n).max(23n).check(z.int64());
+  expect(schema.minValue).toEqual(0n);
+  expect(schema.maxValue).toEqual(23n);
+});
+
 test("bigint formats are distinct at the type level", () => {
   expectTypeOf(z.int64()._zod.def.format).toEqualTypeOf<"int64">();
   expectTypeOf(z.uint64()._zod.def.format).toEqualTypeOf<"uint64">();
--- a/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
+++ b/packages/zod/src/v4/classic/tests/to-json-schema.test.ts
@@ -757,6 +757,26 @@ describe("toJSONSchema", () => {
     `);
   });
 
+  test("number constraints order independence with int()", () => {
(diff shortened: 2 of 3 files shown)


Relequestual:
Fast work! Thanks @colinhacks!

PR #6559, opened 2026-09-04
feat(compile): compiled issue generation by default, on per-subclass emitters

colinhacks:
`z.compile` now produces issues itself on invalid input instead of re-parsing through the interpreter, and the compiler is decomposed into per-subclass emitters in `core/jit.ts`.

**How it works.** Every emitter owns a fast generator and, where one exists, an issue generator. `compile()` builds the fast parser up front and the issue parser lazily on the first rejection, so a schema that never fails pays nothing for it. Failure sites never hand-build issues: a failed predicate calls the node's own runtime `parse` or `check` cold, so the issue shapes cannot drift from the interpreter's; codegen owns only the walk — path prefixes, abort gates, sibling continuation, partial output, the pipe stop rule. Anything without a native emitter (intersection, map, set, lazy, catch, codecs, coercion, xor) runs as an island through its own interpreter run. `{ issues: false }` keeps today's runtime re-parse, and `z.validate` still stops at the first failure.

**Why a table and not JIT constructors.** The `tuple-jit` branch had classic construct `$ZodXJIT` twins. Measured, that retains the whole compiler in every classic bundle (+5.7 KB gz on a bundle that never calls `z.compile`) and leaves mini's `z.compile` compiling nothing, which is presumably why bf7f5632 reverted it. Here classic and mini construct the plain core classes as before; `emitterFor(schema)` resolves the most specific recorded trait against the table only inside `compile()`, and a per-instance `_zod.codegen` override still wins.

**Parity.** The differential suite asserts deep-equal `error.issues` between the interpreter and the compiled parser on every fixture, and the compile-mode project re-runs the entire corpus under global compilation with issue mode on. Two additions close the holes that left: generated issue code that fails to evaluate is now a loud failure under strict mode (the global shim's), where before it fell back to the runtime re-parse and kept every parity test green over a broken emitter; and a seeded generative differential builds random schemas over every emitter-backed node type and the islands, corrupts valid values, and asserts the interpreter and the compiled schema agree on verdict, output, key order, issues and callback counts across 400 seeds. Its first run found that a rest tuple's fixed-item issues come after the rest elements' in the interpreter and came first from the compiler; fixed and pinned. 8,130 tests green.

**Three axes.**

- Bundle (gz, esbuild `--minify`): bundles that never call `z.compile` are unchanged (classic-object 25,459 → 25,487, mini-object 4,669 → 4,662). Bundles that do: classic-object 32,753 → 39,202 (+6.4 KB), mini-object 14,038 → 20,502 (+6.4 KB). If that is too much for compiler users who never read issues, the emitters can move behind their own entry; `issues: false` is already the runtime switch.
- Memory (retained per schema, measured as a slope over distinct schemas): unchanged until the first rejection (5-key compiled 15 KB, 243-leaf compiled 551 KB, same as main's compiler). A schema that has rejected once retains its issue parser: 5-key 15 → 18 KB, 243-leaf 551 → 724 KB. Generated issue code was cut 45% from the prototype (243-leaf: 237 → 131 KB of source); the remainder is bytecode.
- Runtime: the valid path is byte-identical to main's fast path. Failing parses: the compiled walk is 3–6x the interpreter's on nested objects, arrays and tuples. Wide schemas needed one more piece: V8 sizes a function's optimization budget by its bytecode, so the 243-leaf issue parser as one 131 KB function ran unoptimized for thousands of rejections and lost to the interpreter until then. Subtrees over a size threshold are now hoisted into functions of their own (union branches and isolated children too, which also drops a closure allocation per parse), and the 243-leaf failing parse goes from 3.5 µs to 1.0–1.3 µs at peak, 8–11x the interpreter, with the first 2,000 rejections costing the same as the interpreter instead of 2x. Small shapes were at or below 1x until two more pieces: a compiled `safeParse` that rejected re-entered its runtime method through a symbol-keyed params object, paying a context spread and six flag reads in the wrapper, so the methods now raise a module flag the wrapper consumes on entry (5-key object 1.08 → 1.20x, object union 0.95 → 1.03x); and the record fast path materialized every key through `Reflect.ownKeys`, now a for-in over a symbol snapshot taken before any value is read, with the same per-key enumerability recheck as the runtime (two valid keys 259 → ~90 ns, about 3x the interpreter; a proxy input sees one more descriptor lookup per key than it does from the interpreter). What is left on small shapes is the result object in `failure()` — a CPU profile puts it at 38% of self time on both paths — which is a public-shape question, not a codegen one; a bare `z.string()` stays at ~0.87x on a failing parse because that parse is the interpreter's own leaf parse cold-called plus dispatch.

Failing `safeParse` on a quiet machine (load 5–7, interleaved, min of 7 fixed-iteration rounds), interpreter vs `z.compile(s, { issues: false })` (today's runtime re-parse) vs compiled issues:

| shape | interpreter | `issues: false` | compiled issues |
| --- | --- | --- | --- |
| string | 238 ns | 272 (0.87x) | 254 (0.93x) |
| object-5key | 357 | 463 (0.77x) | 307 (1.16x) |
| nested-3deep | 408 | 571 (0.71x) | 352 (1.16x) |
| array-20 | 597 | 682 (0.88x) | 386 (1.55x) |
| tuple-4 | 482 | 545 (0.88x) | 336 (1.43x) |
| union-3obj | 1,465 | 1,197 (1.22x) | 1,312 (1.12x) |
| discunion-3 | 362 | 425 (0.85x) | 332 (1.09x) |
| leaf-243 | 11,463 | 10,229 (1.12x) | 3,862 (2.97x, 12x once tiered up) |

The compiled walk itself is 2–4x the interpreter's on containers (nested 100 vs 205 ns, array 143 vs 418, tuple 100 vs 204) and at parity on unions, where both sides spend the time finalizing per-branch issues; the rest of each row is the result object.

Two mode-invariant cuts rode along, both contract-neutral (every own-property descriptor of the error classes was diffed against main): `finalizeIssue` copies with a loop instead of object rest (139 → 99 ns), and the `$ZodError` initializer no longer redefines `_zod` with a descriptor `$constructor` had already applied (`new ZodError` 403 → 335 ns).

Follows on from #6538 and #6544.

### DIFF
--- a/packages/bench/compile-cold-call.ts
+++ b/packages/bench/compile-cold-call.ts
@@ -1,0 +1,128 @@
+// Where a compiled failing walk spends its time on the shapes that sit below their ceiling: the fast parser's rejection, the issue parser's walk, and the cold call that produces one issue. Fixed iterations, gc between samples, min of 7 interleaved rounds.
+import * as z from "zod";
+import { INVALID, coldParse, compileFn } from "../zod/src/v4/core/compile.js";
+
+declare global {
+  var gc: undefined | (() => void);
+}
+if (!globalThis.gc) {
+  console.error("run with --expose-gc");
+  process.exit(1);
+}
+
+interface Case {
+  name: string;
+  schema: z.ZodType;
+  valid: unknown;
+  invalid: unknown;
+  iters: number;
+}
+const cases: Case[] = [
+  {
+    name: "nested-3deep",
+    schema: z.object({ a: z.object({ b: z.object({ c: z.string() }) }) }),
+    valid: { a: { b: { c: "ok" } } },
+    invalid: { a: { b: { c: 42 } } },
+    iters: 200_000,
+  },
+  {
+    name: "array-20",
+    schema: z.array(z.number()),
+    valid: Array.from({ length: 20 }, (_, i) => i),
+    invalid: Array.from({ length: 20 }, (_, i) => (i === 10 ? "x" : i)),
+    iters: 100_000,
+  },
+  {
+    name: "tuple-4",
+    schema: z.tuple([z.string(), z.number(), z.boolean(), z.string()]),
+    valid: ["a", 1, true, "b"],
+    invalid: ["a", "no", true, "b"],
+    iters: 200_000,
+  },
+];
+
+let sink = 0;
+function sample(fn: () => unknown, iters: number): number {
+  globalThis.gc!();
+  const t0 = process.hrtime.bigint();
+  for (let i = 0; i < iters; i++) if (fn()) sink++;
+  return Number(process.hrtime.bigint() - t0) / iters;
+}
+function minOf(variants: Array<[string, () => unknown]>, iters: number): Map<string, number> {
+  const mins = new Map<string, number>();
+  for (const [, fn] of variants) sample(fn, iters / 10);
+  for (let r = 0; r < 7; r++) {
+    for (const [name, fn] of variants) {
+      const v = sample(fn, iters);
+      if (!mins.has(name) || v < mins.get(name)!) mins.set(name, v);
+    }
+  }
+  return mins;
+}
+
+for (const c of cases) {
+  const fast = compileFn(c.schema as never) as (v: unknown) => unknown;
+  const issues = compileFn(c.schema as never, { issues: true } as never) as unknown as (
+    v: unknown,
+    p: { value: unknown; issues: unknown[] },
+    ctx: undefined
+  ) => unknown;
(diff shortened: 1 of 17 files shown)


colinhacks:
Closing this out. Compiled issue generation doubles what `z.compile` adds to a bundle (7.3 → 13.7 KB gz on classic, 9.4 → 15.8 KB on mini) for a gain that only shows on failing parses, and with the record walk fixed the runtime fallback is at parity with the interpreter on everything but records. That leaves a second implementation of issue construction to keep aligned with the interpreter forever, for a path `z.validate` already serves when speed on rejection matters. The mode-independent pieces — the record walk, the cheaper issue finalization, the generative differential — land in #6567.

PR #6564, opened 2026-09-07
perf: seal the CommonJS exports so require("zod") stops reading through a getter

colinhacks:
TypeScript emits every re-export as a `__createBinding` accessor, so calling an API off the namespace — `z.validate(...)` — reads the function through a getter every time. The engine never sees a constant callee and the call is not inlined. 254 of the 257 exports on `index.cjs` were accessors:

```
require("zod")   z.validate(...)   ~26M ops/s
import "zod"     z.validate(...)   ~49M ops/s
```

Hoisting `const validate = z.validate` and calling that recovers all of it under `require`, which is what points at the namespace read rather than at the function behind it.

## The fix is in the build tool

The accessors come from TypeScript's CommonJS emit, which zshy applies to every library it builds, so the fix belongs there rather than in a zod postbuild script walking `dist`. [colinhacks/zshy#73](https://github.com/colinhacks/zshy/pull/73) adds `sealCjsExports`, released in zshy 0.8.0.

This PR is that option turned on for `zod` and `@zod/mini`, plus the dependency bump. `scripts/seal-cjs-exports.ts` and both postbuild hooks come out.

Every entrypoint goes from all-accessors to all-data-properties:

| entrypoint | before | after |
| --- | --- | --- |
| `index.cjs` | 254 getters / 257 | **0 / 257** |
| `mini/index.cjs` | 252 / 255 | **0 / 255** |
| `v4/core/index.cjs` | 305 / 308 | **0 / 308** |
| `v3/index.cjs` | 107 / 110 | **0 / 110** |

The ESM build is untouched, so tree-shaken bundle size does not move. The CommonJS output grows by the epilogue, ~37 KB across the 60 modules that receive it — the 63 locale files are single default exports, where zshy rebinds `module.exports` and skips the epilogue entirely.

`packages/resolution` gains an assertion that `validate` is a data property, so this cannot silently regress. attw's report is byte-identical to an unsealed build.


colinhacks:
Superseded by colinhacks/zshy#73. The accessors come from TypeScript's CommonJS emit, which zshy applies to every library it builds, so the fix belongs in the build tool rather than in a zod postbuild script.

Marking this a draft. Once zshy ships the option, this becomes `"sealCjsExports": true` in the zshy config of `packages/zod` and `packages/mini`, and `scripts/seal-cjs-exports.ts` comes out. The measurements and the `packages/resolution` assertion carry over unchanged.

colinhacks:
Checked the v3 claim before dropping it. `zod/v3`'s entrypoint ships 107 accessors and 3 data properties today, and assigning to a setter-less accessor already fails exactly the way a non-writable data property does. Sealing changes assignability for 2 of 110 top-level exports, and prototype extension — the v3 plugin surface — is unaffected, verified against the sealed build.

Both comments land on `scripts/seal-cjs-exports.ts`, which this PR is dropping. The fix moved to colinhacks/zshy#73, where the walk is mechanical and per-package; a path exclusion would be zod-specific config the build tool should not grow for this.

PR #6567, opened 2026-09-08
perf(compile): for-in record walk, cheaper issue finalization, and a generative compile differential

colinhacks:
The mode-independent half of #6559, which I'm closing in favor of this. `z.compile` keeps today's runtime re-parse on invalid input; what lands here is the part of that branch that helps regardless of how issues get built.

**Record fast path.** The compiled record walked keys through `Reflect.ownKeys` with a `propertyIsEnumerable` call per key, 154 ns on two keys before any value was checked. It now snapshots the string keys with `getOwnPropertyNames` and the symbol keys with `getOwnPropertySymbols` before any value is read, and rechecks each key's enumerability when it is visited, which is the runtime's `Reflect.ownKeys` walk exactly, order included: a getter that adds, deletes, hides or reveals a key mid-walk gets the runtime's verdict. Pullfrog took two cheaper snapshots off the table on the way here (for-in enumerates the prototype chain after the own keys, `Object.keys` drops a key that is non-enumerable at snapshot time) and both are pinned. Two valid keys 259 → ~106 ns against the interpreter's 214, ten keys 971 → ~496 against 864. A proxy input sees two `ownKeys` traps instead of the interpreter's one; that is the one accepted difference.

**Cheaper issue finalization.** `finalizeIssue` copies the raw issue's own keys with a loop over `Object.keys` instead of object rest with excluded keys, which V8 routes through a generic runtime call (175 → 147 ns in-process; a for-in walk was faster still but also copied inherited enumerable keys off an issue pushed as a class instance, which Copilot caught). Two deliberate differences from object rest: an own `__proto__` key on a raw issue, as `JSON.parse` produces, is dropped rather than assigned through the setter, the same as the record walk already does; and own symbol-keyed fields are not copied, since nothing in zod produces one and reading `getOwnPropertySymbols` on every issue would cost an allocation per issue plus bytes in every bundle to preserve a case that JSON serialization drops anyway; both are pinned. And the `$ZodError` initializer no longer redefines `_zod` with the descriptor `$constructor` already applied (`new ZodError` 403 → 335 ns). Every own-property descriptor of `ZodError`, `ZodRealError`, `$ZodError` and `$ZodRealError` was dumped on both revisions and is identical.

**Harness.** A seeded generative differential builds random schemas over the compiled node types, corrupts valid values, and asserts the interpreter and the compiled schema agree on verdict, output, key order, issues and per-callback run counts across 400 seeds (every callback the interpreter reaches, the compiled schema reaches at least once and at most twice), reporting every failing seed at once. The fixture suites only cover the compositions somebody wrote down; this is what caught the rest-tuple issue order divergence on the other branch. A memory bench measures retained bytes per compiled schema as a slope over distinct schemas, since a single-instance delta also carries one-time costs of whatever the schema touched first.

**Three axes.** Bundle (gz, esbuild `--minify`): classic-object+compile 32,978 → 33,032 (+54 B), mini-object+compile 13,963 → 14,036 (+73 B); bundles that never call `z.compile` move by 10 B or less, all of it the finalization loop; the `zod-mini-boolean` and `zod-mini-string` ceilings in the bundle-size test are re-measured for it. Memory: retained bytes per compiled schema unchanged (5-key 15 KB, 243-leaf 551 KB on the slope bench). Runtime: valid-path record parses about 2.4x faster as above, valid path otherwise byte-identical; a failing `safeParse` through the compiled fallback sits within 5% of the interpreter on every shape measured (string, flat and nested objects, arrays, tuples, unions) and at 0.90x on records, where the walk before the failing key is the difference.

### DIFF
--- a/packages/bench/error-construction-breakdown.ts
+++ b/packages/bench/error-construction-breakdown.ts
@@ -1,0 +1,83 @@
+// Where the time goes when a caller reads `.error` on a failing safeParse: finalizeIssue per issue, the ZodError construction, and the plain Error baseline. Fixed iterations, gc between samples, min of 7 interleaved rounds.
+import * as z from "zod";
+import * as core from "../zod/src/v4/core/index.js";
+import * as util from "../zod/src/v4/core/util.js";
+
+declare global {
+  var gc: undefined | (() => void);
+}
+if (!globalThis.gc) {
+  console.error("run with --expose-gc");
+  process.exit(1);
+}
+
+const inst = z.string();
+const raw = () => ({ code: "invalid_type", expected: "string", input: 42, inst }) as unknown as core.$ZodRawIssue;
+const ctx = { async: false } as core.ParseContextInternal;
+const finalized = util.finalizeIssue(raw(), ctx, core.config());
+const finalizedList = [finalized];
+
+let sink = 0;
+function sample(fn: () => unknown, iters: number): number {
+  globalThis.gc!();
+  const t0 = process.hrtime.bigint();
+  for (let i = 0; i < iters; i++) if (fn()) sink++;
+  return Number(process.hrtime.bigint() - t0) / iters;
+}
+
+// candidate finalizeIssue bodies, same output: object rest (shipped) vs an explicit copy loop
+const SKIP = new Set(["inst", "schema", "continue", "input"]);
+function finalizeLoop(iss: any, c: any, cfg: any): any {
+  const traits: Set<string> | undefined = iss.inst?._zod?.traits;
+  if (traits?.has("$ZodType")) {
+    if (traits.has("$ZodCheck")) iss.schema ??= iss.inst;
+    else iss.schema = iss.inst;
+  }
+  const schemaError = iss.schema !== iss.inst ? iss.schema?._zod.def?.error : undefined;
+  const message = iss.message
+    ? iss.message
+    : (util.unwrapMessage(iss.inst?._zod.def?.error?.(iss)) ??
+      util.unwrapMessage(schemaError?.(iss)) ??
+      util.unwrapMessage(c?.error?.(iss)) ??
+      util.unwrapMessage(cfg.customError?.(iss)) ??
+      util.unwrapMessage(cfg.localeError?.(iss)) ??
+      "Invalid input");
+  const out: any = {};
+  for (const k in iss) if (!SKIP.has(k)) out[k] = iss[k];
+  out.path ??= [];
+  out.message = message;
+  if (c?.reportInput) out.input = iss.input;
+  return out;
+}
+const localeError = core.config().localeError!;
+
+const variants: Array<[string, () => unknown]> = [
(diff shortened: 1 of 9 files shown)


wahajahmed010:
Solid perf PR, and the security angle is a real bonus. A few things I want to flag:

**1. Symbol-keyed own properties are dropped — silent behavior change.** The previous `const { ..., ...rest } = iss` spread includes *both* string and symbol own keys (object rest uses `OwnPropertyKeys`, not `Object.keys`). The new `for (const k of Object.keys(iss))` only iterates string keys. If a user-defined check ever pushes an issue with a symbol-keyed field (e.g. `ctx.issues.push({ [Symbol.toPrimitive]: ..., code: 'custom' })`), that field will silently disappear from the finalized issue.

The benchmark comment mentions this is intentional ("inherited keys stay out"), but inherited is not the same as own-symbol. Worth either:
- Switching to `Reflect.ownKeys(iss)` and filtering the symbol branch the same way (`continue` on `__proto__`/`inst`/etc.), or
- Adding an explicit test that an issue with a symbol-keyed own property is preserved, to lock in the chosen behavior.

**2. The `__proto__` skip is good — but `Object.keys` already excludes it.** A JSON-parsed payload like `{"__proto__": ...}` produces an own `__proto__` *data* property, which `Object.keys` does include. The new code handles that case (the new test confirms it). However: assigning `full[k] = iss[k]` on any string key that happens to be a name like `toString`/`constructor` is harmless here because `full` is a fresh `{}`, so there is no setter on the destination. The explicit `__proto__` filter is therefore defensive and worth keeping. Good.

**3. The "object rest routes through a generic runtime call" comment — is this observable?** The new bench file measures ~300k iterations with `--expose-gc` and reports `min of 7 interleaved rounds`. Two suggestions:
- Show the actual numbers in the PR description (or link to a gist) so reviewers can sanity-check the headline claim rather than re-running locally.
- Watch out for the `core.config()` call in the bench loop — `config()` is a memoized getter but if it does any work the first time it could dominate the per-iter cost. A quick `console.log(config() === config())` check or a hoist into a const would make the measurement cleaner.

**4. `errors.ts` removes the `_messageDesc` — please confirm GC behavior.** The diff drops one of the property descriptors and an export; if `_messageDesc` was being held alive by something other than the module top-level (e.g. a captured closure in an installed formatter), removing it could change when message strings get collected. Skim the deleted site to make sure nothing else references it; the test suite will too, but it is the kind of thing CI may not catch.

**5. New `compile-generative.test.ts` (445 lines) — looks well-scoped from the diff stats but I have not read every case.** One thing to confirm: the generative tests compare against the runtime path. If the codegen change in `generateRecordCheck` produces a different *enumeration order* than `Reflect.ownKeys(input)`, the test needs to sort keys before comparing. Worth a glance.

Overall: ship-able after addressing #1. The proto-pollution hardening is the most valuable part of this PR for downstream users.

PR #6587, opened 2026-09-10
Fix undefined prefault outputs and object keys

### LINKED ISSUE #6585: prefault(undefined) drops object key and removes undefined from inferred output
Using `zod 4.6.1`

I was trying to make a parser returns `undefined` for keys that wasn't in the parsed data at all using `prefault(undefined)` or `prefault(() => undefined)`

```typescript
const undefinedPrefault = z.object({
  a: z.union([z.string(), z.undefined()]).prefault(() => undefined),
});

undefinedPrefault.parse({}); // Should return {a: undefined}
```

And it seems that there are 2 problems with this code:
 - The inferred type of`undefinedPrefault` here is just `{a: string}`, I would expect it to be `{a: string | undefined}`
 - and the prefault value is not applied so the parse returns `{}`

I tried with null instead of undefined and it seems to work:

```typescript
import { z } from "zod";

type Equal<T, U> = [T] extends [U] ? ([U] extends [T] ? true : false) : false;

function assertType<T extends true>() {
  return true as T;
}

const nullPrefault = z.object({
  a: z.union([z.string(), z.null()]).prefault(() => null),
});

const undefinedPrefault = z.object({
  a: z.union([z.string(), z.undefined()]).prefault(() => undefined),
});

// Pass: This is right ✅
assertType<
  Equal<z.infer<typeof nullPrefault>, { a: string | null }>
>();

console.log("Parsed -->", nullPrefault.parse({}));
// Parsed  --> { a: null } ✅

// Should pass but doesn't: wrong ❌
assertType<
  Equal<
    z.infer<typeof undefinedPrefault>,
    { a: string | undefined }
  >
>();

type Debug = z.infer<typeof undefinedPrefault>;
//    ^? type Debug = { a: string } ❌

console.log("Parse -->", undefinedPrefault.parse({}));
// Parsed  --> { }  ❌
```

you can test this example here https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgLzgXzgMyhEcBEyEAJvgNwBQFMAnmAKZwCiAjgK4CGANgDwAqAGjgBVAHxwAvHADafALpx6ADxj0AdsQDOM4QoD8cABTTdileq0z5cAzChtGALizdN9AJRxnmV-UoVMNjUAYxhgCDU4Dk03WD46en4zVQ1tOwdRQ08ECjg4KHoYNihI9MZouD5KNCpgiM14NTYuLgAFAp9m+ClkADoIACMAK3pQwxy8jmc+oPC1Yz6GqGA1AHMsoT6mlqy5d16wDo4uwyzJcW2udwEKNHd-OrUGuCDiekwV+mJ29+OubpQ-WGoxg41yUWmvVmEQWvSWK3W10Br3en2Iu32h1+JzOEnEKI+ai+11u9yoAHpyXBWtFNM4+AALYDaZn5YCrBnwQCg5BRafQ4gkeKxOLw+itMPyeLQGBBMHBLj9Ov9REIkFM4PC1nAAD7y5pcdCiTJkiiPTQQLj0XpcCDrfA0qBuYhwAC0LtE+CECqOXQOHEd9HGdzJlOp-qdeTd4jVzku6DgPIooYAygyIM1nWBaXABmx4MQIPRNGoAOQwZwAdxwWsAMuS8mL8mDxBg8cHC7g8MVqCVQKUJWUvDSoonfH3K1UQjV2BE6wdvQlfQ0UY3+aWMAAi9Fzq0kgPFkrXA4JaMVfxgokoobycAAerYEnBN9vdzGp8stRg66b6harTa7Q6biuu6npzsOXynr6WYBkG7ghlSgGLsB0bxnAdZAA

colinhacks:
Fixes #6585.

Preserve `undefined` in prefault output types and retain missing object keys when a defaulted field has required output. This also fixes transforms that return `undefined` after receiving a defined prefault. Mini now accepts the same prefault inputs as Classic.

The object change uses existing optionality metadata across interpreted, JIT, and compiled parsing. Optional output fields still omit absent keys. Regression tests cover types, callbacks, wrappers, transforms, async parsing, and direct compiled output.

Retained schema memory is unchanged. The gzipped object fixture grows by 6 bytes for Classic and 14 bytes for Mini; boolean fixtures are unchanged. Local runtime samples were rejected because the host was saturated, so no performance claim is made.

### DIFF
--- a/packages/zod/src/v4/classic/tests/prefault.test.ts
+++ b/packages/zod/src/v4/classic/tests/prefault.test.ts
@@ -1,6 +1,76 @@
 import { expect, expectTypeOf, test } from "vitest";
 import { z } from "zod/v4";
 
+test.each([false, true])("undefined prefault preserves object keys (jitless: %s)", async (jitless) => {
+  const previous = z.config().jitless;
+  z.config({ jitless });
+  try {
+    let calls = 0;
+    const field = z.union([z.string(), z.undefined()]).prefault(() => {
+      calls++;
+      return undefined;
+    });
+    const schema = z.object({ a: field });
+    expectTypeOf<z.output<typeof field>>().toEqualTypeOf<string | undefined>();
+    expectTypeOf<z.input<typeof schema>>().toEqualTypeOf<{ a?: string | undefined }>();
+    expectTypeOf<z.output<typeof schema>>().toEqualTypeOf<{ a: string | undefined }>();
+    expect(schema.parse({})).toStrictEqual({ a: undefined });
+    expect(calls).toBe(1);
+    expect(schema.parse({ a: undefined })).toStrictEqual({ a: undefined });
+    expect(calls).toBe(2);
+    expect(await schema.parseAsync({})).toStrictEqual({ a: undefined });
+    expect(calls).toBe(3);
+    expect(schema.parse({ a: "value" })).toStrictEqual({ a: "value" });
+    expect(schema.safeParse({ a: 123 }).success).toBe(false);
+    expect(calls).toBe(3);
+
+    const literal = z.prefault(z.string().optional(), undefined);
+    expect(z.core.compileFn(z.object({ a: literal }))({})).toStrictEqual({ a: undefined });
+    for (const wrapped of [literal, literal.readonly(), literal.nullable(), z.lazy(() => literal)]) {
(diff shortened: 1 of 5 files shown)


colinhacks:
Released in [v4.6.2](https://github.com/colinhacks/zod/releases/tag/v4.6.2).

PR #6594, opened 2026-09-12
Make z.properties() a check again

colinhacks:
Reverts the schema role that #6536 gave `z.properties()`. It is a check again, as in 4.5: it spreads into `.check()`, and `z.instanceof().properties()` stays as sugar over it. The standalone form is gone — `z.properties(shape).parse()`, `z.infer` of it, `z.toJSONSchema()` and `z.compile()` of it, and the cycle tracking only the schema role had.

The schema form was a trap with transforms. A transform inside the shape ran and was discarded, so the parsed value and the inferred type were both the input side, and writing the result back is not an option for the objects the feature exists for: assigning to a getter-only property such as `Response.status` throws. As a check, "the value is never changed" is what a check already means.

What stays: every failing key reports, symbol keys, the `params` argument, and compile support. The check now lives next to `$ZodCheckProperty` as `$ZodCheckProperties` and installs its `Symbol.iterator` as an own property, so the symbol-key loop in `util.members` and the memoizer's callable arm go with the schema role. The API reference drops the standalone paragraph; the 4.6 post already led with the method.

Bundle, gzipped from the built package: mini −17 / −16 / −16 on boolean / string / object, classic −35 / −38 / −44, the named import −447. Runtime and memory: the check body is the same code moved, construction pays one `Object.defineProperty` per `z.properties()` call, and each check instance carries one extra own property.

This removes an API that shipped in 4.6.0.

### DIFF
--- a/packages/docs/content/api.mdx
+++ b/packages/docs/content/api.mdx
@@ -2352,19 +2352,9 @@ type OkResponse = z.infer<typeof okResponse>;
 // => Response & { status: number; redirected: false }
 ```
 
-The underlying `z.properties()` is a standalone schema. It asserts the named properties in place and returns its input untouched — unlike `z.object()`, it never builds a new object, so prototypes, unlisted keys, and object identity survive. Schemas inside the shape are validated but their results are discarded, so a transform or a default never changes the parsed value.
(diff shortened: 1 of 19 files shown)


colinhacks:
Landed in [v4.6.3](https://github.com/colinhacks/zod/releases/tag/v4.6.3).

PR #6595, opened 2026-09-12
Add z.currencyCode() over a vendored ISO 4217 list with a weekly refresh

colinhacks:
Adds `z.currencyCode()`, a format validator for three-letter ISO 4217 currency codes. The list is vendored rather than taken from the `currency-codes` npm package: that package still ships the June 2024 list and its maintainer has been silent since, so it has no XCG or XAD and still treats BGN as active.

The source is SIX Group's list one, the XML the ISO maintenance agency publishes. A script (`nub run update:iso4217`) fetches it and rewrites the regex in `core/regexes.ts`, and a weekly workflow reruns the script and opens a PR whenever the active codes change, naming the added and removed codes. The workflow opens that PR with the default token, which never triggers the test workflow, so it runs the tests that read the list before it pushes.

Placement is top-level on purpose. Under esbuild a member of a nested namespace is paid for by everyone who touches that namespace, so `z.iso.currency()` would have cost every `z.iso.date()` user the whole list, while a top-level export is free for anyone who does not call it. The same mechanism was already charging `z.hostname()`, `z.hex()` and `z.httpUrl()` for every regex in the module, because they reached it as `core.regexes.x`. They now import the module directly, and a treeshake test pins that.

Bundle, esbuild and gzipped:

- a mini bundle that calls `z.currencyCode()`: +944 B over `z.string()` alone, of which the regex is 435 B
- `z.hostname()` mini 5449 → 3390, `z.httpUrl()` mini 5703 → 3650, classic `z.hostname()` 21033 → 19136
- bundles that use none of these: unchanged, apart from +18 B on full classic for the export itself

Runtime is a 178-way anchored alternation, about 20 ns per test, under the E.164 regex. Memory matches `z.hostname()` at 1420 retained bytes per schema. Every locale gets a `currency_code` noun, which the parity test requires.


colinhacks:
Landed in [v4.6.4](https://github.com/colinhacks/zod/releases/tag/v4.6.4).

PR #6599, opened 2026-09-13
perf(v4): spike a novalue path for the interpreted validate

colinhacks:
`z.validate()` on an uncompiled schema is barely faster than `safeParse().success` on valid input — median 1.1x across the sweep, and 0.89x on a plain `z.string()`. The reason is that its one optimization, the early exit from #6544, only fires when something fails. On a passing parse nothing aborts, so `validate` runs the same `_zod.run` and builds the same output object that nobody will read.

This branch prototypes the other half. `ctx.noValue` tells a container that nothing reads what it produces, so it skips the stores into its result. Two useful commits: a spike that measures the ceiling, then the propagation that makes it correct.

### The rule, at runtime

The compiler already decides this at codegen with one line — `buildsValue = needsValue || !!def.checks?.length`, threaded top-down through the walk. At runtime the same rule has to travel per node while the parse context is per call, so a node that reads what it built clears the flag for the subtree below it by cloning the context. There are three, and the clone only ever happens under `validate`:

- **the checks wrapper**, one site in `$ZodType.init`, which needs no knowledge of the type it wraps
- **a pipe's input side**, whose value is what the output side parses
- **both intersection operands**, which `mergeValues` reads

`abortEarly` and `noValue` stay separate flags because of that clone: `abortEarly` is never cleared, `noValue` is, so a subtree below a checked node runs with the early exit on and the skip off.

A container still allocates its placeholder. Classic installs a memoizer by default and `memo.alloc` registers what a back-edge resolves to, so skipping the allocation breaks a cyclic input — and guarding on `!memo` instead silently disables the whole optimization, since `memo` is always set there. Allocating an empty `{}`/`[]` and skipping only the per-key stores keeps cycles working, and it keeps `payload.value` off the caller's input, which is what would otherwise let a loose object's catchall write into it and `.readonly()` freeze it.

### What it buys

Uncompiled `z.validate()`, valid input, minimum of three alternating process runs per side against current `main`:

| schema | main | branch | |
| --- | --- | --- | --- |
| `z.object()`, 50 keys | 599ns | 367ns | −38.7% |
| `z.object()`, 20 keys | 193ns | 156ns | −19.2% |
| `z.array(z.string())`, 50 | 779ns | 679ns | −12.9% |
| `z.array(z.string())`, 10 | 193ns | 175ns | −9.3% |
| nested object | 98ns | 93ns | −5.0% |
| `z.array(z.object())`, 10 | 357ns | 344ns | −3.7% |
| `z.object()`, 5 keys | 60ns | 59ns | −2.5% |

The win is the stores a container skips, so it scales with how much the container writes and disappears on leaves. Compiled schemas move 0.3% median, which is the control — this branch never reaches the compiled path.

### What it costs

- **Uncompiled `safeParse` pays about 2.5%** median on valid input, worst 7.9% on the 20-key object, against a −0.3% median on the untouched compiled rows. That is the branch on `ctx.noValue` in each container, on the call everybody makes.
- **Bundle: +138 B gzipped** on the classic object fixture and **+40 B on mini**, measured by removing one piece at a time:

| piece | mini boolean | mini string | mini object | classic object |
| --- | --- | --- | --- | --- |
| the checks-wrapper clone | 22 B | 18 B | 12 B | 14 B |
| the interpreted object skip | 0 | 0 | 22 B | 30 B |
| the second generated fastpass | 0 | 0 | 0 | 46 B |
| pipe + intersection clears | 0 | 0 | 0 | 28 B |
| the array skip | 0 | 0 | 0 | 24 B |
| the flag and its type field | 0 | 0 | 0 | 10 B |
| **total** | **+22 B** | **+18 B** | **+40 B** | **+138 B** |

  The rows do not sum to the totals, because gzip is not additive; each is what that piece costs with the rest present. The first row is the uncomfortable one — those two lines ship in every bundle, so a `zod/mini` caller with `z.boolean()` and no `validate` anywhere pays 22 B, 0.7% of that bundle, for machinery it cannot reach.
- **Memory: +121 B retained per object schema** called both ways, for the second generated fastpass.
- On invalid input the cross-revision deltas are noise — the untouched compiled rows move 5.4% median — so this run supports no claim there either way. `abortEarly` already owns that path.

### Coverage

The generative differential now runs every random schema through `z.validate` as well: same verdict as `safeParse`, same value seen by a check at the root, and an input that comes back unmutated and unfrozen. Removing the propagation fails it at seed 1. Before that test existed the whole 8124-test suite passed against the spike, which returns wrong booleans and throws.

### Not done

`validateAsync` sets no flag, so async validate gains nothing. Only `z.object()` and `z.array()` honor it; record, map, set and tuple still build what nobody reads.


colinhacks:
Not taking this. The win is concentrated in objects wide enough that most people never write one — 1.63x at fifty keys, 1.24x at twenty, nothing distinguishable from noise at five — while the cost is a couple of percent on every uncompiled `safeParse` and 22 gzipped bytes in every Zod Mini bundle, including bundles that can't reach `validate` at all. Anyone who needs a fast `validate` gets far more from `z.compile()`, and only pays for it if they ask for it.

Leaving the branch up as the measurement. It records the ceiling for a runtime-side skip, and three things that make it harder than it looks: Classic installs a memoizer by default, so guarding on its absence silently disables the whole optimization; a container that skips its allocation writes into the caller's input and freezes it under `.readonly()`; and the whole suite passed against a version that returned wrong booleans and threw.

